PR #4, opened 2023-04-18
attempt xformers hijack attention

winglian:


winglian:
already cherry-picked or rebased from somewhere else

PR #9, opened 2023-04-25
feature dump into main

winglian:
- support for Galactica @ehartford 
- support specific data files from hf datasets
- drops data points from dataset if input, attention mask and label tensors don't match in length
- ignore system prompt from sharegpt dataset
- cache prepared datasets right after tokenization to allow for more reconfiguration without rebuilding cache
- support dataset shards for training
- require explicit 8bit Adam in config for optimizer


ehartford:
Thank you!

ehartford:
addresses https://github.com/winglian/axolotl/issues/7

PR #12, opened 2023-05-06
Add eval_batch_size for evaluation

NanoCode012:
Problem:

I would OOM during the evaluation stage despite setting `micro_batch_size` to be low. Turns out that batch size for eval using a different parameter.

Proposal:

New config for eval stage for lower VRAM GPUs.

Edit:

Since it's now a somewhat required argument, I'm not sure whether we should
- assert value exist
- set some default if not in config


winglian (approved):
thanks!

PR #23, opened 2023-05-09
Fix: Save adapter for lora

NanoCode012:
In the latest commit, the code has changed to `save_model`. This would not save Loras. I added an extra check if it's training lora, to also `save_pretrained`.

### DIFF
--- a/scripts/finetune.py
+++ b/scripts/finetune.py
@@ -228,11 +228,12 @@ def train(
     logging.info(
         f"Training Completed!!! Saving pre-trained model to {cfg.output_dir}"
     )
-    # TODO do we need this fix? https://huggingface.co/docs/accelerate/usage_guides/fsdp#saving-and-loading
-    trainer.save_model(cfg.output_dir)
     
     if cfg.adapter == 'lora':
         trainer.save_pretrained(cfg.output_dir)
+    else:
+        # TODO do we need this fix? https://huggingface.co/docs/accelerate/usage_guides/fsdp#saving-and-loading
+        trainer.save_model(cfg.output_dir)
 
 
 if __name__ == "__main__":
--- a/scripts/finetune.py
+++ b/scripts/finetune.py
@@ -229,11 +229,10 @@ def train(
         f"Training Completed!!! Saving pre-trained model to {cfg.output_dir}"
     )
     
-    if cfg.adapter == 'lora':
-        trainer.save_pretrained(cfg.output_dir)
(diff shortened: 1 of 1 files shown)


scripts/finetune.py
```diff
@@ -230,6 +230,9 @@ def train(
     )
     # TODO do we need this fix? https://huggingface.co/docs/accelerate/usage_guides/fsdp#saving-and-loading
     trainer.save_model(cfg.output_dir)
```
winglian: should we if/else with `trainer.save_model(cfg.output_dir)` ? I expect saving the model if using lora is a pointless exercise since that would simply be the base model?

NanoCode012: Ah, I see! I thought `save_model` would at least output the checkpoint which could be used to train further.

NanoCode012: Would this be ok?

winglian: so https://github.com/tloen/alpaca-lora/blob/main/finetune.py#L275 and https://github.com/johnsmith0031/alpaca_lora_4bit/blob/main/finetune.py#L187 both use save_pretained, but I forget why I changed it to `save_model`. thoughts?

winglian: looks like I changed it here when I refactored https://github.com/OpenAccess-AI-Collective/axolotl/commit/9105935b0065e18466d7aaff4d51fde05353ca27#diff-8f286b88d68f56540a5fd24b9e81b2d6b0f4a1bd97bb62204bd32a8f92893f93R222

winglian: let's just change it back to save_pretrained and lose the seemingly broken save_model

winglian: actually, let me dig into the trainer code more again

NanoCode012: Feel free to edit the PR :) It's open to edits.

Edit: I’m not sure why it got changed. I just noticed it got updated one day and forgot to ask.

winglian:
@NanoCode012 btw, this works with LoRAs right? I ran a full tuned model and got this error:
```
  File "/workspace/axolotl/scripts/finetune.py", line 234, in train
    trainer.save_pretrained(cfg.output_dir)      
AttributeError: 'Trainer' object has no attribute 'save_pretrained'
```

NanoCode012:
Oh, this is completely my bad. It should've been `model.save_pretrained`. Could you try resume checkpoint with this new code?

PR #36, opened 2023-05-23
Qlora

winglian:
based on https://github.com/huggingface/peft/pull/476/files

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,10 +1,10 @@
 peft @ git+https://github.com/huggingface/peft.git
 transformers @ git+https://github.com/huggingface/transformers.git
+bitsandbytes @ git+https://github.com/TimDettmers/bitsandbytes.git
 attrdict
 fire
 PyYAML==6.0
 black
-bitsandbytes==0.37.2
 datasets
 accelerate>=0.19.0
 sentencepiece
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 peft @ git+https://github.com/huggingface/peft.git
 transformers @ git+https://github.com/huggingface/transformers.git
-bitsandbytes @ git+https://github.com/TimDettmers/bitsandbytes.git
+bitsandbytes>=0.39.0
 attrdict
 fire
 PyYAML==6.0
--- a/src/axolotl/utils/models.py
+++ b/src/axolotl/utils/models.py
@@ -6,11 +6,12 @@ from typing import Optional, Tuple, TYPE_CHECKING
 
 import torch
 import transformers
+from torch import nn
 from transformers import (
     AutoModelForCausalLM,
     AutoTokenizer,
     PreTrainedModel,
-    AutoConfig,
+    AutoConfig, BitsAndBytesConfig,
 )
 
 try:
@@ -81,6 +82,16 @@ def load_model(
         logging.exception(e)
         raise e
 
+    model_kwargs = {}
+    if cfg.adapter == "qlora":
+        model_kwargs["quantization_config"] = BitsAndBytesConfig(
+            load_in_4bit=True,
+            llm_int8_threshold=6.0,
+            llm_int8_has_fp16_weight=False,
+            bnb_4bit_compute_dtype=torch.float16,
+            bnb_4bit_use_double_quant=True,
+            bnb_4bit_quant_type="nf4",
+        )
     try:
         if cfg.load_4bit and is_llama_derived_model:
             from alpaca_lora_4bit.autograd_4bit import load_llama_model_4bit_low_ram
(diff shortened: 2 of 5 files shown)


requirements.txt
```diff
@@ -1,10 +1,10 @@
 peft @ git+https://github.com/huggingface/peft.git
 transformers @ git+https://github.com/huggingface/transformers.git
+bitsandbytes @ git+https://github.com/TimDettmers/bitsandbytes.git
```
NanoCode012: `pip install bitsandbytes>=0.39.0` based on his new PR to huggingface transformers.

The main repo is `38.1`. I am not sure how his main repo version is older than the required version. Seems like he has a `closed_beta` branch.

Ref: https://github.com/huggingface/transformers/blob/0b050922b70fd9864b4b751e83f02b7f7cbd9e2d/docs/source/en/main_classes/quantization.mdx

```suggestion
bitsandbytes>=0.39.0
```

src/axolotl/utils/models.py
```diff
@@ -81,6 +82,16 @@ def load_model(
         logging.exception(e)
         raise e
```
NanoCode012, line 84: We need to also
- assert that both `cfg.load_in_8bit` and `qlora` and `load_4bit` are not True. Only one can be True at a time.
- Fix the `torch_dtype` since it does not factor in for qlora 4bit.
- Would `bf16` affect the training? Looking at the sample fine-tuning code, they used `fp16`.

(I cannot reference those lines in the review UI)

src/axolotl/utils/models.py
```diff
@@ -184,8 +198,24 @@ def load_model(
             torch_dtype=torch_dtype,
             device_map=cfg.device_map,
             trust_remote_code=True if cfg.trust_remote_code is True else False,
+            **model_kwargs,
         )
 
+    """### Post-processing on the model
+        Finally, we need to apply some post-processing on the 8-bit model to enable training, let's freeze all our layers, and cast the layer-norm in `float32` for stability. We also cast the output of the last layer in `float32` for the same reasons.
+    """
+    if cfg.adapter == "qlora":
```
NanoCode012: Maybe move this processing into `load_lora`  for cleaner code

NanoCode012 (commented):
Hello, I have added some thoughts. I have not tested it myself yet

NanoCode012:
From paper and qlora repo, 

>  we find that the most critical LoRA hyper-parameter is how many LoRA adapters are used in total and that LoRA on all linear transformer block layers are required to match full finetuning performance.

```python
def find_all_linear_names(args, model):
    cls = bnb.nn.Linear4bit if args.bits == 4 else (bnb.nn.Linear8bitLt if args.bits == 8 else torch.nn.Linear)
    lora_module_names = set()
        if isinstance(module, cls):
            names = name.split('.')
            lora_module_names.add(names[0] if len(names) == 1 else names[-1])


    if 'lm_head' in lora_module_names: # needed for 16-bit
        lora_module_names.remove('lm_head')
    return list(lora_module_names)

modules = find_all_linear_names(args, model)

config = LoraConfig(
        ...
        target_modules=modules,
        ...
    )
```

Edit: If we add this, we need to modify to also consider `cfg.lora_modules_to_save`

src/axolotl/utils/models.py
winglian: I addressed #1, I didn't see anything that needed to be addressed for #2 (was there something else you caught that I missed?) I think bf16 would be better for the parameters that use that since it would avoid potential overflows.

winglian:
> From paper and qlora repo,
> 
> > we find that the most critical LoRA hyper-parameter is how many LoRA adapters are used in total and that LoRA on all linear transformer block layers are required to match full finetuning performance.
> 
> ```python
> def find_all_linear_names(args, model):
>     cls = bnb.nn.Linear4bit if args.bits == 4 else (bnb.nn.Linear8bitLt if args.bits == 8 else torch.nn.Linear)
>     lora_module_names = set()
>     for name, module in model.named_modules():
>         if isinstance(module, cls):
>             names = name.split('.')
>             lora_module_names.add(names[0] if len(names) == 1 else names[-1])
> 
> 
>     if 'lm_head' in lora_module_names: # needed for 16-bit
>         lora_module_names.remove('lm_head')
>     return list(lora_module_names)
> 
> modules = find_all_linear_names(args, model)
> 
> config = LoraConfig(
>         ...
>         target_modules=modules,
>         ...
>     )
> ```
> 
> Edit: If we add this, we need to modify to also consider `cfg.lora_modules_to_save`

nice! can we add this to a separate PR?

NanoCode012:
> nice! can we add this to a separate PR?

Yes! I think this should be tested first especially the part that removes `lm_head`

How were qLORA results?

PR #37, opened 2023-05-24
fix: handles AutoTokenizer from untrusted source

Thytu:
Set `trust_remote_code` param depending of `cfg.trust_remote_code` when calling `AutoTokenizer.from_pretrained`.

Prevents the following error:
```sh
Loading SOME_NAME requires you to execute the configuration file in that repo on your local machine. Make sure you have read the code there to avoid malicious use, then set the option `trust_remote_code=True` to remove this error.
```


winglian (approved):

winglian:
thanks!

PR #41, opened 2023-05-25
attempt to find linear modules for qlora

winglian:


src/axolotl/utils/models.py
```diff
@@ -343,12 +362,15 @@ def load_lora(model, cfg):
         PeftModel,
     )
 
-    lora_config = None
+    bits = 4 if cfg.load_in_4bits else 8 if cfg.load_in_8bits else None
```
NanoCode012: Simplify code to make it easier to read
```suggestion
    bits = None
    if cfg.cfg.load_in_4bits: 
        bits = 4
    elif cfg.cfg.load_in_8bits: 
        bits = 8
```

Edit: Not sure why `elif` is not highlighted. Did I make typo?

NanoCode012 (commented):
Looking good! I'm hopeful for this.

PR #42, opened 2023-05-25
Feat: Add torch to requirements

NanoCode012:
Error with `pip install -e .` in new venv due to missing torch.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,3 +11,4 @@ sentencepiece
 wandb
 einops
 xformers
+torch


winglian:
torch gets installed in the[ base docker env](https://github.com/OpenAccess-AI-Collective/axolotl/blob/main/docker/Dockerfile-base#L32), but it definitely should exists in requirements. But I think when pip install runs, it tries to delete the version that's already installed. thoughts?

NanoCode012:
Hm, when I ran `pip install -e .` in my docker image. It does not install torch again.
```
Requirement already satisfied: torch>=1.6.0 in /root/miniconda3/envs/py3.9/lib/python3.9/site-packages (from accelerate) (2.0.0)
```

winglian:
thanks, what about different version of torch? for example we build images with 1.13.1 and 2.0.0 
https://github.com/OpenAccess-AI-Collective/axolotl/blob/main/.github/workflows/base.yml#L17-L22

NanoCode012:
I will build the Dockerfile following your dev image for 1.13.1 from docker hub and let you know.

NanoCode012:
You're right. It does try to install torch 2.0

NanoCode012:
Shall we update the Readme to require Pytorch installed and add a commented to QuickStart 

```bash
# pip3 install torch
```

and requirements?

```bash
# torch
```

winglian:
> Shall we update the Readme to require Pytorch installed and add a commented to QuickStart
> 
> ```shell
> # pip3 install torch
> ```
> 
> and requirements?
> 
> ```shell
> # torch
> ```

yeah, I think adding it to the docs is sufficient, something like `if torch isn't installed yet...`

winglian:
what about `torch>=1.13.1` ?

NanoCode012:
> what about `torch>=1.13.1` ?

```bash
=> => # Collecting torch>=1.13.1 (from axolotl==0.1)                                                                                                                                                                 
 => => #   Downloading torch-2.0.1-cp39-cp39-manylinux1_x86_64.whl (619.9 MB)
```

PR #45, opened 2023-05-25
Fix: Incorrect config name

NanoCode012:
I made a mistake calling cfg.cfg while copy pasting.

Another thing I noticed is 4bit config being named differently from 8bit load. Should we change them to be consistent?

### DIFF
--- a/src/axolotl/utils/models.py
+++ b/src/axolotl/utils/models.py
@@ -137,7 +137,7 @@ def load_model(
             model = LlamaForCausalLM.from_pretrained(
                 base_model,
                 load_in_8bit=cfg.load_in_8bit and cfg.adapter is not None,
-                load_in_4bit=cfg.load_in_4bit and cfg.adapter is not None,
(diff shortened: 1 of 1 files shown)


NanoCode012:
Before merge, I want to confirm another potential mistake.

https://github.com/OpenAccess-AI-Collective/axolotl/blob/a617f1b65eb3d986ab7844630944fe4c979158fe/src/axolotl/utils/models.py#L49

is only modified here

https://github.com/OpenAccess-AI-Collective/axolotl/blob/a617f1b65eb3d986ab7844630944fe4c979158fe/src/axolotl/utils/models.py#L135

and used here

https://github.com/OpenAccess-AI-Collective/axolotl/blob/a617f1b65eb3d986ab7844630944fe4c979158fe/src/axolotl/utils/models.py#L257

Do you mean to use `cfg.load_in_8bit` instead of `load_in_8bit`?

PR #57, opened 2023-05-25
fixes w/ example for super basic lora starter

winglian:


src/axolotl/prompters.py
```diff
@@ -18,7 +18,7 @@ class AlpacaPrompter:
     prompt_style = None
 
     def __init__(self, prompt_style="instruct"):
```
NanoCode012: Is it possible to 
```suggestion
    def __init__(self, prompt_style=PromptStyle.instruct.value):
```

Then, we would not need it in the next line?

src/axolotl/utils/data.py
```diff
@@ -60,10 +60,12 @@ def load_tokenized_prepared_datasets(
         else Path(default_dataset_prepared_path) / ds_hash
     )
     dataset = None
+    use_auth_token = False
```
NanoCode012: Can we just keep it to `True`? I do not see the benefit to setting it to `False`.

examples/lora-alpaca-7b/config.yml
```diff
@@ -0,0 +1,67 @@
...
+strict: false
+push_dataset_to_hub:
+datasets:
+  - path: teknium/GPT4-LLM-Cleaned
+    type: alpaca
+dataset_prepared_path: last_run_prepared
+val_set_size: 0.02
+adapter: lora
+lora_model_dir:
+sequence_len: 512
```
NanoCode012: Maybe for starter. This can be reduced for a quick run.
```suggestion
sequence_len: 256
```

examples/lora-alpaca-7b/config.yml
```diff
@@ -0,0 +1,67 @@
...
+num_epochs: 4
+optimizer: adamw_bnb_8bit
+torchdistx_path:
+lr_scheduler: cosine
+learning_rate: 0.0002
+train_on_inputs: false
+group_by_length: false
+bf16: false
+fp16: true
+tf32: true
```
NanoCode012: This will only work for ampere. Should we set this to False?

NanoCode012 (commented):
Some more ideas

src/axolotl/prompters.py
winglian: it get's called with None in data.py
```
            d_prompt_style = d_type_split[1] if len(d_type_split) > 1 else None
...
ds_strategy = AlpacaPromptTokenizingStrategy(
                    AlpacaPrompter(d_prompt_style),
                    tokenizer,
                    cfg.train_on_inputs,
                    cfg.sequence_len,
                )
```

examples/lora-alpaca-7b/config.yml
winglian: which one, tf32?

NanoCode012: > which one, tf32?

Yep. it requires ampere.

src/axolotl/utils/data.py
winglian: if it's true, the user needs to log into huggingface-cli before running

NanoCode012: This would mean. If it's a private repo, the auth token will not be included for Line 68? A user might not be able to access private dataset?

winglian: good call, i'll revert this part

src/axolotl/prompters.py
NanoCode012: Gotcha. I was thinking: would it be better to use a different config `d.prompt_style` instead of packing it with `d.type`?

PR #67, opened 2023-05-26
load the tokenizer seperately from the model

### LINKED ISSUE #59: load tokenizer separately and before the models
this way the datasets can be tokenized without the models for data preparation

winglian:
loading the tokenizer separately allows us to only load the tokenizer and prepare the dataset without the slow step of loading ht model. We can also skip loading the dataset for inference or shard cases
fixes #59

### DIFF
--- a/scripts/finetune.py
+++ b/scripts/finetune.py
@@ -5,7 +5,7 @@ import random
 import signal
 import sys
 from pathlib import Path
-from typing import Optional
+from typing import Optional, List, Dict, Any, Union
 
 import fire
 import torch
@@ -117,6 +117,10 @@ def choose_config(path: Path):
     return chosen_file
 
 
+def check_not_in(list1: List[str], list2: Union[Dict[str, Any], List[str]]) -> bool:
+    return not any(el in list2 for el in list1)
+
+
 def train(
     config: Path = Path("configs/"),
     prepare_ds_only: bool = False,
@@ -169,7 +173,7 @@ def train(
         cfg
     )
 
-    if "inference" not in kwargs and "shard" not in kwargs:  # don't need to load dataset for these
+    if check_not_in(["inference", "shard", "merge_lora"], kwargs):  # don't need to load dataset for these
         train_dataset, eval_dataset = load_prepare_datasets(
             tokenizer, cfg, DEFAULT_DATASET_PREPARED_PATH
         )
--- a/scripts/finetune.py
+++ b/scripts/finetune.py
@@ -21,7 +21,7 @@ src_dir = os.path.join(project_root, "src")
 sys.path.insert(0, src_dir)
 
 from axolotl.utils.data import load_prepare_datasets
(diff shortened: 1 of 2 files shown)


scripts/finetune.py
```diff
@@ -161,13 +161,30 @@ def train(
...
 
+    # load the tokenizer first
+    logging.info("loading tokenizer...")
+    tokenizer = load_tokenizer(
+        cfg.base_model_config,
+        cfg.tokenizer_type,
+        cfg
+    )
+
+    if "inference" not in kwargs and "shard" not in kwargs:  # don't need to load dataset for these
```
NanoCode012: You need to also add `merge_lora` here.

Perhaps can be refactored to use the below as the list expand.. Any element in first list is not in second.

```python
def check_lists(list1, list2):
    return not any(element in list2 for element in list1)
```

src/axolotl/utils/models.py
```diff
@@ -34,20 +33,56 @@
...
 
+def load_tokenizer(
+    base_model_config,
+    tokenizer_type,
+    cfg,
+):
+    if tokenizer_type:
+        tokenizer = getattr(transformers, tokenizer_type).from_pretrained(
+            base_model_config,
+            trust_remote_code=True if cfg.trust_remote_code is True else False,
```
```suggestion
            trust_remote_code=cfg.trust_remote_code or False,
```

Syntax Ref: https://stackoverflow.com/questions/4978738/is-there-a-python-equivalent-of-the-c-sharp-null-coalescing-operator

src/axolotl/utils/models.py
```diff
@@ -34,20 +33,56 @@
...
+        )
+    else:
+        tokenizer = AutoTokenizer.from_pretrained(
+            base_model_config,
+            trust_remote_code=True if cfg.trust_remote_code is True else False,
```
```suggestion
            trust_remote_code=cfg.trust_remote_code or False,
```

src/axolotl/utils/models.py
```diff
@@ -122,7 +157,7 @@ def load_model(
                     model_path = str(cache_model_path)
             except:
                 model_path = cfg.base_model
-            model, tokenizer = load_llama_model_4bit_low_ram(
+            model, _ = load_llama_model_4bit_low_ram(
```
NanoCode012, line 160: Is the tokenizer for 4bit ram different? It should be the same right and safe to discard right?

NanoCode012 (commented):
Looks nice!

src/axolotl/utils/models.py
winglian: correct, same tokenizer

scripts/finetune.py
winglian: so like `check_lists(["inference", "shard", "merge_lora"], kwargs.keys()]`?

NanoCode012: Yes or just `kwargs`! And also with a better function name :')

PR #69, opened 2023-05-26
Fix: Disable xformers for QuickStart config

NanoCode012:


NanoCode012:
This crashed if `true` running `examples/lora-openllama-3b/config.yml` on 3090.

PR #72, opened 2023-05-27
refactor: prompter

Thytu:
* Set system prompts as private attributes
* Rename `*AlpacaPrompter.match_prompt_style` as `*AlpacaPrompter.__init_prompt_template`
* `*AlpacaPrompter.__init_prompt_template` will now raise a `ValueError` if an invalide prompte style is provided

I will submit small PRs (instead of a single big one) in order to make it easier for you to review.


NanoCode012:
Hey, thanks for PR. I would like to point out some parts.

- `src/axolotl/prompters.py` : since you have modified the local variable. Classes that inherit the base require change as well
- `src/axolotl/prompt_strategies/creative_acr.py` <- You would also need to rename `system_prompt`
- ~I am not sure whether `raise ValueError(f"Received an unknown prompt style ({self.prompt_style})")` would work for all scenarios. For example, the `CompletionPrompter` does not require a prompt style. One alternative would be to not inherit `AlpacaPrompter` anymore.~ Fixed in #88

winglian (approved):
looks good to me, let me see if I can validate it

winglian:
Also, would you mind rebasing this against main? I added a unit test for the original behavior for this feature. https://github.com/OpenAccess-AI-Collective/axolotl/pull/87

thanks!

Thytu:
Didn't see that you're using inheritance from `AlpacaPrompter`, it thus need to be changed from private (`__foo`) to protected (`_foo`).

Thytu:
Cleaning my PR history

Issue #73, opened 2023-05-27
Unusable early_stopping_patience param


Thytu:
Whenever user uses `early_stopping_patience` it results in the following error: `AssertionError: EarlyStoppingCallback requires load_best_model_at_end = True`

NanoCode012:
Related to https://github.com/OpenAccess-AI-Collective/axolotl/issues/28 and this comment https://github.com/OpenAccess-AI-Collective/axolotl/issues/24#issuecomment-1541337225

winglian:
commented on your pr. lmk if that sounds reasonable. https://github.com/OpenAccess-AI-Collective/axolotl/pull/74#issuecomment-1565424811

NanoCode012:
Please use #28  for this problem.

Issue #28, opened 2023-05-11
early stopping callback requires load_best_model_at_end to be True


winglian:
there are a lot of factors that affect how `load_best_model_at_end` gets set from eval steps and save_steps. we need to figure out what the best way to handle this is.

winglian:
I think a good solution is allow explicit disabling of `load_best_model_at_end` via configuration to override all those, and if it's not specifically disabled, then add some warning in validation https://github.com/OpenAccess-AI-Collective/axolotl/blob/main/src/axolotl/utils/validation.py
`if cfg.load_best_model_at_end is not False and cfg. early_stopping_patience:`

flotos:
I also got this error, my config worked well with eval_steps: 5, but when changed to eval_steps:30 I receive this error.

winglian:
this can be set explicitly now.

PR #74, opened 2023-05-27
refactor(params): momentarily disable early_stopping_patience until issue #73 is fixed

Thytu:


winglian:
`load_best_model_at_end` is derived from several factors, see
https://github.com/OpenAccess-AI-Collective/axolotl/blob/main/src/axolotl/utils/trainer.py#L115-L120

I think a good solution is allow explicit disabling of `load_best_model_at_end` via configuration to override all those, and if it's not specifically disabled, then add some warning in validation https://github.com/OpenAccess-AI-Collective/axolotl/blob/main/src/axolotl/utils/validation.py
`if cfg.load_best_model_at_end is not False and cfg. early_stopping_patience:`

NanoCode012:
Closed for now. Please refer to discussion at #28 where we will try to think of better method.

PR #92, opened 2023-05-27
add support for opimum bettertransformers

winglian:
https://pytorch.org/blog/out-of-the-box-acceleration/

testing initial support for gpt-neox arch


NanoCode012:
I checked that Pytorch link. This is not a torch 2.0 feature only right?

Edit: I also see a `float16`. This is different from `fp16`?

winglian:
so I had to add float16 as an option b/c there is no way to have the model load as float16 currently without enabling automatic mixed precision which kicks in when you pass fp16 or bf16 to the trainer.

winglian:
might be easiest to warn or raise a ValueError if they are using `torch.__version__.split(".")[0] < 2` ?

NanoCode012:
> might be easiest to warn or raise a ValueError if they are using `torch.__version__.split(".")[0] < 2` ?

I think this is good. 

> so I had to add float16 as an option b/c there is no way to have the model load as float16 currently without enabling automatic mixed precision which kicks in when you pass fp16 or bf16 to the trainer.

I just worry this will be confusing.. Could we just check for `cfg.flash_optimum` then handle it better instead of creating a new config?

winglian:
what about `cfg.flash_optimum_float32`? this way we default to float16 when `cfg.flash_optimum` is set (the sane default), but if they really wanted float32, they can set it here.

NanoCode012:
> what about `cfg.flash_optimum_float32`? this way we default to float16 when `cfg.flash_optimum` is set (the sane default), but if they really wanted float32, they can set it here.

Hm, by default, I think people would expect their code to be float32. Would `cfg.flash_optimum_float16` be better?

Edit: Could you also add it to Readme in case I forget to later?

winglian:
> Hm, by default, I think people would expect their code to be float32. Would `cfg.flash_optimum_float16` be better?

so in doing some additional experiments, there are cases where I need to explicitly load in `float16` where I'm not using `bettertransformers`

examples/pythia-12b/config.yml
```diff
@@ -1,39 +1,49 @@
...
+lora_r: 64
 lora_alpha: 32
-lora_dropout: 0.05
+lora_dropout: 0.0
 lora_target_modules:
-  - query_key_value
+lora_target_linear: true
 lora_fan_in_fan_out: true  # pythia/GPTNeoX lora specific
-wandb_project: gpt4all-neox-20b
+wandb_project: pythia-12b
```
NanoCode012: Maybe this could be removed so it doesn't create a new project for someone unintentionally?

examples/pythia-12b/config.yml
```diff
@@ -1,39 +1,49 @@
...
+fp16: false
+float16: true
+tf32: true
+flash_optimum: true
 early_stopping_patience:
 resume_from_checkpoint:
 local_rank:
+gradient_checkpointing: true
+fsdp:
+fsdp_transformer_layer_cls_to_wrap:
```
NanoCode012: Should this have been indented inside `fsdp_config` or removed entirely?

based off https://huggingface.co/openaccess-ai-collective/manticore-13b/blob/aed786b0200251c9962ac200c50f7e367f264b46/configs/manticore.yml#L99-L100

scripts/finetune.py
```diff
@@ -242,6 +255,21 @@ def train(
...
+                [random.randrange(0, len(train_dataset) - 1) for i in range(5)]  # nosec
+            ),
+            tokenizer,
+        )
+
+    if prepare_ds_only:
+        logging.info("Finished preparing dataset. Exiting...")
+        return
+
+    model.train()
```
NanoCode012:
```suggestion
```

Not needed as stated in previous Issue, Huggingface Trainer sets the mode to train on each training step.

scripts/finetune.py
```diff
@@ -242,6 +255,21 @@ def train(
...
+        check_dataset_labels(
+            train_dataset.select(
+                [random.randrange(0, len(train_dataset) - 1) for i in range(5)]  # nosec
+            ),
+            tokenizer,
+        )
+
+    if prepare_ds_only:
+        logging.info("Finished preparing dataset. Exiting...")
+        return
```
NanoCode012: This seems duplicate code. There is a prepare above.

scripts/finetune.py
```diff
@@ -242,6 +255,21 @@ def train(
...
         return
 
+    if cfg.debug:
+        logging.info("check_dataset_labels...")
+        check_dataset_labels(
+            train_dataset.select(
+                [random.randrange(0, len(train_dataset) - 1) for i in range(5)]  # nosec
+            ),
+            tokenizer,
+        )
```
NanoCode012: This seems duplicate code. There is a debug above.

scripts/finetune.py
```diff
@@ -285,13 +316,21 @@ def train(
...
     if not Path(cfg.output_dir).is_dir():
         os.makedirs(cfg.output_dir, exist_ok=True)
-    trainer.train(resume_from_checkpoint=resume_from_checkpoint)
+    if cfg.flash_optimum:
+        with torch.backends.cuda.sdp_kernel(
+            enable_flash=True, enable_math=True, enable_mem_efficient=True
+        ):
+            trainer.train(resume_from_checkpoint=resume_from_checkpoint)
+    else:
+        trainer.train(resume_from_checkpoint=resume_from_checkpoint)
```
NanoCode012, line 337: I wonder if there's a cleaner way of writing this,  although not necessary.

src/axolotl/utils/callbacks.py
```diff
@@ -30,3 +31,39 @@ def on_save(
...
 
         return control
+
+
+class SaveBetterTransformerModelCallback(
+    TrainerCallback
+):  # pylint: disable=too-few-public-methods
+    """Callback to save the BetterTransformer wrapped model"""
+
+    def on_step_end(
```
NanoCode012, line 41: Maybe you should change to `on_save` like above. Then clean the save checks below.
```suggestion
    def on_save(
```

Edit: After reading your comments below. I'm not sure if this suggestion is correct.

src/axolotl/utils/data.py
```diff
@@ -387,8 +388,43 @@ def load_prepare_datasets(
...
+        self.dataset_path = dataset_path
+        self.max_tokens = max_tokens
+
+    def __iter__(self):
+        buffer = []
+        for sample in load_dataset(
+            self.dataset_path,
+        )["train"].shuffle():
+            buffer += self.tokenizer(sample["text"])["input_ids"]
+            buffer += [self.tokenizer.eos_token_id]
```
NanoCode012: Would it be possible to combine these two line above to prevent `buffer` from having to extend twice?

src/axolotl/utils/models.py
```diff
@@ -115,9 +116,9 @@ def load_model(
         logging.info("patching with sdp attention")
         hijack_llama_sdp_attention()
 
-    if cfg.bf16:
+    if cfg.bf16 or cfg.bfloat16:
```
NanoCode012, line 125: Why is there bloat16?

src/axolotl/utils/models.py
```diff
@@ -115,9 +116,9 @@ def load_model(
...
         torch_dtype = torch.bfloat16
-    elif cfg.load_in_8bit or cfg.fp16:
+    elif cfg.load_in_8bit or cfg.fp16 or cfg.float16:
```
NanoCode012, line 127: What is float16? Not sure if I asked this before and should've been some optimum_float16?

src/axolotl/utils/trainer.py
```diff
@@ -227,6 +231,10 @@ def setup_trainer(cfg, train_dataset, eval_dataset, model, tokenizer):
     ]:  # only save in rank 0
         callbacks.append(SavePeftModelCallback)
 
+    if hasattr(model, "use_bettertransformer") and model.use_bettertransformer is True:
+        logging.info("Setting up SaveBetterTransformerModelCallback.")
```
NanoCode012:
```suggestion
```

Maybe to reduce unnecessary logs?

src/axolotl/utils/validation.py
```diff
@@ -57,7 +59,32 @@ def validate_config(cfg):
     if (cfg.base_model and "falcon" in cfg.base_model.lower()) and cfg.fsdp:
         raise ValueError("FSDP is not supported for falcon models")
 
+    if cfg.flash_optimum is True:
+        if cfg.adapter:
+            logging.warning(
+                "BetterTransformers probably doesn't work with PEFT adapters"
+            )
+        if cfg.fp16 or cfg.bf16:
+            raise ValueError("AMP is not supported with BetterTransformer")
```
NanoCode012, line 73: This may be added to tests?

NanoCode012 (commented):
I think some parts have been discussed before, but not sure if we decided on something.

examples/pythia-12b/README.md
```diff
@@ -0,0 +1,9 @@
+# Python 12B
```
utensil:
```suggestion
# Pythia 12B
```

src/axolotl/utils/data.py
```diff
@@ -399,32 +399,116 @@ def load_prepare_datasets(
...
+
+    logging.debug(len(ret["input_ids"]))
+    return ret
 
 
 def load_pretraining_dataset(path, tokenizer, max_tokens=2048):
-    return PretrainingDatasetWrapper(tokenizer, path, max_tokens=max_tokens)
+    encode = functools.partial(encode_pretraining, tokenizer, max_tokens)
+    dataset = load_dataset(path, streaming=True, split="train")
+    dataset = dataset.shuffle(seed=42, buffer_size=10_000)
```
NanoCode012: Use cfg.seed if possible

src/axolotl/utils/models.py
winglian: when setting bf16, that's a HF AMP trainer specific thing, whereas we can alternatively use bloat16 without enabling HF AMP

src/axolotl/utils/models.py
winglian: related to fp16 and bf16 be HF and AMP related args,

PR #98, opened 2023-05-28
Add pre-commit: black+flake8+pylint+mypy+isort+bandit

### LINKED ISSUE #43: add pre-commit hook with pylint, flake8 and black
mypy would be nice too, but that might be a big ask

NanoCode012:
Closes #43 

- [x] black
- [x] pylint
- [x] flake8
- [x] mypy
- [x] isort
- [x] bandit 
- [x] Run on previous files
- [x] Test


NanoCode012:
- [x] Successfully ran a simple fine-tune using QuickStart openllama lora.
- [ ] Verify all configs are accurate.

Make sure this PR is squashed or things will be messy

src/axolotl/prompters.py
```diff
@@ -187,18 +231,18 @@ class Conversation:
...
-        for i, (role, message) in enumerate(self.messages):
+        for _, (role, message) in enumerate(self.messages):
             if message:
-                yield (role + ":", " " + message)
+                yield role + ":" + " " + message
             else:
-                logging.warning("role with empty message: " + role)
-                yield (role + ":",)
+                logging.warning(f"role with empty message: {role}")
+                yield role + ":"
```
NanoCode012: To satisfy `mypy`, I had to change from two `yield (str, str)` to just `yield str`. I am not sure if this broke anything. The only place that seem to call this seems to expect one value only.

NanoCode012:
~I could not apply `isort` to pre-commit. `black` and `isort` kept circular fixing each other despite setting profile in `isort`.~

Fixed in https://github.com/OpenAccess-AI-Collective/axolotl/pull/98/commits/c97a31cd2d164b36895b7e5577989319deb5d2b8

src/axolotl/prompt_strategies/alpaca_instruct.py
```diff
@@ -1,10 +1,12 @@
+"""Module loading the AlpacaInstructPromptTokenizingStrategy class"""
+
 from axolotl.prompt_tokenizers import AlpacaPromptTokenizingStrategy
 from axolotl.prompters import AlpacaPrompter, PromptStyle
 
 
 def load(tokenizer, cfg):
     return AlpacaPromptTokenizingStrategy(
-        AlpacaPrompter(PromptStyle.instruct),
+        AlpacaPrompter(PromptStyle.INSTRUCT),
```
winglian: may as well fix this bug
```suggestion
        AlpacaPrompter(PromptStyle.INSTRUCT.value),
```

src/axolotl/prompters.py
winglian: so I think `get_prompt` needs to return `Generator[Tuple[str, Optional[str]], None, None]` 
(or maybe `Generator[Union[str, Tuple[str, Optional[str]]], None, None]` ? )
then `ShareGPTPrompter.build_prompt` also needs to return `Generator[Tuple[str, Optional[str]], None, None]`, because if you look at `ShareGPTPromptTokenizingStrategy.tokenize_prompt` , it calls:
```
            for i, part in enumerate(
                self.prompter.build_prompt(self.get_conversation_thread(prompt))
            ):
```
it expects either a Tuple or str 
```
if isinstance(part, tuple):
    ...
else:
    ...
```

NanoCode012: If so, the `yield ensemble` just a few lines before also needs to be swapped since that returns one string.

NanoCode012:
I will rebase before squash merge when you have finished review or when you think it's proper. I don't want to break the branch since you may be testing on it now.

winglian:
I ran this dataset (Set shards high so it runs a quicker)
```
  - path: ehartford/wizard_vicuna_70k_unfiltered
    type: sharegpt:chat
    shards: 20
```
it's still correct on main
Screenshot 2023-05-30 at 9 37 48 AM
but the entire conversation is attention masked out on this branch
Screenshot 2023-05-30 at 9 37 08 AM

src/axolotl/prompt_tokenizers.py
```diff
@@ -349,3 +416,40 @@ def _tokenize(self, prompt, add_eos_token=True, strip_bos_token=False):
...
+    input_ids = res["input_ids"]
+    input_len = len(input_ids)
+    result["input_ids"][current_len : current_len + input_len] = input_ids
+    result["attention_mask"][current_len : current_len + input_len] = [
+        1 if x != pad_token_id else 0 for x in input_ids
+    ]
+    result["labels"][current_len : current_len + input_len] = labels
+    current_len += input_len
+
+    return result, current_len
```
NanoCode012, line 455: I refactored this function out due to mypy. However, would this be a performance loss since we call this function, then return the result and overwrite previous result object?

NanoCode012:
Tested that it can finetune openllama config to step 25. Please make sure to squash into 1 commit.

PR #99, opened 2023-05-28
new hf_use_auth_token setting so login to hf isn't required

### LINKED ISSUE #84: [Bug] Add `cfg.hf_use_auth_token` to set whether to attach auth token
As discussed in Discord, if a user is not authenticated to huggingface, the code would error as it expects the token.

We would like to swap to look for whether to attach using a config instead `cfg.hf_use_auth_token`.

- [ ] Set all `use_auth_token=True` to load from cfg `use_auth_token=cfg.hf_use_auth_token` instead
https://github.com/OpenAccess-AI-Collective/axolotl/blob/87dffbc451fcd129f143c64a3ff4ea9336aaa3a5/src/axolotl/utils/data.py#L67

- [ ] Add it to the docs under all yaml options

winglian:
fixes #84

### DIFF
--- a/tests/test_validation.py
+++ b/tests/test_validation.py
@@ -93,3 +93,29 @@ class ValidationTest(unittest.TestCase):
 
         with pytest.raises(ValueError, match=r".*4bit.*"):
             validate_config(cfg)
+
+    def test_hf_use_auth_token(self):
+        base_cfg = DictDefault(
+            {
+                "push_dataset_to_hub": None,
+                "hf_use_auth_token": None,
+            }
+        )
+
+        cfg = base_cfg | DictDefault(
+            {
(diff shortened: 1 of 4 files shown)


tests/test_validation.py
```diff
@@ -93,3 +93,29 @@ def test_qlora_merge(self):
 
         with pytest.raises(ValueError, match=r".*4bit.*"):
             validate_config(cfg)
+
+    def test_hf_use_auth_token(self):
+        base_cfg = DictDefault(
```
NanoCode012: I think you don't need base_cfg here and the merge below. It was useful in others because I wanted to inherit similar base configs.

NanoCode012 (commented):
Looks good!

PR #102, opened 2023-05-29
add 9.0 for h100 support and move debug higher up so it prints when preparing the datasets

winglian:


winglian:
looks like this is going to be a slog to figure out the necessary libs to get it working with H100. https://github.com/OpenAccess-AI-Collective/axolotl/actions/runs/5108207548
closing for now to revisit another day

PR #108, opened 2023-05-30
default to qlora support, make gptq specific image

winglian:
peft with qlora and gptq are mutually exclusive. we should default to bleeding edge with qlora support, and add a separate image for gptq support for those who want it


README.md
```diff
@@ -30,7 +30,7 @@
 ```bash
 git clone https://github.com/OpenAccess-AI-Collective/axolotl
 
-pip3 install -e .[int4]
+pip3 install -e .[gptq]
```
NanoCode012: I think default in QuickStart should not be gptq if you recommend the `-e .` below.

README.md
```diff
@@ -57,9 +57,9 @@ accelerate launch scripts/finetune.py examples/lora-openllama-3b/config.yml \
   1. Install python **3.9**
 
   2. Install python dependencies with ONE of the following:
-      - `pip3 install -e .[int4]` (recommended)
-      - `pip3 install -e .[int4_triton]`
-      - `pip3 install -e .`
+      - `pip3 install -e .` (recommended, supports qlora, no gptq/int4 support)
+      - `pip3 install -e .[gptq_triton]`
+      - `pip3 install -e .[gptq]`
```
NanoCode012: Which is preferred `triton` or `non-triton` version? Should the one that's more preferred be on top?

.github/workflows/base.yml
```diff
@@ -18,10 +18,22 @@ jobs:
...
+          - cuda: cu118
+            cuda_version: 11.8.0
+            cuda_version_bnb: "118"
+            pytorch: 2.0.0
+            axolotl_extras: gptq
+          - cuda: cu117
+            cuda_version: 11.7.0
+            cuda_version_bnb: "117"
+            pytorch: 1.13.1
+            axolotl_extras: gptq
```
NanoCode012: Do we need to think about python>3.9 or let it be only 3.9 for now? It may be a waste of resources if we support too many kinds.

I also remember getting an image depreciation warning with cuda:117 and torch1.13.1. Not sure if the image that you base on will suddenly be deleted later

winglian: yeah, the image count could really explode fast. maybe only gptq on a single group of versions and python?

winglian: yeah, would like to add python 3.10 too at some point, maybe just for the newer PyTorch

README.md
winglian: probably non-triton, will adust

.github/workflows/base.yml
```diff
@@ -18,10 +18,22 @@ jobs:
             cuda_version: 11.8.0
             cuda_version_bnb: "118"
             pytorch: 2.0.0
+            axolotl_extras:
           - cuda: cu117
             cuda_version: 11.7.0
             cuda_version_bnb: "117"
             pytorch: 1.13.1
+            axolotl_extras:
+          - cuda: cu118
```
NanoCode012: I think cuda and cuda_version_bnb can be combined. In the image, you can just add “cu” in front of the cuda version.

PR #115, opened 2023-05-30
docker fixes: py310, fix cuda arg in deepspeed

winglian:
adds a new py310 build for latest PyTorch/cuda only, no gptq support
adds a fix for cuda args w deepspeed


winglian:
fyi, after merging this into main, push this to main-base, wait for base images to build, and then we need to update downstream images to pull the py3x tagged images

PR #119, opened 2023-05-31
Feat(inference): Swap to GenerationConfig

### LINKED ISSUE #66: [Question] Should inference instruction have stripped last new line?
https://github.com/OpenAccess-AI-Collective/axolotl/blob/bbfc333a0136bfcf3f2129986253a8141c6c64d5/scripts/finetune.py#L68-L71

The inference script requires typing enter to pass the input even it's only one line. The result is that a `\n` is appended to that line.

```python
>>> get_multi_line_input()
Give me an instruction (Ctrl + D to finish): 
test
'test\n'
```

Should this be changed to `instruction=instruction.strip('\n')`?

I am not sure about other prompting style, but for `completion`, we want the text to be continued `test is a word`. Instead of having a `\n` after the input `test\n is a word`.

An alternative solution would be to add strip inside of the `CompletionPrompter`.

NanoCode012:
Closes https://github.com/OpenAccess-AI-Collective/axolotl/pull/70

Fixes: https://github.com/OpenAccess-AI-Collective/axolotl/issues/66

### DIFF
--- a/scripts/finetune.py
+++ b/scripts/finetune.py
@@ -83,7 +83,7 @@ def do_inference(cfg, model, tokenizer, prompter="AlpacaPrompter"):
         with torch.no_grad():
             generation_config = GenerationConfig(
                 repetition_penalty=1.1,
-                max_new_tokens=100,
+                max_new_tokens=1024,
                 temperature=0.9,
                 top_p=0.95,
                 top_k=40,
--- a/scripts/finetune.py
+++ b/scripts/finetune.py
@@ -12,6 +12,7 @@ from typing import Any, Dict, List, Optional, Union
 import fire
 import torch
 import yaml
+from transformers import GenerationConfig
 
 from axolotl.utils.data import load_prepare_datasets
 from axolotl.utils.dict import DictDefault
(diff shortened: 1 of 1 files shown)


scripts/finetune.py
```diff
@@ -73,26 +74,33 @@ def do_inference(cfg, model, tokenizer, prompter="AlpacaPrompter"):
...
         model.eval()
         with torch.no_grad():
-            # gc = GenerationConfig()  # TODO swap out and use this
-            generated = model.generate(
-                inputs=batch["input_ids"].to(cfg.device),
-                do_sample=True,
-                use_cache=True,
+            generation_config = GenerationConfig(
                 repetition_penalty=1.1,
                 max_new_tokens=100,
```
winglian:
```suggestion
                max_new_tokens=1024,
```

winglian: let's go ahead and bump this up. I find myself manually editing this when I need to test things. I don't know if it's worth a config setting for.

winglian (approved):

NanoCode012:
@winglian , as a future PR, I just realized that we can "load" this config from huggingface. This will allow us to select certain configs and keep using it.

```py
generation_config.save_pretrained("my_account/my_model", push_to_hub=True)

model = AutoModelForCausalLM.from_pretrained("my_account/my_model")
model.generation_config
```

PR #120, opened 2023-05-31
split up llama model loading so config can be loaded from base config and models can be loaded from a path

winglian:
just ran into this issue, this worked for me to get it working

### DIFF
--- a/src/axolotl/utils/models.py
+++ b/src/axolotl/utils/models.py
@@ -10,7 +10,7 @@ from typing import TYPE_CHECKING, Optional, Tuple  # noqa: F401
 import bitsandbytes as bnb
 import torch
 import transformers
-from transformers import AutoModelForCausalLM  # noqa: F401
+from transformers import AutoModelForCausalLM, LlamaConfig  # noqa: F401
 from transformers import PreTrainedModel  # noqa: F401
 from transformers import AutoConfig, AutoTokenizer, BitsAndBytesConfig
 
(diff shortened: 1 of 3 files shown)


NanoCode012:
I'm not sure about others, but sometimes I use this base_model_config to point for tokenizer in a separate repo (maybe I'm using the bug as a feature).

Should we have a `cfg.tokenizer_config` instead?

src/axolotl/utils/models.py
```diff
@@ -10,7 +10,7 @@
 import bitsandbytes as bnb
 import torch
 import transformers
-from transformers import AutoModelForCausalLM  # noqa: F401
+from transformers import AutoModelForCausalLM, LlamaConfig  # noqa: F401
```
NanoCode012: Moved to below and remove F401.

```suggestion
```

src/axolotl/utils/models.py
```diff
@@ -10,7 +10,7 @@
...
 from transformers import PreTrainedModel  # noqa: F401
 from transformers import AutoConfig, AutoTokenizer, BitsAndBytesConfig
```
NanoCode012: Fix lint
```suggestion
from transformers import (
    AutoConfig,
    AutoModelForCausalLM,
    AutoTokenizer,
    BitsAndBytesConfig,
    LlamaConfig,
)
```

NanoCode012:
There is a redundant import nearby as well.

```py
if TYPE_CHECKING:
    from peft import PeftConfig  # noqa: F401
    from transformers import PreTrainedTokenizer  # noqa: F401
```

to

```py
if TYPE_CHECKING:
    from peft import PeftConfig  # noqa: F401
```

The PreTrainedTokenizer is never used.

---

And if it does not inconvenience you, could you also please add the cfg to readme?

PR #134, opened 2023-05-31
fix batch size calculation

winglian:
so gradient accumulation steps shouldn't change regardless of number of gpus. 

what should change is the calculated batch size (which we use elsewhere to calculate total steps)

also, added a fix to raise when it can't figure out how to prepare a dataset.


ehartford:
lgtm

PR #146, opened 2023-06-02
added docker-compose file

FarisHijazi:
simple docker file for development. It has a volume to use the code and also a volume for ~/.cache/huggingface so that it can use downloaded models

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,0 +1,20 @@
+# version: '3.8'
+services:
+  axolotl:
+    build:
+      context: .
+      dockerfile: ./docker/Dockerfile
+    volumes:
+      - .:/workspace/axolotl
+      - ~/.cache/huggingface/:/root/.cache/huggingface/
+    # set environment variables
+    environment:
+      - WANDB_API_KEY=${WANDB_API_KEY}
+    deploy:
+      resources:
+        reservations:
+          devices:
+            - driver: nvidia
+              # count: 1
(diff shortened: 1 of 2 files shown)


ehartford (commented):
I don't see the purpose of a docker-compose when we only have/need a single container

winglian (approved):
lgtm

docker-compose.yaml
```diff
@@ -0,0 +1,20 @@
...
+    environment:
+      - WANDB_API_KEY=${WANDB_API_KEY}
+    deploy:
+      resources:
+        reservations:
+          devices:
+            - driver: nvidia
+              # count: 1
+              capabilities: [gpu]
+    command: tail -f /dev/null
```
winglian, line 20: is this the correct command for local dev?

FarisHijazi: `command: tail -f /dev/null` keeps the container running otherwise it will shutdown immediately

FarisHijazi:
> I don't see the purpose of a docker-compose when we only have/need a single container

if you have a way to just use a docker build/run command I'm more than happy to use that

docker-compose.yaml
winglian: 👍 sgtm, I may change it to use sleep infinity down the line, https://stackoverflow.com/a/35770783

winglian:
Thanks @FarisHijazi !!!

PR #155, opened 2023-06-06
new prompters, misc fixes for output dir missing using fsdp, and changing max seq len

winglian:


scripts/finetune.py
```diff
@@ -279,6 +279,9 @@ def train(
             logging.info(
                 f"Using Auto-resume functionality to start with checkpoint at {resume_from_checkpoint}"
             )
+
+    if not Path(cfg.output_dir).is_dir():
+        os.makedirs(cfg.output_dir, exist_ok=True)
```
winglian, line 284: when using FSDP, it chokes if this doesn't already exist

src/axolotl/utils/models.py
```diff
@@ -234,6 +234,10 @@ def load_model(
                 base_model,
                 trust_remote_code=cfg.trust_remote_code or False,
             )
+            # Shouldn't be a problem most of the time. will obviously error if the model doesn't support this
+            # when training starts
+            if config.max_seq_len and cfg.sequence_len > config.max_seq_len:
```
winglian, line 239: this will allow for specific cases to extend the context length (like MPT)

NanoCode012: I recommend adding a warning so people know this is happening

winglian: thanks, I'll add it in the next round

PR #159, opened 2023-06-08
Fix training over existing lora

AngainorDev:
When training with Lora, and starting with an existing lora weights, current code produces a model with 0 trainable params and training can't work. Adding the "is_trainable" param allows the loaded peft to be trained and fixes the bug.

### DIFF
--- a/scripts/finetune.py
+++ b/scripts/finetune.py
@@ -158,7 +158,7 @@ def train(
     cfg_keys = cfg.keys()
     for k, _ in kwargs.items():
         # if not strict, allow writing to cfg even if it's not in the yml already
-        if k in cfg_keys or cfg.strict is False:
+        if k in cfg_keys or not cfg.strict:
             # handle booleans
             if isinstance(cfg[k], bool):
                 cfg[k] = bool(kwargs[k])
@@ -198,9 +198,9 @@ def train(
     logging.info(f"loading tokenizer... {tokenizer_config}")
     tokenizer = load_tokenizer(tokenizer_config, cfg.tokenizer_type, cfg)
 
-    if check_not_in(
-        ["shard", "merge_lora"], kwargs
-    ) and not cfg.inference:  # don't need to load dataset for these
+    if (
+        check_not_in(["shard", "merge_lora"], kwargs) and not cfg.inference
+    ):  # don't need to load dataset for these
         train_dataset, eval_dataset = load_prepare_datasets(
             tokenizer, cfg, DEFAULT_DATASET_PREPARED_PATH
         )
@@ -226,7 +226,7 @@ def train(
         cfg.model_type,
         tokenizer,
         cfg,
-        adapter=cfg.adapter
+        adapter=cfg.adapter,
     )
 
     if "merge_lora" in kwargs and cfg.adapter is not None:
--- a/scripts/finetune.py
+++ b/scripts/finetune.py
@@ -182,6 +182,9 @@ def train(
         if cfg.bf16:
             cfg.fp16 = True
         cfg.bf16 = False
+        
+    # Store inference mode into cfg when passed via args
+    cfg.inference = True if "inference" in kwargs else cfg.get("inference", False)
 
     # load the tokenizer first
     tokenizer_config = cfg.tokenizer_config or cfg.base_model_config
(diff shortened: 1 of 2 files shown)


NanoCode012:
Hi @AngainorDev , thanks for the PR. May I ask a few questions?

1) Does this impact inference or inference speed?
2) Is this also applicable for `llama-adapter`?
3) Since `lora` and `qlora` uses the same function for loading, does this config work as intended for `qlora`?
4) If it won't take too long of your time, could you provide output before/after the change?

AngainorDev:
1. No visible impact on inference. If we want to be on the safe side, The setting could be activated for training only.
2. I did not test, but given the code and default params, yes.
3. I did not test qlora, but since we get 0 trainable params if it's not set, it can't work without anyway.
4. The most visible change in the output is the result of print_trainable_params:
If you set a "lora_model_dir: " in the yml config, and try to train, you'll see 0 trainable params right away, and a flat eval loss curve.
When passing True, you get the same trainable param count as when you don't provide an existing lora_model_dir, and training goes as intended.

NanoCode012:
Thanks for getting back!

1. Regarding this, it would seem that it affects `inference_mode`. Perhaps another config will be needed? This will need some discussion. https://github.com/huggingface/peft/blob/e45529b149c7f91ec1d4d82a5a152ef56c56cb94/src/peft/peft_model.py#L149-L180
2. We can leave `llama_adapter` for later if not sure
3. Perhaps, this needs to be tested once in case it breaks for `qlora`

AngainorDev:
1. Yes, that's what I meant: No need for another config, just set this param to True if we are training, and False when we are in `--inference` mode.

I'll try to test qlora as well.

AngainorDev:
inference is passed as a param to load_model()
https://github.com/OpenAccess-AI-Collective/axolotl/blob/193c73bce040fe965f5ea66d235e8823bd19e5e7/src/axolotl/utils/models.py#L84

Instead of passing it all the way down to load_lora(), I'd rather suggest to store it in the cfg dict, from finetune.py
This way, it's already accessible by any function that has cfg, more flexible if we need it elsewhere later on.

So this would boil down to a 
`is_trainable = not cfg.inference` in models.py

Shall I do that ?

NanoCode012:
Yes! That can work!

AngainorDev:
Inference can now also be optionaly specified from the yml config.
`--inference` on the command line takes precedence.

Harmonised the use of inference to rely on cfg.inference where it was used.

scripts/finetune.py
```diff
@@ -182,15 +182,18 @@ def train(
         if cfg.bf16:
             cfg.fp16 = True
         cfg.bf16 = False
+        
+    # Store inference mode into cfg when passed via args
+    cfg.inference = True if "inference" in kwargs else cfg.get("inference", False)
```
NanoCode012: I don't think you need to do this as the following should've covered it https://github.com/OpenAccess-AI-Collective/axolotl/blob/febe902517b11e5280b4325530ddc26fd3553694/scripts/finetune.py#L149-L159

AngainorDev: Indeed. Tested and this works (cfg.strict==False)

The only difference without this line is cfg.inference == None instead of False when specified neither in the config nor in the args.
No consequence if tested with "not", but not formally correct.

This also hints that the 2 other bool switches tested via "in kwargs" could be handled as cfg. as well, to harmonize everything.

NanoCode012: Hey

> No consequence if tested with "not", but not formally correct.

Yes, that would be fine, I think. The code base right now uses a lot of these, although not formally correct as you've said.

> This also hints that the 2 other bool switches tested via "in kwargs" could be handled as cfg. as well, to harmonize everything.

Uhm, for this PR, I would try to keep things small as possible. Also, those kwargs are not used elsewhere, so it's ok.

src/axolotl/utils/models.py
```diff
@@ -82,8 +82,7 @@ def load_model(
     model_type,
     tokenizer,
     cfg,
-    adapter="lora",
-    inference=False,
```
winglian, line 86: removing this breaks line 99 below since `inference` is no longer defined
`if cfg.device not in ["mps", "cpu"] and inference is False:`

winglian (changes requested):

src/axolotl/utils/models.py
AngainorDev: Sorry, fixed.

winglian:
A couple of things, because we now rely on getting `inference` into `cfg` rather than simply checking whether it's in `kwargs`, we need to ensure that the config will even allow that, either we would need to update the example configs to all have `strict: false` or change the behavior so that strict behavior is not the default. I think the latter is preferred. 

For this, let's change:
```
if k in cfg_keys or cfg.strict is False: 
```
to 
```
if k in cfg_keys or not cfg.strict: 
```

you'll also want to re-run `pre-commit run --all-files` to fix any linting issues

winglian (approved):
lgtm! Thanks @AngainorDev

NanoCode012:
Thank you for the PR!

PR #163, opened 2023-06-08
Feat: Set matmul tf32=True when tf32 passed

### LINKED ISSUE #145: torch.backends.cuda.matmul.allow_tf32 = True
Find a good place in the code to add this if cfg.tf32 is set

NanoCode012:
Closes https://github.com/OpenAccess-AI-Collective/axolotl/issues/145

To be aware: 
- Pytorch keeps this `False` by default. https://pytorch.org/docs/stable/notes/cuda.html#tensorfloat-32-tf32-on-ampere-devices following this RFC https://github.com/pytorch/pytorch/issues/67384
- Huggingface recommends it as nvidia has shown small differences https://huggingface.co/docs/transformers/perf_train_gpu_one#tf32

### DIFF
--- a/scripts/finetune.py
+++ b/scripts/finetune.py
@@ -183,6 +183,9 @@ def train(
             cfg.fp16 = True
         cfg.bf16 = False
 
+    if cfg.tf32:
+        torch.backends.cuda.matmul.allow_tf32 = True
+
     # load the tokenizer first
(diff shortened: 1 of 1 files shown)


NanoCode012:
Confirmed that it trains with 

```bash
accelerate launch scripts/finetune.py examples/lora-openllama-3b/config.yml --tf32=True
```

PR #166, opened 2023-06-08
Fix: Set to use cfg.seed or 42 for seed

### LINKED ISSUE #47: [Bug] Seed does not always load from `cfg.seed`
I think this is an easy Issue to tackle for anyone interested.

It should be set to a default value if not defined, somewhere at start hopefully (maybe when loading config).

- [x] Update below

https://github.com/OpenAccess-AI-Collective/axolotl/blob/a617f1b65eb3d986ab7844630944fe4c979158fe/src/axolotl/utils/data.py#L115

- [x] Update below

https://github.com/OpenAccess-AI-Collective/axolotl/blob/a617f1b65eb3d986ab7844630944fe4c979158fe/src/axolotl/utils/data.py#L220

- [x] Pass seed to Trainer
- [ ] Pass seed to any function that has it available

NanoCode012:
Closes https://github.com/OpenAccess-AI-Collective/axolotl/issues/47

### DIFF
--- a/src/axolotl/utils/data.py
+++ b/src/axolotl/utils/data.py
@@ -78,6 +78,13 @@ def load_tokenized_prepared_datasets(
     else:
         logging.info(f"Unable to find prepared dataset in {prepared_ds_path}")
         logging.info("Loading raw datasets...")
+
+        if cfg.seed:
+            seed = cfg.seed
+        else:
+            logging.info("No seed provided, using default seed of 42")
+            seed = 42
+
         datasets = []
         # pylint: disable=invalid-name
         for d in cfg.datasets:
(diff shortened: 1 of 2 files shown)


NanoCode012:
Datasets shuffle: https://huggingface.co/docs/datasets/v1.2.1/package_reference/main_classes.html?highlight=shuffle#datasets.Dataset.shuffle

I used `42` instead of `None` if there is no `cfg.seed` for backward compatibility and better control.

Huggingface trainer arg: https://huggingface.co/docs/transformers/main_classes/trainer#transformers.TrainingArguments

Training works for a few global steps on open llama.

PR #167, opened 2023-06-08
Fix: Refactor out unmodified save_steps and eval_steps

### LINKED ISSUE #82: [Refactor] Remove use of local variables `save_steps` and `eval_steps` as they are not modified
https://github.com/OpenAccess-AI-Collective/axolotl/blob/87dffbc451fcd129f143c64a3ff4ea9336aaa3a5/src/axolotl/utils/trainer.py#L53-L54

When I looked into the code and saw this, I thought it was saved to a local variable to be modified / compared. However, it does not seem so.

It might be better to remove these and use the original cfg counterparts to remove the preconception that the local variable is different. Of course, it's also ok to leave them as is.

NanoCode012:
Close https://github.com/OpenAccess-AI-Collective/axolotl/issues/82


NanoCode012:
open llama lora runs

PR #168, opened 2023-06-09
Disable Wandb if no wandb project is specified

bratao:
This fix when no wandb api and project is provided


winglian (approved):

winglian:
thanks @bratao !

NanoCode012:
```
Using the `WANDB_DISABLED` environment variable is deprecated and will be removed in v5. Use the --report_to flag to control the integrations used for logging result (for instance --report_to none).
```

I'm getting this as a result.

PR #178, opened 2023-06-09
Update README.md to reflect current gradient checkpointing support

PocketDocLabs:
Previously the readme stated gradient checkpointing was incompatible with 4-bit lora in the current implementation however this is no longer the case. I have replaced the warning with a link to the hugging face documentation on gradient checkpointing.


NanoCode012:
Edit: Ignore my previous comment. I think it's good to go. Using main's link might break in future revision.

winglian (approved):

PR #179, opened 2023-06-10
fix for max sequence len across different model types

winglian:


NanoCode012:
I think this may not be clean solution. Would it be better to add it to validate_config to only use of these two config and error on the other?

The one kept should be placed in the docs.

PR #180, opened 2023-06-10
Add streaming inference & fix stopping at EOS

Glavin001:
## What's new?
- [x] Add streaming inference
- [x] Fix stopping at eos_token

When I was testing inference with the Falcon config file it  used eos_token=<|endoftext|> instead of what was previously being set in the do_inference() code was `<s/>` so it wouldn’t stop early it would always use the entire 1024 max new tokens.

## Demo

https://www.loom.com/share/03f8c8903b42484ea7aad72ab07a6340

### DIFF
--- a/scripts/finetune.py
+++ b/scripts/finetune.py
@@ -12,7 +12,7 @@ from typing import Any, Dict, List, Optional, Union
 import fire
 import torch
 import yaml
-from transformers import GenerationConfig
+from transformers import GenerationConfig, TextStreamer
 
 from axolotl.utils.data import load_prepare_datasets
 from axolotl.utils.dict import DictDefault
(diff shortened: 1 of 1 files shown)


scripts/finetune.py
```diff
@@ -64,13 +64,21 @@ def get_multi_line_input() -> Optional[str]:
...
+
+    for token, symbol in default_tokens.items():
+        # If the token isn't already specified in the config, add it
+        if not (cfg.special_tokens and token in cfg.special_tokens):
+            tokenizer.add_special_tokens({token: symbol})
 
     prompter_module = getattr(importlib.import_module("axolotl.prompters"), prompter)
 
     while True:
+        print("=" * 80)
```
Glavin001, line 77: Let me know if you like or dislike the separators. Can remove.

NanoCode012:
Thank you for PR. I am currently still thinking how to handle this. Since we might not only use alpaca format for inference, I'm thinking of how to make the `prompt` dynamic/passable and setting the appropriate default tokens.

Edit: as winglian has approved, I guess we can deal with the above at a later point. The streaming is a great addition!

winglian (approved):

scripts/finetune.py
```diff
@@ -64,13 +64,21 @@ def get_multi_line_input() -> Optional[str]:
...
 
 def do_inference(cfg, model, tokenizer, prompter="AlpacaPrompter"):
-    tokenizer.add_special_tokens({"unk_token": "<unk>"})
-    tokenizer.add_special_tokens({"bos_token": "<s>"})
-    tokenizer.add_special_tokens({"eos_token": "</s>"})
+    default_tokens = {
+        "unk_token": "<unk>",
+        "bos_token": "<s>",
+        "eos_token": "</s>"
+    }
```
winglian:
```suggestion
    default_tokens = {"unk_token": "<unk>", "bos_token": "<s>", "eos_token": "</s>"}
```

PR #181, opened 2023-06-10
add support to extend context with xpos rope

winglian:
I'm able to generate a PoC with openllama extended to 4k with this (will release soon)

### DIFF
--- a/src/axolotl/utils/models.py
+++ b/src/axolotl/utils/models.py
@@ -127,6 +127,14 @@ def load_model(
         # TODO: Check if this would overwrite previous additional_special_tokens
         tokenizer.add_special_tokens({"additional_special_tokens": [MEM_TOKEN]})
 
+    if cfg.is_llama_derived_model and cfg.xpos_rope:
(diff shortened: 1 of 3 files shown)


src/axolotl/utils/models.py
```diff
@@ -127,6 +127,14 @@ def load_model(
         # TODO: Check if this would overwrite previous additional_special_tokens
         tokenizer.add_special_tokens({"additional_special_tokens": [MEM_TOKEN]})
 
+    if cfg.is_llama_derived_model and cfg.xpos_rope:
```
NanoCode012, line 130: Perhaps it should be `elif` from above. We should maybe not allow multiple patches?

winglian: it works with xformers_attention,it's specific to the rotary embedding layer which most of the above doesn't touch

NanoCode012: Oh ok!

NanoCode012: Could you also add this to Readme please?

PR #183, opened 2023-06-10
pass a prompt in from stdin for inference

winglian:
it code as it mostly already handles a prompt from stdin, this cleans it up to skip the prompter if it's included in the prompt txt file


NanoCode012:
~I noticed that you have a `prompter` arg, but did not use it. Was there a reason?~

Edit: I misread the code using the UI. Looking in the file made sense.

PR #189, opened 2023-06-11
various fixes

winglian:
- match up gradient checkpointing when using lora w config
- mpt doesn't support gradient checkpointing
- get rid of some configs, formalize pythioa lora config


NanoCode012 (approved):
As of now, the checks work for falcon/mpt within the base_model's name. However, we might consider to improve this when people use their own model names based on falcon/mpt.

winglian:
> As of now, the checks work for falcon/mpt within the base_model's name. However, we might consider to improve this when people use their own model names based on falcon/mpt.

yeah, It's hard in this step of validation since we haven't loaded the model yet and can't inspect the model class

PR #197, opened 2023-06-12
chore: Fix inference README.

mhenrichsen:
Seems like the README wasn't updated when python file execution changed to Fire.

May relate to: https://github.com/OpenAccess-AI-Collective/axolotl/issues/160


NanoCode012:
Thank you!

Edit: That error is something else. It's due to transformers code.

PR #201, opened 2023-06-13
Fix landmark inference

NanoCode012:
- [x] Test with generate vs pipeline reach parity
- [ ] Test confirm in ooba


NanoCode012:
It's the wrong config. We should target add_special_token=False

PR #204, opened 2023-06-14
fix packing for tokenizers that don't use a bos_token when the bos token and eos token are both the same

winglian:
when packing with tokenizers such as gpt2, the tokenizer doesn't generate a bos token, so we need to reset the attention on the next token.


NanoCode012:
Closing as stale

Issue #205, opened 2023-06-14
need to add `if config.max_sequence_length` check too


winglian:
ERROR:root:'>' not supported between instances of 'int' and 'NoneType'
Traceback (most recent call last):
  File "/workspace/axolotl/src/axolotl/utils/models.py", line 259, in load_model
    and cfg.sequence_len > config.max_sequence_length
TypeError: '>' not supported between instances of 'int' and 'NoneType'
ERROR

PR #206, opened 2023-06-14
issue #205 bugfix

MaciejKarasek:
Added if statements to check if value of attribute is not None. Function `hasattr` returns True when atrribute exists even if its value is None.

Related to issue #205

### DIFF
--- a/src/axolotl/utils/models.py
+++ b/src/axolotl/utils/models.py
@@ -252,11 +252,16 @@ def load_model(
             )
             # Shouldn't be a problem most of the time. will obviously error if the model doesn't support this
             # when training starts
-            if hasattr(config, "max_seq_len") and config.max_seq_len and cfg.sequence_len > config.max_seq_len:
+            if (
+                hasattr(config, "max_seq_len")
+                and config.max_seq_len
+                and cfg.sequence_len > config.max_seq_len
+            ):
                 config.max_seq_len = cfg.sequence_len
                 logging.warning(f"increasing context length to {cfg.sequence_len}")
             elif (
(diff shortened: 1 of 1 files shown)


NanoCode012:
Hello, thanks for the PR!

I'm curious on the effect of `hasattr(config, "max_seq_len")`. Would it work without this since we're performing the check you have newly added?

MaciejKarasek:
> I'm curious on the effect of `hasattr(config, "max_seq_len")`. Would it work without this since we're performing the check you have newly added?

It would work if we are sure that attribute like max_seq_len is always a part of the config class. In case there is no attribute like that and if statement would look like this `if config.max_seq_len:` there's gonna be an error `AttributeError: type object 'config' has no attribute 'max_seq_len'`

NanoCode012:
Ah, I made the mistake. Thank you for clarifying. I confused `config` with `cfg`. I thought they were the same.

MaciejKarasek:
Seems I forgot to install pre-commit, but i think it not gonna work on mine python 3.11 cause in .pre-commit-config.yaml there is line `python: python3.9` so anyway :/

NanoCode012:
Hey @MaciejKarasek , no worries. Actually, my python (for local dev) is also 3.11, so it might work as well :)

However, if you do not want to install, that's fine also! Feel free to look at the CI Details, and it would show you the issue and maybe a potential fix.

winglian (approved):
thank you! lmk if you need any help with the pre-commit hooks

MaciejKarasek:
Thanks guys

MaciejKarasek:
Should be all right now.

PR #214, opened 2023-06-15
Fix tokenizing labels

winglian:
when tokening the full prompt and then changing the labels by tokenizing the system and instruction only, leads to potential off by one errors on the labels depending on whitespace. This simply tokenizes them separately like we do for chat

### DIFF
--- a/tests/test_prompt_tokenizers.py
+++ b/tests/test_prompt_tokenizers.py
@@ -61,6 +61,7 @@ class TestPromptTokenizationStrategies(unittest.TestCase):
         tests the interface between the user and assistant parts
         """
         prompter = NoSystemPrompter()
+        # pylint: disable=duplicate-code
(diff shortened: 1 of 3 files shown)


tests/test_prompt_tokenizers.py
```diff
@@ -53,6 +56,45 @@ def test_sharegpt_integration(self):
             self.assertEqual(len(example[fields]), len(tokenized_conversation[fields]))
             self.assertEqual(example[fields], tokenized_conversation[fields])
 
+    def test_completion(self):
```
NanoCode012: I think this function name could be renamed as the below does not use the `completion` prompter?

winglian: thanks, I had it in a completion class, then renamed and omitted things from the PR. will adjust

PR #215, opened 2023-06-15
support adamw and grad norm hyperparams

winglian:


tests/test_validation.py
```diff
@@ -263,3 +263,45 @@ def test_flash_optimum(self):
 
         with pytest.raises(ValueError, match=regex_exp):
             validate_config(cfg)
+
+    def test_adamw_hyperparams(self):
```
NanoCode012, line 267: Could you add one separate test or include into this test, a normal optimizer without these extra hyperparam would pass? (for future proofing)

PR #221, opened 2023-06-17
[WIP] Support loading data files from a local directory

utensil:
Per discord discussion https://discord.com/channels/1104757954588196865/1104757955204743201/1119515088324395129 and hf doc https://huggingface.co/docs/datasets/v2.13.0/en/package_reference/loading_methods#datasets.load_dataset.path .

Work-in-progress, [passed initial tests](https://colab.research.google.com/github/utensil/llm-playground/blob/main/notebooks/datasets/load_local_dataset_from_dir.ipynb), need more testing and to improve pr doc.

### DIFF
--- a/src/axolotl/utils/data.py
+++ b/src/axolotl/utils/data.py
@@ -102,13 +102,26 @@ def load_tokenized_prepared_datasets(
                 pass
 
             # prefer local dataset, even if hub exists
-            if Path(d.path).exists():
-                ds = load_dataset(
-                    "json",
-                    data_files=d.path,
-                    streaming=False,
-                    split=None,
-                )
+            local_path = Path(d.path)
+            if local_path.exists():
+                if local_path.is_dir():
+                    ds = load_dataset(
+                        d.path,
+                        data_files=d.data_files,
+                        streaming=False,
+                        split=None,
+                    )
+                elif local_path.is_file():
+                    ds = load_dataset(
+                        "json",
+                        data_files=d.path,
+                        streaming=False,
+                        split=None,
+                    )
+                else:
+                    raise ValueError(
+                        "unhandled dataset load: local path exists, but is neither a directory or a file"
+                    )
             elif ds_from_hub:
                 if d.data_files:
                     ds = load_dataset(


src/axolotl/utils/data.py
```diff
@@ -102,13 +102,22 @@ def load_tokenized_prepared_datasets(
...
+            local_path = Path(d.path)
+            if local_path.exists():
+                if local_path.is_dir():
+                    ds = load_dataset(
+                        d.path,
+                        data_files=d.data_files,
+                        streaming=False,
+                        split=None,
+                    )
+                else: # if local_path.is_file():
```
utensil:
```suggestion
                else:  # if local_path.is_file():
```

src/axolotl/utils/data.py
```diff
@@ -102,13 +102,22 @@ def load_tokenized_prepared_datasets(
...
+            local_path = Path(d.path)
+            if local_path.exists():
+                if local_path.is_dir():
+                    ds = load_dataset(
+                        d.path,
+                        data_files=d.data_files,
+                        streaming=False,
+                        split=None,
+                    )
+                else:  # if local_path.is_file():
```
NanoCode012: Perhaps change to elif?

samuelazran:
Is it possible to pass train and test files?

src/axolotl/utils/data.py
utensil: I was a bit hesitated, because I'm used to writing `if...elif...else` but there seems to be nothing to write for `else`

NanoCode012: Would it ever reach an else afterward? Do we need to raise Error?

utensil:
> Is it possible to pass train and test files?

It's feasible. And a reasonable feature request. But I don't know how it interacts with `val_set_size` yet. I think axolotl doesn't support pulling train and test splits for train and eval respectively at the moment. It's feasible to add though.

@winglian can you weigh in on this?

src/axolotl/utils/data.py
winglian: I'd be happy with having the else raise some sort of unhandled exception

utensil: OK

winglian:
> > Is it possible to pass train and test files?
> 
> 
> 
> It's feasible. And a reasonable feature request. But I don't know how it interacts with `val_set_size` yet. I think axolotl doesn't support pulling train and test splits for train and eval respectively at the moment. It's feasible to add though.
> 
> 
> 
> @winglian can you weigh in on this?

As far as train and test splits, we would need a bit of a refactoring, but it's on the list of things to do. Lots of datasets dont have splits, and oftentimes it's nice to be able to dynamically size the splits at runtime without needing to re process/retokenize everything.

winglian:
> > Is it possible to pass train and test files?
> 
> 
> 
> It's feasible. And a reasonable feature request. But I don't know how it interacts with `val_set_size` yet. I think axolotl doesn't support pulling train and test splits for train and eval respectively at the moment. It's feasible to add though.
> 
> 
> 
> @winglian can you weigh in on this?

As far as train and test splits, we would need a bit of a refactoring, but it's on the list of things to do. Lots of datasets dont have splits, and oftentimes it's nice to be able to dynamically size the splits at runtime without needing to re process/retokenize everything. 

I think what needs to happen is we keep track of datasets with train and test splits and tokenizer those appropriately, and then only use the split size on datasets that don't have them. I'd also like to figure out a way to make sure it is deterministic enough that when taking larger Val splits it always includes data that would have been in a smaller Val split.

This should happen at the Individual dataset level before shuffling them together or packing datasets. This ensures that adding new datasets doesn't move data in and out of splits which can be helpful when retraining with a different mix of datasets or partial replication of a previous model.

utensil:
I totally agree the tracking splits and use them when available, as well as ensuring increasing val split won't leave val data out.

I have another concern about the caching of datasets, it's cached in hf and locally, when the content of the file or the directory changes, the previous cache remains valid. This is due to the efficient hashing mechanism used now, which I totally agree, but users might need reminders to invalidate the cache when they change data, and tools to do so easily.

utensil:
Just to mark the status of this PR:

I'm not in a timely need of this feature, so I haven't been able to finish everything that should come with this PR, but it works. If you need such a solution and find this PR, you can pull from this branch for the time being.

I plan to have a 10-day time away from computers in the next week (at least for most of the time), so I might only be able to finish this PR in sometime mid July, but I assume some refactor would already happen by then, in that case, feel free to integrate this simple implementation or leave it out.

NanoCode012 (approved):
I think this PR is ok to go.

PR #224, opened 2023-06-18
System prompt data

### LINKED ISSUE #222: support alpaca datasets with the system prompt baked into the dataset

winglian:
adds support for pulling the system prompt from the data files. Useful for some CoT datasets.

Also adds more test coverage

### DIFF
--- a/src/axolotl/prompters.py
+++ b/src/axolotl/prompters.py
@@ -63,29 +63,6 @@ class AlpacaPrompter:
         yield res
 
 
-class SystemDataPrompter(AlpacaPrompter):
-    """
-    Alpaca Style Prompter that uses system prompts from the dataset
-    """
-
-    def build_prompt_w_system(
-        self,
-        system: str,
-        instruction: str,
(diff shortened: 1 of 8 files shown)


src/axolotl/prompters.py
```diff
@@ -128,9 +132,6 @@ def build_prompt(
     ) -> Generator[str, None, None]:
         yield instruction
 
-    def get_response(self, output: str) -> str:
-        return output.strip()
```
NanoCode012, line 132: Are these `get_response` removed because they are unused?

NanoCode012 (commented):
It looks alright. I couldn't test it myself, but seem like the change is small for people not using system prompts.

I'm just worried for increased complexity. Optional system prompts, optional inputs etc.

winglian:
> It looks alright. I couldn't test it myself, but seem like the change is small for people not using system prompts.
> 
> 
> 
> I'm just worried for increased complexity. Optional system prompts, optional inputs etc.

Yeah. That was dead code that was copied from alpaca repo.

PR #248, opened 2023-06-29
Feat: Add SuperHOT

NanoCode012:
Source: https://kaiokendev.github.io/til
Patch from https://huggingface.co/kaiokendev/superhot-13b-8k-no-rlhf-test

Thanks https://huggingface.co/emozilla/open_llama_7b-scaled for the instruction for setup

Note: According to the doc, they specifically mention `8192` as context. Therefore, I have left it as such. We can change this if there are different info to back them up.


ssmi153:
Just a note that Huggingface Transformers natively supports this now: https://github.com/huggingface/transformers/commit/34d94094279d2c903d9d8a51a65edb265f22c849. Does this make it an easier implementation on Axolotl?

NanoCode012:
> Just a note that Huggingface Transformers natively supports this now: [huggingface/transformers@34d9409](https://github.com/huggingface/transformers/commit/34d94094279d2c903d9d8a51a65edb265f22c849). Does this make it an easier implementation on Axolotl?

@ssmi153  Thanks for the info! It is definitely better to use `transformer` than ours as that'll mean more eyes on that code and less for axolotl to maintain. If you're interested in making a PR to integrate this, please let me know!

PR #251, opened 2023-06-29
Feat: Add NTKAware scaled

NanoCode012:
Source: https://twitter.com/yampeleg/status/1674430869828956161
Notebook: https://colab.research.google.com/drive/1VI2nhlyKvd5cw4-zHvAIk00cAVj2lCCC

TODO
- [x] Test

If anyone is available to test, feel free to share the results.

### DIFF
--- a/README.md
+++ b/README.md
@@ -446,7 +446,7 @@ landmark_attention:
 # llama only
 xpos_rope:
 # Non-linear interpolation scheme that adjusts the RoPE's base. See https://twitter.com/yampeleg/status/1674430869828956161
-ntlk_scaled:
+ntk_scaled:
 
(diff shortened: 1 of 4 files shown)


NanoCode012:
I tested that the patch works on a custom notebook and noticed slightly faster runtime for < 1k prompts.

However, the VRAM usage for a 7B model spiked to 35GB and did not accurately retrieve the data (not much testing for now).

Edit: Need more testing. Unlike other rope patches, this one does not change model's seq length nor tokenizer. I followed their notebook's methodology.

NanoCode012:
Following discussion with emozilla, I have also set the seq length.

PR #261, opened 2023-07-06
Added a support for Vicuna 1.1 format

Nan-Do:
I'm using this for my own purposes and I thought it would be helpful to other people. In this patch I have added support to train models using the Vicuna 1.1 dataset format https://lmsys.org/blog/2023-03-30-vicuna/ which to this day is the most efficient model at 7B.

### DIFF
--- a/src/axolotl/prompt_tokenizers.py
+++ b/src/axolotl/prompt_tokenizers.py
@@ -242,6 +242,23 @@ class CompletionPromptTokenizingStrategy(InstructionPromptTokenizingStrategy):
         return next(iter(self.prompter.build_prompt(instruction, input, response)))
 
 
+class VicunaPromptTokenizingStrategy(InstructionPromptTokenizingStrategy):
+    """
+    Tokenizing strategy for Vicuna-1.1 prompts.
+    """
+
+    def tokenize_prompt(self, prompt):
+        full_prompt = self._build_full_prompt(prompt["instruction"], None, prompt["output"])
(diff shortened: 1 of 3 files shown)


NanoCode012:
Hello! Thank you for the PR.

Currently, we're trying to migrate and stop fitting more code within the `prompter` and `prompt_tokenizer` files now as it is not scalable.

If you're interested in moving the code into the prompt_strategies folder (for ex: https://github.com/OpenAccess-AI-Collective/axolotl/blob/main/src/axolotl/prompt_strategies/sharegpt_simple.py) , it would be great for us! That link should be very similar to what you want. To load files from that dir, you can use the `type: <filename>.load_<load fn name>`

Nan-Do:
Sure I'll be happy to take a look at that as soon as I have time

jphme:
@Nan-Do did you successfully finetune Vicuna 1.3 with your changes?

I just opened an issue ( #262 ) due to this.

Looking at your implementation  2 Questions:
- it seems like your template can't do multi-turn conversations (unlike official vicuna) but just 1 question, 1 answer, correct?
- in connection with that, is there any specific reason you went with the InstructionPromptTokenizingStrategy which doesn't mask user input? 
Thanks a lot!

Nan-Do:
@jpdus no I haven't had the opportunity yet to try v1.3 but I'll try next.

About the implementation my training data doesn't include multi-turn conversation so I didn't include it, as for the tokenizing strategy that's a mistake on my part again I was interested on training on the inputs so I didn't realize that option wasn't being honored.

PR #266, opened 2023-07-08
don't use llama if trust_remote_code is set since that needs to use AutoModel path

winglian:
per emozilla, we shouldn't be using these models when trust_remote_code is set


NanoCode012:
I would think:
1) Since we warn if this is on, users can be careful and may be responsible for it.
2) If someone were to customize llama, that would make it incompatible with our code (which can be both pros and cons)
3) I think if you were to check, we should move this to validate_config function to prevent more validation code in codebase spread out

PR #269, opened 2023-07-08
compute perplexity from cross entropy

winglian:


casper-hansen:
@winglian How long until we can get this metric running in axolotl and log it in wandb while training?

casper-hansen:
If we want this to match other libraries, we probably want to do something like 

```
ppl = exp(sum(loss_all_batches) / num_of_batch)
```

Huggingface also has its own. I think you can set batch_size=512

```python
from evaluate import load
perplexity = load("perplexity", module_type="metric")

def compute_metrics(eval_pred):
    logits, labels = eval_pred
    predictions = np.argmax(logits, axis=-1)
    return perplexity.compute(predictions=predictions, model_id='llama')
```

NanoCode012:
Closing as stale. For future readers, we have a Perplexity eval callback via `eval_causal_lm_metrics: ["perplexity"]`

PR #277, opened 2023-07-14
Allow non-default dataset configurations

cg123:
Just a tiny change to make it possible to use HuggingFace datasets with multiple configurations. For example, to train on the 'enron_emails' configuration of The Pile, you could use this configuration:
```yml
datasets:
  - path: EleutherAI/pile
    name: enron_emails
    type: completion
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -350,6 +350,7 @@ datasets:
     type: alpaca # format | format:<prompt_style> (chat/instruct) | <prompt_strategies>.load_<load_fn>
     data_files: # path to source data files
     shards: # number of shards to split data into
+    name: # name of dataset configuration to load
 
 # axolotl attempts to save the dataset as an arrow after packing the data together so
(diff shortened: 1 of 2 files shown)


winglian:
🔥 This has been on my todo list, thank you! Does this breaks functionality if the dataset doesnt have a named configuration?

cg123:
Nope! This should still work exactly the same as before if the config field is not present.

The default value of the `name` argument is None, so `load_datasets` should get the exact same arguments for existing configs. I've tested a partial run of the openllama 3b config to be sure.

NanoCode012:
Hey, nice work! Could you also please add this somewhere in the Readme for visibility?

README.md
NanoCode012, line 1: Thank you for this example. Could you also add this `name:` parameter within the "all yaml options" below?

PR #279, opened 2023-07-15
Feat(readme): improve docs on multi-gpu

NanoCode012:
- Remove `accelerate config`
- Add fsdp sample
- Add method to use deepspeed


winglian (approved):
thanks!

PR #285, opened 2023-07-17
Attention mask and position id fixes for packing

winglian:


src/axolotl/utils/dataloader.py
```diff
@@ -0,0 +1,202 @@
...
+        result.append(batch[rank])
+        # add total seqs for all ranks
+        result_totseqs.append(tot_seqs)
+
+    return result, result_totseqs, s, len(result) * c * n
+
+
+class MultipackDistributedDataloader:
+    """Unpadded data loading using Multipack.
+    Approximate (at most ~1.22x) the optimal solution of the identical-machines scheduling problem, which is NP-hard.
```
NanoCode012, line 139: I think it would be good to give link to One's repo for source

src/axolotl/utils/dataloader.py
```diff
@@ -0,0 +1,202 @@
...
+        # add local rank
+        result.append(batch[rank])
+        # add total seqs for all ranks
+        result_totseqs.append(tot_seqs)
+
+    return result, result_totseqs, s, len(result) * c * n
+
+
+class MultipackDistributedDataloader:
+    """Unpadded data loading using Multipack.
```
winglian, line 137:
```suggestion
    """Unpadded data loading using Multipack.
    Adapted from https://github.com/imoneoi/openchat/blob/v3_fix_mle_loss/ochat/training_deepspeed/multipack_dataloader.py
```

scripts/finetune.py
```diff
@@ -345,14 +365,12 @@ def terminate_handler(_, __, model):
     # TODO do we need this fix? https://huggingface.co/docs/accelerate/usage_guides/fsdp#saving-and-loading
     # only save on rank 0, otherwise it corrupts output on multi-GPU when multiple processes attempt to write the same file
     if cfg.fsdp:
-        model.save_pretrained(cfg.output_dir)
+        trainer.save_model(cfg.output_dir)
```
winglian, line 370: this fixes an issue where the `.save_pretrained` call doesn't properly unwrap the model and leaves `_flat_param` modules in the saved model.

src/axolotl/datasets.py
```diff
@@ -148,13 +156,17 @@ def __iter__(self):
                             input_ids, dtype=self.tokens_dtype
                         )
                         attention_mask_with_concat = torch.tensor(
-                            attention_mask, dtype=self.tokens_dtype
+                            [idx * m for m in attention_mask], dtype=torch.int16
```
winglian, line 158: this is needed to create the correct expanded attention mask when using the legacy packing and SDP or flash-attn

src/axolotl/monkeypatch/llama_attn_hijack_flash.py
```diff
@@ -84,35 +92,15 @@ def forward(
...
-            h=nheads,
+        qkv = rearrange(qkv, "b s ... -> (b s) ...")
+        cu_q_lens, max_s = get_cu_seqlens(key_padding_mask)
+        cu_q_lens = cu_q_lens.squeeze()
+
+        output = flash_attn_varlen_qkvpacked_func(
+            qkv, cu_q_lens, max_s, 0.0, softmax_scale=None, causal=True
         )
+        output = rearrange(output, "(b s) ... -> b s ...", b=bsz)
+
```
winglian: not sure exactly what the old logic does, but new logic handles sample packing

src/axolotl/utils/data.py
```diff
@@ -413,7 +417,51 @@ def load_prepare_datasets(
...
+            + "test"
+            + "|"
+            + str(cfg.seed or 42)
+        )
+        train_fingerprint = hashlib.md5(
+            to_hash_train.encode(), usedforsecurity=False
+        ).hexdigest()
+        test_fingerprint = hashlib.md5(
+            to_hash_test.encode(), usedforsecurity=False
+        ).hexdigest()
```
winglian, line 441: pre-calculating a fingerprint fixes a bug where each process would otherwise calculate it's own different fingerprint for the same transform on the original dataset

src/axolotl/utils/dataloader.py
```diff
@@ -0,0 +1,301 @@
...
+    def __len__(self):
+        # this doesn't return the actual length b/c with distributed samplers, not all dataloaders get
+        # the same share of total tokens
+        # if not self.eff_total_used:
+        #     batches, _ = self.generate_batches(set_stats=True)
+        # LOG.info(
+        #     f"packing_efficiency_estimate: {self.packing_efficiency_estimate} "
+        #     f"actual packing efficiency: {self.efficiency()}"
+        # )
+        return max(1, self._len_est())
```
winglian, line 276: this prevents a but where training hangs near the end of an epoch since each process can end up with a different length based on it's random sampling.

src/axolotl/utils/trainer.py
```diff
@@ -76,6 +159,63 @@ def create_scheduler(
...
+                )
+            )
+        return super().get_eval_dataloader(eval_dataset)
+
+    def compute_loss(self, model, inputs, return_outputs=False):
+        # if self.args.sample_packing:
+        #     labels = inputs.pop("labels")
+        #     outputs = model(**inputs)
+        #     loss = trainer_weighted_loss(outputs, labels, shift_labels=True)
+        #     return (loss, outputs) if return_outputs else loss
```
winglian, line 219: this is one's weighted cross entropy loss. I think it's identical to HF's loss calc, but I need to verify another time

src/axolotl/datasets.py
```diff
@@ -30,19 +30,18 @@ def __init__(  # pylint: disable=super-init-not-called
...
     ):
         self.prompt_tokenizer = prompt_tokenizer
-        self.dataset = dataset
+        super().__init__(self.process(dataset).data, **kwargs)
 
-    def __iter__(self):
-        features = self.dataset.features.keys()
+    def process(self, dataset):
+        features = dataset.features.keys()
         num_proc = os.cpu_count()
```
NanoCode012:
```suggestion
        num_proc = min(64, os.cpu_count())
```

README.md
```diff
@@ -375,7 +375,10 @@ dataset_shard_idx:
 sequence_len: 2048
 # max sequence length to concatenate training samples together up to
 # inspired by StackLLaMA. see https://huggingface.co/blog/stackllama#supervised-fine-tuning
+# soon to be DEPRECATED
```
NanoCode012:
```suggestion
# FutureWarning: This will soon to be DEPRECATED
```

Also, what would be the con of leaving this alone, but just warn that there is a better option?

README.md
```diff
@@ -375,7 +375,10 @@ dataset_shard_idx:
...
 max_packed_sequence_len: 1024
+# use efficient multi-packing with block diagonal attention and per sequence position_ids
```
NanoCode012:
```suggestion
# use efficient multi-packing with block diagonal attention and per sequence position_ids. Recommend set to 'true'
```

scripts/finetune.py
```diff
@@ -232,12 +237,25 @@ def train(
...
                 max_tokens=cfg.sequence_len,
-                seed=cfg.seed,
+                seed=cfg.seed or 42,
             )
             # https://discuss.huggingface.co/t/how-to-use-huggingface-trainer-streaming-datasets-without-wrapping-it-with-torchdatas-iterablewrapper/25230
             train_dataset = train_dataset.with_format("torch")
             eval_dataset = None
 
+        if is_main_process():
+            # process on rank 0 first so it gets cached so other ranks load from cache
```
NanoCode012, line 247: We need to remove the tip to pre-tokenize from Readme or change it to say "You can pre-tokenize" instead of "recommending pre-tokenization now"

src/axolotl/prompt_strategies/alpaca_w_system.py
```diff
@@ -86,12 +90,20 @@ class OpenOrcaSystemDataPrompter(SystemDataPrompter):
...
 
     def match_prompt_style(self):
+        # pylint: disable=duplicate-code
         if self.prompt_style == PromptStyle.INSTRUCT.value:
             self.turn_format = "### User:\n{instruction}\n\n### Additional Context:\n{input}\n\n### Assistant:\n"
             self.turn_no_input_format = "### User:\n{instruction}\n\n### Assistant:\n"
         if self.prompt_style == PromptStyle.CHAT.value:
-            self.turn_format = "USER: {instruction}\n{input}\nASSISTANT:"
-            self.turn_no_input_format = "USER: {instruction}\nASSISTANT:"
+            self.turn_format = "User: {instruction}\n{input}\nAssistant:"
```
NanoCode012: The format got changed from upper caps! Do we want to just change this? Vicuna 1.1 uses `USER`.

src/axolotl/utils/data.py
```diff
@@ -413,7 +414,51 @@ def load_prepare_datasets(
...
+
+        if is_main_process():
+            dataset = dataset.train_test_split(
```
NanoCode012, line 444: We can extract this into a yield generic function since I see this pattern twice now.

Do X in main first, then do Y in other workers.

Sample code: https://github.com/ultralytics/yolov5/blob/dd104811c2e0419529aa7177ce160321e9694837/utils/torch_utils.py#L89

src/axolotl/utils/models.py
```diff
@@ -90,8 +96,10 @@ def load_model(
 
     # TODO refactor as a kwarg
     load_in_8bit = cfg.load_in_8bit
-    cfg.is_llama_derived_model = "llama" in base_model or (
-        cfg.model_type and "llama" in cfg.model_type.lower()
+    cfg.is_llama_derived_model = (
+        "llama" in base_model
+        or (cfg.model_type and "llama" in cfg.model_type.lower())
+        or cfg.is_llama_derived_model is True
```
```suggestion
        or cfg.is_llama_derived_model
```

I would recommend moving this arg first though.

src/axolotl/utils/trainer.py
```diff
@@ -190,7 +443,14 @@ def setup_trainer(cfg, train_dataset, eval_dataset, model, tokenizer):
     if cfg.save_safetensors:
         training_arguments_kwargs["save_safetensors"] = cfg.save_safetensors
 
+    if cfg.sample_packing_eff_est:
+        training_arguments_kwargs[
+            "sample_packing_efficiency"
+        ] = cfg.sample_packing_eff_est
```
NanoCode012, line 449: Need to add this cfg to readme

src/axolotl/utils/trainer.py
```diff
@@ -204,7 +464,7 @@ def setup_trainer(cfg, train_dataset, eval_dataset, model, tokenizer):
         eval_steps=cfg.eval_steps if cfg.val_set_size > 0 else None,
         save_steps=cfg.save_steps,
         output_dir=cfg.output_dir,
-        save_total_limit=3,
+        save_total_limit=cfg.save_total_limit if cfg.save_total_limit else 4,
```
NanoCode012, line 467: Need to add this cfg to readme

src/axolotl/utils/trainer.py
```diff
@@ -222,6 +482,8 @@ def setup_trainer(cfg, train_dataset, eval_dataset, model, tokenizer):
         if cfg.lr_scheduler and cfg.lr_scheduler not in ("one_cycle", "log_sweep")
         else "cosine",
         weight_decay=cfg.weight_decay if cfg.weight_decay is not None else 0.0,
+        sample_packing=cfg.sample_packing if cfg.sample_packing else False,
+        sample_packing_seq_len_multiplier=cfg.micro_batch_size or 1,
```
NanoCode012: We should not set default here. We must error if this cfg.micro_batch_size is not provided since it's just universally needed.

src/axolotl/utils/collators.py
```diff
@@ -0,0 +1,121 @@
...
+from dataclasses import dataclass
+from typing import Any, Optional, Union
+
+import numpy as np
+from transformers import PreTrainedTokenizerBase
+from transformers.utils import PaddingStrategy
+
+
+@dataclass
+class DataCollatorForSeq2Seq:
```
NanoCode012, line 13: Should this inherit from Transformers?

winglian: It was a full copy/paste and I don't think there is anything I wanted to inherit

src/axolotl/utils/data.py
winglian: let's make this a follow up issue

scripts/finetune.py
winglian: updated

tmm1:
Some resources that inspired this work:

https://www.graphcore.ai/posts/packedbert-how-to-accelerate-nlp-tasks-for-transformers-with-packing

https://arxiv.org/pdf/2107.02027.pdf

https://github.com/imoneoi/multipack_sampler

PR #293, opened 2023-07-19
Fix(tokenizing): Use multi-core

NanoCode012:
This improves speed for tokenizer and adds a progress bar.

Benchmark:

| dataset | old | new | 
| --- | --- | --- |
| teknium/GPT4-LLM-Cleaned | ~70s | < 10s |
| WizardLM/WizardLM_evol_instruct_V2_196k | ~5 min | ~55s | 

Note: just internally counting, nothing strict.

Future PR for more speed can:
- use batching. (requires changing class to not use `iter`)
- fix Dataset.from_list and shuffle (do we need to save to a list? why not save to a dataset directly to cut off this part)
- ConstantIterList can be improved

**Full credit to**: neverendingtoast

### DIFF
--- a/src/axolotl/datasets.py
+++ b/src/axolotl/datasets.py
@@ -1,12 +1,13 @@
 """Module containing Dataset functionality"""
 
 import logging
+import os
 from typing import List
 
 import torch
 from datasets import IterableDataset
 
-from .prompt_tokenizers import InvalidDataException, PromptTokenizingStrategy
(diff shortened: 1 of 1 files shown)


winglian (approved):
🔥

NanoCode012:
One note: this removes the error if dataset is empty. We can simply just add into it the `data.py` file instead.

PR #265, opened 2023-07-06
Added experimental vicuna 1.1+ support

jphme:
As discussed in #262 and #261 , I implemented support for Vicuna 1.1+, following the [original code](https://github.com/lm-sys/FastChat/blob/main/fastchat/train/train.py) as close as possible.

This should be even more flexible than Vicunas own finetuning/training code, as my code allows for flexible system messages.

Dataset should look like 
```
[{"conversation":{"from": "human", "value": "Who are you?"}, {"from": "gpt", "value": "I am Vicuna"},...}]
```
as .jsonl file; the first "from" can be "system" instead and will be used as system message if present.

Submitting as draft as the code is still untested, but will report back.

PR #296, opened 2023-07-19
experimental llama 2 chat support

jphme:
Heavily based on my Vicuna 1.3 implementation in #265 and leaning on FastChats implementation of Llama2 Chat tokenization.

Still experimental and untested, happy for feedback/improvements (I am not sure whether the eos/bos token handling is 100% correct here...).


winglian:
thanks! would you mind adding some unit tests? for example for sharegpt/vicuna we use some fixtures https://github.com/OpenAccess-AI-Collective/axolotl/blob/main/tests/test_prompt_tokenizers.py#L39-L61

jphme:
> thanks! would you mind adding some unit tests? for example for sharegpt/vicuna we use some fixtures https://github.com/OpenAccess-AI-Collective/axolotl/blob/main/tests/test_prompt_tokenizers.py#L39-L61

Hey @winglian thank you very much. I wanted to finish yersterday but was too tired in the end - want to get everything in a working state (and this was a lot of debugging, as even the original LLAMA implementation and the HF transformers implementation had different tokenization, see .e.g. https://github.com/huggingface/transformers/pull/24935 .

Attention Masks are still off by one in some cases (working on it now), but eveything else should be consistent with the reference implementation now.

If everything works and I did a successful finetune, I will try to add some fixtures. 
A little problem is that I currently have no GPU locally or on a fixed dev server and thus testing is always difficult outside of finetuning runs on Lambdalabs/runpod (and I am still quite new so don't really have a good setup with sufficient automation).

jphme:
Should be ready to use now, I double checked the tokenization with [Transformers](https://github.com/huggingface/transformers/blob/7381987f906114a3b3c1297f2954df5a63c3eb69/src/transformers/models/llama/tokenization_llama.py#L334) and the original Llama implementation and spend a lot of time to be sure that the masking works correctly.

Will try to add tests/fixtures but help would be appreciated

jphme:
added fixtures now (and even an additional test that compares to transformers tokenization. Unfortunately they don't implement attention masking, otherwise we could've just used their method....).

jphme:
somehow the tokenized file messed up pre-commits again? Should be fixed (and my finetune with this implementation is running now...)

winglian (approved):

minosvasilias:
Why is this using `human` and `gpt` roles instead of the `user` and `assistant` ones outlined in the [original Llama repo](https://github.com/facebookresearch/llama/blob/main/example_chat_completion.py)?

jphme:
> Why is this using `human` and `gpt` roles instead of the `user` and `assistant` ones outlined in the [original Llama repo](https://github.com/facebookresearch/llama/blob/main/example_chat_completion.py)?

To keep it consistent with the ShareGPT and Vicuna training format and make it more convenient to finetune different models without changing the (conversation) data format every time.
From my own experience, its way more convenient to have some kind of standard data format when finetuning (because you will often try to finetune different models with the same data). The Prompt Strategy should be the layer that customizes that standard conversation format for the model architecture. Imho there are different trade-offs for tuning and inference...

Changing that is also trivial (just adapt the prompter to the different role names) but changing the role names in your dataset is even more trivial.

(But that all my personal opinion, happy to change that if people finde it confusing)

minosvasilias:
Makes sense, thanks for explaining.
If the role names were retained/injected into the prompt, it'd be a different story, but since it's just SYS and INST tokens it makes no practical difference and i agree consistency across models is nice. 👍

tmm1:
This is great, thanks @jphme!

PR #300, opened 2023-07-20
Pytorch 2.0.1

winglian:
since we can't use flash attention 2 yet, let's make sure we pin the base model to v1.0.9 for now, pin to PyTorch 2.0.1 and fix the PyTorch scaled dot product attention.


NanoCode012:
I vaguely remember some package that only works on torch 2.0.0 and not torch 2.0.1. Was it xformers?

PR #313, opened 2023-07-22
don't resize embeddings to multiples of 32x by default

winglian:
resizing the embeddings to larger than the actual number of tokens causes problems with llama.cpp downstream. let's not enable this by default for safety.


NanoCode012:
YES! I had to modify downstream code for this 😂

NanoCode012:
While on this topic, I was wondering whether to consider using the "smart resize" code floating around in fast chat and other repos.

PR #318, opened 2023-07-22
add support for H100/CUDA 12.2

winglian:


winglian:
Closing since main seems to work consistently on h100s now.

PR #321, opened 2023-07-24
latest HEAD of accelerate causes 0 loss immediately w FSDP

winglian:
this PR pins accelerate to a known good version since HEAD results in immediate 0.0 loss


winglian:
Issue filed with accelerate https://github.com/huggingface/accelerate/issues/1763

PR #322, opened 2023-07-25
ReLoRA implementation (with quantization)

cg123:
This implements most of the method of ReLoRA, from this paper: https://arxiv.org/abs/2307.05695
It works with both lora + qlora adapters, unquantized or with 4/8 bit quantization.  It's slower and uses more disk space if quantization is enabled, as it has to save full weight checkpoints in order to apply the weight differences and re-quantize. But it works! Eventually-approximately-maybe-almost-full rank training on a consumer GPU.

Currently very ugly. I don't have time to clean this up for a proper pull request at the moment, but will in a few weeks. In the meanwhile here it is as a proof of concept.

Currently limited to models on disk, and only tested to work with the `adamw_bnb_8bit` optimizer.

Example config:
```yml
base_model: /root/llama-7b/
base_model_config: /root/llama-7b/
load_in_4bit: true
adapter: qlora
lora_r: 8
lora_alpha: 16
lora_target_linear: true
relora_steps: 100
relora_warmup_steps: 10
relora_cpu_offload: false
```


src/axolotl/utils/trainer.py
```diff
@@ -292,6 +293,22 @@ def setup_trainer(cfg, train_dataset, eval_dataset, model, tokenizer):
         trainer_kwargs["optimizers"] = (optimizer, lr_scheduler)
 
     callbacks = []
+
+    if cfg.relora_steps:
+        assert cfg.adapter in (
+            "lora",
+            "qlora",
+        ), "Adapter must be lora or qlora to use ReLoRA"
```
winglian: ~~this should get moved to `validation.py`~~ Since this is already in `validation.py`, this can be removed.

mhenrichsen:
Looks really interesting :)
@cg123 can you add some documentation?

winglian:
@cg123 from what I can tell, the result of this is a full weights model and not an adapter, right? Would it be possible to have the final output be an adapter?

winglian:
[relora.patch](https://github.com/OpenAccess-AI-Collective/axolotl/files/12328580/relora.patch)

after rebasing this on main, here's a patch to move this into it's own trainer class with a method override to create the proper scheduler

cg123:
> @cg123 from what I can tell, the result of this is a full weights model and not an adapter, right? Would it be possible to have the final output be an adapter?

The result of the ReLoRA procedure can't be modeled by a single LORA. I suppose it would be possible to output a set of N LORAs (one for each restart that occurred,) but that would add more bookkeeping, use even more disk space, and I'm not sure if that's actually useful for anyone.

Thanks for that patch! It's definitely cleaner that way. I'll do a bit more cleanup and write some documentation in the next couple of days.

cg123:
This should be pretty much good to go now. Two notes:
* ReLoRA uses a lot of disk space - checkpoints are full-parameter fp16 saves. I've been allocating 300gb instances to fine-tune 13b Llama models.
* Resuming from checkpoint does not work, as it will load the latest lora but not the latest base model. `validate_config` now checks this. I might do a followup pr to add a workaround.

cg123:
I lied, I couldn't let that stand. Resuming from checkpoints works now.

scripts/finetune.py
```diff
@@ -301,13 +302,17 @@ def terminate_handler(_, __, model):
 
     LOG.info(f"Training Completed!!! Saving pre-trained model to {cfg.output_dir}")
 
+    if cfg.adapter == "lora" and cfg.relora_steps:
+        model = model.merge_and_unload()
```
NanoCode012: Is this necessary for only lora + relora setting? How about qlora?

scripts/finetune.py
```diff
@@ -242,6 +242,21 @@ def train(
...
+            sorted_paths = sorted(
+                possible_checkpoints,
+                key=lambda path: int(path.split("-")[-1]),
+            )
+            cfg.resume_from_checkpoint = sorted_paths[-1]
+            LOG.info(
+                f"Using Auto-resume functionality to start with checkpoint at {cfg.resume_from_checkpoint}"
+            )
+    resume_from_checkpoint = cfg.resume_from_checkpoint
+
```
NanoCode012, line 259: Just curious on why this is moved. Did it fix the resume issue?

src/axolotl/utils/trainer.py
```diff
@@ -608,6 +656,8 @@ def setup_trainer(cfg, train_dataset, eval_dataset, model, tokenizer, total_num_
     trainer_cls = (
         OneCycleLRSchedulerTrainer
         if cfg.lr_scheduler == "one_cycle" and (cfg.fsdp or cfg.adapter == "qlora")
+        else ReLoRATrainer
+        if cfg.relora_steps
```
NanoCode012: I recommend refactor to a normal if elif instead of this long chain

NanoCode012 (commented):
Hey there! Thanks for this PR. I just added a few points that I've noticed.

scripts/finetune.py
cg123: It is only necessary for lora + relora, and only when the base weights are unquantized. Thanks for poking at this, I hadn't noticed that the final save for 8-bit lora+relora was incorrect. I've made it more explicit what's going on here.

scripts/finetune.py
cg123: Yes, this was part of it. `ReLoRACallback` needs the path to the checkpoint being resumed from, so it needs to have been selected before the callback gets constructed (immediately below in `setup_trainer`.)

src/axolotl/utils/trainer.py
cg123: Done, and I've added a check for overlapping one_cycle + relora in `validate_config`.

ehartford:
Can't wait to try it out!

ehartford:
Is there anything special when I use ReLoRA?  Do I need to increase learning rate?

cg123:
> Is there anything special when I use ReLoRA? Do I need to increase learning rate?

In general the hyperparameters for a regular QLoRA/LoRA run should be about right for ReLoRA as long as `relora_steps` isn't set too low.

A larger number of restarts means a higher-rank cumulative update, and thus a potentially lower loss minimum, but also slower convergence. I've been trying to aim for 3 to 10 restarts for the whole training run, which has landed around 150 steps per restart for most of my testing configurations. With the exact same learning rate, batch sizes, etc. that I would for training with QLoRA, but with smaller values of `lora_r`, I've seen in the neighborhood of 15% slower convergence but lower final losses.

I would bet you could be a little more aggressive with the learning rate and get even better results. I don't have the compute to explore that properly, though, so I've been playing it safe.

Hope this helps!

mhenrichsen:
In regards to LR, it has been suggested to double the standard learning rate.
https://twitter.com/guitaricet/status/1680647706048430087

PR #339, opened 2023-08-05
Update XFormers Attention Monkeypatch to handle Llama-2 70B (GQA)

ssmi153:
This is a fix for this issue: https://github.com/OpenAccess-AI-Collective/axolotl/issues/338 . It fixes the XFormersAttention monkeypatch. I've tested this with LLama-2 70B and Llama-2 13B and both are able to start training appropriately without any errors. 

Notes: 
1) I haven't yet tested this with an complete finetune from start to finish - this will take a number of days to complete.
2) We still also need to fix the FlashAttention monkeyscript. I've been working on this but have run into a number of errors, so I'm pushing this out in the meantime.

### DIFF
--- a/src/axolotl/monkeypatch/llama_attn_hijack_xformers.py
+++ b/src/axolotl/monkeypatch/llama_attn_hijack_xformers.py
@@ -156,7 +156,9 @@ def xformers_forward(
                 f" {attn_output.size()}"
             )
 
-    attn_output = attn_output.transpose(1, 2).contiguous()
+        attn_output = attn_output.transpose(1, 2).contiguous()
+        #end x-formers vs. not x-formers if-else block
+
     attn_output = attn_output.reshape(bsz, q_len, self.hidden_size)
 
     if self.pretraining_tp > 1:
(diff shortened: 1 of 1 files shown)


ssmi153:
Maybe hold off on committing this - I just found that while it runs LLama-2-13b properly, the losses are completely different (starts at 8 rather than starting at 1). I need to look more closely into why this is, as it shouldn't be changing anything for the 13B model.

ssmi153:
OK, this most recent commit fixes the difference in losses. It was a whitespace issue - darn Python! This now performs exactly like it did previously for 13B (and 7B) models without GQA, and also now works for 70B models with GQA. It's ready for review and merging.

winglian (approved):
thank you!

winglian:
@ssmi153 looks like there are some pre-commit formatting issues. `pre-commit run --all-files` should fix it

ssmi153:
OK, I'll look into this. Struggling to get pre-commit to work on Windows at
the moment. I'll sort this out as soon as I can.

winglian:
@ssmi153 you could cherry-pick this commit https://github.com/OpenAccess-AI-Collective/axolotl/commit/9793faf6dc0c68908aee620c6dbe3a18322861c7

ssmi153:
Champion! I've pulled your changes into the pull request (in probably in the most awkward possible way, sorry - it's been a few years since I last commited code, and our last setup used mercurial, so I'm just getting my head around Github's workflows).

winglian:
Thanks @ssmi153. This is very much appreciated and wanted.

PR #343, opened 2023-08-05
Feat: Add rope scaling

### LINKED ISSUE #342: [Feature] Add config for RoPE scaling
Seems like we can open two config
- rope_scaling_type:
- rope_scaling_factor:

Validation is done in `transformer` side.

Ref: https://github.com/gante/transformers/blob/30409af6e1b2b5efb6d9932b3e3b4ce20cfdb30e/src/transformers/models/llama/configuration_llama.py#L155-L174

NanoCode012:
Closes #342 

Note: 
- This only applies it to `llama` arch. It should also support `neox`. Should I add specific if condition for this?
- It was stated on docs to NOT increase the max_position_embeddings ourselves for this. Let's say, if we want 8k, just change the rope factor.
- We leave validation to `transformer`.

- [x] Test

### DIFF
--- a/README.md
+++ b/README.md
@@ -474,6 +474,10 @@ landmark_attention:
 # xpos RoPE see https://github.com/kaiokendev/cutoff-len-is-context-len/blob/main/util/xpos_rope_llama_monkey_patch.py
 # llama only
 xpos_rope:
+# RoPE Scaling https://github.com/huggingface/transformers/pull/24653
+rope_scaling:
+  type: # linear | dynamic
+  factor: # float
 
 # resume from a specific checkpoint dir
 resume_from_checkpoint:
--- a/src/axolotl/utils/models.py
+++ b/src/axolotl/utils/models.py
@@ -219,7 +219,9 @@ def load_model(
         elif cfg.is_llama_derived_model and not cfg.trust_remote_code:
             from transformers import LlamaForCausalLM
 
-            config = LlamaConfig.from_pretrained(base_model_config)
+            config = LlamaConfig.from_pretrained(
+                base_model_config, rope_scaling=cfg.rope_scaling
+            )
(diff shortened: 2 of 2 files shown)


winglian (approved):
lgtm

winglian:
> * It was stated on docs to NOT increase the max_position_embeddings ourselves for this. Let's say, if we want 8k, just change the rope factor.

has everyone else been doing it incorrectly? https://huggingface.co/togethercomputer/LLaMA-2-7B-32K/blob/main/config.json#L14

NanoCode012:
> > * It was stated on docs to NOT increase the max_position_embeddings ourselves for this. Let's say, if we want 8k, just change the rope factor.
> 
> has everyone else been doing it incorrectly? https://huggingface.co/togethercomputer/LLaMA-2-7B-32K/blob/main/config.json#L14

It could be seen here on docs to not set it ourselves: https://github.com/gante/transformers/blob/30409af6e1b2b5efb6d9932b3e3b4ce20cfdb30e/src/transformers/models/llama/configuration_llama.py#L80-L87

At the same time, we also change this if the seq_len > the model's context, so I'm not sure if we should add a validation config check for that to make sure it does not happen.

Regarding the together's model, they have their own custom code for modelling_llama, since they require trust_remote_code. It's not exactly the same, so I can't say..

NanoCode012:
This brought up another question in my mind, what about dataset packing? Do we just pass a 4k dataset to a "16k" dataset? If we do not set 16k seq_len, the dataset would be at most 4k? Do we need to add a condition to Not change max_pos_embedding if rope_scaling is true?

NanoCode012:
Tested that it runs:
- with rope scaling
- with rope scaling , type: burp . Error as expected
- without rope scaling

PR #350, opened 2023-08-07
set `group_by_length` to false in all examples

tmm1:
the README discourages its use

and it makes running the examples slower


mhenrichsen:
We discussed the use of group_by_length in the discord channel a week ago or so, and concluded that it was worth setting it to true even though training loss will be a bit more wonky throughout training, it should be worth the better RAM utilization.

winglian (approved):

tmm1:
@mhenrichsen thanks for that pointer, i updated the README based on the summary of that discussion

PR #351, opened 2023-08-07
smart resize embeddings

winglian:


README.md
```diff
@@ -322,9 +322,9 @@ tokenizer_type: AutoTokenizer
 trust_remote_code:
 # use_fast option for tokenizer loading from_pretrained, default to True
 tokenizer_use_fast:
-# resize the model embeddings when new tokens are added to multiples of 32
-# this is reported to improve training speed on some models
-resize_token_embeddings_to_32x:
+# resize the model embeddings when new tokens are added to multiples of N
+# multiples of 32 are reported to improve training speed on some models
+resize_token_embeddings_multiple:
```
NanoCode012, line 331: I think this description is a bit confusing. Need to make it clear that we expect the `N`.

src/axolotl/utils/models.py
```diff
@@ -31,6 +31,45 @@
...
+    embedding matrix of the model to match the new size of the tokenizer. If any new special tokens
+    have been added, the function computes the average embedding values of the existing embeddings
+    and sets those values for the new special token embeddings. This is done separately for the input
+    embeddings and output embeddings of the model.
+    """
+
+    old_tokens = model.get_input_embeddings().weight.data.shape[0]
+    num_new_tokens = len(tokenizer) - old_tokens
+    embeddings_len = (
+        math.ceil(len(tokenizer) / resize_token_embeddings_multiple)
```
NanoCode012, line 52: I think we may need some validation for this `resize_token_embeddings_multiple` value (maybe in validate_config).
- is it a number? is it allowed to be a float?
- is it smaller than length of tokenizer..?
- any other edge case?

src/axolotl/utils/models.py
```diff
@@ -31,6 +31,45 @@
...
+        * resize_token_embeddings_multiple
+        if resize_token_embeddings_multiple
+        else len(tokenizer)
+    )
+    model.resize_token_embeddings(embeddings_len)
```
NanoCode012, line 57: Something I wondered, is there a reason to `resize` if num_new_tokens <= 0?

Could we move this line into `if` block?

One thing: this `resize` function calls `tie_weights` internally. If we do move it into below `if`, would there be any loss/side effect?

winglian:
@NanoCode012 this is roughly based on https://github.com/tatsu-lab/stanford_alpaca/blob/761dc5bfbdeeffa89b8bff5d038781a4055f796a/train.py#L65-L85 adapted to how we handle things

NanoCode012:
@winglian , ah yes. I guess we can leave the main logic untouched for safety. Tho, I think the naming or doc of the config should be a bit adjusted a bit to make it clear that it expects a value.

README.md
NanoCode012: `pad_to_multiple_of` HF plan to support this arg. Would be good to use this name. (PR not merged yet) https://github.com/huggingface/transformers/pull/25088/files#diff-6b72b98c4c2dcfc6cc606843917733f5d858374fbc22a735ff483bbc0c1e63ea

NanoCode012:
This is a bit stale, so closing it. We do have `resize_token_embeddings_to_32x: bool`, but I'm not aware people even use it.

PR #353, opened 2023-08-08
Added a simple inference webserver

enn-nafnlaus:
Usage example:

accelerate launch scripts/finetune.py summarize.yaml --inference --base_model=path/to/my/model --load_in_8bit=True --server --server_port 1567 --server_addr 127.0.0.1

Then in another terminal:

curl -X POST -d "$(cat test_text.txt)" http://localhost:1567/

### DIFF
--- a/scripts/finetune.py
+++ b/scripts/finetune.py
@@ -13,9 +13,13 @@ import fire
 import torch
 import yaml
 
+import http.server
+import socketserver
+import urllib.parse
+
 # add src to the pythonpath so we don't need to pip install this
 from optimum.bettertransformer import BetterTransformer
-from transformers import GenerationConfig, TextStreamer
+from transformers import GenerationConfig, TextStreamer, TextDataset
 
 from axolotl.logging_config import configure_logging
 from axolotl.utils.data import load_prepare_datasets, load_pretraining_dataset
@@ -36,7 +40,6 @@ LOG = logging.getLogger("axolotl.scripts")
 
 DEFAULT_DATASET_PREPARED_PATH = "last_run_prepared"
 os.environ["HF_HUB_ENABLE_HF_TRANSFER"] = "1"
-
 
 def choose_device(cfg):
     def get_device():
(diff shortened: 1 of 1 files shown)


utensil:
Is it possible to allow submitting inference request during training? Just like running a manual eval.

enn-nafnlaus:
> Is it possible to allow submitting inference request during training? Just like running a manual eval.

Could always run a separate server while training.

utensil:
> > Is it possible to allow submitting inference request during training? Just like running a manual eval.
> 
> Could always run a separate server while training.

The idea is to fully integrated into training. The server accepts requests and put it into a queue, the model being trained is register with a callback to dequeue the request and inference it, just as running evals during training.

In the case, we can reuse the model in training to do live inference at will.

I admit that this might be a separate PR.

enn-nafnlaus:
I agree, that's a separate PR.  :)  This is a very simple lightweight server, just a modification of the existing inference code to get instructions from HTTP requests rather than the terminal, and return them to the requesting client rather than the screen.  If you want to take the time to configure it to run training and inference through a queue, go ahead  :)

I implemented it because to run test inferences, you either have to do it completely manually at present (typing / pasting into a terminal for each one), or you have a long overhead of waiting for the inference server to start up for every single test.  By starting up a HTTP server and accepting requests on it, you can automatically run many different tests without a ton of overhead.  Or use it for non-testing / production purposes, for that matter.

utensil:
Yes, currently it's not able to reuse the loaded model to inference a batch of inputs, very inconvenient.

scripts/finetune.py
```diff
@@ -163,6 +166,82 @@ def check_not_in(list1: List[str], list2: Union[Dict[str, Any], List[str]]) -> b
...
+
+    def do_POST(self):
+        content_length = int(self.headers['Content-Length'])
+        post_data = self.rfile.read(content_length)
+        instruction = post_data.decode('utf-8')
+        
+        if not instruction:
+            response = ""
+        else:
+            default_tokens = {"unk_token": "<unk>", "bos_token": "<s>", "eos_token": "</s>"}
```
winglian, line 185: we could probably grab these from the tokenizer, as any special tokens defined in the config are added to the tokenizer when it is instantiated.

enn-nafnlaus: Patches welcome.  :)  Again, this is just redoing a copy of the preexisting inference section as a webserver. I didn't change the logic.

scripts/finetune.py
```diff
@@ -279,7 +358,14 @@ def train(
                 prompter = None
             else:
                 prompter = kwargs["prompter"]
-        do_inference(cfg, model, tokenizer, prompter=prompter)
+
+        if cfg.server:
+            server_address = (cfg.server_addr, cfg.server_port)
+            httpd = socketserver.TCPServer(server_address, lambda *args, **kwargs: HttpHandler(*args, cfg=cfg, prompter=prompter, tokenizer=tokenizer, model=model, **kwargs))
+            print(f"Server running on port {cfg.server_port}")
+            httpd.serve_forever()
```
winglian, line 366: should this be explicitly killed at the end of training?

enn-nafnlaus: Can you train and run inference at the same time?  This only runs if you're in --inference mode.

Stillerman:
Might be able to use Gradio as a web server #812

NanoCode012:
Closing as stale, though I do agree our inference cli should be revamped.

PR #357, opened 2023-08-10
benchmark harness

tmm1:
this is the very beginning of a simple benchmark harness

it uses pytest plugins to run generate different cfgs, and can collect arbitrary data during the test about things like vram usage, forward/backward pass speeds, etc

for example:

```
test_id                                status      duration_ms    vram_baseline
-------------------------------------  --------  -------------  ---------------
test_benchmark_attn[llama2]            passed          7252.83         0.314697
test_benchmark_attn[llama2_xformers]   passed          6924.83         0.608582
test_benchmark_attn[llama2_flashattn]  passed          6879.68         0.6203
```

right now this is only wired up to `load_model` and `load_tokenizer`, but i will continue to flesh it out


benchmarks/test_bench.py
```diff
@@ -0,0 +1,267 @@
...
+                    results_bag[key] = val
+
+    finally:
+        if trainer is not None:
+            opt = trainer.optimizer
+            trainer.optimizer = None
+            opt.zero_grad()
+            del opt
+            del trainer
+        del tokenizer
```
winglian: wrap in try / except UnboundLocalError:

PR #363, opened 2023-08-12
fix(save): save as safetensors

NanoCode012:
Please merge when approved.


winglian (approved):
@NanoCode012 can you rebase against main pls?

Issue #346, opened 2023-08-06
OSError: Error no file named pytorch_model.bin, tf_model.h5, model.ckpt.index or flax_model.msgpack found in directory


SuperMasterBlasterLaser:
Hello everyone.

I'm trying to train LoRA by using this [model](https://huggingface.co/TheBloke/airoboros-l2-70B-GPT4-2.0-GPTQ)

I've downloaded this model at directory and placed my training files. My config looks like this:

```yaml
base_model: /root/TheBloke/airoboros-l2-70B-GPT4-2.0-GPTQ
base_model_config: /root/TheBloke/airoboros-l2-70B-GPT4-2.0-GPTQ
model_type: AutoModelForCausalLM
tokenizer_type: AutoTokenizer

gptq: true
gptq_groupsize: 128 # group size
gptq_model_v1: false # v1 or v2

load_in_8bit: false
load_in_4bit: true
strict: false

datasets:
  - path: json
    data_files: /root/lines.jsonl
    type: completion
dataset_prepared_path: last_run_prepared
val_set_size: 0.01
output_dir: /root/my_lora

sequence_len: 4096
max_packed_sequence_len: 4096

adapter: lora
lora_model_dir:
lora_r: 32
lora_alpha: 16
lora_dropout: 0.05
lora_target_linear: true
lora_fan_in_fan_out:

wandb_project:
wandb_watch:
wandb_run_id:
wandb_log_model:

gradient_accumulation_steps: 4
micro_batch_size: 2
num_epochs: 3
optimizer: adamw_bnb_8bit
lr_scheduler: cosine
learning_rate: 0.0002

train_on_inputs: false
group_by_length: true
bf16: true
fp16: false
tf32: false

gradient_checkpointing: true
early_stopping_patience:
resume_from_checkpoint:
local_rank:
logging_steps: 1
xformers_attention: true
flash_attention:

warmup_steps: 10
eval_steps: 20
save_steps:
debug:
deepspeed:
weight_decay: 0.0
fsdp:
fsdp_config:
special_tokens:
  bos_token: "<s>"
  eos_token: "</s>"
  unk_token: "<unk>"
  pad_token: "<pad>"

```

When I launched it with command

    accelerate launch scripts/finetune.py /root/lora.yml

It returned this error:

```
Traceback (most recent call last):
  File "/root/axolotl/scripts/finetune.py", line 358, in <module>
    fire.Fire(train)
[...]

PR #364, opened 2023-08-12
fix(model loading): warn when model revision is passed to gptq

NanoCode012:
Adds a check if `model_revision` is passed when using gptq and depending on axolotl to download from HF.

I chose to raise an error instead of modifying upstream repo as I don't think that's preferred right now.


winglian (approved):

NanoCode012:
Relates to #346

PR #365, opened 2023-08-12
Fix(message): Improve error message for bad format

### LINKED ISSUE #325: ShareGPT datasets very strict -- data rows with non-standard role throw AssertionError
I found that one of the datasets I was consuming from HuggingFace with multi-language entries occasionally had roles other than "user" and "gpt". For example, "bing" or "chatgpt". This results in the entire dataset being rejected with an Assertion error and no context.

The offending code is:
```
  File "src/axolotl/prompters.py", line 319, in build_prompt
    assert role == conv.roles[j % 2]
AssertionError
```

Ideally additional context would be given to allow for debugging, such as the offending line or perhaps the rest of the datafile can be processed and a warning thrown.

People running into this issue can either remove or edit the lines with abnormal role names or add those roles to line 302 of prompters.py:
`        roles = {"human": conv.roles[0], "gpt": conv.roles[1]}`

Note that the particular file I was using then had other issues (e.g. system messages as "gpt") that resulted in IndexError as well

 In general would be good for this parsing to provide more context when issues are encountered, but opening this issue in part so that others can know that these issues are likely specific types of data issues.

NanoCode012:
Closes #325

### DIFF
--- a/src/axolotl/prompt_strategies/llama2_chat.py
+++ b/src/axolotl/prompt_strategies/llama2_chat.py
@@ -29,7 +29,7 @@ from dataclasses import dataclass, field
 from typing import Generator, List, Sequence
 
 from axolotl.prompt_tokenizers import PromptTokenizingStrategy
-from axolotl.prompters import IGNORE_TOKEN_ID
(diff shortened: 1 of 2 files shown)


winglian (approved):
🔥

PR #366, opened 2023-08-12
fix(tokenize): fix list index out of range by swapping dataset

### LINKED ISSUE #314: [Bug] IndexError: list index out of range in falcon config
I received this while testing the falcon config.

Reproduce:

`accelerate launch scripts/finetune.py examples/falcon/config-7b.yml`

Fix:

```diff
datasets:
  - path: teknium/GPT4-LLM-Cleaned
    type: alpaca:chat
+    data_files:  alpaca_gpt4_data_unfiltered.json
```

There is some issue with the [unnatural_instructions_unfiltered.json](https://huggingface.co/datasets/teknium/GPT4-LLM-Cleaned/blob/main/unnatural_instructions_unfiltered.json) file that is causing the conflict.

NanoCode012:
Fix #314 . 

This is only a workaround. The other workaround is to point to the exact file path, but I find this solution a bit cleaner.

It's still unclear on why this problem exists for falcon+pythia, but ok for llama.

### DIFF
--- a/examples/falcon/config-7b-lora.yml
+++ b/examples/falcon/config-7b-lora.yml
@@ -9,8 +9,8 @@ gptq: false
 strict: false
 push_dataset_to_hub:
 datasets:
-  - path: teknium/GPT4-LLM-Cleaned
-    type: alpaca:chat
+  - path: mhenrichsen/alpaca_2k_test
+    type: alpaca
 dataset_prepared_path: last_run_prepared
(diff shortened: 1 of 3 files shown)


NanoCode012:
Should we just merge this? Teknium might be busy to fix his dataset. Using a simple 2k dataset also simplifies testing for others.

winglian:
superseded by #395

PR #375, opened 2023-08-13
simplify `load_tokenizer`

tmm1:


scripts/finetune.py
```diff
@@ -178,9 +178,8 @@ def train(
     setup_wandb_env_vars(cfg)
 
     # load the tokenizer first
-    tokenizer_config = cfg.tokenizer_config or cfg.base_model_config
-    LOG.info(f"loading tokenizer... {tokenizer_config}")
-    tokenizer = load_tokenizer(tokenizer_config, cfg.tokenizer_type, cfg)
+    LOG.info(f"loading tokenizer... {cfg.tokenizer_config or cfg.base_model_config}")
```
tmm1, line 181: maybe `normalize_config` should set `cfg.tokenizer_config` so we don't need this `or` here?

winglian (approved):

PR #381, opened 2023-08-13
standardize attn hijack patches

tmm1:
all three monkeypatch files now follow the same format and comment structure

you can diff the implementations to see whats different, using `diff -uNdr` or `vim -d`

### DIFF
--- a/src/axolotl/monkeypatch/llama_attn_hijack_flash.py
+++ b/src/axolotl/monkeypatch/llama_attn_hijack_flash.py
@@ -149,7 +149,7 @@ def flashattn_forward(
         # only on first autoregressive step q,k,v have same seqlen
         is_causal = past_key_value is not None
 
-    if self.training and attention_mask.shape[0] == 1:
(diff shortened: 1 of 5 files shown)


winglian:
I've tested the flash-attn variant, but not xformers or sdp yet. That one looked fine

ehartford:
This is really nice.  Thank you so much.

ehartford:
I was curious did we just switch from flash attention 1 to flash attention 2, or do we support either one?  (some people use flash attention 1 because it's more stable, they are still fixing bugs in flash attention 2)

tmm1:
Both versions were supported before and after this PR, but I don't know if anyone has tested with v1 and these changes. The ImportError fallback at the top is what tries to load the older symbols (`unpadded` vs `varlen`)

I think flash-attn v2 is stable now, and with prebuilt wheels its easy to install. FWIW the entire xformers test suite now passes with flash-attn 2.0.8

PR #384, opened 2023-08-13
Fix crash when running without CUDA

cg123:
`log_gpu_memory_usage` was failing hard when running with `CUDA_VISIBLE_DEVICES=""`, preventing doing things like merging adapters on the CPU. This simply skips the GPU memory logging when cuda is not available.


winglian (approved):
thanks!

PR #388, opened 2023-08-14
Added "epoch" evaluation_strategy

flotos:
Enable running eval on epoch count instead of steps

### DIFF
--- a/src/axolotl/utils/trainer.py
+++ b/src/axolotl/utils/trainer.py
@@ -451,6 +451,15 @@ def setup_trainer(cfg, train_dataset, eval_dataset, model, tokenizer, total_num_
             "sample_packing_efficiency"
         ] = cfg.sample_packing_eff_est
 
+    if cfg.val_set_size == 0:
+        evaluation_strategy = "no"
+    elif cfg.eval_steps < 1:
+        # eval every epoch
+        evaluation_strategy = "epoch"
+    else:
+        # eval every eval_steps steps
+        evaluation_strategy = "steps"
+
     training_args = AxolotlTrainingArguments(  # pylint: disable=unexpected-keyword-arg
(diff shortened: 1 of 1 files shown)


NanoCode012:
Hey! I would prefer extracting this into its own section/variable than doing two ifs like that for readability.

teknium1:
Hello this is not working for me

flotos:
The current parameter is a bit weird (hence why they planned a refactor I think) because it will save every epoch if the number is below 1, and will ignore the number you used. (it won't save every 0.2 epoch if that's the number you used for example).
Did it not work even after 1 epoch?

ehartford:
I would be happy to use a 0.2 epoch setting

flotos:
According to HF trainer this can't be done with basic Trainers arguments : https://huggingface.co/docs/transformers/main_classes/trainer#transformers.TrainingArguments.evaluation_strategy

So we would need to calculate an `eval_step = (trainDatasetSize / batch_size) * cfg.eval_step` if we are using an epoch strategy.
However I don't know how micro_batch_size work and if it would affect the above calculation, or if other parameters too.

NanoCode012:
> According to HF trainer this can't be done with basic Trainers arguments : https://huggingface.co/docs/transformers/main_classes/trainer#transformers.TrainingArguments.evaluation_strategy
> 
> So we would need to calculate an `eval_step = (trainDatasetSize / batch_size) * cfg.eval_step` if we are using an epoch strategy. However I don't know how micro_batch_size work and if it would affect the above calculation, or if other parameters too.

Hello, I think this may not be true. This PR has a bit weird logic as said and there is a follow up PR addressing this #403 .

> eval_steps: If smaller than 1, will be interpreted as ratio of total training steps.
https://huggingface.co/docs/transformers/main_classes/trainer#transformers.TrainingArguments.eval_steps

This may be what @ehartford wanted.

PR #393, opened 2023-08-15
recast loralayer, norm, lmhead + embed token weights per original qlora

winglian:
see https://github.com/artidoro/qlora/blob/main/qlora.py#L396-L405

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,5 +29,8 @@ setup(
             "flash-attn",
             "deepspeed",
         ],
+        "peft": [
+            "peft @ git+https://github.com/huggingface/peft.git",
+        ],
     },
 )
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -24,7 +24,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          pip install -e .
+          pip install -e .[peft]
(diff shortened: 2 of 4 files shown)


NanoCode012:
Should this be for qlora only or lora too and difference from the one near kbit?

winglian:
I think both, since lora and lora only differ really in having a quantized frozen base model.

setup.py
```diff
@@ -29,5 +29,8 @@
             "flash-attn",
             "deepspeed",
         ],
+        "peft": [
+            "peft @ git+https://github.com/huggingface/peft.git",
+        ],
```
winglian, line 34: we do it this way b/c we don't install Peft with setup.py as the gptq variant has a different Peft version pinned.

NanoCode012: Should we just add the pip install line to the `test.yml` instead of adding it here?

winglian: I really want to deprecate GPTQ support 😭 , then this wouldn't be an issue

NanoCode012: I'm wondering, how many people actually do use it?

mhenrichsen: Time to do another poll on github maybe?

tmm1 (approved):

PR #395, opened 2023-08-15
better handling of empty input ids when tokenizing

### LINKED ISSUE #314: [Bug] IndexError: list index out of range in falcon config

winglian:
fixes #314 , supercedes #366

### DIFF
--- a/src/axolotl/prompt_tokenizers.py
+++ b/src/axolotl/prompt_tokenizers.py
@@ -74,7 +74,7 @@ class PromptTokenizingStrategy(abc.ABC):
             padding=False,
             return_tensors=None,
         )
-        if not len(result["input_ids"]):
+        if len(result["input_ids"]) == 0:
             LOG.warning("Tokenizer result is empty. You may want to audit your dataset")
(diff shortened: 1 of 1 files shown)


NanoCode012:
This would fix the issue linked. But, I think it's a good cause to error and give a reason (or do a validation first). For example, why would the output ever be empty?

winglian:
maybe a warning and continue?

NanoCode012 (approved):
sounds good!

PR #399, opened 2023-08-15
WIP setup model_init to prepare for wandb sweeps

winglian:
according to this we need to setup a model_init "factory" so it starts with a fresh model on each sweep run.

https://wandb.ai/matt24/vit-snacks-sweeps/reports/Hyperparameter-Search-for-HuggingFace-Transformer-Models--VmlldzoyMTUxNTg0

### DIFF
--- a/src/axolotl/utils/trainer.py
+++ b/src/axolotl/utils/trainer.py
@@ -364,7 +364,7 @@ def setup_fsdp_envs(cfg):
 
 def model_init_factory(tokenizer, cfg):
     def model_init(tokenizer, cfg):
-        model, peft_config = load_model(cfg, tokenizer)
+        model, _ = load_model(cfg, tokenizer)
         model.config.use_cache = False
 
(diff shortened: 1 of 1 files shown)


src/axolotl/utils/trainer.py
```diff
@@ -617,8 +632,10 @@ def setup_trainer(cfg, train_dataset, eval_dataset, model, tokenizer, total_num_
         if cfg.lr_scheduler == "one_cycle" and (cfg.fsdp or cfg.adapter == "qlora")
         else AxolotlTrainer
     )
+
     trainer = trainer_cls(
-        model=model,
+        # model=model,
+        model_init=model_init_factory(tokenizer, cfg),
```
NanoCode012, line 638: Seem like they do not invoke the function

```
model_init=model_init,
```

Is that why `partial` is used?

winglian:
Closing. Spoke with @morganmcg1 from wandb and he says this isn't necessary

PR #400, opened 2023-08-15
add sync_model_states parameter to fix resume from checkpoint with fsdp

winglian:


NanoCode012:
Nice! I never got this to work for resume before.

One note: this adds a slight overhead. Perhaps, we only need to check for this IF we need to resume training then we error ?

```
if cfg.fsdp and cfg.resume and not has_attr(cfg.fsdp_config, 'sync_module..'):
   raise ValueError("Please set ... to true")
```

or, we can just force it to True in that case.

https://github.com/pytorch/pytorch/blob/fd214aa8be7160cba5c6efec7a1b06c9c33b37a4/torch/distributed/fsdp/fully_sharded_data_parallel.py#L348-L354

winglian:
@NanoCode012 So I think it has something to do with the weights too when you save them. This is still a WIP as I need to test it, but I've tried it with these settings to do the resume and it didn't work, but a hunch says I need it when training too.

NanoCode012:
Interesting it says here that it's for loading checkpoint to broadcast from rank 0 to others. Do we need to change model loading script I wonder?

https://github.com/pytorch/pytorch/blob/fd214aa8be7160cba5c6efec7a1b06c9c33b37a4/torch/distributed/fsdp/api.py#L296-L321

winglian:
Interesting, unfortunately all the load from checkpoint logic is baked into the HF trainer iirc.

this part, right?
```
        >>> if dist.get_rank() == 0:
        >>>     # Load checkpoint only on rank 0 to avoid memory redundancy
        >>>     state_dict = torch.load("my_checkpoint.pt")
        >>>     model.load_state_dict(state_dict)
```

winglian:
here are some notes for tracking what I've found https://github.com/huggingface/transformers/issues/25100#issuecomment-1669712422

winglian:
alright, fixed for unpacked resume from FSDP checkpoints. packing needs some other fix. see #406

winglian:
should we add a warning that resume functionality with packing and FSDP doesn't work currently? (when the trainer is tarted with fsdp and sample_packing) This way the user is aware before they start training and it isn't a surprise.

NanoCode012:
> should we add a warning that resume functionality with packing and FSDP doesn't work currently? (when the trainer is tarted with fsdp and sample_packing) This way the user is aware before they start training and it isn't a surprise.

Sounds good!

NanoCode012 (approved):

PR #401, opened 2023-08-15
update docs for tokenizer_legacy

winglian:


NanoCode012:
We should clarify the default is True?

NanoCode012 (approved):

tongyx361:
> We should clarify the default is True?

May I ask why you default `tokenizer_legacy` as True?

NanoCode012:
@tongyx361 , hey! This was from a year ago. If I recall correctly, it was that the new behavior was not correct/created some issues esp for llama. It's very likely they fixed it and this should be reverted. If you would like, please create a new Issue on this, so we can track this better.

PR #403, opened 2023-08-15
fix eval steps and strategy

winglian:
@teknium1 I think this is the correct logic, since the trainer automatically handles when eval_steps < 1


NanoCode012:
This logic makes more sense. However, should we just migrate to allow passing the evaluation_strategy and save_strategy instead to start getting parity with HF Trainer?

PR #404, opened 2023-08-15
Fix(config): Update handling of deepspeed config

NanoCode012:
As the ds_config.json has moved, I think it's a good chance to refactor the code a bit.

- Validation should be moved to the `validate_config` function instead.
- Docs are improved to reflect it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -568,7 +568,7 @@ fsdp_config:
   fsdp_transformer_layer_cls_to_wrap: LlamaDecoderLayer
 ```
 
-- llama Deepspeed: append `ACCELERATE_USE_DEEPSPEED=true` in front of finetune command and set the following:
+- llama Deepspeed
 ```yaml
 deepspeed: # path to config
 ```
--- a/README.md
+++ b/README.md
@@ -517,7 +517,7 @@ tokens:
 fsdp:
 fsdp_config:
 
-# Deepspeed
+# Deepspeed config path
 deepspeed:
 
 # Path to torch distx for optim 'adamw_anyprecision'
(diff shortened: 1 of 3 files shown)


src/axolotl/utils/config.py
```diff
@@ -186,6 +186,16 @@ def validate_config(cfg):
             "sample_packing not compatible with xformers_attention. Use flash_attention"
         )
 
+    if (
+        os.environ.get("ACCELERATE_USE_DEEPSPEED") == "true"
```
winglian: rather than seeing if it is set, perhaps we automatically set this for them if `--deepspeed` is passed in. Similar to how I did it for FSDP?

README.md
```diff
@@ -568,7 +568,10 @@ fsdp_config:
   fsdp_transformer_layer_cls_to_wrap: LlamaDecoderLayer
 ```
 
-- llama Deepspeed: append `ACCELERATE_USE_DEEPSPEED=true` in front of finetune command
+- llama Deepspeed: append `ACCELERATE_USE_DEEPSPEED=true` in front of finetune command and set the following:
```
```suggestion
- llama Deepspeed: prepend `ACCELERATE_USE_DEEPSPEED=true` in front of finetune command and set the following:
```

src/axolotl/utils/config.py
NanoCode012: Sure! That sounds good. I have done so

README.md
NanoCode012: Resolved due to setting this automatically

winglian (approved):

PR #405, opened 2023-08-15
Ax art

winglian:


scripts/finetune.py
```diff
@@ -33,10 +33,23 @@
...
+    ascii_art = """
+                           dP            dP   dP 
+                           88            88   88 
+.d8888b. dP.  .dP .d8888b. 88 .d8888b. d8888P 88 
+88'  `88  `8bd8'  88'  `88 88 88'  `88   88   88 
+88.  .88  .d88b.  88.  .88 88 88.  .88   88   88 
+`88888P8 dP'  `dP `88888P' dP `88888P'   dP   dP 
+"""
+
+    if os.environ.get("LOCAL_RANK", "0") == 0:
```
NanoCode012: use `is_main`

scripts/finetune.py
```diff
@@ -146,6 +159,7 @@ def train(
     prepare_ds_only: bool = False,
     **kwargs,
 ):
+    print_axolotl_text_art()
```
NanoCode012, line 163: why not print at the start of finetune, not only inference :)

NanoCode012: sorry, the thing was collapsed and didn't see

PR #422, opened 2023-08-16
fix orca prompts

winglian:


src/axolotl/prompt_strategies/alpaca_w_system.py
```diff
@@ -92,8 +92,9 @@ class OpenOrcaSystemDataPrompter(SystemDataPrompter):
     def match_prompt_style(self):
         # pylint: disable=duplicate-code
         if self.prompt_style == PromptStyle.INSTRUCT.value:
-            self.turn_format = "### User:\n{instruction}\n\n### Additional Context:\n{input}\n\n### Assistant:\n"
-            self.turn_no_input_format = "### User:\n{instruction}\n\n### Assistant:\n"
+            self.turn_format = "### Human:\n{instruction}\n### Additional Context:\n{input}\n### Assistant:\n"
```
NanoCode012, line 95: Is this following the Orca paper's prompt? I do not see where "Additional Context" was said.

winglian: there wasn't one, so in most cases this won't get used, but wanted to include some variant that at least accounted for it.

NanoCode012 (approved):

Issue #438, opened 2023-08-19
Tokenizer issue with Cerebras' BTLM 3B


AlpinDale:
### Please check that this issue hasn't been reported before.

- [X] I searched previous [Bug Reports](https://github.com/OpenAccess-AI-Collective/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

Attempting to tokenize a `oasst` style dataset with BTLM's `GPT2Tokenizer` results in the following warning being displayed during tokenization:

```
[WARNING] [axolotl._tokenize:78] [PID:3175928] Tokenizer result is empty. You may want to audit your dataset.
```

Ignoring this will lead to:
```sh
Traceback (most recent call last):                                                                                                                                                                │
  File "/home/alpin/miniconda3/envs/axolotl/lib/python3.10/site-packages/multiprocess/pool.py", line 125, in worker                                                                               │
    result = (True, func(*args, **kwds))                                                                                                                                                          │
  File "/home/alpin/miniconda3/envs/axolotl/lib/python3.10/site-packages/datasets/utils/py_utils.py", line 1347, in _write_generator_to_queue                                                     │
    for i, result in enumerate(func(**kwargs)):                                                                                                                                                   │
  File "/home/alpin/miniconda3/envs/axolotl/lib/python3.10/site-packages/datasets/arrow_dataset.py", line 3450, in _map_single                                                                    │
    example = apply_function_on_filtered_inputs(example, i, offset=offset)                                                                                                                        │
  File "/home/alpin/miniconda3/envs/axolotl/lib/python3.10/site-packages/datasets/arrow_dataset.py", line 3353, in apply_function_on_filtered_inputs                                              │
    processed_inputs = function(*fn_args, *additional_args, **fn_kwargs)                                                                                                                          │
  File "/home/alpin/axolotl/src/axolotl/prompt_tokenizers.py", line 125, in tokenize_prompt                                                                                                       │
    tokenized_res_prompt = self._tokenize(                                                                                                                                                        │
  File "/home/alpin/axolotl/src/axolotl/prompt_tokenizers.py", line 88, in _tokenize                                                                                                              │
    if result["input_ids"][0] == self.tokenizer.bos_token_id and strip_bos_token:                                                                                                                 │
IndexError: list index out of range
```

Along with a thousand more lines of error.

### Current behaviour

Apparently works with Alpaca format.

### Steps to reproduce

Add this to the config YAML:

```yaml
[...]

PR #442, opened 2023-08-20
gracefully handle empty input

winglian:


winglian:
fixes #438

src/axolotl/prompt_tokenizers.py
```diff
@@ -85,7 +85,11 @@ def _tokenize(self, prompt: str, add_eos_token=True, strip_bos_token=False):
             result["input_ids"].append(self.tokenizer.eos_token_id)
             result["attention_mask"].append(1)
 
-        if result["input_ids"][0] == self.tokenizer.bos_token_id and strip_bos_token:
+        if (
+            len(result["input_ids"]) > 0
```
NanoCode012, line 89: Could we also add a LOG.info / warn if length is 0? Same like a previous PR?

winglian: we already warn about in line 78

NanoCode012, line 89: I guess if we wanted to reuse variables, we can optionally extract this to a variable and reuse the variable instead. This PR is good enough though.

NanoCode012 (approved):

PR #446, opened 2023-08-20
feat: add Metharme prompt strategy

TearGosling:
This is my first time contributing to someone else's project, forgive me if I've missed anything. This PR adds a prompt strategy for our **Metharme** series of instructional models. More specifically, it follows a JSONL format with keys `prompt` and `generation` - our special, ChatML-like tokens are treated normally for the time being. However, we do a special trick to reduce model generation length (people have said that our models have outputs which are *too* lengthy) by adding additional EOS tokens (3 in total by default) to the end of `input_ids` rather than just one.

### DIFF
--- a/src/axolotl/prompt_strategies/metharme.py
+++ b/src/axolotl/prompt_strategies/metharme.py
@@ -39,12 +39,9 @@ class MetharmePromptTokenizingStrategy(InstructionPromptTokenizingStrategy):
         if result["input_ids"][-1] == self.tokenizer.eos_token_id:
             num_eos_tokens -= 1
 
-        if num_eos_tokens > 0 and add_eos_token:
+        if (num_eos_tokens > 0 and add_eos_token and len(result["input_ids"]) > 0):
             for _ in range(num_eos_tokens):
-                if (
-                    len(result["input_ids"]) > 0
-                    and len(result["input_ids"]) < self.sequence_len
-                ):
+                if len(result["input_ids"]) < self.sequence_len:
                     result["input_ids"].append(self.tokenizer.eos_token_id)
                     result["attention_mask"].append(1)
 
--- a/src/axolotl/prompt_strategies/metharme.py
+++ b/src/axolotl/prompt_strategies/metharme.py
@@ -3,9 +3,7 @@
 import logging
 from typing import Tuple
 
-from axolotl.prompt_tokenizers import (
-    InstructionPromptTokenizingStrategy,
-)
+from axolotl.prompt_tokenizers import InstructionPromptTokenizingStrategy
 from axolotl.prompters import AlpacaPrompter
 
 LOG = logging.getLogger("axolotl")
@@ -19,13 +17,15 @@ class MetharmePromptTokenizingStrategy(InstructionPromptTokenizingStrategy):
     """
 
     def parse_instruction_fields(self, prompt) -> Tuple[str, str, str]:
-        return (
-            prompt["prompt"],
-            "",
-            prompt["generation"]
-        )
+        return (prompt["prompt"], "", prompt["generation"])
 
-    def _tokenize(self, prompt: str, add_eos_token: bool = True, strip_bos_token: bool = False, num_eos_tokens: int = 3):
+    def _tokenize(
+        self,
+        prompt: str,
+        add_eos_token: bool = True,
+        strip_bos_token: bool = False,
+        num_eos_tokens: int = 3,
+    ):
         result = self.tokenizer(
             prompt,
             truncation=True,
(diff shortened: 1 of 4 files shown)


src/axolotl/prompt_strategies/metharme.py
```diff
@@ -0,0 +1,77 @@
...
+        if len(result["input_ids"]) == 0:
+            LOG.warning("Tokenizer result is empty. You may want to audit your dataset")
+        # If there's already an EOS token there, subtract from the number added
+        if result["input_ids"][-1] == self.tokenizer.eos_token_id:
+            num_eos_tokens -= 1
+
+        if num_eos_tokens > 0 and add_eos_token:
+            for _ in range(num_eos_tokens):
+                if (
+                    len(result["input_ids"]) > 0
```
NanoCode012: I think you can move this to line 42 instead in the outer if block.

src/axolotl/prompt_strategies/metharme.py
```diff
@@ -0,0 +1,77 @@
...
+    """
+
+    system_prompt = ""
+    system_no_input_prompt = ""
+    system_format = ""
+    turn_format = "{instruction}"
+    turn_no_input_format = "{instruction}"
+
+    def __init__(self, *args, **kwargs):
+        pass
```
NanoCode012, line 70: Might not need this init function.

NanoCode012 (commented):
Hello there! Thanks for the PR! I just added some minor comments. I think the PR looks very nice.

I really find the concept interesting using 3 EOS.

src/axolotl/prompt_strategies/metharme.py
```diff
@@ -0,0 +1,77 @@
...
+class MetharmePrompter(AlpacaPrompter):
+    """
+    Prompter for the Metharme models.
+    """
+
+    system_prompt = ""
+    system_no_input_prompt = ""
+    system_format = ""
+    turn_format = "{instruction}"
+    turn_no_input_format = "{instruction}"
```
NanoCode012, line 67: May I ask if there's any samples of formatting you have for Metharme? I think this part might not look right. An example of alpaca is here: https://github.com/TearGosling/axolotl/blob/eba46c261c3d57e6f10dbb2b59daa1225f372781/src/axolotl/prompters.py#L22-L45

src/axolotl/prompt_strategies/metharme.py
TearGosling: Nice catch, thank you!

src/axolotl/prompt_strategies/metharme.py
TearGosling: I see what you mean. The reason we don't provide a system format here is that the system prompt is already provided within the `prompt` field. Metharme uses a ChatML-like system, like so:
```{"prompt": "<|system|>Goal: Assistant\nObjective: Follow the user's instructions<|user|>A train travels from Town A to Town B at a speed of 60 miles per hour. The return trip from Town B back to Town A is made at a speed of 40 miles per hour. What is the average speed for the entire round trip?<|model|>", "generation": "The average speed for the entire round trip is 48 miles per hour.\n\nHere's the reasoning behind this:\n\n1. To find the average speed for the entire round trip, we need to consider the harmonic mean of the two speeds, not the arithmetic mean.\n2. The formula for the harmonic mean of two numbers, x and y, is H = 2xy / (x + y).\n3. Plugging in the values for the speeds during the two trips, we get: H = 2 * 60 * 40 / (60 + 40) = 4800 / 100 = 48.\n4. Therefore, the average speed for the entire round trip is 48 miles per hour.\n\nIt's important to note that simply averaging the two speeds (60 + 40) / 2 = 50 would not give the correct answer because the time spent traveling at each speed is different due to the varying distances covered."}```

Admittedly, I am aware there is a ChatML-like system already implemented within Axolotl but due to time constraints, I think this is easier for now and works just as well.

File: src/axolotl/prompt_strategies/metharme.py (hunk 2 above)
Comment (TearGosling) [reply src/axolotl/prompt_strategies/metharme.py:70]: See comment on line 65 of `metharme.py`

File: src/axolotl/prompt_strategies/metharme.py (hunk 3 above)
Comment (TearGosling) [reply src/axolotl/prompt_strategies/metharme.py:67]: Additionally, this is also why we have an `__init__` function which only does a pass: the prompt is already built.

Comment (NanoCode012) [reply src/axolotl/prompt_strategies/metharme.py:67]: Ok, this sounds good! Thanks for the clarification.

**Review by NanoCode012 — APPROVED** [review]
Hey, this PR is good to go. Please let me know if you need help with the linting. You can either see the fix via the Github failed Actions or run using 

```bash
pip3 install -r requirements-dev.txt -r requirements-tests.txt
pre-commit install
```

NanoCode012:
Thank you for fixing lint. There's only one minor warning. You can disable it using `# pylint: disable=super-init-not-called` since it's ok in your case.

TearGosling:
👍

AlpinDale:
Is this ready for merging @NanoCode012?

NanoCode012:
I really apologize with the hassle of lint. It appears that there's a space issue. If you'd like, I can fix whatever's left then merge it.

TearGosling:
That would be good, thank you so much

NanoCode012:
Thank you for the PR!

PR #460, opened 2023-08-23
merge lora on train completion

winglian:
for almost all lora use cases, users are going to want to merge the lora with the base model. let's make it easier. Still needs to be tested.

### DIFF
--- a/scripts/finetune.py
+++ b/scripts/finetune.py
@@ -158,7 +158,8 @@ def check_not_in(list1: List[str], list2: Union[Dict[str, Any], List[str]]) -> b
 def merge_lora(model, tokenizer, cfg):
     LOG.info("running merge of LoRA with base model")
     model = model.merge_and_unload()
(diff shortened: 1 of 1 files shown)


scripts/finetune.py
```diff
@@ -155,6 +155,22 @@ def check_not_in(list1: List[str], list2: Union[Dict[str, Any], List[str]]) -> b
     return not any(el in list2 for el in list1)
 
 
+def merge_lora(model, tokenizer, cfg):
+    LOG.info("running merge of LoRA with base model")
+    model = model.merge_and_unload()
+    model.to(dtype=torch.float16)
```
NanoCode012: We may need to set dtype based on config?

NanoCode012 (approved):

winglian:
so this won't work as expected, since the model is loaded in 4/8 bit and can't be merged with the lora. We might have to reload the model w full weights before merging

PR #462, opened 2023-08-23
clear the dataset cache on explicit pre-tokenization

winglian:
TheBloke reported that when experimenting with custom prompt strategies, the transformed dataset gets cached and has to be manually cleared. This will clear it automatically when `--prepare_ds_only` option is set.

### DIFF
--- a/src/axolotl/cli/__init__.py
+++ b/src/axolotl/cli/__init__.py
@@ -219,7 +219,11 @@ def load_datasets(
 ) -> TrainDatasetMeta:
     tokenizer = load_tokenizer(cfg)
 
-    train_dataset, eval_dataset, total_num_steps = prepare_dataset(cfg, tokenizer)
+    skip_cache = cli_args.prepare_ds_only
+
+    train_dataset, eval_dataset, total_num_steps = prepare_dataset(
+        cfg, tokenizer, skip_cache=skip_cache
+    )
 
     if cli_args.debug or cfg.debug:
         LOG.info("check_dataset_labels...")
--- a/src/axolotl/cli/prepare_ds.py
+++ b/src/axolotl/cli/prepare_ds.py
@@ -1,0 +1,33 @@
+"""
+CLI to prepare the datasets for training
+"""
+from pathlib import Path
+
+import fire
+import transformers
+
+from axolotl.cli import (
+    check_accelerate_default_config,
+    load_cfg,
+    load_datasets,
+    print_axolotl_text_art,
+)
+from axolotl.common.cli import TrainerCliArgs
+
+
+def do_cli(config: Path = Path("examples/"), **kwargs):
+    # pylint: disable=duplicate-code
+    print_axolotl_text_art()
+    check_accelerate_default_config()
+    parser = transformers.HfArgumentParser((TrainerCliArgs))
+    parsed_cli_args, _ = parser.parse_args_into_dataclasses(
+        return_remaining_strings=True
+    )
+    parsed_cli_args.prepare_ds_only = True
+    parsed_cfg = load_cfg(config, **kwargs)
+
+    load_datasets(cfg=parsed_cfg, cli_args=parsed_cli_args)
+
+
+if __name__ == "__main__":
+    fire.Fire(do_cli)
--- a/src/axolotl/utils/data.py
+++ b/src/axolotl/utils/data.py
@@ -56,11 +56,11 @@ def md5(to_hash: str, encoding: str = "utf-8") -> str:
         return hashlib.md5(to_hash.encode(encoding)).hexdigest()  # nosec
 
 
-def prepare_dataset(cfg, tokenizer):
+def prepare_dataset(cfg, tokenizer, skip_cache=False):
     if not cfg.pretraining_dataset:
         with zero_first(is_main_process()):
             train_dataset, eval_dataset = load_prepare_datasets(
-                tokenizer, cfg, DEFAULT_DATASET_PREPARED_PATH
+                tokenizer, cfg, DEFAULT_DATASET_PREPARED_PATH, skip_cache=skip_cache
             )
     else:
         train_dataset = load_pretraining_dataset(
@@ -89,7 +89,7 @@ def prepare_dataset(cfg, tokenizer):
 
 
 def load_tokenized_prepared_datasets(
-    tokenizer, cfg, default_dataset_prepared_path
+    tokenizer, cfg, default_dataset_prepared_path, skip_cache=False
 ) -> DatasetDict:
     tokenizer_name = tokenizer.__class__.__name__
     ds_hash = str(
@@ -150,7 +150,7 @@ def load_tokenized_prepared_datasets(
 
         # pylint: disable=invalid-name
         for d in for_d_in_datasets(cfg.datasets):
-            ds: Union[Dataset, DatasetDict] = None
+            ds: Union[Dataset, DatasetDict]
             ds_from_hub = False
             try:
                 load_dataset(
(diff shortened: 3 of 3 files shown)


NanoCode012:
Hm, would it be better to make it to not load from cache (or clear all cache) on load? I think we had the discussion about caching pre-tokenization? Can we take this time to base it off the lack of `dataset_prepared_path`?

NanoCode012:
Hm, would it be better to make it to not load from cache (or clear all cache) on load? I think we had the discussion about caching pre-tokenization? Can we take this time to base it off the lack of `dataset_prepared_path`?

winglian:
> Hm, would it be better to make it to not load from cache (or clear all cache) on load? I think we had the discussion about caching pre-tokenization? Can we take this time to base it off the lack of `dataset_prepared_path`?

In this case, `dataset_prepared_path` wouldn't matter. The caching is the native datasets caching that caches transforms on the dataset.

NanoCode012:
I'm not clear if using `prepare_ds_only` to indicate no cache makes sense. Should we always clear instead?

TheBloke:
Thanks Wing!

> Could we, `if not cfg.dataset_prepared_path: clearcache()`? I'm not clear if using `prepare_ds_only` to indicate no cache makes sense.

Yeah that sounds good. Yesterday I was doing training and I left out `dataset_prepared_path` thinking it might solve the problem, so it's logical to link those IMHO.

I know they're different caching processes, but it's logical to think that all caching will either happen or not happen together.

Also, leaving out `dataset_prepared_path` doesn't stop it saving its prepared dataset - it just creates `last_run_prepared` instead.

In total I would love to see:
1. A way to disable it saving prepared runs - either `dataset_prepared_path` is blank, or else a new boolean. I prefer the former; I didn't like that it added `last_prepared_run` in the current folder, especially as it's not mentioned in `.gitignore`
2. Some way to disable the HF dataset caching as Wing mentioned - I think that's best linked to 1, or it could be a new boolean perhaps.

Thanks very much

TheBloke:
To add: I do agree that just linking it to `prepared_ds_only` isn't quite enough.  For example if I aborted training, or I wanted to re-run training with tweaked prompt_strategy code, I would still need to manually delete both the `datasets` and `last_run_prepared` entries

Ideally I would like a solution where I don't have to remember to manually delete the prepared and cached datasets, both when I've just done `prepare_ds_only` (which I do only to test my prompt_strategy code with `--debug`), and when a previous run has been done on this same dataset JSON.

winglian:
@TheBloke this new PR will clear the cache whenever prepare_ds is called. it also adds a new cli `python -m axolotl.cli.prepare_ds examples/your.yml`  (you can still use the old cli too, and it will use the new behavior)

NanoCode012:
Hello caseus, what if we have a new cfg as @TheBloke mentioned? `dataset_use_cache: false`? This would be very explicit.

flexchar:
I'd vote for a much more explicit config parameter. `prepare_ds_only` implies absolutely nothing about cache in my world.

TheBloke:
Yeah I agree that a specific config param would be best.

While `prepare_ds_only` has been brought up again, I've been meaning to mention for a while that this param is no longer functional, in the sense that it no longer only prepares the dataset.  

Running `python3 -m axolotl.cli.train config.yaml --prepare_ds_only --debug` will now proceed on to training, necessitating a Control-C to abort this manually.

(I've not checked to see if `--prepare_ds_only` still does something useful, like interacting with `--debug`, or whether it's now completely non-functional.)

Here's a log of that happening - you can see the final `example_labels` message, followed by it then starting to train

```
[2023-11-24 22:41:47,011] [INFO] [axolotl.check_example_labels:36] [PID:17755] [RANK:0]



[2023-11-24 22:41:47,011] [INFO] [axolotl.scripts.load_datasets:330] [PID:17755] [RANK:0] printing prompters...
[2023-11-24 22:41:47,011] [INFO] [axolotl.scripts.load_datasets:332] [PID:17755] [RANK:0] Pre-tokenized or custom dataset types are unsupported for logging
[2023-11-24 22:41:47,024] [DEBUG] [axolotl.train.log:60] [PID:17755] [RANK:0] loading tokenizer... Intel/neural-chat-7b-v3-1
[2023-11-24 22:41:47,338] [DEBUG] [axolotl.load_tokenizer:100] [PID:17755] [RANK:0] EOS: 2 / </s>
[2023-11-24 22:41:47,338] [DEBUG] [axolotl.load_tokenizer:101] [PID:17755] [RANK:0] BOS: 1 / <s>
[2023-11-24 22:41:47,338] [DEBUG] [axolotl.load_tokenizer:102] [PID:17755] [RANK:0] PAD: 2 / </s>
[2023-11-24 22:41:47,338] [DEBUG] [axolotl.load_tokenizer:103] [PID:17755] [RANK:0] UNK: 0 / <unk>
[2023-11-24 22:41:47,338] [DEBUG] [axolotl.train.log:60] [PID:17755] [RANK:0] loading model
Downloading shards: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 2/2 [00:00<00:00, 14.39it/s]
Loading checkpoint shards:   0%|                                                                                                                                                                                                                                                                                                                     | 0/2 [00:00<?, ?it/s]Loading checkpoint shards:   0%|                                                                                                                                                                                                                                                                                                                     | 0/2 [00:14<?, ?it/s]
  File "/root/miniconda3/envs/py3.10/lib/python3.10/runpy.py", line 196, in _run_module_as_main
(CONTROL-C)
```

NanoCode012:
@TheBloke , regarding your second point, the cli has been updated to `axolotl.cli.preprocess`  https://github.com/OpenAccess-AI-Collective/axolotl/tree/main#preprocess-dataset

Concerning this PR, you can now leave `dataset_prepared_path:` as empty, and it won't load from cache anymore, but will still save to cache in case you need it later. Does that solve the problem? We wish to make this clearer.

TheBloke:
@NanoCode012 ahhh I see! I missed that change. That explains it then. My bad.  I tried `axolotl.cli.preprocess` just now and it worked great, thank you!

Could I suggest that if `--prepare_ds_only` is not going to do anything, it should generate an error if used on `axolotl.cli.train`?

And that's great to know regarding `dataset_prepared_path`.  That behaviour sounds fine.  Thanks for the info!

By the way, it's slightly off-topic, but as we're talking about preparing datasets and rather than open an Issue specifically for it:  do you know why Wing implemented this change in `utils/tokenization.py` https://github.com/OpenAccess-AI-Collective/axolotl/blob/main/src/axolotl/utils/tokenization.py#L34 :

image

Specifically, telling it not to output a space between tokens when `text_only` is used?

It makes the output really hard to read, and means you can't spot the token boundaries.   I know I could just not use `text_only` and see the token boundaries clearly, but the reason I implemented text_only was so I could clearly read the plain text output, which is hard to read in the normal mode - and in text_only I could still see when a word was two or three tokens instead of one.  Now it's much harder to read, and impossible to tell how many tokens are involved.

At first I assumed it must be a bug but then I looked at the commit history and realised it was done deliberately.  I can PR a change to remove it but maybe there's some good reason for it I've not realised?

NanoCode012:
@TheBloke , sorry for late reply. 

> By the way, it's slightly off-topic, but as we're talking about preparing datasets and rather than open an Issue specifically for it: do you know why Wing implemented this change in utils/tokenization.py https://github.com/OpenAccess-AI-Collective/axolotl/blob/main/src/axolotl/utils/tokenization.py#L34 :

I am not aware of why as well. It does make things hard to read. Perhaps, we can move this to a new issue/PR and hopefully ask for a change?

TheBloke:
OK sure, I will open a little PR for it and then if there's some reason, Wing can tell us there.  Maybe it was just a mistake.

Thanks for your help!

NanoCode012:
Closing this as we fixed this with `cli.preprocess`

PR #463, opened 2023-08-23
gather/broadcast the max value of the packing efficiency automatically

winglian:


src/axolotl/utils/distributed.py
```diff
@@ -143,8 +142,35 @@ def reduce_and_broadcast(fn1, fn2):
...
     # and then broadcast it to all ranks
     return compute_and_broadcast(lambda: fn2(gathered_values))
+
+  
+def broadcast_dict(vals: dict):
+    if not is_distributed():
+        return vals
+
+    if is_main_process():
+        data_byte = pickle.dumps(vals)
```
NanoCode012: is it normal to use pickle? I heard that it could be insecure? Edit: My bad, seem like I was on an older commit.

casper-hansen:
This PR makes multi-GPU work on custom datasets. Previously, training got stuck at epoch 1.0 when `micro_batch_size` was 1. However, with this PR, that problem is resolved.

image

PR #466, opened 2023-08-23
Fix(doc): Clarify config

NanoCode012:
Removes the top `configs` directory as it wasn't used anymore.


winglian (approved):
probably need to rebase against main

PR #468, opened 2023-08-24
Add support for GPTQ using native transformers/peft

winglian:


examples/gptq/lora.yml
```diff
@@ -0,0 +1,76 @@
+base_model: TheBloke/Llama-2-7B-GPTQ
```
NanoCode012: this file should maybe be in `examples/llama2/gptq-lora.yml`

requirements.txt
```diff
@@ -1,3 +1,5 @@
+--extra-index-url https://huggingface.github.io/autogptq-index/whl/cu117/
```
NanoCode012: is this cuda version ok?

src/axolotl/utils/models.py
```diff
@@ -147,32 +148,15 @@ def load_model(
...
-            # For backward compatibility
-            from peft import (
-                prepare_model_for_int8_training as prepare_model_for_kbit_training,
-            )
-
     model_kwargs = {}
     if cfg.model_revision:
         model_kwargs["revision"] = cfg.model_revision
+    if cfg.gptq:
+        # TODO we should figure out how read the models config.json first
```
NanoCode012: We can load from config with AutoConfig.from_pretrained?

winglian: yeah, I might tackle this in a refactor. not sure how easily this is handled for cases where it points to a local model path.

Issue #470, opened 2023-08-24
ReLoRA example config


ehartford:
### What piece of documentation is affected?

can there please be an example config for ReLoRA?

### What part(s) of the article would you like to see updated?

can there please be an example config for ReLoRA?

### Additional Information

can there please be an example config for ReLoRA?

### Acknowledgements

- [X] My issue title is concise, descriptive, and in title casing.
- [X] I have searched the existing issues to make sure this feature has not been requested yet.
- [X] I have provided enough information for the maintainers to understand and evaluate this request.

NanoCode012:
Seems like the author of the PR https://github.com/OpenAccess-AI-Collective/axolotl/pull/322#issue-1819539344 provided some sample

ehartford:
that isn't a full config file though.  I don't know what other things are needed or not needed.

ehartford:
and also - a PR comment isn't documentation, it would be good to have it in a real example file that can be executed and functions

NanoCode012:
That's a good point. Perhaps @cg123 might be interested in a PR to add this config since they already have one.

PR #471, opened 2023-08-24
Add example Llama 2 ReLoRA config

cg123:
As per #470, adds an example configuration file that I've used to test locally on a 3090.

### DIFF
--- a/examples/llama-2/relora.yml
+++ b/examples/llama-2/relora.yml
@@ -41,7 +41,7 @@ wandb_log_model:
 gradient_accumulation_steps: 4
 micro_batch_size: 4
 num_epochs: 3
-optimizer: paged_adamw_32bit
+optimizer: adamw_bnb_8bit
 lr_scheduler: cosine
 learning_rate: 0.0002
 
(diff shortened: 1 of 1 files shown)


mhenrichsen:
Looks good. Maybe default to a smaller dataset?

I'm not sure if `paged_adamw_32bit` is the way to go.

Edit:
You also mentioned that you only tested with `adamw_bnb_8bit` in the PR. Has this changed?

cg123:
Ah yes, good point - I have tested it with `paged_adamw_32bit` as well now, and it works fine. Happy to change it if you think there's a better default here though.

As for the smaller dataset: I purposefully chose this one as being on the smaller side of what I would use recommending ReLoRA with, as there need to be enough steps to get a few restarts in before you start to see any benefit. A smaller dataset would definitely be fine for just testing that you can run ReLoRA at all though, if that's more useful?

mhenrichsen:
Sounds good.
I get the idea behind using a larger dataset to see the benefits of ReLoRA. 
Personally I've used smaller datasets to test that it can complete multiple epochs, saves and evaluates without running out of VRAM. Using a larger dataset would make those take longer time.

NanoCode012 (approved):

cg123:
I have been struck down for my hubris. `paged_adamw_32bit` does not, in fact, work fine in all configurations.

Switched to `adamw_bnb_8bit` for the example.

ehartford:
That is great to know!  Thank you for your excellent work!

tmm1 (approved):

Issue #385, opened 2023-08-14
[Bug] Do not override llama pad token if exists


NanoCode012:
https://github.com/OpenAccess-AI-Collective/axolotl/blob/919246fbc1f6d925e792703d8fe989d05e200029/src/axolotl/utils/models.py#L57-L61

```py
        if not (tokenizer.pad_token or cfg.special_tokens.get("pad_token")):
            LOG.warning(
                "LlamaTokenizer does not have a pad token. Setting to LLAMA_DEFAULT_PAD_TOKEN"
            )
            tokenizer.pad_token = LLAMA_DEFAULT_PAD_TOKEN
```

winglian:
@NanoCode012 is this still something we need to fix?

NanoCode012:
This is now fixed with #475 and #477

PR #475, opened 2023-08-24
improve llama pad token handling

winglian:


NanoCode012:
Related to #385 

I'm not sure we should set it to EOS as a fallback.

Secondly, I would say: we should check if the pad token is set already to prevent overriding any previous setting. 
```
config = AutoConfig.from_pretrained
hasattr(config, 'pad_token') 
```
In fact, we should only set it to the default PAD token if the user also does not specify the pad_token within the `special_tokens`

PR #478, opened 2023-08-25
fix types w lora

winglian:
per the discussion in the MoE channel, we should definitely be keeping the norm, lm_head and embed layers as fp32 unless absolutely necessary. This backs out some of the changes, but I don't think we need to have exact parity with the qlora implementation.

### DIFF
--- a/src/axolotl/utils/models.py
+++ b/src/axolotl/utils/models.py
@@ -11,7 +11,6 @@ import bitsandbytes as bnb
 import torch
 import transformers
 from optimum.bettertransformer import BetterTransformer
-from peft.tuners.lora import LoraLayer
 from transformers import (  # noqa: F401
     AutoConfig,
     AutoModelForCausalLM,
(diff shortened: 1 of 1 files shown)


src/axolotl/utils/models.py
```diff
@@ -352,6 +359,16 @@ def load_model(
             model, use_gradient_checkpointing=cfg.gradient_checkpointing
         )
 
+        # LlamaRMSNorm layers are in fp32 after kbit_training, so we need to
+        # convert them back to fp16/bf16 for flash-attn compatibility.
+        if cfg.flash_attention and cfg.is_llama_derived_model:
+            for name, module in model.named_modules():
```
NanoCode012, line 365: Is this supposed to be before or after load_adapter?

winglian: this is where it was before we moved it around. The adapter doesn't actually touch any of these, it just loads a new LoraModel w LoraLayers and freezes these.

PR #479, opened 2023-08-25
Feat(cfg): Add code-llama configs for all sizes

mhenrichsen:
Added configs for all sizes of code-llama.

Only tested 7b qlora.
https://wandb.ai/mhenrichsen/codellama-test/runs/cs2oacp8?workspace=user-mhenrichsen

### DIFF
--- a/examples/code-llama/13b/lora.yml
+++ b/examples/code-llama/13b/lora.yml
@@ -1,7 +1,7 @@
 base_model: codellama/CodeLlama-13b-hf
 base_model_config: codellama/CodeLlama-13b-hf
 model_type: LlamaForCausalLM
-tokenizer_type: LlamaTokenizer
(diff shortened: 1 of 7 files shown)


NanoCode012:
Can you verify that the type is correct? Is it not a separate model arch/type called LlamaCode?

it is fine tuned from llama2 though

mhenrichsen:
It seems to train fine through same architechture as llama-2. I just assumed that they reused the achitechture and trained on code related data.

NanoCode012:
Not sure if tokenizer uses a different class? https://github.com/huggingface/transformers/pull/25740/files#diff-2fa8024f3177ae77ca87d3030c95e4fb1b05b39cd2b016b7a703d076841b3746

NanoCode012 (approved):

PR #480, opened 2023-08-25
support for datasets with multiple names

winglian:
some datasets like https://huggingface.co/datasets/bigcode/commitpackft have 100+ named parts. normally, handling that would require adding multiple blocks like:
```
datasets:
  - path: bigcode/commitpackft
    name: python
    type: ...
  - path: bigcode/commitpackft
    name: ruby
    type: ...
  - path: bigcode/commitpackft
    name: typescript
    type: ...
etc...
```

This PR let's us at least simplify this to 
```
datasets:
  - path: bigcode/commitpackft
    name: 
      - ruby
      - python
      - typescript
    type: ...
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -328,6 +328,15 @@ See [examples](examples) for quick start. It is recommended to duplicate and mod
       name: enron_emails
       type: completion # format from earlier
 
+  # huggingface repo with multiple named configurations/subsets
+  datasets:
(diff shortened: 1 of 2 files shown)


NanoCode012 (commented):
Should we add an example  of this on the docs?

PR #483, opened 2023-08-26
fix: inference did not move the model to the correct device

maximegmd:


winglian (commented):
If this is an inference only issue, we should wrap it in a conditional. Moving it around for training might have unintended consequences.

maximegmd:
That's in the do_inference function, can it ever do anything else if we get to this function?

tmm1 (approved):

A-Raafat:
Could you please do the same in `do_merge_lora` function? it doesn't load in the gpu

PR #485, opened 2023-08-26
fix: finetune model inference needs the dtype fix to work with flash-attn

maximegmd:


src/axolotl/utils/models.py
```diff
@@ -363,16 +364,19 @@ def load_model(
...
+        fix_dtype = True
 
-        # LlamaRMSNorm layers are in fp32 after kbit_training, so we need to
-        # convert them back to fp16/bf16 for flash-attn compatibility.
-        if cfg.flash_attention and cfg.is_llama_derived_model:
-            for name, module in model.named_modules():
-                if "norm" in name:
+    # LlamaRMSNorm layers are in fp32 after kbit_training or full finetune, so we need to
+    # convert them back to fp16/bf16 for flash-attn compatibility.
+    if (fix_dtype or cfg.adapter == "" or cfg.adapter is None) and (
```
winglian: `if (fix_dtype or cfg.adapter == "" or cfg.adapter is None) and (`
could be simplified to 
`if (fix_dtype or not cfg.adapter) and (`

winglian (approved):
lgtm

src/axolotl/utils/models.py
```diff
@@ -355,6 +355,7 @@ def load_model(
             if hasattr(module, "weight"):
                 module.to(torch.float32)
 
+    fix_dtype = False
```
tmm1:
```suggestion
    fix_dtype = not cfg.adapter
```

src/axolotl/utils/models.py
```diff
@@ -363,16 +364,19 @@ def load_model(
...
+        fix_dtype = True
 
-        # LlamaRMSNorm layers are in fp32 after kbit_training, so we need to
-        # convert them back to fp16/bf16 for flash-attn compatibility.
-        if cfg.flash_attention and cfg.is_llama_derived_model:
-            for name, module in model.named_modules():
-                if "norm" in name:
+    # LlamaRMSNorm layers are in fp32 after kbit_training or full finetune, so we need to
+    # convert them back to fp16/bf16 for flash-attn compatibility.
+    if (fix_dtype or not cfg.adapter) and (
```
tmm1:
```suggestion
    if fix_dtype and (
```

src/axolotl/utils/models.py
tmm1: let's also rename this, something like need_fa2_dtype

PR #498, opened 2023-08-28
`pad_to_sequence_len`, for reduced VRAM peak usage due to memory fragmentation

Birch-san:
config option to help with measuring worst-case memory usage.

forces data collator to always output batches that max out your configured context length.

this will give a somewhat optimistic reading, because the amount of fragmentation will be minimal (as all batches will be able to re-use previously-allocated buffers). but it's a lot better than finding out the impact of a long prompt super late in your run.

### DIFF
--- a/src/axolotl/utils/trainer.py
+++ b/src/axolotl/utils/trainer.py
@@ -585,16 +585,14 @@ def setup_trainer(cfg, train_dataset, eval_dataset, model, tokenizer, total_num_
         callbacks.append(SaveBetterTransformerModelCallback)
 
     data_collator_kwargs = {
-        "padding": True, # True/"longest" is the default
+        "padding": True,  # True/"longest" is the default
     }
-    if cfg.collator_pad_to_longest:
-        pass
+    if cfg.pad_to_sequence_len:
+        data_collator_kwargs["pad_to_multiple_of"] = 64 * round(cfg.sequence_len / 64)
     else:
         # A100 is best at 64, while others at 8. Let's use the larger so we don't have to check
         # https://docs.nvidia.com/deeplearning/performance/dl-performance-matrix-multiplication/index.html
         data_collator_kwargs["pad_to_multiple_of"] = 64
-    if cfg.pad_to_worst_case_seq_len:
-        data_collator_kwargs["pad_to_multiple_of"] = cfg.sequence_len
 
     if cfg.is_llama_derived_model and cfg.landmark_attention:
         from axolotl.monkeypatch.llama_landmark_attn import (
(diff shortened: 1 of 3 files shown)


src/axolotl/utils/trainer.py
```diff
@@ -593,6 +593,8 @@ def setup_trainer(cfg, train_dataset, eval_dataset, model, tokenizer, total_num_
         # A100 is best at 64, while others at 8. Let's use the larger so we don't have to check
         # https://docs.nvidia.com/deeplearning/performance/dl-performance-matrix-multiplication/index.html
         data_collator_kwargs["pad_to_multiple_of"] = 64
+    if cfg.pad_to_worst_case_seq_len:
```
tmm1: @winglian @NanoCode012 any thoughts on naming here? Maybe just `pad_inputs` or `pad_to_sequence_len`?

tmm1:
looking at the docs (https://huggingface.co/docs/transformers/main_classes/data_collator#transformers.DataCollatorWithPadding.padding), it seems padding=True and padding="longest" mean the same thing

so i've removed the unnecessary `collator_pad_to_longest`, which is already the default behavior

instead, there's a new `pad_to_sequence_len` which will control padding. FYI, I tried padding="max_length" but it doesn't work correctly when micro_batch_size>1, so using pad_to_multiple_of= is the best solution. i ensured that the padding is still also a multiple of 64 (and 8) to ensure best performance.

tmm1:
also, note to my future self: i tested this with the following patch to transformers trainer to see what shape the inputs are on each step:

```diff
diff --git a/src/transformers/trainer.py b/src/transformers/trainer.py
index 85d4fd5a5..a9e7729ae 100755
--- a/src/transformers/trainer.py
+++ b/src/transformers/trainer.py
@@ -1813,6 +1813,7 @@ class Trainer:
 
             step = -1
             for step, inputs in enumerate(epoch_iterator):
+                print(os.getpid(), step, {k:v.shape for k, v in inputs.items()})
                 total_batched_samples += 1
                 if rng_to_sync:
                     self._load_rng_state(resume_from_checkpoint)
```

before, with `examples/llama-2/lora.yml` (which uses sequence_len=4096 and micro_batch_size=2), i saw iterations of sizes 8091, 8178, 7787, etc

after this patch, the sizes are consistently 8192 each step and the memory usage is much more stable.

tmm1:
> config option to help with measuring worst-case memory usage.

ironically using this option actually reduces memory usage in all my tests. without it, fragmentation is a problem and peak VRAM usage is higher. (this makes sense, because when each step is exactly the same size the same few memory buffers can be re-used every time).

tmm1:
testing `examples/llama-2/lora.yml` again, before patch:

> GPU memory usage while training: 7.309GB (+8.193GB cache, +0.848GB misc)

after patch, and adding `pad_to_sequence_len: true`:

> GPU memory usage while training: 7.310GB (+6.086GB cache, +0.848GB misc)

so peak VRAM usage is more than 2GB lower!

PR #499, opened 2023-08-28
Refactor train cfg cli

winglian:
Splitting out the non-training specific arguments into its own arguments dataclass. 
making the train function only accept pre-parsed config object so it can be called as a standalone option later on

### DIFF
--- a/scripts/finetune.py
+++ b/scripts/finetune.py
@@ -184,9 +184,9 @@ def train(
     LOG.info(f"loading tokenizer... {cfg.tokenizer_config or cfg.base_model_config}")
     tokenizer = load_tokenizer(cfg)
 
-    if (
-        cli_args.shard or cli_args.merge_lora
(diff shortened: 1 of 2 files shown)


scripts/finetune.py
```diff
@@ -61,6 +77,8 @@ def get_multi_line_input() -> Optional[str]:
 
 
 def do_inference(cfg, model, tokenizer, prompter: Optional[str]):
+    if prompter == "None":
+        prompter = None
```
tmm1, line 81: Why is this needed?

tmm1 (approved):
Great idea!

scripts/finetune.py
winglian: we have in the existing docs --prompter=None which I believe translates to a literal string. we could alternatively update the doc to be simply `--prompter` or alternatively default the value to None (which is what we're doing here)

PR #500, opened 2023-08-29
tweak: improve base builder for smaller layers

maximegmd:
By merging layers we don't save the apt cache

### DIFF
--- a/docker/Dockerfile-base
+++ b/docker/Dockerfile-base
@@ -13,16 +13,14 @@ ARG CUDA="118"
 
 ENV PYTHON_VERSION=$PYTHON_VERSION
 
-RUN apt-get update
-RUN apt-get install -y wget git build-essential ninja-build git-lfs libaio-dev && rm -rf /var/lib/apt/lists/*
-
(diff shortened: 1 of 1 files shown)


docker/Dockerfile-base
```diff
@@ -13,16 +13,14 @@ ARG CUDA="118"
 
 ENV PYTHON_VERSION=$PYTHON_VERSION
 
-RUN apt-get update
-RUN apt-get install -y wget git build-essential ninja-build git-lfs libaio-dev && rm -rf /var/lib/apt/lists/*
-
-RUN wget \
+RUN apt-get update \
+    && apt-get install -y wget git build-essential ninja-build git-lfs libaio-dev && rm -rf /var/lib/apt/lists/*
+    && wget \
```
tmm1, line 18: Can't the RUN wget stay as a separate layer still?

NanoCode012:
I think I talked with winglian about this topic, and we plan to swap the base layer after gptq merge, so I don't know if this is needed.

PR #504, opened 2023-08-29
fix: bad dtype for full finetune

maximegmd:
It seems that every configuration that requires flash attention and is llama derived requires this change.

### DIFF
--- a/src/axolotl/utils/models.py
+++ b/src/axolotl/utils/models.py
@@ -369,7 +369,7 @@ def load_model(
 
     # LlamaRMSNorm layers are in fp32 after kbit_training or full finetune, so we need to
     # convert them back to fp16/bf16 for flash-attn compatibility.
-    if cfg.flash_attention and cfg.is_llama_derived_model:
+    if needs_fa2_dtype or (cfg.flash_attention and cfg.is_llama_derived_model):
         LOG.info("converting modules to %s for flash attention", cfg.torch_dtype)
         for name, module in model.named_modules():
             if "norm" in name:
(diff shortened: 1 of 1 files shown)


src/axolotl/utils/models.py
```diff
@@ -367,11 +366,10 @@ def load_model(
         model = prepare_model_for_kbit_training(
             model, use_gradient_checkpointing=cfg.gradient_checkpointing
         )
-        needs_fa2_dtype = True
 
     # LlamaRMSNorm layers are in fp32 after kbit_training or full finetune, so we need to
     # convert them back to fp16/bf16 for flash-attn compatibility.
-    if needs_fa2_dtype and (cfg.flash_attention and cfg.is_llama_derived_model):
+    if cfg.flash_attention and cfg.is_llama_derived_model:
```
winglian: I think it's should be when it's a lora OR flash attention. (still need to add back the removed lines above)
```suggestion
    if needs_fa2_dtype or (cfg.flash_attention and cfg.is_llama_derived_model):
```

src/axolotl/utils/models.py
```diff
@@ -358,7 +358,6 @@ def load_model(
             if hasattr(module, "weight"):
                 module.to(torch.float32)
 
-    needs_fa2_dtype = cfg.adapter or cfg.fsdp
```
winglian: can you restore this line too pls?

src/axolotl/utils/models.py
```diff
@@ -367,11 +366,10 @@ def load_model(
         model = prepare_model_for_kbit_training(
             model, use_gradient_checkpointing=cfg.gradient_checkpointing
         )
-        needs_fa2_dtype = True
```
winglian: and this one

PR #510, opened 2023-08-30
chore(callback): Remove old peft saving code

NanoCode012:
Previously, peft does not save the adapter, hence we needed this code. After peft updated a while back, this callback unnecessarily saves the adapter another time.

We can now remove it.

### DIFF
--- a/src/axolotl/utils/callbacks.py
+++ b/src/axolotl/utils/callbacks.py
@@ -43,29 +43,6 @@ LOG = logging.getLogger("axolotl.callbacks")
 IGNORE_INDEX = -100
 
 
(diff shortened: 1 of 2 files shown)


tmm1 (approved):

NanoCode012:
Per winglian's comment in Discord, relora might need this code. I think I need some confirmation what relora usually saves at as I haven't trained one before. And if it's necessary for only relora, we can change the condition for only then.

NanoCode012:
Seems like relora already saves its own adapter_model

image

NanoCode012:
Rebased!

PR #513, opened 2023-08-31
misc fixes/improvements

winglian:


tmm1 (approved):

scripts/finetune.py
```diff
@@ -262,6 +262,8 @@ def train(
     if peft_config:
         LOG.info(f"Pre-saving adapter config to {cfg.output_dir}")
         peft_config.save_pretrained(cfg.output_dir)
+    tokenizer.save_pretrained(str(Path(cfg.output_dir)))
```
NanoCode012: Duplicate with line 287?

winglian: This is so the tokenizer is available in case someone wants to use the model at a checkpoint.

NanoCode012: It's these lines I'm confused about. https://github.com/OpenAccess-AI-Collective/axolotl/blob/6d261422ebfcceed47e0c147cf4fd0811040486a/scripts/finetune.py#L285-L287

- It seems that `output_dir` is only created in link above, but you also seem to save tokenizer and model to that `output_dir` before creation.
- I do not see a difference between the two tokenizer saves.

NanoCode012 (approved):

PR #516, opened 2023-08-31
let hf trainer handle torch compile

winglian:


src/axolotl/utils/trainer.py
```diff
@@ -579,6 +579,9 @@ def setup_trainer(cfg, train_dataset, eval_dataset, model, tokenizer, total_num_
         if cfg.bench_dataset:
             training_arguments_kwargs["bench_dataset"] = cfg.bench_dataset
 
+    if torch.__version__ >= "2" and sys.platform != "win32":
+        training_arguments_kwargs["torch_compile"] = True
```
tmm1: I think the conditional is probably unnecessary if HF is doing the work

But we will need to expose this to cfg because many things still fail trying to compile and train.

We may need the other compile options like backend exposed as cfg too

winglian: thanks! updated

tmm1 (approved):

README.md
```diff
@@ -526,6 +526,10 @@ wandb_log_model: # "checkpoint" to log model to wandb Artifacts every `save_step
 # where to save the finished model to
 output_dir: ./completed-model
 
+# whether to use torch.compile and which backend to use
+torch_compile:  # bool
+torch_compile_backend:  # Optional[str]
```
NanoCode012, line 531: Is it necessary to open this "backend" config?

> Refer to the PyTorch doc for possible values and note that they may change across PyTorch versions.

and reading the source of the torch compile, it seems to only have a condition for `inductor`.

src/axolotl/train.py
```diff
@@ -80,10 +80,6 @@ def train(
 
     model.config.use_cache = False
 
-    if torch.__version__ >= "2" and sys.platform != "win32":
```
NanoCode012, line 83: do we still need to add this check on validate config?

NanoCode012:
It seems like there may be small benefits to leaving it to default on. Is there any case having it off is better?

src/axolotl/train.py
tmm1: No I think we can let HF handle it

README.md
tmm1: Not sure; there are other backends but maybe they're not useful?

```
torch._dynamo.list_backends()

['aot_ts_nvfuser', 'cudagraphs', 'inductor', 'ipex', 'nvprims_nvfuser', 'onnxrt', 'tvm']
```

tmm1:
> Is there any case having it off is better?

it cannot default on because it doesn't work. if you try to train it errors out.

src/axolotl/utils/trainer.py
```diff
@@ -579,6 +580,21 @@ def setup_trainer(cfg, train_dataset, eval_dataset, model, tokenizer, total_num_
         if cfg.bench_dataset:
             training_arguments_kwargs["bench_dataset"] = cfg.bench_dataset
 
+    if cfg.torch_compile:
+        if torch.__version__ < "2.1.0":  # pylint: disable=protected-access
```
winglian, line 584: @tmm1 lmk if you think this is good enough for now

Issue #494, opened 2023-08-28
 Training hangs with sample_packing for specific datasets [multi-GPU]


jphme:
### Please check that this issue hasn't been reported before.

- [X] I searched previous [Bug Reports](https://github.com/OpenAccess-AI-Collective/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

Training should run normally.

### Current behaviour

With specific datasets and a multi-GPU setup (2*RTX 4090s in my case), training hangs (after the first training step or after the fist epoch) until a timeout error occurs.

The bug only happens:

- with 2 (or more?) GPUs
- with specific datasets (the average/maximum number of tokens per example seems to be important)
- with `sequence_len: 4096` (2048 runs fine)
- with `sample_packing: true`

With `micro_batch_size: 1` instead of 2 (which is also used for `sample_packing_seq_len_mulitpler`!), training starts fine but hangs after epoch 1 has finished.

When training hangs, both GPUs are stuck below 1/3 TWP (~90W in my case) and VRAM is still allocated.
P2P settings like `NCCL_P2P_DISABLE=1` don't change anything. 

### Steps to reproduce

Steps to reproduce (with config from [Llama-2/examples/lora.yml](https://github.com/OpenAccess-AI-Collective/axolotl/blob/main/examples/llama-2/lora.yml) below.
Minimal example dataset to reproduce the issue is here [jphme/test_ds_2gpu_axolotl_bug](https://huggingface.co/datasets/jphme/test_ds_2gpu_axolotl_bug).

```bash
#works fine
accelerate launch scripts/finetune.py examples/llama-2/lora.yml

#hangs with 2gpus, nothing except datasegt changed
accelerate launch scripts/finetune.py examples/llama-2/lora_w_dataset_above.yml
```

Reducing sequence_len to 2048 or disabling sample_packing prevents the hanging; when setting `micro_batch_size: 1` training runs fine until epoch 1 has finished and then hangs. 

### Possible solution

I suspect that the bug has something todo with the `sample_packing` implementation and is only triggered for examples with many tokens:

Numer of tokens of the original Llama-2 lora.yml dataset `mhenrichsen/alpaca_2k_test` which runs fine:
```bash
count    2000.000000
mean      217.408000
std       146.209145
min        49.000000
25%        95.000000
50%       167.000000
75%       313.000000
max       762.000000
dtype: float64
```

Token length of my custom dataset:
```
count      25.000000
mean     1805.000000
[...]

PR #517, opened 2023-08-31
Alternative sample_packing implementation

jphme:
I dont want to merge in the current state, just open a draft PR to make people aware of this (and probably someone with better programming skills than me can implement a `real` solution).

The current implementation of `sample_packing` causes the training to hang when `micro_batch_size`>1  (and sometimes even with mbs=1) - see issue #494 for a detailed report.

This implementation is closer to the original Openchat implementation and doesn't use the distributed sampler. The advantage is that packing is more efficient (the algorithm knows the whole dataset) and estimated regarding num_steps are better. The disadvantage is that every process has to load the whole dataset - for very large datasets probably not an option.

Detailed sweep with my real-world dataset (which im unfortunately not allowed to share); consisting of comparatively many long examples:
[WAND Report](https://api.wandb.ai/links/jan-harries/jyib3z7m).

A *good* solution would probably use some modified BatchSampler, so that accelerate would be able to either distribute the batches from one central process or to all nodes as a whole, depending on the dataset (see [here](https://github.com/huggingface/accelerate/blob/7befe580c2393b0a3d75a8aa75af79b8e57f84a4/src/accelerate/data_loader.py) ). 

Another solution could be to selectively ignore the buffers causing the hanging during gradient accumulation (using [this](https://github.com/pytorch/pytorch/blob/7ec439206853ec1a7ba74cfeeaa15d45d30475cc/torch/nn/parallel/distributed.py#L2149) but here Im out of my depth). 

Unfortunately I don't have the time + skills to implement a clean/good solution for this currently - just dropping here so it´s  not lost, if Im the only one with this problem happy to have the PR closed ;).

### DIFF
--- a/src/axolotl/utils/dataloader.py
+++ b/src/axolotl/utils/dataloader.py
@@ -3,11 +3,10 @@ import hashlib
 import itertools
 import logging
 import math
-from typing import Any, Callable, List, Union
+from typing import Any, Callable, List
 
 import numba
 import numpy as np
-from torch.utils.data import DistributedSampler, Sampler
 
 LOG = logging.getLogger("axolotl.utils.dataloader")
 
@@ -145,7 +144,6 @@ class MultipackDistributedDataloader:
         collate_fn: Callable,
         seq_max_length: int = 2048,
         batch_size: int = 1,
-        sampler: Union[Sampler, DistributedSampler] = None,
         packing_efficiency_estimate: float = 1.0,
         sample_packing_seq_len_multiplier: int = 1,
         num_replicas: int = 1,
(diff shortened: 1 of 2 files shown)


teknium1:
Have you tested this on various datasets? Did you do the stuff to get the per gpu token size?

I trained with what I think is still the latest packing iteration just a week or two ago and it completed fine with >1 bs

jphme:
> Did you do the stuff to get the per gpu token size?

What exactly do you mean? The length/efficiency estimation runs well, the problem happens on gradient accumulation.

For datasets: Datasets where many small examples (with few tokens) dominate, so that the batches are very even, work without any problems; the hangs only happen with datasets with large token-lengths, where the batches are often uneven and estimates change frequently (see the random example dataset in the issue).

But imho there have to be multiple sources of the problem - one software problem (buffer synchronization) and one hardware problem (hangs somewhere in the nvidia stack without p2p)...

jphme:
Seems very closely related with the problem/solution reported here (also RTX 4090s, problems go away when the batch size is known pre-distribution:
https://github.com/Lightning-AI/lightning/issues/11910#issuecomment-1546796342

jphme:
I think this has been solved in the meantime.

PR #524, opened 2023-09-04
move is_llama_derived_model into normalize_config

tmm1:


winglian (approved):
lgtm, I think later we should address running the config validation step after the normalization step so that we can check other setting combinations that are llama specific

PR #525, opened 2023-09-04
use flash_attn xentropy when available

tmm1:
uses xentropy module if installed

see memory improvement numbers in https://github.com/Dao-AILab/flash-attention/issues/475#issuecomment-1704571675

### DIFF
--- a/src/axolotl/monkeypatch/llama_attn_hijack_flash.py
+++ b/src/axolotl/monkeypatch/llama_attn_hijack_flash.py
@@ -57,7 +57,9 @@ def replace_llama_attn_with_flash_attn(packed: Optional[bool] = False):
             CrossEntropyLoss, inplace_backward=True
         )
     except ImportError:
-        pass
+        LOG.info(
+            "optimized flash-attention CrossEntropyLoss not found (run `pip install git+https://github.com/Dao-AILab/flash-attention.git#egg=xentropy_cuda_lib&subdirectory=csrc/xentropy`)"
+        )
 
 
 # Disable the transformation of the attention mask in LlamaModel as the flash attention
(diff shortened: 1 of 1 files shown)


src/axolotl/monkeypatch/llama_attn_hijack_flash.py
```diff
@@ -44,6 +49,16 @@ def replace_llama_attn_with_flash_attn(packed: Optional[bool] = False):
...
 
+    try:
+        from flash_attn.losses.cross_entropy import CrossEntropyLoss
+
+        LOG.info("patching with flash_attn.losses.cross_entropy")
+        transformers.models.llama.modeling_llama.CrossEntropyLoss = partial(
+            CrossEntropyLoss, inplace_backward=True
+        )
+    except ImportError:
+        pass
```
mhenrichsen: Maybe add some logging to announce it's not installed?

winglian:
```suggestion
        LOG.warning("unable to import flash_attn.losses.cross_entropy.CrossEntropyLoss")
```

winglian (approved):
good to go after adding a warning.

src/axolotl/monkeypatch/llama_attn_hijack_flash.py
tmm1: since most users won't have this installed, i wasn't sure about using a warning. but i agree it makes sense to log something.

```suggestion
        LOG.info("optimized flash-attention CrossEntropyLoss not found (run `pip install git+https://github.com/Dao-AILab/flash-attention.git#egg=xentropy_cuda_lib&subdirectory=csrc/xentropy`)"
```

IgnacioFDM:
Maybe this and #526 should be added to the Dockerfile?

PR #527, opened 2023-09-05
log rank

winglian:
this is helpful for multi-gpu logging since tracking things down by PID isn't ideal

### DIFF
--- a/src/axolotl/logging_config.py
+++ b/src/axolotl/logging_config.py
@@ -23,6 +23,7 @@ class ColorfulFormatter(Formatter):
     }
 
     def format(self, record):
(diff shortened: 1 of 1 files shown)


src/axolotl/logging_config.py
```diff
@@ -23,6 +23,7 @@ class ColorfulFormatter(Formatter):
     }
 
     def format(self, record):
+        record.rank = int(os.getenv("LOCAL_RANK", "0"))
```
NanoCode012, line 26: Seems good. I was a bit confused sometimes which rank the output comes from.

When this is multi-node, this can have duplicate values, therefore, we might need global_rank when we move into multi-node.

NanoCode012 (approved):

PR #537, opened 2023-09-08
Early stopping metric

winglian:
fix to set the benchmarks to use for early stopping


winglian:
tested and validated with 2x gpus, needed to broadcast the metrics to all ranks so they have the metrics available, and also reorder the callbacks so the early stopping check happens after the benchmarks

PR #538, opened 2023-09-08
Model parallel

winglian:
supports falcon-180b now! using naive model parallel. correctly sets the device when launching on multi-gpu without accelerate


src/axolotl/utils/bench.py
```diff
@@ -28,7 +28,7 @@ def gpu_memory_usage_smi(device=0):
 
 
 def log_gpu_memory_usage(log, msg, device):
-    if not torch.cuda.is_available():
+    if not torch.cuda.is_available() or device == "auto":
```
winglian, line 31: not able to determine the device to log gpu stats for

PR #539, opened 2023-09-08
[WIP][Feature] Tensor parallel

winglian:


thincal:
It seems that tensor_parallel from BlackSamorez is inactive for maintenance ? any plan for this support in future @winglian

PR #547, opened 2023-09-10
improve how we setup eval/save strategies and steps

winglian:


src/axolotl/utils/trainer.py
```diff
@@ -556,21 +556,30 @@ def setup_trainer(cfg, train_dataset, eval_dataset, model, tokenizer, total_num_
...
-    else:
-        # we have an eval set, but no steps defined, use epoch
-        training_arguments_kwargs["evaluation_strategy"] = "epoch"
 
-    if cfg.save_strategy:
+    if cfg.save_steps:
+        # save_steps implies save_strategy of steps
+        training_arguments_kwargs["save_strategy"] = "steps"
+        training_arguments_kwargs["save_steps"] = cfg.save_steps
+    elif cfg.save_strategy:
```
NanoCode012, line 581: One side effect would be that: individuals who depend on `save_strategy: epoch/no` and forget to remove `save_step` would get a different result. However, this change is good as it's more logical.

NanoCode012 (approved):

PR #559, opened 2023-09-12
document that packaging needs to be installed before flash-attn

winglian:


README.md
```diff
@@ -140,7 +142,8 @@ accelerate launch scripts/finetune.py examples/openllama-3b/lora.yml \
   git clone https://github.com/OpenAccess-AI-Collective/axolotl
   cd axolotl
 
-  pip3 install -e .
+  pip3 install packaging
+  pip3 install -e .[flash-attn]
   pip3 install protobuf==3.20.3
   pip3 install -U --ignore-installed requests Pillow psutil scipy
```
NanoCode012, line 148: You can merge these other installs together.

NanoCode012 (approved):

NanoCode012:
Do we also need to mention that `ninja` is required?

README.md
winglian: the problem is that if packaging isn't installed first, trying to install at the same time fails for flash-attn.

winglian:
> Do we also need to mention that `ninja` is required?

I haven't seen that ninja has been problematic, but we can address in another pr

PR #561, opened 2023-09-13
check for the existence of the default accelerate config that can create headaches

winglian:


NanoCode012 (approved):
Didn't know this was possible! Much needed.

PR #577, opened 2023-09-15
don't resize embeddings if it's already large enough

winglian:
for example, we don't actually need to resize phi b/c the embeddings size is already much larger than the tokenizer len

```python
>>> from transformers import AutoModelForCausalLM, AutoTokenizer
>>> tokenizer = AutoTokenizer.from_pretrained("microsoft/phi-1_5", trust_remote_code=True)
>>> len(tokenizer)
50295
>>> model = AutoModelForCausalLM.from_pretrained("microsoft/phi-1_5", trust_remote_code=True)
>>> model.get_input_embeddings().num_embeddings
51200
>>> 
```

### DIFF
--- a/src/axolotl/utils/models.py
+++ b/src/axolotl/utils/models.py
@@ -291,7 +291,8 @@ def load_model(
         if cfg.resize_token_embeddings_to_32x
         else len(tokenizer)
     )
-    model.resize_token_embeddings(embeddings_len)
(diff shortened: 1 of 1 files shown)


NanoCode012:
As I rmb, model.resize implicitly calls model.tie_weights. By doing this, we're not tying weights. I'm not clear whether this is still important as it was a warning message while finetuning a month ago.

It is also weird that the tokenizer's length and the model's weights have different length..

PR #578, opened 2023-09-15
use fastchat conversations template

winglian:
use the `vicuna_v1.1` Conversations template from the fastchat package. This will make it easier to extend additional conversation types since we can simply load an already registered conversation format template.


README.md
```diff
@@ -180,7 +180,7 @@ Have dataset(s) in one of the following format (JSONL recommended):
   ```json
   {"instruction": "...", "input": "...", "output": "..."}
   ```
-- `sharegpt:chat`: conversations where `from` is `human`/`gpt`
+- `sharegpt`: conversations where `from` is `human`/`gpt`
```
NanoCode012, line 183: We need to ensure smooth transition. There are a lot of configs with the old names.

I think we should for now in config auto convert the old names to the new ones and log a deprecation warning.

README.md
```diff
@@ -269,11 +269,11 @@ Have dataset(s) in one of the following format (JSONL recommended):
   ```json
   {"prompt": "...", "generation": "..."}
   ```
-- `sharegpt_simple.load_role`: conversations where `role` is used instead of `from`
+- `sharegpt.load_role`: conversations where `role` is used instead of `from`
```
NanoCode012, line 272: Same for this. We need to smoothly transition this change.

requirements.txt
```diff
@@ -31,3 +31,5 @@ scipy
 scikit-learn==1.2.2
 pynvml
 art
+wandb
+fschat
```
NanoCode012: we should set version for fschat in case a change upstream causes a change with prompting without our knowledge.

Also, I am not quite a fan of installing an entire package just for their conversation script, but it seems like a clean integration.

NanoCode012 (commented):
Just a quick look first.

README.md
```diff
@@ -443,6 +443,7 @@ datasets:
     data_files: # Optional[str] path to source data files
     shards: # Optional[int] number of shards to split data into
     name: # Optional[str] name of dataset configuration to load
+    conversation:  # Optional[str] fastchat conversation type, only used with type: sharegpt
```
NanoCode012, line 446:
```suggestion
    conversation:  # Optional[str] fastchat conversation type, only used with type: sharegpt. See options: https://github.com/lm-sys/FastChat/blob/main/fastchat/conversation.py
```

src/axolotl/monkeypatch/fastchat_conversation_turns.py
```diff
@@ -0,0 +1,174 @@
...
+                yield role + ":", ""
+    else:
+        raise ValueError(f"Invalid style: {self.sep_style}")
+
+
+def add_get_turns_to_conversation():
+    import fastchat.conversation
+
+    fastchat.conversation.Conversation.get_turns = get_turns
+    fastchat.conversation.Conversation.get_prompt = get_prompt
```
NanoCode012, line 174: May I ask what changed to need this patch? I see a lot of similarity Conversation from fschat.

NanoCode012:
Oh, didn't see it got merged while I was checking it

src/axolotl/monkeypatch/fastchat_conversation_turns.py
winglian: Yeah. Mostly the ability to reuse most of the predefined conversation types defined thwre

PR #580, opened 2023-09-15
support custom field for completion from yml

winglian:
you can now specify an alternate field from the dataset to use instead of `text` as the completion text

### DIFF
--- a/README.md
+++ b/README.md
@@ -445,6 +445,9 @@ datasets:
       # 'no_input_format' cannot include {input}
       no_input_format: "{instruction} "
 
+      # for completions datsets, uses the provided field if not `text`
+      field:
+
 # axolotl attempts to save the dataset as an arrow after packing the data together so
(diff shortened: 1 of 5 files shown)


winglian:
verified this against the tiny-textbooks dataset

README.md
```diff
@@ -322,6 +322,7 @@ See [examples](examples) for quick start. It is recommended to duplicate and mod
     - path: EleutherAI/pile
       name: enron_emails
       type: completion # format from earlier
+      field: text # Optional[str] default: text, field to use for completion data
```
NanoCode012, line 325: This property would need to be put in "all yaml options" and specified for only "completion".

NanoCode012 (commented):
While it's a very clean code, I think it seems quite complicated just to change the key for the `completion` dataset.

Personally, I would just write a simple python script to rename the key of the dataset..

PR #587, opened 2023-09-15
add bf16 check

winglian:


src/axolotl/utils/config.py
```diff
@@ -4,6 +4,7 @@
 import os
 
 import torch
+from transformers.utils import is_torch_bf16_gpu_available
```
AlpinDale, line 7: `torch` has a similar function, `torch.cuda.is_bf16_supported()`. Would it be better to use it directly through torch instead of transformers?

src/axolotl/utils/config.py
```diff
@@ -88,6 +89,14 @@ def normalize_config(cfg):
 
 
 def validate_config(cfg):
+    if is_torch_bf16_gpu_available():
+        if not cfg.bf16 and not cfg.bfloat16:
+            LOG.info("bf16 support detected, but not enabled for this configuration.")
+    else:
+        if cfg.bf16 or cfg.bfloat16:
+            raise ValueError(
+                "bf16 requested, but not available on this GPU. Requires 30XX series or above."
```
NanoCode012:
```suggestion
                "bf16 requested, but not available on this GPU. Requires Ampere series or above."
```

src/axolotl/utils/config.py
winglian: thanks, updated the check to use that.

winglian: actually, using the torch native check breaks the unit tests since cuda isn't available. transformers utils has a cuda check built in.

PR #594, opened 2023-09-17
Feat(data): Allow loading local csv and text

NanoCode012:
We forgot these two.


NanoCode012:
Seems like we don't even use `ds_type` from the cfg in the code? We seem to load it from the file extension

```
    ds_type: # Optional[str] (json|arrow|parquet|text|csv) defines the datatype when path is a file
```

Should this be removed?

winglian:
See line 180image

NanoCode012:
Ah I see. That would mean this PR changes might not be necessary. Only Readme is new.

winglian (approved):

PR #597, opened 2023-09-18
minor tweaks to simplify

winglian:
don't debug the attention mask value since that is almost always exclusively `1` for all cases.
simplify config by not asking users to update the total_num_tokens and sample_packing_eff_est since that computation has been optimized and the optimization to set that ahead of time can cause headaches

### DIFF
--- a/src/axolotl/utils/trainer.py
+++ b/src/axolotl/utils/trainer.py
@@ -429,7 +429,7 @@ def calculate_total_num_steps(cfg, train_dataset, tokenizer):
                 .apply(lambda x: len(x))  # pylint: disable=unnecessary-lambda
                 .values
             )
-            LOG.info(f"📝 UPDATE CONFIG WITH: `total_num_tokens: {total_num_tokens}`")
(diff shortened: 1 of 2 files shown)


src/axolotl/utils/trainer.py
```diff
@@ -489,6 +489,8 @@ def calculate_total_num_steps(cfg, train_dataset, tokenizer):
             data_loader_len = data_loader.len_w_stats()
             actual_eff = data_loader.efficiency()
             LOG.info(f"data_loader_len: {data_loader_len}")
+            # FIXME: is there a bug here somewhere? the total num steps depends
+            # on the agreed on value for sample_packing_eff_est
```
NanoCode012, line 493: what does this mean? Do you mean the total num steps should be affected by sample efficiency or some other var?

winglian: exactly. because if the various processes have different calculations, we choose the max value to find the smallest steps that they all share. It doesn't seem to be a major issue atm, but should get addressed.

PR #602, opened 2023-09-19
more sane defaults for openllama 3b used for quickstarts

winglian:
Since we point to open llama 3b in the quickstart section, let's use some sane defaults


winglian:
verified on single gpu. vram usage recorded
lora - 7gb vram
qlora - 4.5gb vram
full ft - 21gb vram

PR #609, opened 2023-09-19
skip the gpu memory checks if the device is set to 'auto'

winglian:
resolves #456

Issue #599, opened 2023-09-18
Llama2 GPTQ training does not work


Napuh:
### Please check that this issue hasn't been reported before.

- [X] I searched previous [Bug Reports](https://github.com/OpenAccess-AI-Collective/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

execute `finetune.py` with `examples/llama-2/gptq-lora.yml`. 

Execution does not throw any error and the model trains fine.

### Current behaviour

Execution throws error after a while. It does not start the trainer whatsoever.

Error thrown:
```
[2023-09-18 11:25:48,695] [INFO] [axolotl.train.train:57] [PID:6348] [RANK:0] loading model and (optionally) peft_config...                                                                                                                     [2023-09-18 11:28:06,557] [ERROR] [axolotl.load_model:321] [PID:6348] [RANK:0] Exception raised attempting to load model, retrying with AutoModelForCausalLM                                                                                    [2023-09-18 11:28:06,557] [ERROR] [axolotl.load_model:324] [PID:6348] [RANK:0] Found modules on cpu/disk. Using Exllama backend requires all the modules to be on GPU.You can deactivate exllama backend by setting `disable_exllama=True` in the quantization config object                                                                                            Traceback (most recent call last):                                                                                        File "/root/axolotl/src/axolotl/utils/models.py", line 272, in load_model                                                 model = AutoModelForCausalLM.from_pretrained(                                                                         File "/opt/conda/lib/python3.10/site-packages/transformers/models/auto/auto_factory.py", line 563, in from_pretrained 
    return model_class.from_pretrained(
  File "/opt/conda/lib/python3.10/site-packages/transformers/modeling_utils.py", line 3287, in from_pretrained
    model = quantizer.post_init_model(model)
  File "/opt/conda/lib/python3.10/site-packages/optimum/gptq/quantizer.py", line 482, in post_init_model
    raise ValueError(
ValueError: Found modules on cpu/disk. Using Exllama backend requires all the modules to be on GPU.You can deactivate exllama backend by setting `disable_exllama=True` in the quantization config object
Traceback (most recent call last):
  File "/root/axolotl/src/axolotl/utils/models.py", line 272, in load_model
    model = AutoModelForCausalLM.from_pretrained(
  File "/opt/conda/lib/python3.10/site-packages/transformers/models/auto/auto_factory.py", line 563, in from_pretrained
    return model_class.from_pretrained(
[...]

PR #611, opened 2023-09-20
override device_map if the model is gptq-based

Napuh:
Override `device_map` in config if the model is GPTQ-based.

This solves #599, which wasn't entirely addressed by #609.

After this PR, example `gptq-lora.yml` trains correctly.

Finetuning with 2xRTX3090 works too, but I'm not really sure on how it will perform on custom accelerate configs.


winglian (approved):

winglian:
@Napuh This should be fixed with #621 now

PR #614, opened 2023-09-20
Only run tests when a change to python files is made

maximegmd:
There is no need to waste resources when we edit readmes ^^

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - "main"
+    paths:
+      - '**.py'
   pull_request:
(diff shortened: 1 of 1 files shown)


winglian (commented):
Can you add that under pull request too? Otherwise it will also run it for README changes in prs. Thanks!

.github/workflows/tests.yml
```diff
@@ -4,6 +4,8 @@ on:
   push:
     branches:
       - "main"
+    paths:
+      - '**.py'
   pull_request:
```
winglian, line 9:
```suggestion
  pull_request:
      paths:
       - '**.py'
```

PR #616, opened 2023-09-21
split completion text to sequence_len

winglian:
This adds support for splitting completion text to the sequence_len rather than dropping or truncating.

### DIFF
--- a/src/axolotl/prompt_tokenizers.py
+++ b/src/axolotl/prompt_tokenizers.py
@@ -41,10 +41,15 @@ class PromptTokenizingStrategy(abc.ABC):
         self.tokenizer: PreTrainedTokenizer = tokenizer
         self.train_on_inputs = train_on_inputs
         self.sequence_len = sequence_len
+        self.max_length = sequence_len
 
     @abc.abstractmethod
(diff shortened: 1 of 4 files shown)


src/axolotl/prompt_tokenizers.py
```diff
@@ -77,7 +82,7 @@ def _tokenize(
             result = self.tokenizer(
                 prompt,
                 truncation=True,
-                max_length=self.sequence_len,
+                max_length=self.max_length,
```
NanoCode012, line 85: What is max_length for? Since you're splitting long texts in the prompter, I don't quite understand what this is for.

winglian: the tokenizer no longer emits the entire tokenized string (it used to emit a warning), but now truncates it to max_length.

winglian: without this, we would end up with truncated completion data, and we wouldn't be able to split it later on.

NanoCode012: ah I see. You split after tokenize, not before.

PR #632, opened 2023-09-25
Added quotes to the pip install -e command in the documentation to fix an incompatibility …

Nan-Do:
…with shells that do glob expansion like zsh a more comprehensive description can be found at https://github.com/mu-editor/mu/issues/852


winglian (approved):
👍 thank you!

PR #633, opened 2023-09-25
Fix(cfg): Add validation for save_strategy and eval_strategy

NanoCode012:
This PR should fix the ambiguity in configs when users have conflicting save_strategy with save_steps or eval_strategy and eval_steps.

This also cleans the check for eval and save to use same condition format.


winglian (approved):

NanoCode012:
Rebased. Can merge when tests done.

PR #644, opened 2023-09-27
Feat: Add example for Mistral

NanoCode012:
Todo:
- [x] Update transformer commit

In future PR, allow sample_packing.

Upstream PR: https://github.com/huggingface/transformers/pull/26447

### DIFF
--- a/examples/mistral/config.yml
+++ b/examples/mistral/config.yml
@@ -1,0 +1,69 @@
+base_model: mistralai/Mistral-7B-v0.1
+base_model_config: mistralai/Mistral-7B-v0.1
+model_type: MistralForCausalLM
+tokenizer_type: LlamaTokenizer
+
+load_in_8bit: false
+load_in_4bit: false
+strict: false
+
+datasets:
+  - path: mhenrichsen/alpaca_2k_test
+    type: alpaca
+dataset_prepared_path: last_run_prepared
+val_set_size: 0.01
+output_dir: ./out
+
+sequence_len: 4096
+sample_packing:
+pad_to_sequence_len:
+
+adapter:
+lora_model_dir:
+lora_r: 32
+lora_alpha: 16
+lora_dropout: 0.05
+lora_target_linear: true
+lora_fan_in_fan_out:
+
+wandb_project:
+wandb_entity:
+wandb_watch:
+wandb_run_id:
+wandb_log_model:
+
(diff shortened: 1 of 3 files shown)


examples/mistral/config.yml
```diff
@@ -0,0 +1,69 @@
+base_model: mistralai/Mistral-7B-v0.1
+base_model_config: mistralai/Mistral-7B-v0.1
+model_type: MistralForCausalLM
```
winglian, line 3: Should be able to use AutoModelForCausalLM

winglian:
let's use `[78dd120](https://github.com/huggingface/transformers/commit/78dd1202823ca035b9609ddbcdaac2945a6530ff)`

winglian:
this PR should fix the issues we were having w transformers main that we had to pin it in the first place https://github.com/OpenAccess-AI-Collective/axolotl/pull/636

examples/mistral/config.yml
```diff
@@ -0,0 +1,69 @@
...
+strict: false
+
+datasets:
+  - path: mhenrichsen/alpaca_2k_test
+    type: alpaca
+dataset_prepared_path: last_run_prepared
+val_set_size: 0.01
+output_dir: ./out
+
+sequence_len: 4096
```
mhenrichsen: Isn't mistral 8k?

examples/mistral/config.yml
```diff
@@ -0,0 +1,69 @@
...
+sample_packing:
+pad_to_sequence_len:
+
+adapter:
+lora_model_dir:
+lora_r: 32
+lora_alpha: 16
+lora_dropout: 0.05
+lora_target_linear: true
+lora_fan_in_fan_out:
```
mhenrichsen: Maybe we should remove these if this is a full finetune example.

winglian:
also, see #646

examples/mistral/config.yml
NanoCode012: Thanks! As of that time, there was no flash, so I just thought to use a reasonable balanced number. However, I will now update it now that we have FA patch.

examples/mistral/config.yml
NanoCode012: Yes, but I think it would be better to specify for later parsing (aka is_mistral_derived..)

PR #646, opened 2023-09-27
Mistral flash attn packing

winglian:


ehartford:
I can't wait to try this!  I'll start a build tonight

PR #651, opened 2023-09-28
Make dataset_processes configurable

corbt:
I'm using the Axolotl script to train models on modal.com serverless GPUs. Unfortunately, their environment seems to have some kind of bug where if I try to run `datasets.filter` with too high a `num_proc`, it throws an error and dies.

This PR adds a new configuration option `dataset_processes`, which lets you explicitly set the number of processes used to map/filter the dataset. If not included, this defaults to the current behavior of setting that to `os.cpu_count()`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -477,6 +477,9 @@ datasets:
 dataset_prepared_path: data/last_run_prepared
 # push prepared dataset to hub
 push_dataset_to_hub: # repo path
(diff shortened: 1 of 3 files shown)


winglian (approved):

corbt:
Just pushed a new commit that fixes black.

PR #652, opened 2023-09-28
Fix bug when using pretokenized datasets

ein-ich:
Fix bug because type is not set when using pretokenized datasets.
Also improve Readme.


winglian (approved):
Thanks!

PR #665, opened 2023-10-02
prepared dataset caching, other misc fixes

winglian:
this will only explicitly cache when `dataset_prepared_path` is defined in the yml
other small fixes like wording for inference and not adding extra spaces when debugging tokenization


NanoCode012 (approved):
One more thing would be to add a message on readme near that config to say "leave empty for no cache".

PR #669, opened 2023-10-03
Feat: Allow usage of native Mistral FA when no sample_packing

NanoCode012:


winglian:
@NanoCode012 don't forget to pin transformers to the new commit in `requirements.txt`

winglian:
@NanoCode012 seems like we need to split the tests to run independently. Once the patch is loaded for the sample packing case, it continues to use that for the next test that is supposed to use the transformers fa2 patch w/o sample packing

winglian (approved):

PR #671, opened 2023-10-04
flash_attention + sample packing for stablelm 3b

winglian:


src/axolotl/monkeypatch/btlm_attn_hijack_flash.py
```diff
@@ -17,7 +18,8 @@ def replace_btlm_attn_with_flash_attn(model_name="cerebras/btlm-3b-8k-base"):
     # this is a wonky hack to get the remotely loaded module
     model_config = AutoConfig.from_pretrained(model_name, trust_remote_code=True)
     # we need to load the model here in order for modeling_btlm to be available
-    AutoModelForCausalLM.from_pretrained(model_name, trust_remote_code=True)
+    with init_empty_weights():
```
NanoCode012, line 21: while using this, do you need to load the weights after?

winglian: nah, what happens with it is if you haven't ever loaded the model before, b/c it uses trust_remote_code, the modeling code isn't available yet to modify, so we need to pseudo-load an empty model

NanoCode012: Oh I see. Weirdly, never saw people use this before. Except if it's a large model and need to load one part first

Haoxiang-Wang:
Thank you for integrating StableLM into this codebase! @winglian 
Can you provide your config YAML file to run this StableLM (stabilityai/stablelm-3b-4e1t) model?

PR #675, opened 2023-10-05
Adding qlora config for Mistral

tokenbender:
Contains fix for Mistral FA issue - ValueError: You are attempting to perform batched generation with padding_side='right' this may lead to unexpected behaviour for Flash Attention version of Mistral. Make sure to  call tokenizer.padding_side  = 'left' before tokenizing the input.

Fix for now is to set sample_packing: true and pad_to_sequence_len: true


tokenbender:
The workaround can also go in config.yml but I don't think sample_packing: true/pad_to_sequence_len: true should be forced on the default full FT config.

seungduk-yanolja:
can we just add a config to set padding_side: left?

NanoCode012:
The main issue will be fixed in #681 . However, we can still get this config in :)

NanoCode012 (commented):
Could you rename the file to `qlora.yml` to follow the same format as others?

tokenbender:
> Could you rename the file to `qlora.yml` to follow the same format as others?

Done

NanoCode012 (approved):

PR #679, opened 2023-10-05
Feat: Set WORKDIR to /workspace/axolotl

NanoCode012:
This is just to get rid of unneeded cd. Whenever I enter into the docker container, I keep needing to cd into axolotl. We should just start off inside the repo.

Untested, but should work?


winglian (approved):
seems fine, I checked the Dockerfile-runpod too since that depends on this and it seems fine.

PR #681, opened 2023-10-05
Fix: ValueError when FA + Mistral when padding_side=right

NanoCode012:
- [ ] Test


ghost (approved):

NanoCode012:
Ok, it should work. Unfortunately, don't have the GPU to test. Keep oom-ing locally.

NanoCode012:
I'm a bit unclear why the Error did not crop up during E2E tests. We should look into this later.

winglian (approved):

PR #687, opened 2023-10-06
feature: add aligned samples for completion prompt strategy

kallewoof:
By default, data will be chopped into samples, with the last sample being right-padded to fill up the context. This will often result in last-samples-of-a-text that have a short text followed by a bunch of pads. While this is fine from a training perspective, it is far more common in real life to encounter a completion task where you are given a very short amount of the *starting text*. This PR flips the padding side, when possible, so that the very first sample begins with padding tokens, followed by the starting text, and the remaining samples are then perfectly aligned with the sequence length so that no more padding is required.

In other words, we end up with filled up context sizes for all samples except the first ones in each input, which now looks like

```[PAD] [PAD] [PAD] ... [PAD] Once upon a time, there was a```

whereas we would normally end up with filled up context sizes for all except the last one in each input, which looks like

```happily ever after.[PAD] [PAD] [PAD] ...[PAD] [PAD] [PAD]```

I should note that this approach would most likely benefit instruction format prompt strategies as well, with the caveat that it at minimum gets to the Response part.

### DIFF
--- a/README.md
+++ b/README.md
@@ -422,6 +422,8 @@ tokenizer_legacy:
 # resize the model embeddings when new tokens are added to multiples of 32
 # this is reported to improve training speed on some models
 resize_token_embeddings_to_32x:
+# Whether to use left or right padding, defaults to right
+tokenizer_padding_side: right
 
 # used to identify which the model is based on
 is_falcon_derived_model:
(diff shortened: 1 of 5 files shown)


kallewoof:
Tested. This code works as intended. It is giving me `[PAD] [PAD] [PAD] [PAD] [PAD] Once upon a time` for starting samples and everything else looks unpadded.

winglian:
would you be able to add a unit test to validate this behavior?

kallewoof:
Done. Converting to draft though, as I am seeing some odd losses when using this in training. Digging.

kallewoof:
The odd losses I was seeing were unrelated to this PR (verified both by seeing the weird loss on main branch, and by seeing the weird loss go away when I fixed my settings). I think this is RFM.

PR #693, opened 2023-10-06
update mistral lr, sample pack

mhenrichsen:


winglian (approved):

NanoCode012:
Could you apply linter to that file? Seems like you're missing newline at the end.

mhenrichsen:
> Could you apply linter to that file? Seems like you're missing newline at the end.

fixed.

PR #701, opened 2023-10-08
improve handling of the prepared ds path and other cfg defaults

winglian:
Defaults to saving the prepared dataset, even if it's not going to be used unless explicitly called
warns in red when calling prepare_ds_only without a prepared dataset path set
disable sample packing when doing inference


NanoCode012:
Doesn't this revert the no cache behavior when prepared dataset config is missing?

winglian:
it writes the cache, it just doesn't use it. Which I think is safer in most cases, especially if you're explicitly running prepare_ds_only. Like if you're running that command, and not caching the dataset, it serves no purpose, since that's the expectation of running prepare.

PR #707, opened 2023-10-09
improve: Enhance code readability of prompt_tokenizers.py

seungduk-yanolja:
# Description

## Motivation and Context

As many tokenzer variations are added, its complexity is increasing and it is hard to understand what each code block means.
We better clean up duplicated code and enhance code readability.

## How has this been tested?

I don't have Cuda devices so cannot run tests. It would be great if someone can run tests for me.

## Changes

1. **Removed Duplication**:
    - Removed the redundant `_tokenize` method which was present twice.
    - Unified the logic of handling user and assistant roles in tokenization.
    
2. **Simplified Empty Text Handling**:
    - Streamlined the empty text tokenization process.
    
3. **Code Clarifications**:
    - Simplified the checks for adding eos tokens and stripping bos tokens, making it more readable.
    - Unified the constant `IGNORE_INDEX` replacing varied constants like `-100`.
    
4. **Handling Unexpected Cases**:
    - Enhanced error handling for unexpected roles.

5. **Clean-up**:
    - Removed unnecessary imports.
    - Simplified several parts of the code to improve readability.

## Types of changes

Readability improvement

### DIFF
--- a/src/axolotl/prompt_strategies/completion.py
+++ b/src/axolotl/prompt_strategies/completion.py
@@ -53,8 +53,8 @@ class CompletionPromptTokenizingStrategy(InstructionPromptTokenizingStrategy):
             tokenized_full_prompt = self._tokenize(full_prompt)
 
             for key, val in tokenized_full_prompt.items():
-                for i in range(0, len(val), self.max_length):
-                    res[key].append(val[i : i + self.max_length])
+                for i in range(0, len(val), self.sequence_len):
+                    res[key].append(val[i : i + self.sequence_len])
 
         return dict(res)
 
--- a/src/axolotl/prompt_strategies/completion.py
+++ b/src/axolotl/prompt_strategies/completion.py
@@ -53,8 +53,8 @@ class CompletionPromptTokenizingStrategy(InstructionPromptTokenizingStrategy):
             tokenized_full_prompt = self._tokenize(full_prompt)
 
             for key, val in tokenized_full_prompt.items():
-                for i in range(0, len(val), self.sequence_len):
-                    res[key].append(val[i : i + self.sequence_len])
+                for i in range(0, len(val), self.max_length):
+                    res[key].append(val[i : i + self.max_length])
 
         return dict(res)
 
--- a/src/axolotl/prompt_tokenizers.py
+++ b/src/axolotl/prompt_tokenizers.py
@@ -45,7 +45,6 @@ class PromptTokenizingStrategy(abc.ABC):
         self.prompter = prompter
         self.tokenizer: PreTrainedTokenizer = tokenizer
         self.train_on_inputs = train_on_inputs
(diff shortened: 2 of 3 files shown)


src/axolotl/prompt_strategies/completion.py
```diff
@@ -53,8 +53,8 @@ def tokenize_prompt(self, prompt):
             tokenized_full_prompt = self._tokenize(full_prompt)
 
             for key, val in tokenized_full_prompt.items():
-                for i in range(0, len(val), self.sequence_len):
-                    res[key].append(val[i : i + self.sequence_len])
+                for i in range(0, len(val), self.max_length):
+                    res[key].append(val[i : i + self.max_length])
```
winglian: there is a difference in the completion prompts between the sequence len and the max length as they aren't the same/interchangeable in this case. the completion prompter tokenizes at a longer length and then splits the text into correct length

NanoCode012: Perhaps, we should also document this (as comments?) as it would be confusing to others as they read the code.

seungduk-yanolja: Could you please provide more details? I'm still having trouble understanding. I will leave a comment here if needed. If the difference exists only in the completion tokenizing strategy, doesn't it have to be here instead of its parent class?

seungduk-yanolja: reverted the change that I unified sequence_len and max_length. PTAL

src/axolotl/prompt_tokenizers.py
```diff
@@ -422,38 +425,6 @@ def tokenize_prompt(self, prompt):
...
 
-    def _tokenize(self, prompt, add_eos_token=True, strip_bos_token=False):
-        if not prompt.strip():
-            LOG.warning("Empty text requested for tokenization.")
-            result = BatchEncoding(data={"input_ids": [], "attention_mask": []})
-        else:
-            result = self.tokenizer(
-                prompt,
-                truncation=True,
-                max_length=self.sequence_len,
```
seungduk-yanolja, line 435: Please note that it was `sequence_len` before but the merged `_tokenize` method will use `max_length`. They were different.

src/axolotl/prompt_strategies/completion.py
seungduk-yanolja: @winglian do you still have a concern or can you review this PR?

winglian:
@seungduk-yanolja looks good! thank you! I'll work on resolving the merge conflicts tomorrow.

winglian (approved):

winglian:
@seungduk-yanolja I rebased your PR over current main, so will get this merged once the tests pass. thanks again!

seungduk-yanolja:
I had to sort out these things before sorry and thank you 🙏

PR #708, opened 2023-10-09
Get qlora mistral-7b fine tuning working on a single 4090

lukemarsden:
Sets micro_batch_size from 4 to 2 in the `mistral/qlora.yml` example. With this change, qlora fine tuning of mistral-7b fits on a single 4090, per https://twitter.com/Teknium1/status/1709750388528939473.

I tried a few different values of `micro_batch_size` and `gradient_accumulation_steps`, as suggested by the readme, but I'm not an expert so please advise if there's a better way to do this.

image
image
Screenshot from 2023-10-09 17-15-49

### DIFF
--- a/examples/mistral/qlora.yml
+++ b/examples/mistral/qlora.yml
@@ -43,7 +43,7 @@ wandb_run_id:
 wandb_log_model:
 
 gradient_accumulation_steps: 4
-micro_batch_size: 4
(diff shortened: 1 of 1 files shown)


NanoCode012 (approved):
Sure, this would be helpful for most home users.

PR #710, opened 2023-10-09
Added couple switches for inference:

githubbar:
--prompt_file="prompt_test.json" specify a json file with multiple prompts. Format {"instruction": "answer"}
{"instruction2": "answer2"}

generation_cfg="generation_config.json". JSON file  to control generation parameters
as per https://huggingface.co/docs/transformers/main_classes/text_generation

# Description


src/axolotl/cli/__init__.py
```diff
@@ -85,6 +86,14 @@ def do_inference(
     cfg: DictDefault,
     cli_args: TrainerCliArgs,
 ):
+    
+    # Support for several multiline prompts from JSON file
+    nPrompt = 0
```
seungduk-yanolja: This codebase is using the snake case for variable names

src/axolotl/cli/__init__.py
```diff
@@ -112,37 +121,52 @@ def do_inference(
...
+        if prompts:
+            if nPrompt < len(prompts):
+                instruction = prompts[nPrompt]
+                nPrompt += 1
+            else:
+                return
+        else:
+            instruction = get_multi_line_input()
+        # Run inference only on one device
+        if not instruction or cfg.device != "cuda:0":
```
seungduk-yanolja: what is the justification running inference only on a single GPU?

src/axolotl/cli/__init__.py
```diff
@@ -112,37 +121,52 @@ def do_inference(
 
     while True:
         print("=" * 80)
-        # support for multiline inputs
-        instruction = get_multi_line_input()
-        if not instruction:
+        if prompts:
+            if nPrompt < len(prompts):
+                instruction = prompts[nPrompt]
+                nPrompt += 1
```
seungduk-yanolja: looks like it is not a number of prompts but an index instead

src/axolotl/cli/__init__.py
githubbar: I get inference performed multiple times while running it with 4 GPUS: accelerate launch --multi_gpu --num_processes=4 --num_machines=1 \
--mixed_precision='no' --dynamo_backend='no' \
--module axolotl.cli.inference lora13b-chat.yml \

scripts/finetune.py
winglian, line 1: We're not ready to delete this.

scripts/runpod-entrypoint.sh
winglian, line 1: why was this deleted?

src/axolotl/common/cli.py
```diff
@@ -27,6 +27,8 @@ class TrainerCliArgs:
     merge_lora: bool = field(default=False)
     prepare_ds_only: bool = field(default=False)
     prompter: Optional[str] = field(default=None)
+    prompt_file: Optional[str] = field(default=None)
+    generation_cfg: Optional[str] = field(default=None)
```
winglian, line 31: what should the value of generation_cfg look like? this doesn't seem obvious atm

src/axolotl/cli/__init__.py
```diff
@@ -112,37 +121,52 @@ def do_inference(
 
     while True:
         print("=" * 80)
-        # support for multiline inputs
-        instruction = get_multi_line_input()
-        if not instruction:
+        if prompts:
+            if idx_prompt < len(prompts):
+                instruction = prompts[idx_prompt]
+                idx_prompt += 1
```
winglian: perhaps something like `instruction = next(prompts)` instead?

winglian:
I think the ability to quickly test against prompts from a json file is great. Let's keep this PR scoped to that and address the other issues separately. Thanks!

scripts/finetune.py
githubbar: deleted by  mistake

scripts/runpod-entrypoint.sh
githubbar: deleted by  mistake

src/axolotl/common/cli.py
githubbar: it should point to the config file "generation_config.json"

contents of the file could be:

{
  "bos_token_id": 1,
  "do_sample": true,
  "eos_token_id": 2,
  "pad_token_id": 0,
  "temperature": 0.6,  // 'randomness' of outputs, 0.0 is the min and 1.0 the max
  "top_p": 0.95,
  "repetition_penalty": 1.1,
  "max_new_tokens": 1024,  // max number of tokens to generate in the output
  "top_k": 40
}

src/axolotl/cli/__init__.py
githubbar: yes

src/axolotl/cli/__init__.py
githubbar, line 1: when using with the promtfile.json this just prints the prompt the second time.

PR #713, opened 2023-10-10
Fix: lowercase `True` values in config

atgctg:
Corrected the Mistral example to use lowercase 'true' instead of 'True' since YAML is case-sensitive.


NanoCode012:
Hey, nice catch. While on this topic: could you also fix the one in `examples/pythia/lora.yml`

atgctg:
Done!

NanoCode012:
Thank you for the PR

PR #716, opened 2023-10-11
Save Axolotl config as WandB artifact

jphme:
As discussed on discord, tested it and works fine for me (but happy for suggestions/improvements, im not a WandB expert).

### DIFF
--- a/src/axolotl/utils/callbacks.py
+++ b/src/axolotl/utils/callbacks.py
@@ -530,7 +530,11 @@ class SaveAxolotlConfigtoWandBCallback(TrainerCallback):
         **kwargs,  # pylint: disable=unused-argument
     ):
         if is_main_process():
-            artifact = wandb.Artifact(name="axolotl-config", type="config")
-            artifact.add_file(local_path=self.cfg.axolotl_config_path)
-            wandb.run.log_artifact(artifact)
+            try:
+                artifact = wandb.Artifact(name="axolotl-config", type="config")
+                artifact.add_file(local_path=self.cfg.axolotl_config_path)
+                wandb.run.log_artifact(artifact)
+                LOG.info("Axolotl config has been saved to WandB as an artifact.")
+            except (FileNotFoundError, ConnectionError) as err:
+                LOG.warning(f"Error while saving Axolotl config to WandB: {err}")
         return control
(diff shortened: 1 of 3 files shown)


src/axolotl/utils/trainer.py
```diff
@@ -775,6 +776,9 @@ def setup_trainer(cfg, train_dataset, eval_dataset, model, tokenizer, total_num_
         LogPredictionCallback = log_prediction_callback_factory(trainer, tokenizer)
         trainer.add_callback(LogPredictionCallback(cfg))
 
+    if cfg.use_wandb:
+        trainer.add_callback(SaveAxolotlConfigtoWandBCallback(cfg))
```
NanoCode012: I think you only need to pass the path in, not entire config. The class init would have to be updated too.

src/axolotl/utils/callbacks.py
```diff
@@ -514,3 +514,27 @@ def log_table_from_dataloader(name: str, table_dataloader):
...
+        state: TrainerState,  # pylint: disable=unused-argument
+        control: TrainerControl,
+        **kwargs,  # pylint: disable=unused-argument
+    ):
+        if is_main_process():
+            try:
+                artifact = wandb.Artifact(name="axolotl-config", type="config")
+                artifact.add_file(local_path=self.cfg.axolotl_config_path)
+                wandb.run.log_artifact(artifact)
+                LOG.info("Axolotl config has been saved to WandB as an artifact.")
```
NanoCode012, line 537: maybe log.debug is enough

src/axolotl/utils/trainer.py
jphme: youre right, fixed

src/axolotl/utils/callbacks.py
jphme: Im indifferent, we log/print 20+ lines of Wandb initialization (and hundreds of lines on training begin) anyway, wouldn't it make sense to leave it at info to make people aware that their config is archived - how should anyone (except us) ever find/know it otherwise?
With an explicit (short) statement after successful archivation in the log, people then know that they can e.g safely overwrite their config.yml for a new experiment.

But happy to change that as well, whats your opinion @winglian ?

PR #729, opened 2023-10-14
Clarify custom format example

casper-hansen:
The current README seems a bit unclear on specifying a custom prompt format. In this PR, I modify the example to show a common scenario where you want a generation pair of `prompt/completion`.

I get rid of the YAML formatting of `|-` as it can quickly lead to unexpected behavior. One should just use a string, although it looks less fancy, it will be easier to ensure correctness.

```
    type:
      system_prompt: ""
      field_system: system
      format: "[INST] {instruction} [/INST]"
      no_input_format: "[INST] {instruction} [/INST]"
```

There are other more complicated formats that I choose to omit as you are better off using something that is not error-prone, especially when you are already using a custom dataset. Here is one example with Alpaca instruct, perhaps this can be included somewhere else in the docs:

```
    type:
      system_prompt: ""
      field_system: system
      format: "\n### Instruction: \n{instruction}\n### Output: \n"
      no_input_format: "\n### Instruction: \n{instruction}\n### Output: \n"

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -297,7 +297,13 @@ Have dataset(s) in one of the following format (JSONL recommended):
 
 #### How to add custom prompts
 
-If you want an input/output type of format, you can define a custom prompt in your YAML:
+For a dataset that is preprocessed for instruction purposes:
+
+```json
+{"instruction": "...", "output": "..."}
+```
+
+You can use this example in your YAML config:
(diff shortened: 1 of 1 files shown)


winglian (approved):
thanks!

PR #765, opened 2023-10-22
Feat: Add dataset loading from S3, GCS

### LINKED ISSUE #750: Support Cloud Storage datasets

### 🔖 Feature description

https://huggingface.co/docs/datasets/filesystems#cloud-storage

### ✔️ Solution

TBD

### ❓ Alternatives

_No response_

### 📝 Additional Context

_No response_

### Acknowledgements

NanoCode012:
Closes #750

This allows loading from S3/GCS.

Usage: `path: s3://path/to/data.jsonl` or `path: s3://path/to/parquet/dir/`. Use same convention as local dataset, passing `ds_type` where appropriate.

Test:
- [x] S3
- [ ] GCS

Note: Azure is commented out till we figure out how to pass creds securely

### DIFF
--- a/src/axolotl/utils/data.py
+++ b/src/axolotl/utils/data.py
@@ -170,30 +170,74 @@ def load_tokenized_prepared_datasets(
             except (FileNotFoundError, ConnectionError):
                 pass
 
+            ds_from_cloud = False
+            storage_options = {}
+            remote_file_system = None
+            if config_dataset.path.startswith("s3://"):
+                try:
+                    import aiobotocore.session  # type: ignore
+                    import s3fs  # type: ignore
+                except ImportError as exc:
+                    raise ImportError(
+                        "s3:// paths require aiobotocore and s3fs to be installed"
+                    ) from exc
+
+                # Takes credentials from ~/.aws/credentials for default profile
+                s3_session = aiobotocore.session.AioSession(profile="default")
+                storage_options = {"session": s3_session}
+                remote_file_system = s3fs.S3FileSystem(**storage_options)
+            elif config_dataset.path.startswith(
+                "gs://"
+            ) or config_dataset.path.startswith("gcs://"):
+                try:
+                    import gcsfs  # type: ignore
+                except ImportError as exc:
+                    raise ImportError(
+                        "gs:// or gcs:// paths require gcsfs to be installed"
+                    ) from exc
+
+                # gcsfs will use default credentials from the environment else anon
+                # https://gcsfs.readthedocs.io/en/latest/#credentials
+                storage_options = {"token": None}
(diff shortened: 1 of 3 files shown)


src/axolotl/utils/data.py
```diff
@@ -161,30 +161,69 @@ def for_d_in_datasets(dataset_configs):
...
                 if local_path.is_dir():
-                    # TODO dirs with arrow or parquet files could be loaded with `load_from_disk`
-                    ds = load_dataset(
-                        d.path,
-                        name=d.name,
-                        data_files=d.data_files,
-                        streaming=False,
-                        split=None,
-                    )
+                    ds = load_from_disk(d.path)
```
NanoCode012: I saw a TODO here and swapped to use `load_from_disk`. I don't think the previous code was correct?

src/axolotl/utils/data.py
```diff
@@ -165,30 +165,74 @@ def for_d_in_datasets(dataset_configs):
...
                 if local_path.is_dir():
-                    # TODO dirs with arrow or parquet files could be loaded with `load_from_disk`
-                    ds = load_dataset(
-                        config_dataset.path,
-                        name=config_dataset.name,
-                        data_files=config_dataset.data_files,
-                        streaming=False,
-                        split=None,
-                    )
+                    ds = load_from_disk(config_dataset.path)
```
winglian, line 237: I don't believe this is equivalent, since the previous `load_dataset` call also has a data_files component to only load specific files from the path.

jphme:
Very cool feature! Would it in theory be possible to extend the same code/approach also for model loading and checkpoint saving?

NanoCode012:
@jphme , thanks! That's my next plan. However, this approach relies partially on `dataset` implementation for cloud loading. We would need a separate (maybe similar) pipeline for model saving.

casper-hansen:
How will we differentiate between downloading vs streaming from your cloud storage in the future when streaming is added? Could we just use the same path but differentiate between the two somehow?

NanoCode012:
> How will we differentiate between downloading vs streaming from your cloud storage in the future when streaming is added? Could we just use the same path but differentiate between the two somehow?

Perhaps, we might have a `stream: true` parameter? Also, if I recall correctly, streaming would be with the `Streaming` library, right?

src/axolotl/utils/data.py
gordicaleksa: This indeed broke my workflow: I have a directory with txt files inside, the `load_dataset` knew how to handle that, `load_from_disk` raises an exception

NanoCode012: @gordicaleksa , could you please open an Issue along with a sample config? It'll make it easier to remember this.

PR #767, opened 2023-10-22
Feat(wandb): Refactor to be more flexible

### LINKED ISSUE #236: Improve documentation and implementation of W&B options
1. These options are essentially mapping to the W&B environment variables, for their usage, users need to search and check https://docs.wandb.ai/guides/track/environment-variables#optional-environment-variables , https://docs.wandb.ai/ref/python/watch etc. It might be good to add this link to the readme, and suggest a common usage
2. The implementation could be optimized to just automatically map anything that starts with `wandb_` to the same env var

discord discussion: https://discord.com/channels/1104757954588196865/1111279858136383509/1121703569029091412

NanoCode012:
Closes https://github.com/OpenAccess-AI-Collective/axolotl/issues/236

Breaking change:
- Swap `wandb_run_id` -> `wandb_name` following wandb eng's recommendation https://github.com/OpenAccess-AI-Collective/axolotl/issues/236#issuecomment-1728317747 . I have added a warning and backward compatible fix for now.
- Removes`WANDB_DISABLED` env if `wandb_project` passed. This should be the expected behavior.

New feature:
- Now allows any `wandb_` config to be passed to env for more flexibility
- More tests!

### DIFF
--- a/README.md
+++ b/README.md
@@ -659,7 +659,7 @@ wandb_mode: # "offline" to save run metadata locally and not sync to the server,
 wandb_project: # Your wandb project name
 wandb_entity: # A wandb Team name if using a Team
 wandb_watch:
-wandb_run_id: # Set the name of your wandb run
+wandb_name: # Set the name of your wandb run
 wandb_log_model: # "checkpoint" to log model to wandb Artifacts every `save_steps` or "end" to log only at the end of training
 
 # Where to save the full-finetuned model to
@@ -952,7 +952,7 @@ wandb_mode:
 wandb_project:
 wandb_entity:
 wandb_watch:
-wandb_run_id:
+wandb_name:
 wandb_log_model:
 ```
 
--- a/README.md
+++ b/README.md
@@ -660,6 +660,7 @@ wandb_project: # Your wandb project name
 wandb_entity: # A wandb Team name if using a Team
 wandb_watch:
 wandb_name: # Set the name of your wandb run
+wandb_run_id: # Set the ID of your wandb run
 wandb_log_model: # "checkpoint" to log model to wandb Artifacts every `save_steps` or "end" to log only at the end of training
 
 # Where to save the full-finetuned model to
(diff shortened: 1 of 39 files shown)


winglian (commented):
Does this PR make a major difference since it still points to the same keyword arg here?

NanoCode012:
Yes!

- it should allow other wandb env without us having to manually add.
- we use run_id and instead of letting wandb automatically generate that. The Engineer said this wasn't good to do. They recommended using name.
- fixed an unexpected issue with conflict of `WANDB_DISABLED`

Edit: I noticed one edge case from reading their docs on run_id. During resume, it might not continue in the same training and show as new/separate one.

IgnacioFDM:
Is resume fixed with this? Currently it starts overriding old data (because step begins from 0 after resuming I think).

NanoCode012:
@IgnacioFDM , yes! That should fix this. Previously, we set `run_id` which is their internal tracking method. Now, we use `name`. However, I'm not sure if this will create a new separate run or actually resume. If anyone has the bandwidth, feel free to do a quick test.

IgnacioFDM:
Just tested it and each time you resume, it creates a new run (with the same name) on wandb.

NanoCode012:
@tcapelle , thanks for your code snippet a while back and sorry for this late PR. Following your feedback to use `wandb_name` instead of `wandb_run_id`, may I ask if you have any idea how to perform a resume in such a case? According to the comment above, it creates a new run instead.

tcapelle:
Ohh great! Yeah, for resuming, you need the id, so you would need to keep both attributed. Sorry for missing that.

IgnacioFDM:
I tried setting the WANDB_RUN_ID and it works like it did before this PR: it continues the run, but still leads to wrong data (either new steps don't show up, or they show up but it starts deleting earlier steps).

I think the issue arises from the fact that step (not global_step) resets to 0 when you resume.

NanoCode012:
Rebased! Updated it to support both `wandb_name` and `wandb_run_id` instead of forcing all to name following idea from #824

winglian (approved):

PR #768, opened 2023-10-22
Feat: Add tensorboard support

### LINKED ISSUE #763: Support tensorboard

### 🔖 Feature description

Support tensorboard as an alternative to wandb.

### ✔️ Solution

Before I started using axolotl I swore never to use wandb again. Needs access to server, annoying retry loop that sometimes breaks, doesn't log all steps and misses spikes. I was just reminded that I'd like to use tensorboard when it couldn't save the trained model because no memory left on device (wandb .cache folder is 500GB).

### ❓ Alternatives

_No response_

### 📝 Additional Context

_No response_

### Acknowledgements

NanoCode012:
Should close #763 

Untested, but should work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -573,9 +573,12 @@ relora_cpu_offload: # True to perform lora weight merges on cpu during restarts,
 wandb_mode: # "offline" to save run metadata locally and not sync to the server, "disabled" to turn off wandb
 wandb_project: # Your wandb project name
 wandb_entity: # A wandb Team name if using a Team
-wandb_watch:
+wandb_watch: # Set whether to log your models gradients, parameters or neither
 wandb_run_id: # Set the name of your wandb run
 wandb_log_model: # "checkpoint" to log model to wandb Artifacts every `save_steps` or "end" to log only at the end of training
+
+# tensorboard
+tensorboard: # Set to true to enable
 
 # Where to save the full-finetuned model to
 output_dir: ./completed-model
--- a/requirements.txt
+++ b/requirements.txt
@@ -31,3 +31,4 @@ scikit-learn==1.2.2
 pynvml
 art
 fschat==0.2.29
+tensorboard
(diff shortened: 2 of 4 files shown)


winglian (approved):

winglian:
based on the docs, perhaps we should convert this to `None` if it's empty? that way it should default back to the `all` case if not setup
```
report_to (str or List[str], optional, defaults to "all") — The list of integrations to report the 
results and logs to. Supported platforms are "azure_ml", "clearml", "codecarbon", 
"comet_ml", "dagshub", "flyte", "mlflow", "neptune", "tensorboard", and "wandb". Use 
"all" to report to all integrations installed, "none" for no integrations.
```

ein-ich:
Ok I tested it and it works; but I realized it even works on main and without `tensorboard: true`. It seems it always created a tensorboard `runs` logdir in `output_dir`, and I just never noticed?

winglian:
> Ok I tested it and it works; but I realized it even works on main and without `tensorboard: true`. It seems it always created a tensorboard `runs` logdir in `output_dir`, and I just never noticed?

Maybe it never did that until you installed the tensorboard package?

ein-ich:
Yes you're right, it only creates the logs when tensorboard is installed.

NanoCode012:
> Ok I tested it and it works; but I realized it even works on main and without `tensorboard: true`. It seems it always created a tensorboard `runs` logdir in `output_dir`, and I just never noticed?

Hmm, that's weird. It should've been `report_to=None` if not using wandb, which according to the docs would not have any integration. Maybe tensor board logs are just defaulted? In such a case, is this PR needed?

ein-ich:
grafik
(from https://huggingface.co/docs/transformers/v4.34.1/en/main_classes/callback)

winglian:
Yeah, I think the only thing we need is to add tensorboard to the `requirements.txt` or to document that people should install it if wanted. I'm leaning towards just installing it with requirements or maybe with an extras like `pip install -e .[tensorboard]`

NanoCode012:
Closed in favor of adding tensorboard to requirements

fakerybakery:
Hi, sorry I'm a bit confused, how do I use Tensorboard in Axolotl then?

NanoCode012:
@fakerybakery , you can just train normally without setting wandb configs or setting it to disabled

Issue #752, opened 2023-10-21
Keyword "assistant" Error with ShareGPT Datasets


MilesQLi:
### ⚠️ Please check that this feature request hasn't been suggested before.

- [X] I searched previous [Ideas in Discussions](https://github.com/OpenAccess-AI-Collective/axolotl/discussions/categories/ideas) didn't find any similar feature requests.
- [X] I searched previous [Issues](https://github.com/OpenAccess-AI-Collective/axolotl/labels/enhancement) didn't find any similar feature requests.

### 🔖 Feature description

On ShareGPT datasets such as shibing624/sharegpt_gpt4, some "from" values are "assistant" rather than "gpt", so, the program raises the keyword error.

### ✔️ Solution

Maybe take both assistant and gpt as the machine's responses?

### Acknowledgements

- [X] My issue title is concise, descriptive, and in title casing.
- [X] I have searched the existing issues to make sure this feature has not been requested yet.
- [X] I have provided enough information for the maintainers to understand and evaluate this request.

NanoCode012:
I checked that linked dataset. It seems to be human/gpt.

In case you want to swap keys, please check this. 

https://github.com/OpenAccess-AI-Collective/axolotl/blob/21cf09b60840ee03ba3ee4e57c2707d2296f532c/src/axolotl/prompt_strategies/sharegpt.py#L78-L90

It swaps roles using a simple dictionary map.

MilesQLi:
@NanoCode012 I said "some "from" values are "assistant" rather than "gpt"". You checked only some data from that dataset, not all.

NanoCode012:
Oh I see! You can use this kind of role_map then. 

```py
role_map = {"assistant": "gpt", "human": "human", "gpt": "gpt"} 
```

MilesQLi:
@NanoCode012 Thanks! I'm not sure I'm doing it the way you mentioned. I updated the role_map, but still get the error. Please see the two screenshots.
![image](https://github.com/OpenAccess-AI-Collective/axolotl/assets/11975133/3002089d-a351-4b60-84d4-11efa3b56175)
![image](https://github.com/OpenAccess-AI-Collective/axolotl/assets/11975133/0aafd2f3-0ee6-4996-bd5f-121bc0abf0de)

MilesQLi:
I fixed this by updating SimpleShareGPTPromptTokenizingStrategy with the role_map you provided.

PR #774, opened 2023-10-22
Update to adapt to sharegpt datasets with "assistant" rather than "gp…

MilesQLi:
…t" as the machine answers.

This fixed issue #752.


winglian:
My preference is to have this be a separate prompter, since the underlying dataset is the issue here. It's better to flag the dataset as incorrect through this process rather than ignore the issue.

winglian:
@MilesQLi I updated your branch/PR so that simply adding the `strict: false` option will enable this behavior
```
  type: sharegpt
  strict: false
```

PR #785, opened 2023-10-25
Create preprocess CLI

casper-hansen:
- remove `--prepare_ds_only`
- use `python -m axolotl.cli.preprocess your_config.yml` to preprocess
- when using `--debug`, print raw prompt template when it is supported
- modified readme to include new CLI and update the multi-GPU section


winglian:
I noticed that running it the first time prints the prompter,
Screenshot 2023-10-26 at 8 46 20 AM
but running it again right after doesn't
Screenshot 2023-10-26 at 8 46 46 AM

winglian (approved):
Other than the prompter printing, lgtm. thanks!

casper-hansen:
> I noticed that running it the first time prints the prompter,
> Screenshot 2023-10-26 at 8 46 20 AM
> but running it again right after doesn't
> Screenshot 2023-10-26 at 8 46 46 AM

That’s because we can’t print a pre-tokenized dataset. That would take a bit of work to do and considered out of scope for this PR.

PR #786, opened 2023-10-26
support for explicit test_dataset definition for evals

winglian:
first pass for supporting different datasets for evals rather than splitting the test dataset.

### DIFF
--- a/src/axolotl/utils/data.py
+++ b/src/axolotl/utils/data.py
@@ -4,7 +4,7 @@ import hashlib
 import logging
 from collections import defaultdict
 from pathlib import Path
-from typing import Dict, List, Tuple, Union
+from typing import Dict, List, Optional, Tuple, Union
 
 import torch
 from datasets import (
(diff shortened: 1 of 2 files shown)


src/axolotl/utils/data.py
```diff
@@ -239,8 +251,9 @@ def for_d_in_datasets(dataset_configs):
                 raise ValueError("unhandled dataset load")
             # support for using a subset of the data
             if config_dataset.shards:
-                if "train" in ds:
-                    ds = ds.shuffle(seed=seed)["train"].shard(
+                shard_split = config_dataset.split if config_dataset.split else split
```
NanoCode012: Maybe we can refactor this code to split first (using the code around L270-275), then shard after?

NanoCode012 (commented):
Hmm, our data pipeline is getting a bit complex. I had a bit of hard time following the flow

NanoCode012 (commented):
Adding this config to Readme would be helpful

winglian:
This will address #875

NanoCode012:
I think will need to add to doc about this and also, whether it would be appropriate to hardcode the `train`/`test` split or allow more open control?

winglian:
> I think will need to add to doc about this and also, whether it would be appropriate to hardcode the `train`/`test` split or allow more open control?

@NanoCode012 What do you mean about hardcoding?

DreamGenX:
Do you have an example or documentation of how this can be used?

DreamGenX:
I was trying to reserve engineer how it's supposed to work, and maybe there's a bug here:

https://github.com/OpenAccess-AI-Collective/axolotl/blob/18f811978c01d567c2294140f53abcf8c086e337/src/axolotl/utils/data.py#L443

```py
    dataset, prompters = load_tokenized_prepared_datasets(
        tokenizer, cfg, default_dataset_prepared_path
    )
```

Shouldn't you pass `split=split` to `load_tokenized_prepared_datasets`? Otherwise it will never load from `cfg.test_datasets`.

PR #787, opened 2023-10-26
Feat: Warns to add to modules_to_save when adding tokens or switching special_tokens

NanoCode012:


src/axolotl/utils/config.py
```diff
@@ -369,6 +369,22 @@ def validate_config(cfg):
             "If you want to full finetune, please turn off load_in_8bit and load_in_4bit."
         )
 
+    if (
+        cfg.adapter
+        and cfg.tokens
```
winglian, line 453: so this gets tricky, since I think in the case of `<|im_start|>` and `<|im_end|>`; `<|im_end|>` is usually a special token and doesn't need to be defined under `tokens`

NanoCode012: That is tricky.. That means we need to detect if the special_tokens is new also..

NanoCode012: This is now checked in latest update

NanoCode012:
The code now errors when the config is incorrect in these adapter training cases:
- adding new tokens without proper modules_to_save
- changing special tokens without proper modules_to_save

**Breaking change**: This means that all training configs without proper modules_to_save won't work now, which should be as intended as they may have been incorrect.

winglian (approved):

PR #791, opened 2023-10-27
Fix Deepspeed Zero3 Config

teknium1:
Update DS Zero 3 conf to not use CPU Offload (because it dramatically slows things down, just reduce batchsize unless you cant fit bs 1) and replace LR Scheduler with a properly decaying one - with constant LR scheduler, it will cause loss to increase after each epoch

Compare dark pink (decaying LR Schedule) with all others, including blue, which used constant LR Schedule (the default in this conf)
image

### DIFF
--- a/deepspeed/zero3.json
+++ b/deepspeed/zero3.json
@@ -1,14 +1,5 @@
-{
   "zero_optimization": {
     "stage": 3,
-    "offload_optimizer": {
-      "device": "cpu",
-      "pin_memory": true
-    },
-    "offload_param": {
-      "device": "cpu",
-      "pin_memory": true
-    },
     "overlap_comm": true,
     "contiguous_gradients": true,
     "sub_group_size": 0,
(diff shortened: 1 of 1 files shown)


casper-hansen:
Probably a good idea to remove CPU offload in zero2 as well. In my tests, the offloading is very minimal and if you turn it off, you get 2x training speed.

https://github.com/OpenAccess-AI-Collective/axolotl/blob/c25ba7939b35dbd9589bc694ea06c3490e8f9b54/deepspeed/zero2.json#L4-L6

winglian (approved):

teknium1:
> Probably a good idea to remove CPU offload in zero2 as well. In my tests, the offloading is very minimal and if you turn it off, you get 2x training speed.
> 
> https://github.com/OpenAccess-AI-Collective/axolotl/blob/c25ba7939b35dbd9589bc694ea06c3490e8f9b54/deepspeed/zero2.json#L4-L6

Yea I think it is only a good idea if and when you cant fit even bs 1 on vram

casper-hansen:
> Yea I think it is only a good idea if and when you cant fit even bs 1 on vram

Yes, then it makes sense. Perhaps an idea is to remove it from zero2 and add a disclaimer about it in the REAMDE?

PR #795, opened 2023-10-27
multipack w batch sampler 

### LINKED ISSUE #406: sample packing with resume_from_checkpoint

- [X] I searched previous [Bug Reports](https://github.com/OpenAccess-AI-Collective/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

resume should step over the number of batches already trained and resume from the same point in the data

### Current behaviour

fails to load due to the MultipackDataloader not having a batch sampler, Setting the batch sampler to be the same as the sampler results in an issue in acclerate.data_loader reloading the dataset into a new dataloader in an attempt to skip over the batches
```
    dataloader = DataLoader(dataset, batch_sampler=new_batch_sampler, **kwargs)                                                                                                                                                                                                        
  File "/root/miniconda3/envs/py3.10/lib/python3.10/site-packages/torch/utils/data/dataloader.py", line 245, in __init__                   
    raise ValueError('prefetch_factor option could only be specified in multiprocessing.'                                                  
ValueError: prefetch_factor option could only be specified in multiprocessing.let num_workers > 0 to enable multiprocessing, otherwise set prefetch_factor to None.                   
```

### Steps to reproduce

see above

### Possible solution

need to figure out a way to refactor the current dataloader either as a better sampler or extend the real torch dataloader

- [ ] Android
- [ ] iPhone/iPad
- [X] Linux
- [ ] macOS
- [ ] Windows

### Python Version

3.10

### axolotl branch-commit

main

### Acknowledgements

- [X] I have searched the existing issues to make sure this bug has not been reported yet.
- [X] I am using the latest version of axolotl.
- [X] I have provided enough information for the maintainers to reproduce and diagnose the issue.

winglian:
resolves #406


casper-hansen:
Could this and should this replace the DataLoader? The sampler looks like a simple version of the Multipack DataLoader.

winglian:
Yeah. I want to rip out the previous dataloader. Yeah, it's basically the sampler but adapted to subclass the batch sampler and the various fixes needed to use uneven batches

casper-hansen:
So Multipack can be implemented as a sampler and then we can use the standard DataLoader from PyTorch? Probably good to make the full adjustment in this PR if we go that way

winglian:
- eval sample packing issues
- 32k context issues
- remove old dataloader
- adamw_bnb_8bit optimizer issue (??)

PR #799, opened 2023-10-28
cleanup verbosity a bit

winglian:
some of the logging is a bit excessive. output some things on main rank only, move some to debug level


NanoCode012:
Would it be cleaner to have a `log_zero_only` function instead?

PR #804, opened 2023-10-30
adalomo optimizer support

winglian:


AlpinDale:
I'm not terribly familiar with AdaLOMO (took a quick look at the algorithm 1 in the paper), but it seems like you're detaching `tr_loss` and then deleting it:

```py
           should_stop = False
            if self.control.should_epoch_stop or self.control.should_training_stop:
                should_stop = True
            if self.control.should_log:
                device = tr_loss.device
                tr_loss.detach()
                del tr_loss
                tr_loss = torch.tensor(0.0, requires_grad=True).to(device)

            return tr_loss, should_stop
```

This removes `tr_loss` from the computation graph but doesn't reset/remove the entire graph itself. Then you create a new `tr_loss` tensor:

```py
tr_loss = torch.tensor(0.0, requires_grad=True).to(device)
```

But it isn't connected to the rest of the computation graph in any way (since it's just a standalone tensor you've created from scratch), so no resetting is done. As a consequence of this, when you call `loss.backward()` later on, it still backpropagates through the original compute graph and not just the `tr_loss` tensor. To reset, I would assume you'll need to make sure all tensors/operations that contribute to `tr_loss` are also reset. You can do this (probably) by:

```diff
    def fused_backward(self, loss, lr):
        """
        执行一步反向传播并更新模型的梯度。
        :param loss: 模型的loss值
        :param lr: 学习率
        """
        self.lr = lr
        if self.loss_scale:
            loss = loss * self.loss_scale
        self.step_num += 1
-        loss.backward()
+        loss.backward(retain_graph=True)
 ```
 
 Or, you can set `requires_grad=False` instead of `True` to completely exclude it from the compute graph?

winglian:
We should test, but this may no longer be needed as lomo support us upstreamed in https://github.com/huggingface/transformers/issues/29649 and it seems we already have the lomo optimizer package as an extra in setup.py

winglian:
closing since this is now supported upstream

PR #820, opened 2023-11-04
Added Mezo, revised commit

thistleknot:
realized I cut out an important section, testing to see if it's updating weights properly


thistleknot:
doesn't seem to improve tr_loss

PR #824, opened 2023-11-05
Add the option `wandb_run_name`

Haoxiang-Wang:
## Issue

Currently, `axolotl` only allows users to set `wandb_run_id`, which serves as both the run ID and run name. However, if a user runs the same script multiple times (e.g., with different random seeds), the new run will overwrite the old one on WandB since the `wandb_run_id` is the same and WandB only allows unique `run_id`. Another case occurs when a user deletes a previous run on WandB and launches a new run with the same `wandb_run_id`; WandB will throw errors and prevent the script from running successfully (see https://github.com/wandb/wandb/issues/1192).

## Solution

To resolve these issues, I have added a new option, `wandb_run_name`, which sets only the run name. WandB allows multiple runs with the same run names, so the aforementioned issues are resolved. When `wandb_run_name` is set and `wandb_run_id` is not, WandB will create a random unique new ID for the new run.

## Backward Compatibility

For backward compatibility, that is, when `wandb_run_id` is set and `wandb_run_name` is not, the run name will still be the same as `wandb_run_id`.


winglian:
@NanoCode012 maybe this is a slightly more correct version of #767 since it includes both run name and run id? I think #767 is probably better in some respects since it has more validation and also improved logic to set the env vars.

NanoCode012:
If the author for this PR is interested, they can take the validation tests and env refactor from my previous PR to here. I agree having both id/name is better.

The one thing that I think should change in my old PR is that the config validation should only warn `if wandb_run_id and not wandb_name`, so that old users are alerted to update their old configs.

Haoxiang-Wang:
> If the author for this PR is interested, they can take the validation tests and env refactor from my previous PR to here. I agree having both id/name is better.
> 
> The one thing that I think should change in my old PR is that the config validation should only warn `if wandb_run_id and not wandb_name`, so that old users are alerted to update their old configs.

Thanks for the pointer! I will take a look at your PR later.

NanoCode012:
This is now fixed in https://github.com/OpenAccess-AI-Collective/axolotl/pull/767

PR #828, opened 2023-11-05
mezo

thistleknot:
Tested evaluation metrics print properly without taking too much room whether and regression testing all the way through.
Allows eval loss to be evaluated when save_steps and eval_steps are set to a float or empty.
image


thistleknot:
train loss does not seem to be going down :/

PR #832, opened 2023-11-05
mezo

thistleknot:
confirmed weights were being updated


thistleknot:
doesn't seem to be improving model

PR #841, opened 2023-11-10
cleanup the old multipack dataloader

winglian:


src/axolotl/utils/data.py
```diff
@@ -445,14 +446,13 @@ def load_prepare_datasets(
         train_fingerprint = md5(to_hash_train)
         test_fingerprint = md5(to_hash_test)
 
-        with zero_first(is_main_process()):
```
NanoCode012, line 448: Is there a reason we had this `with` context previously? Is it for caching?

winglian: yeah, because the function it is in is only ever called in a zero_first context, this is a bit redundant.

PR #842, opened 2023-11-10
support for fp8

winglian:
can now train fp8 with `fp8: true`
- doesn't support sample packing
- doesn't work with deepspeed. this is due to how accelerate wraps the model w deepspeed first, so doesn't wrap it later for fp8 (we could probably manually wrap it beforehand)
- doesn't work with flash attention
- the model weights are still in bf16, as torch complains about missing storage when using torch.float8_e5m2
```
  File "/workspace/axolotl/src/axolotl/utils/models.py", line 250, in load_model
    model = LlamaForCausalLM.from_pretrained(               
  File "/root/miniconda3/envs/py3.10/lib/python3.10/site-packages/transformers/modeling_utils.py", line 3051, in from_pretrained
    dtype_orig = cls._set_default_torch_dtype(torch_dtype)
  File "/root/miniconda3/envs/py3.10/lib/python3.10/site-packages/transformers/modeling_utils.py", line 1222, in _set_default_torch_dtype
    torch.set_default_dtype(dtype)
  File "/root/miniconda3/envs/py3.10/lib/python3.10/site-packages/torch/__init__.py", line 661, in set_default_dtype
    _C._set_default_dtype(d)
TypeError: couldn't find storage object Float8_e5m2Storage
```

### DIFF
--- a/src/axolotl/core/trainer_builder.py
+++ b/src/axolotl/core/trainer_builder.py
@@ -483,6 +483,10 @@ class HFCausalTrainerBuilder(TrainerBuilderBase):
         training_arguments_kwargs["fp16"] = (
             self.cfg.fp16 and not self.cfg.bf16
         ) or False
+        if self.cfg.fp8:
+            training_arguments_kwargs["fp16"] = False
+            training_arguments_kwargs["bf16"] = False
+
         training_arguments_kwargs["tf32"] = self.cfg.tf32
         training_arguments_kwargs["warmup_steps"] = warmup_steps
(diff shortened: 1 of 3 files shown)


sbhavani:
@winglian a few updates on TE/FP8, it'd be great to see FP8 integrated into axolotl!

1. DS+FP8 support is in progress in accelerate and should be ready in the next month
2. TE supports FlashAttention-2
3. TE supports model weights in FP8 ([Float8Tensor](https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/pytorch/float8_tensor.py))

PR #851, opened 2023-11-15
Update data.py for signature generation

MilesQLi:
Change of conversation formatting type should also trigger updating the preprocessed dataset, so it should be part of the signature.

### DIFF
--- a/src/axolotl/utils/data.py
+++ b/src/axolotl/utils/data.py
@@ -99,7 +99,12 @@ def load_tokenized_prepared_datasets(
                 str(cfg.sequence_len)
                 + "@"
                 + "|".join(
-                    sorted([f"{d.path}:{d.type}:{d.shards}:{d.conversation}" for d in cfg.datasets])
+                    sorted(
+                        [
+                            f"{d.path}:{d.type}:{d.shards}:{d.conversation}"
+                            for d in cfg.datasets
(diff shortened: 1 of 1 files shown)


winglian (commented):
thanks, this is definitely needed. `conversation` is an optional parameter, so we need to account for when this isn't provided (I think it would be vicuna)

MilesQLi:
@winglian I tested it with a mixture of alpaca and sharegpt datasets. Even for a alpaca dataset without conversation argument, it will go through: d.conversation will be empty rather than raising an error.

src/axolotl/utils/data.py
```diff
@@ -99,7 +99,7 @@ def load_tokenized_prepared_datasets(
                 str(cfg.sequence_len)
                 + "@"
                 + "|".join(
-                    sorted([f"{d.path}:{d.type}:{d.shards}" for d in cfg.datasets])
+                    sorted([f"{d.path}:{d.type}:{d.shards}:{d.conversation}" for d in cfg.datasets])
```
winglian: linter fix
```suggestion
                    sorted(
                        [
                            f"{d.path}:{d.type}:{d.shards}:{d.conversation}"
                            for d in cfg.datasets
                        ]
                    )
```

PR #852, opened 2023-11-15
include the suffix modified string in ascii art

fpreiss:
Previously the suffix was applied to modify the `ascii_text` string. `ascii_text` was however not added to the actual print statement.


winglian (approved):
Good catch. Thanks!

Issue #672, opened 2023-10-04
Any way we can get dropout on full finetune?


enn-nafnlaus:
### ⚠️ Please check that this feature request hasn't been suggested before.

- [X] I searched previous [Ideas in Discussions](https://github.com/OpenAccess-AI-Collective/axolotl/discussions/categories/ideas) didn't find any similar feature requests.
- [X] I searched previous [Issues](https://github.com/OpenAccess-AI-Collective/axolotl/labels/enhancement) didn't find any similar feature requests.

### 🔖 Feature description

Full finetune suffers badly from epoch spikes, making any training lasting past the end of an epoch (and esp. 2 or more epochs) difficult to get any further progress from it. A deeper understanding of the data should be able to be achieved with dropout. But while there's lora_dropout, we don't have any dropout option available for full finetune.  Anyway we could get that added?

### ✔️ Solution

Add dropout

### ❓ Alternatives

lora_dropout only applies to LoRAs.

### Acknowledgements

- [X] My issue title is concise, descriptive, and in title casing.
- [X] I have searched the existing issues to make sure this feature has not been requested yet.
- [X] I have provided enough information for the maintainers to understand and evaluate this request.

NanoCode012:
Dropout feature in LoRA is due to a feature in PEFT upstream. If we want to add dropout, we would need to modify the architecture ourselves since it's not a built in thing. Not sure if that's a good idea.

Alternatively, perhaps, a Lower lr might be good for you or to experiment with schedulers?

enn-nafnlaus:
Lower LR can hide the spike but doesn't help spread out learning / prevent overconcentration of functionality in specific neurons and bottlenecking. And actually if you use a lower LR throughout the whole training (we only have limited tools for nuanced LR control over time) you actually get a worse eval loss because you start hitting epochs when you're less far along in the training process.

Is there an upstream library where it would be better to add dropout?

Another option would be L2 regularization or any of the other dropout alternatives. Through directly dropping out parts of the network (whether through traditional dropout, DropPath, or whatnot) is AFAIK the most effective means.

I know most people are using axolotl to train LoRAs, but these epoch spikes with finetuning are a big problem.

NanoCode012:
You would need to modify the architecture model code itself in modeling code. For ex, llama: https://github.com/huggingface/transformers/blob/main/src/transformers/models/llama/modeling_llama.py

However, I do not have enough expertise in updating this to add layers.

[...]

PR #853, opened 2023-11-15
allow overriding of model_config parameters from the YML

winglian:
This should allow users to update values in the model config without having to edit the JSON.

for example for the attention_dropout for #672 
```
model_config:
    attention_dropout: 0.01
```

### DIFF
--- a/src/axolotl/utils/models.py
+++ b/src/axolotl/utils/models.py
@@ -17,7 +17,6 @@ from transformers import (  # noqa: F401
     AutoTokenizer,
     BitsAndBytesConfig,
     GPTQConfig,
-    LlamaConfig,
     PreTrainedModel,
     PreTrainedTokenizerBase,
 )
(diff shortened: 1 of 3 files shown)


src/axolotl/utils/models.py
```diff
@@ -358,13 +354,7 @@ def load_model(
             "Exception raised attempting to load model, retrying with AutoModelForCausalLM"
```
NanoCode012: Should remove this LOG.error then if you're removing the this AutoModel block.

NanoCode012 (commented):
Nice clean PR! I think you'd want to add this to the docs as well. 

Should we deprecate the current rope configs and move them here also?

src/axolotl/utils/models.py
```diff
@@ -234,20 +232,17 @@ def load_model(
         ):
             model_kwargs["use_flash_attention_2"] = True
 
+    if cfg.model_config:
+        for key, val in cfg.model_config.items():
+            setattr(model_config, key, val)
```
NanoCode012: I forgot to ask if this deals with nested items such as in rope_scaling case?

NanoCode012 (commented):
One more point to update would be the config_validate for backward compatible..

src/axolotl/utils/models.py
winglian: this would simply update the entire value of `rope_scaling`, so it's not recursive

winglian:
> Should we deprecate the current rope configs and move them here also?

which rope configs are you referring to?

NanoCode012:
> > Should we deprecate the current rope configs and move them here also?
> 
> which rope configs are you referring to?

Ah, I meant `rope_scaling`. Since we now expect them in the model_overide, we should add it in our config validation check, so that it doesn't break old configs and warns user of deprecation.

PR #856, opened 2023-11-15
various bugfixes

winglian:
use latest tinyllama release
check if val_set_size is empty first
update sdp and xformers llama patches for updated upstream transformers
fix system prompt when no input
calculate total and total supervised tokens even when not sample packing

### DIFF
--- a/src/axolotl/monkeypatch/llama_attn_hijack_sdp.py
+++ b/src/axolotl/monkeypatch/llama_attn_hijack_sdp.py
@@ -25,6 +25,7 @@ def sdp_attention_forward(
     past_key_value: Optional[Tuple[torch.Tensor]] = None,
     output_attentions: bool = False,
     use_cache: bool = False,
+    padding_mask: Optional[torch.LongTensor] = None,  # pylint: disable=unused-argument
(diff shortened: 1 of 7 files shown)


NanoCode012:
Hey, I recall our conversation about some ValueError: Expected len>=0. Was that issue fixed? If not, can it be combined here.

winglian:
> Hey, I recall our conversation about some ValueError: Expected len>=0. Was that issue fixed? If not, can it be combined here.

just pushed a fix for that too. thanks!

src/axolotl/monkeypatch/llama_attn_hijack_sdp.py
```diff
@@ -25,6 +25,7 @@ def sdp_attention_forward(
     past_key_value: Optional[Tuple[torch.Tensor]] = None,
     output_attentions: bool = False,
     use_cache: bool = False,
+    padding_mask: Optional[torch.LongTensor] = None,  # pylint: disable=unused-argument
```
NanoCode012, line 28: Should you add kwarg to prevent future break on update?

NanoCode012 (approved):
Not sure if the kwargs is also needed on every other patch, but this looks good to me!

PR #859, opened 2023-11-15
pin hf transformers and accelerate to latest release, don't reinstall pytorch

winglian:


NanoCode012:
Can we remove the `pip3 install -U git+https://github.com/huggingface/peft.git` from Readme then since it's in requirements?

Secondly, `setup.py` should remove the torch logic also?

winglian:
> Can we remove the `pip3 install -U git+https://github.com/huggingface/peft.git` from Readme then since it's in requirements?
> 
> Secondly, `setup.py` should remove the torch logic also?

updated. thanks!

casper-hansen:
I think this is a good change. What do you think about pinning transformers to a specific release? Then you can upgrade on a per-need basis. This should limit breakage!

PR #861, opened 2023-11-15
make docker command more robust

winglian:


NanoCode012:
I was told `--ipc=host` would unlock shared memory size (I didn't double check). I think it would be fine to leave the top one as original to not complicate it. 

I'm not so familiar with `--privileged`.

casper-hansen:
LGTM!

PR #862, opened 2023-11-15
Docs: add instructions to 1-click launching on public clouds

concretevitamin:
Hi axolotl team,

Thanks for making this excellent package!  It's great to see Axolotl's adoption accelerating.

The SkyPilot community has been [asking](https://github.com/skypilot-org/skypilot/issues/2744) about how to easily launch Axolotl on the cloud, and the community has contributed an easy-to-use [example](https://github.com/skypilot-org/skypilot/tree/master/llm/axolotl). 

Let us know if you think the quick instructions added in this PR are helpful?

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,28 @@ Use `winglian/axolotl-runpod:main-latest` or use this [direct link](https://runp
 #### Windows
 Please use WSL or Docker!
 
+
+#### Launching on public clouds (GCP, AWS, Azure, OCI, and more)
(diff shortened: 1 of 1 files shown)


winglian (approved):
This is great! btw, I think you could simplify the spot configuration by simply adding `--auto_resume_from_checkpoints=True` to cli in the axolotl-spot.yaml, that way you only need a single config that can be shared across both.

concretevitamin:
@winglian Thanks! Updated the table of contents too.

winglian (approved):

PR #874, opened 2023-11-17
Install from git url

msaroufim:
I'm working on https://llm-efficiency-challenge.github.io/ and evaluating a bunch of submissions where axolotl is quite popular

Problem is since you don't have a pypy package people are resorting to cloning or copying your whole repo over, this is some new syntax that's a bit more convenient and if you need help setting up nightly binaries on github actions please lmk. I'm not sure most people using a library should be doing editable installs


winglian (approved):
thanks! we were hoping to publish a package to pypi, but sometimes we pin dependencies to commits, which pypi doesn't like

PR #876, opened 2023-11-17
Phi update 202311

winglian:
Phi recently changed their model names, so lets update our overrides to support sample packing.


winglian:
ran the remote code w/o sample packing and new patches w sample packing, and everything seems okay.

PR #893, opened 2023-11-24
Feat: Add warmup_ratio

NanoCode012:
Untested!

It is a hassle to do warmup_steps without knowing full steps.

If anyone is available, feel free to give this a try.

### DIFF
--- a/README.md
+++ b/README.md
@@ -675,8 +675,8 @@ gradient_accumulation_steps: 1
 micro_batch_size: 2
 eval_batch_size:
 num_epochs: 4
-warmup_steps: 100
-warmup_ratio: 0.05
+warmup_steps: 100  # cannot use with warmup_ratio
+warmup_ratio: 0.05  # cannot use with warmup_steps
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -676,6 +676,7 @@ micro_batch_size: 2
 eval_batch_size:
 num_epochs: 4
 warmup_steps: 100
+warmup_ratio: 0.05
```
winglian: Should probably remove warmup steps.line before this one then

winglian (approved):
Good to go once the example yaml is updated

README.md
NanoCode012: Hm, would it be better if I add a comment saying that they're exclusive? At the very least, we want to document this config.

NanoCode012: or did you mean, I should also update all the example yaml?

NanoCode012: I have updated readme to say they're exclusive

NanoCode012:
Tested that it runs.

PR #898, opened 2023-11-25
support for json data as completion

winglian:


NanoCode012:
Is this to train a model that outputs in json?

src/axolotl/prompt_strategies/completion.py
```diff
@@ -82,11 +96,23 @@ def load(tokenizer, cfg, ds_cfg: Optional[Dict[str, Any]] = None):
...
+
+def load_json(tokenizer, cfg):
+    strat = CompletionJSONPromptTokenizationStrategy(
+        CompletionPrompter(),
+        tokenizer,
+        True,
+        cfg.sequence_len,
+        max_length=cfg.sequence_len * 64,
+    )
+
```
NanoCode012, line 117: Seems like you're missing

```
    if ds_cfg and "field" in ds_cfg:
         strat.field = ds_cfg["field"]
```

 here as well. 

It would be good to not have to repeat this code though. Maybe it can be inherited from a base?

winglian: this shouldn't be needed for this. the objective here is to train on the entire row/dict as a json string. so it learns the inputs and outputs as json format

NanoCode012: Oh, interesting

PR #899, opened 2023-11-27
feature: loss watchdog for terminating training runs that are failing

kallewoof:
This adds a loss watchdog, which will stop the trainer if the loss exceeds the given threshold for more than `loss_watchdog_patience` steps (in a row). It is useful to prevent using a bunch of resources to keep training when the model has broken down.


winglian (approved):
lgtm

PR #902, opened 2023-11-28
ensure merged model matches the training dtype

winglian:
seems like all the lora merges are fp16. matching the training would open up bfloat16 support which should be better.

### DIFF
--- a/src/axolotl/cli/__init__.py
+++ b/src/axolotl/cli/__init__.py
@@ -72,7 +72,7 @@ def do_merge_lora(
     LOG.info("running merge of LoRA with base model")
     model = model.merge_and_unload()
     if cfg.bf16:
-        model.to(dtype=torch.float16)
+        model.to(dtype=torch.bfloat16)
     else:
         model.to(dtype=torch.float16)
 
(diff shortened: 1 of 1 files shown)


src/axolotl/cli/__init__.py
```diff
@@ -71,7 +71,10 @@ def do_merge_lora(
 
     LOG.info("running merge of LoRA with base model")
     model = model.merge_and_unload()
-    model.to(dtype=torch.float16)
+    if cfg.bf16:
+        model.to(dtype=torch.float16)
```
winglian:
```suggestion
        model.to(dtype=torch.bfloat16)
```

NanoCode012:
Perhaps, some individuals "might" want fp32? We should respect same dtype as in training. You might want to do `model.to(cfg.torch_dtype)`. This would load the one from


https://github.com/OpenAccess-AI-Collective/axolotl/blob/71b7ea3c056f15123b56fef3151b4044c80078b4/src/axolotl/utils/config.py#L73-L78

src/axolotl/cli/__init__.py
```diff
@@ -71,7 +71,10 @@ def do_merge_lora(
 
     LOG.info("running merge of LoRA with base model")
     model = model.merge_and_unload()
-    model.to(dtype=torch.float16)
+    if cfg.bf16:
+        model.to(dtype=torch.bfloat16)
+    else:
+        model.to(dtype=torch.float16)
```
winglian:
```suggestion
    model.to(dtype=cfg.torch_dtype)
```

PR #903, opened 2023-11-28
Support device_map=sequential & max_memory config parameters

brthor:
In order to train a large model, I recently had to use the combination of `device_map=sequential` and `max_memory` in order to balance the memory usage per gpu (in model parallel) manually.

- `device_map=auto` should not be forced when a value has been specified in the config
- `max_memory` allows for setting memory limits per gpu and is most useful with `device_map=sequential`

This was the simplest way I found to implement supporting these two options.

cc @winglian

### DIFF
--- a/README.md
+++ b/README.md
@@ -613,6 +613,7 @@ sample_packing_eff_est:
 total_num_tokens:
 
 # Passed through to transformers when loading the model when launched without accelerate
+# Use `sequential` when training w/ model parallelism to limit memory
 device_map:
 # Defines the max memory usage per gpu on the system. Passed through to transformers when loading the model.
 max_memory:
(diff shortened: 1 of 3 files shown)


winglian (approved):
lgtm

README.md
```diff
@@ -612,6 +612,11 @@ eval_sample_packing:
 sample_packing_eff_est:
 total_num_tokens:
 
+# Passed through to transformers when loading the model when launched without accelerate
+device_map:
```
NanoCode012, line 617: Would it be possible that you add the mention of using `sequential` here if you want to limit memory? Also, since this seems to be only for single gpu. A note here would be good.

```suggestion
# Passed through to transformers when loading the model when launched without accelerate
# Use `sequential` when training w/ model parallelism to limit memory
device_map:
```

winglian: I went ahead and updated this.

brthor: Thanks @winglian 

@NanoCode012 This is indeed for use across multiple gpus and is how I use it to control memory per device.

PR #913, opened 2023-12-04
feat: add check for quantized model

NanoCode012:
There was a case in discord when someone tried to train AWQ model using full finetune.

This adds check that when `quantization_config` is in the model's config to make sure gptq is set.

Inversely, it makes sure that when `gptq: true`, the quantization_config exists and GPTQ model.

### DIFF
--- a/src/axolotl/utils/models.py
+++ b/src/axolotl/utils/models.py
@@ -28,6 +28,27 @@ from axolotl.utils.dict import DictDefault
 LOG = logging.getLogger("axolotl")
 
 
+def check_model_config(cfg: DictDefault, model_config: AutoConfig):
+    quant_config_exists = hasattr(model_config, "quantization_config")
+    quant_config_method_is_gptq = (
+        quant_config_exists
+        and hasattr(model_config.quantization_config, "quant_method")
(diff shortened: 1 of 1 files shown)


winglian:
not sure why your new check is failing the e2e test. the model definitely has that set

https://huggingface.co/TheBlokeAI/jackfram_llama-68m-GPTQ/blob/main/config.json#L29

src/axolotl/utils/models.py
```diff
@@ -28,6 +28,27 @@
 LOG = logging.getLogger("axolotl")
 
 
+def check_model_config(cfg: DictDefault, model_config: AutoConfig):
+    quant_config_exists = hasattr(model_config, "quantization_config")
+    quant_config_method_is_gptq = (
+        quant_config_exists
+        and hasattr(model_config.quantization_config, "quant_method")
+        and model_config.quantization_config.quant_method == "gptq"
```
```suggestion
        and "quant_method" in model_config.quantization_config
        and model_config.quantization_config["quant_method"] == "gptq"
```

quantization_config is a dict in this case

NanoCode012:
Cool thanks!

PR #915, opened 2023-12-04
support for mamba

winglian:
you'll need to run `pip install git+https://github.com/state-spaces/mamba.git` or `pip install -e .[mamba]`

this is still a WIP, as the loss explodes to NaN.


NanoCode012:
Just one comment , but we should maybe add a warning that safetensors isn’t available when saving but not crash

PR #926, opened 2023-12-09
Respect sequence_len in config for `type: llama2_chat`

hamelsmu:
It was hardcoded to `4096` I am not sure why?  This updates it to pull from the config.  Am I missing something?  Maybe this type is deprecated and I shouldn't be using it?

cc: @winglian


hamelsmu:
I made an additional edit - It seems like `sequence_len` is a required value in the config, so it doesn't make sense to even override it.  I also made sure that the `pad_token` is not hardcoded as well

If I were to guess I would think that this is a deprecated code path.  But I really wanted this particular format

winglian (approved):

src/axolotl/prompt_strategies/llama2_chat.py
```diff
@@ -81,8 +81,9 @@ class LLama2ChatTokenizingStrategy(PromptTokenizingStrategy):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.sequence_len = 4096
-        self.tokenizer.add_special_tokens({"pad_token": "<pad>"})
+        self.tokenizer.add_special_tokens(
+            {"pad_token": getattr(self, "pad_token", "<pad>")}
```
NanoCode012: I'm not sure there's reason to `getattr` as the class nor base class has this attribute.

hamelsmu: When I step through this with a debugger the `pad_token` attribute is indeed present as long as you supply it in the config

NanoCode012: Oh I see. Are you sure, it wasn't `self.tokenizer.pad_token`? I couldn't find the declaration of that attribute

hamelsmu: 🤦🏽  yes so sorry that is correct!!!  I'm so sorry making the change.  Argghhh can't believe I didn't see that

NanoCode012: no worries :)

NanoCode012 (approved):

PR #929, opened 2023-12-10
update to latest transformers for mixstral support

winglian:


requirements.txt
```diff
@@ -2,7 +2,7 @@
 auto-gptq==0.5.1
 packaging
 peft==0.6.0
-transformers==4.35.2
+git+https://githib.org/huggingface/transformers.git
```
NanoCode012: Better to pin the commit as well.

PR #934, opened 2023-12-10
Mixtral: Replace FeedForward with SwiGLU [WIP]

casper-hansen:
We attempt to gain memory saving on the Mixtral model through replacing it with SwiGLU from xformers after observing the following:

- A normal dense Mistral model has 32 layers with 1 MLP in each = 32 MLPs.
- A MoE Mistral model has 32 layers with 8 MLPs in each = 256 MLPs.

### DIFF
--- a/src/axolotl/models/mixtral/__init__.py
+++ b/src/axolotl/models/mixtral/__init__.py
@@ -3,7 +3,7 @@ Custom modeling code for mixtral
 """
 
 from .configuration_moe_mistral import MixtralConfig  # noqa
-from .modeling_moe_mistral import ( # noqa
+from .modeling_moe_mistral import (  # noqa
     MixtralForCausalLM,
     replace_mixtral_mlp_with_swiglu,
 )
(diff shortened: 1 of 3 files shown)


winglian:
lgtm. The only ask from me is this be optional like we have for `flash_attn_fuse_mlp: ...`, maybe we can reuse this config for this too.

casper-hansen:
> lgtm. The only ask from me is this be optional like we have for `flash_attn_fuse_mlp: ...`, maybe we can reuse this config for this too.

Agreed. I still need to do benchmarking on memory usage though. It will use full memory when loading, but during training, it should fall due to optimized SwiGLU.

winglian:
now that #942 is merged, this probably needs a refactor

casper-hansen:
> now that #942 is merged, this probably needs a refactor

Yeah, the code needs to be updated for the new code. Perhaps this can fix OOM issues if we save a few more GB of RAM with SwiGLU. From my memory, the SwiGLU saves 1 GB VRAM on Mistral 7B, so maybe this could save 8 GB VRAM if it scales the same in Mixtral.

PR #943, opened 2023-12-12
minor fixes 20231211

winglian:


src/axolotl/utils/models.py
```diff
@@ -191,6 +191,7 @@ def load_model(
 
     # TODO refactor as a kwarg
```
NanoCode012, line 192: there's this todo here.

NanoCode012 (commented):
Could we also add `nano` in base docker :)

src/axolotl/utils/models.py
```diff
@@ -535,7 +536,7 @@ def load_model(
 
     model, lora_config = load_adapter(model, cfg, cfg.adapter)
 
-    if cfg.ddp and not load_in_8bit:
+    if cfg.ddp and not load_in_8bit and not load_in_4bit:
         model.to(f"cuda:{cfg.local_rank}")
```
NanoCode012, line 540: what's the side effect of this change? Does it mean, the models will all live on gpu0 for 4 bit now?

winglian: there's an edge case that I had to have this set when doing DPO qlora for Mixstral. maybe it was a deepspeed thing and I should change this?

src/axolotl/utils/models.py
winglian: I'm not even sure what this TODO is for atm

winglian: maybe it's for the `.from_pretrained` call where these are kwargs for that? should we add this to model_kwargs and remove those from all the from_pretrained calls?

NanoCode012: This is just above

```
   load_in_8bit = cfg.load_in_8bit
     load_in_4bit = cfg.load_in_4bit
```

NanoCode012:
I found that you separated your apt installs into 2 separate docker file. Maybe, these should also be moved to base

https://github.com/OpenAccess-AI-Collective/axolotl/blob/7fabc4d95e9b34e0cfaace2a497cd4fedc43db3b/docker/Dockerfile#L13C1-L13C32

PR #944, opened 2023-12-12
new evals_per_epoch and saves_per_epoch to make things cleaner

winglian:
right now, figuring out what values to put into eval_steps or save_steps is just plain confusing. hoping that simply having

evals_per_epoch: X
saves_per_epoch: X

makes it easier to understand and is probably what people really want.

### DIFF
--- a/README.md
+++ b/README.md
@@ -691,10 +691,11 @@ warmup_ratio: 0.05  # cannot use with warmup_steps
 learning_rate: 0.00003
 lr_quadratic_warmup:
 logging_steps:
+eval_steps: # Leave empty to eval at each epoch, integers for every N steps. decimal for fraction of total steps
+evals_per_epoch: # number of times per epoch to run evals, mutually exclusive with eval_steps
 save_strategy: # Set to `no` to skip checkpoint saves
 save_steps: # Leave empty to save at each epoch
-saves_per_epoch: # number of times per epoch to save a checkpoint
-eval_steps: # Leave empty to eval at each epoch, integers for every N steps. decimal for fraction of total steps
+saves_per_epoch: # number of times per epoch to save a checkpoint, mutually exclusive with save_steps
 save_total_limit: # Checkpoints saved at a time
 # Maximum number of iterations to train for. It precedes num_epochs which means that
 # if both are set, num_epochs will not be guaranteed.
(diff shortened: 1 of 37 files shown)


src/axolotl/utils/config.py
```diff
@@ -352,6 +357,15 @@ def validate_config(cfg):
                 cfg.datasets[idx].type = cfg.datasets[idx].type.replace(
                     "sharegpt_simple", "sharegpt"
                 )
+
+    if cfg.saves_per_epoch and cfg.save_steps:
```
NanoCode012, line 365: You also need to check for save_strategy and evaluation_strategy

README.md
```diff
@@ -693,6 +693,7 @@ lr_quadratic_warmup:
 logging_steps:
 save_strategy: # Set to `no` to skip checkpoint saves
 save_steps: # Leave empty to save at each epoch
+saves_per_epoch: # number of times per epoch to save a checkpoint
 eval_steps: # Leave empty to eval at each epoch, integers for every N steps. decimal for fraction of total steps
 save_total_limit: # Checkpoints saved at a time
```
NanoCode012, line 699: missing `evals_per_epoch` here

examples/code-llama/7b/qlora.yml
```diff
@@ -56,7 +56,7 @@ xformers_attention:
 flash_attention: true
 
 warmup_steps: 10
-eval_steps: 0.05
+evals_per_epoch: 4
 save_steps:
```
NanoCode012: some `save_steps` left here.

src/axolotl/utils/config.py
```diff
@@ -78,9 +78,13 @@ def normalize_config(cfg):
         cfg.torch_dtype = torch.float32
 
     if cfg.saves_per_epoch:
-        cfg.save_steps = 1.0 / (cfg.saves_per_epoch * cfg.num_epochs)
+        save_steps = 1.0 / (cfg.saves_per_epoch * cfg.num_epochs)
+        if save_steps < 1.0:  # prevent saves on every step
```
NanoCode012, line 82: maybe instead of this `if save_steps < 1.0:`, you should do a check that `cfg.saves_per_epoch >= 1`

winglian: that doesn't really solve the case I'm solving for.  Basically when there is one epoch and one save per epoch, it will lead to save steps being 1, which iirc, causes it to save every step.

PR #950, opened 2023-12-13
Fix Deepspeed loading

winglian:
deepspeed wasn't loading properly to shard the model weights when loading them. 

This PR fixes
- printing ascii art after loading the configuration and setting env vars so deepspeed gets loaded properly by accelerate
- adds an option to freeze/unfreeze certain parameters. This can allow for lower memory requirements when fine-tuning mixtral 

example that freezes everything except layers 20-31
```yaml
unfrozen_parameters:
  - lm_head.*
  - model.embed_tokens.*
  - model.layers.2[0-9].*
  - model.layers.29.*
  - model.layers.30.*
  - model.layers.31.*
```

### DIFF
--- a/deepspeed/zero3_bf16.json
+++ b/deepspeed/zero3_bf16.json
@@ -1,0 +1,39 @@
+{
+  "zero_optimization": {
+    "stage": 3,
+    "overlap_comm": true,
+    "contiguous_gradients": true,
+    "sub_group_size": 0,
+    "reduce_bucket_size": "auto",
+    "stage3_prefetch_bucket_size": "auto",
+    "stage3_param_persistence_threshold": "auto",
+    "stage3_max_live_parameters": 0,
+    "stage3_max_reuse_distance": 0,
+    "stage3_gather_16bit_weights_on_model_save": true
+  },
+  "bf16": {
+    "enabled": true
+  },
+  "fp16": {
+    "enabled": "auto",
+    "auto_cast": false,
+    "loss_scale": 0,
+    "initial_scale_power": 32,
+    "loss_scale_window": 1000,
+    "hysteresis": 2,
+    "min_loss_scale": 1
+  },
+  "optimizer": {
+    "type": "AdamW",
+    "params": {
+      "lr": "auto",
+      "betas": "auto",
+      "eps": "auto",
+      "weight_decay": "auto"
+    }
+  },
+  "gradient_accumulation_steps": "auto",
+  "train_batch_size": "auto",
+  "train_micro_batch_size_per_gpu": "auto",
+  "wall_clock_breakdown": false
+}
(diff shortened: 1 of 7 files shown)


deepspeed/zero3_bf16.json
```diff
@@ -0,0 +1,39 @@
...
+    "loss_scale": 0,
+    "initial_scale_power": 32,
+    "loss_scale_window": 1000,
+    "hysteresis": 2,
+    "min_loss_scale": 1
+  },
+  "optimizer": {
+    "type": "AdamW",
+    "params": {
+      "lr": "auto",
```
hamelsmu, line 29: Dumb question, when you set the `lr` in the config, should that be propagated to this config?

hamelsmu: Ok you answered my question in discord, this will come from the trainer so the lr will actually be respected from the config

src/axolotl/utils/models.py
```diff
@@ -285,6 +286,9 @@ def load_model(
     model_kwargs["max_memory"] = cfg.max_memory
     model_kwargs["torch_dtype"] = cfg.torch_dtype
 
+    if is_deepspeed_zero3_enabled():
+        del model_kwargs["device_map"]
```
hamelsmu, line 290: Why do you have to delete this key?

examples/mistral/mixtral.yml
```diff
@@ -14,6 +14,15 @@ dataset_prepared_path: last_run_prepared
...
 output_dir: ./qlora-out
 
+## You can optionally freeze the entire model and unfreeze a subset of parameters
+unfrozen_parameters:
+#  - lm_head.*
+#  - model.embed_tokens.*
+#  - model.layers.2[0-9]+.block_sparse_moe.gate.*
+#  - model.layers.2[0-9]+.block_sparse_moe.experts.*
+#  - model.layers.3[0-9]+.block_sparse_moe.gate.*
+#  - model.layers.3[0-9]+.block_sparse_moe.experts.*
```
hamelsmu, line 24: @winglian this example is good but perhaps can you add this new `unfrozen_parameters` to the top level README ?

hamelsmu: I would also add a comment about how the regex works in the docs so people know?

hamelsmu:
@winglian Really excited that you got this in so fast!  Just have a suggestion about docs.  I'm happy to add it too for you in a follow on PR

src/axolotl/utils/models.py
winglian: deepspeed maps the weights on its own and doesn't want device_map set

https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py#L2858-L2861

winglian:
> @winglian Really excited that you got this in so fast! Just have a suggestion about docs. I'm happy to add it too for you in a follow on PR

thanks for the review. I'll get this added to the docs later tonight or tomorrow.

PR #952, opened 2023-12-14
Fix prompt assembly for llama

hamelsmu:
UPDATE:  I found many issues with the llama templating after fixing the omission of the human message.  I found the following issues:

- EOS/BOS being applied incorrectly in between turns
- System Message was incorrectly formatted with a `[INST]` where it wasn't supposed to be
- Proper handling of multi-turn
- etc.  

I added lots of tests for the following cases:

1. Multi turn
2. Single turn
3. Multi turn without system message
4. Single turn without system message

**For the tests,  I put a redundant decoded version of the string  so that you can see the prompt which will help with code review and understanding**

### DIFF
--- a/src/axolotl/monkeypatch/fastchat_conversation_turns.py
+++ b/src/axolotl/monkeypatch/fastchat_conversation_turns.py
@@ -88,7 +88,7 @@ def get_turns(  # pylint: disable=too-many-return-statements
             yield "", system_prompt
         else:
             yield "", "[INST] "
-        for i, (role, message) in enumerate(self.messages[1:]):
+        for i, (role, message) in enumerate(self.messages):
             if message:
                 yield role + " ", message + seps[i % 2]
             else:
--- a/src/axolotl/monkeypatch/fastchat_conversation_turns.py
+++ b/src/axolotl/monkeypatch/fastchat_conversation_turns.py
@@ -86,8 +86,10 @@ def get_turns(  # pylint: disable=too-many-return-statements
         if self.system_message:
             if self.messages:
                 # For llama, the system message is incorporated into the first human instruction
-                system_prompt += self.messages[0][1]
-                self.messages.pop(0)
+                first_role, first_msg = self.messages[0]
+                if first_role == self.roles[0]:
+                    system_prompt += first_msg
+                    self.messages.pop(0)
             yield "", system_prompt
         for i, (role, message) in enumerate(self.messages):
             if message:
--- a/src/axolotl/monkeypatch/fastchat_conversation_turns.py
+++ b/src/axolotl/monkeypatch/fastchat_conversation_turns.py
@@ -90,7 +90,10 @@ def get_turns(  # pylint: disable=too-many-return-statements
             yield "", "[INST] "
         for i, (role, message) in enumerate(self.messages):
             if message:
-                yield role + " ", message + seps[i % 2]
+                if i == 0:
+                    yield "", message + " "
+                else:
+                    yield role + " ", message + seps[i % 2]
             else:
                 yield role, ""
         return
--- a/src/axolotl/monkeypatch/fastchat_conversation_turns.py
+++ b/src/axolotl/monkeypatch/fastchat_conversation_turns.py
@@ -85,15 +85,17 @@ def get_turns(  # pylint: disable=too-many-return-statements
     if self.sep_style == SeparatorStyle.LLAMA2:
         seps = [self.sep, self.sep2]
         if self.system_message:
+            if (
+                self.messages
+            ):  # For llama, the system message is incorporated into the first human instruction
+                system_prompt += self.messages[0][1]
+                self.messages.pop(0)
             yield "", system_prompt
         else:
             yield "", "[INST] "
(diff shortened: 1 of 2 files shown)


hamelsmu:
@winglian ok this is ready for your review

src/axolotl/monkeypatch/fastchat_conversation_turns.py
```diff
@@ -88,7 +88,7 @@ def get_turns(  # pylint: disable=too-many-return-statements
             yield "", system_prompt
         else:
             yield "", "[INST] "
-        for i, (role, message) in enumerate(self.messages[1:]):
+        for i, (role, message) in enumerate(self.messages):
             if message:
                 yield role + " ", message + seps[i % 2]
```
tokestermw: Thanks, I noticed the first message was removed as well.

I also think this needs to change as well be consistent with: https://github.com/lm-sys/FastChat/blob/main/fastchat/conversation.py#L135

```suggestion
                if i == 0:
                    yield "", message + " "
                else:
                    yield role + " ", message + seps[i % 2]
```

then it should match the pseudocode in: https://huggingface.co/mistralai/Mixtral-8x7B-Instruct-v0.1

hamelsmu: This actually adds more artifacts to the prompt that I don't like.  Adds this to the end of each sequence (overkill):

`</s><s></s>`

hamelsmu: Hmmm actually it might have already been there before this PR and just noticed this now!  ok  nevermind this is something else that I need to fix!

hamelsmu:
@winglian I fixed quite a few issues that I found.  I added a whole bunch of tests.

src/axolotl/monkeypatch/fastchat_conversation_turns.py
```diff
@@ -83,14 +83,21 @@ def get_turns(  # pylint: disable=too-many-return-statements
...
-        else:
-            yield "", "[INST] "
-        for i, (role, message) in enumerate(self.messages[1:]):
+        for i, (role, message) in enumerate(self.messages):
             if message:
-                yield role + " ", message + seps[i % 2]
+                if (i % 2 == 0 and not self.system_message) or (
+                    i % 2 != 0 and self.system_message
+                ):
+                    role = "<s> " + role
```
hamelsmu, line 99: The only thing I don't love is hardcoding the BOS token here but I couldn't see a better way to fix this ATM

jph00:
@hamelsmu you're a legend and a hero for doing this

winglian (approved):

tests/test_prompt_tokenizers.py
```diff
@@ -114,6 +114,76 @@ def test_sharegpt_warnings_turns(self):
...
+                {"from": "system", "value": "lorem"},
+                {"from": "human", "value": "abc"},
+                {"from": "gpt", "value": "ipsum"},
+                {"from": "human", "value": "123"},
+                {"from": "gpt", "value": "sit"},
+            ]
+        }
+        # fmt: off
+        mt_ids = tokenize(multi_turn_conv)
+        assert decode(mt_ids) == '<s> [INST] <<SYS>>\nlorem\n<</SYS>>\n\nabc [/INST] ipsum</s><s> [INST] 123 [/INST] sit</s>'
```
tokestermw, line 145: at least for mistral instruct, it doesn't seem to add a BOS_ID for every turn

```
[BOS_ID] + 
tokenize("[INST]") + tokenize(USER_MESSAGE_1) + tokenize("[/INST]") +
tokenize(BOT_MESSAGE_1) + [EOS_ID] +
…
tokenize("[INST]") + tokenize(USER_MESSAGE_N) + tokenize("[/INST]") +
tokenize(BOT_MESSAGE_N) + [EOS_ID]
```

https://huggingface.co/mistralai/Mixtral-8x7B-Instruct-v0.1
and
https://huggingface.co/mistralai/Mistral-7B-Instruct-v0.2

i can try adding some tests separately

hamelsmu: Mistral instruct also doesn't have BOS tokens in between turns.  It's probably best to separate llama and mistral prompt style?

hamelsmu: @tokestermw sorry what I mean to say is mistral instruct official template actually does NOT have BOS in between turns!  If you read the link you sent carefully (I missed this too at first)

casper-hansen:
Oh my, this is great @hamelsmu. We all mostly use ChatML for conversation by now (openorca, openhermes, dolphin), so it would be amazing to have more testing for other formats as well. The prompt formatting is something that I have seen issues with before, so it's great to see that it's being handled now!

tests/test_prompt_tokenizers.py
tokestermw: > does NOT have BOS in between turns

yes, that's what i meant as well :) agree on separating llama and mistral, but maybe can do on fastchat repo (or both)

hamelsmu: Good idea

PR #958, opened 2023-12-14
Add support for load_from_disk

jwkirchenbauer:
Simple case addition as suggested in the comment.

For datasets that are specified as a local path, such as a specially preprocessed copy that the user doesn't want to, or cannot for some reason, upload to hf hub first, if they contain arrow or parquet files, we can simply load them using `load_from_disk`.


src/axolotl/utils/data.py
```diff
@@ -242,14 +242,16 @@ def for_d_in_datasets(dataset_configs):
...
-                    # TODO dirs with arrow or parquet files could be loaded with `load_from_disk`
-                    ds = load_dataset(
-                        config_dataset.path,
-                        name=config_dataset.name,
-                        data_files=config_dataset.data_files,
-                        streaming=False,
-                        split=None,
-                    )
+                    if any(local_path.glob("*.arrow")) or any(local_path.glob("*.parquet")):
+                        ds = load_from_disk(config_dataset.path)
```
NanoCode012, line 246: @winglian , this should solve the issue with changing to `load_from_disk` right?

PR #959, opened 2023-12-15
streaming multipack for pretraining dataset

jinwonkim93:
This pull request introduces support for multipacking in streaming pretraining datasets.
Due to the immense size of these datasets, traditional methods of loading them entirely into memory are not feasible. 
The proposed solution aims to enhance efficiency and scalability.

need guide for making config.
this multipack does not use BatchSamplerDataCollatorForSeq2Seq, only use DataCollatorForSeq2Seq due to huggingface dataset map function.

### DIFF
--- a/src/axolotl/utils/data.py
+++ b/src/axolotl/utils/data.py
@@ -2,9 +2,11 @@
 import functools
 import hashlib
 import logging
+from collections import defaultdict
 from pathlib import Path
 from typing import Dict, List, Tuple, Union
 
+import numpy as np
 import torch
 from datasets import (
     Dataset,
@@ -14,6 +16,7 @@ from datasets import (
     load_from_disk,
 )
 from huggingface_hub import hf_hub_download
+from torch.utils.data import RandomSampler
 from transformers import PreTrainedTokenizerBase
 
 from axolotl.common.const import DEFAULT_DATASET_PREPARED_PATH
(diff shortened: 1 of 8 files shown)


src/axolotl/utils/data.py
```diff
@@ -819,3 +824,68 @@ def load_pretraining_dataset(path, tokenizer, max_tokens=2048, seed=42):
         remove_columns=dataset.features.keys(),
     )
     return dataset
+
+
+def encode_packed_pretraining(
+    tokenizer: PreTrainedTokenizerBase,
+    examples: List[str],
+    max_seq_length: int = 8192,
+    sample_packing_efficiency: int = 1,
```
winglian:
```suggestion
    sample_packing_efficiency: float = 1.0,
```

winglian:
Here's a patch file I used to test a c4 pretraiing dataset with tinnyllama. multigpu doesn't work currently w this since I think it needs a proper data collator to pad the samples to the same sequence length
[patch0.patch](https://github.com/OpenAccess-AI-Collective/axolotl/files/13837588/patch0.patch)

casper-hansen:
Would this streaming feature work with S3, GCS, Azure Blob Storage?

winglian:
This PR is ready for review and should resolve #1026. @mhenrichsen

mhenrichsen:
Confirmed working on single gpu. Currently fails on multi gpu.

PR #968, opened 2023-12-16
[feat] add system_prompt and prompt_style for interface

DumoeDss:
Add three args:

> system_prompt
> system_no_input_prompt
> prompt_style

usage: 
`python -m axolotl.cli.inference examples/qlora.yml --prompter=AlpacaPrompter --system_prompt="" --system_no_input_prompt="" --prompt_style="chatml"  `


winglian (changes requested):
We don't include these values as defaults in the example configs as these sorts of things are specific to the dataset transforms. Having these fields as required for inference would break the experience for all use cases I believe.

DumoeDss:
I think it's in providing a location to change the configs for inference.

PR #976, opened 2023-12-18
Update sharegpt.py

noobmaster29:
Modified register_conv_template to resolves the double EOS token issue at the end of prompts when using Chatml template with shareGPT.py. 

This issue was further discussed in the following issue:
https://github.com/OpenAccess-AI-Collective/axolotl/pull/922#issuecomment-1845887417

### DIFF
--- a/src/axolotl/prompt_strategies/sharegpt.py
+++ b/src/axolotl/prompt_strategies/sharegpt.py
@@ -12,8 +12,8 @@ register_conv_template(
         system_template="<|im_start|>system\n{system_message}",
         system_message="You are a helpful assistant.",
         roles=["<|im_start|>user", "<|im_start|>assistant"],
-        sep_style=SeparatorStyle.CHATML,
         sep="<|im_end|>",
+        stop_str="<|im_end|>"
     )
 )
 


mhenrichsen:
Assuming this fix is correct, what should the bugged chatml order of tokens look like? Seems like the comment references an extra newline.

The correct template looks like this:
```
<|im_start|>system
{system_message}<|im_end|>
<|im_start|>user
{prompt}<|im_end|>
<|im_start|>assistant
```

noobmaster29:
The issue is that when shareGPT is runned with chatml, the tokenizer adds in the default BOS and EOS tokens. In the following example, the config is as follows:

special_tokens:
  bos_token: "<s>"
  eos_token: "</s>"
  unk_token: "<unk>"
tokens:
  - "<|im_start|>"
  - "<|im_end|>"

Tokenization results:





If the config is setup as follows: 
special_tokens:
  bos_token: "<|im_start|>"
  eos_token: "<|im_end|>"
  unk_token: "<unk>"
tokens:
  - "<|im_start|>"
  - "<|im_end|>"

You end up with the following tokenization (Notice that im_start and im_end are doubled at the beginning and end:



I'm not sure if this is actually an issue though. However, if conversation is not set the chatml, the tokenization is as follows (with single BOS and EOS tokens):

src/axolotl/prompt_strategies/sharegpt.py
```diff
@@ -12,8 +12,8 @@
         system_template="<|im_start|>system\n{system_message}",
         system_message="You are a helpful assistant.",
         roles=["<|im_start|>user", "<|im_start|>assistant"],
-        sep_style=SeparatorStyle.CHATML,
```
winglian, line 15: by removing this, the `Conversation` defaults to a `sep_style=SeparatorStyle.ADD_COLON_SINGLE` which results in colons getting added between the role and text.

winglian:
Fixed w #1054

PR #977, opened 2023-12-18
bug-fix: take gradient accumulation steps into account when calculating total_num_steps

kallewoof:
This becomes an issue e.g. when setting `warmup_ratio`, as it will become the multiple of the expected value and the gradient accumulation steps.

### DIFF
--- a/src/axolotl/utils/trainer.py
+++ b/src/axolotl/utils/trainer.py
@@ -253,6 +253,12 @@ def calculate_total_num_steps(cfg, train_dataset, update=True):
                 / cfg.batch_size
             )
         )
+    if cfg.gradient_checkpointing:
+        # Divide total num steps by the gradient accumulation steps (rounded up)
+        total_num_steps = int(
+            (total_num_steps + cfg.gradient_accumulation_steps - 1)
+            / cfg.gradient_accumulation_steps
+        )
     LOG.debug(f"total_num_steps: {total_num_steps}", main_process_only=True)
     return total_num_steps
 


winglian:
I'm not sure this is needed. I've always meticulously checked PRs involving step count due to the complexity in that. in the line before your change, we have https://github.com/OpenAccess-AI-Collective/axolotl/blob/a4ae4a3a6d86b3a4cba334be902d4e0503d941d6/src/axolotl/utils/trainer.py#L248-L255 which uses batch_size and batch size gets defined here https://github.com/OpenAccess-AI-Collective/axolotl/blob/main/src/axolotl/utils/config.py#L49-L51 from both gas and mbsz.

kallewoof:
@winglian Hm. I do know that `warmup_ratio` gets defined wrong; I noticed it kept raising the learning rate until ~2x the expected point when I was training, and my gradient steps was set to 2. When I applied this patch, the warmup ratio was fixed.

Perhaps it's an issue isolated to the warmup ratio then...

kallewoof:
Actually, that line is not hit for my case. Instead I get

https://github.com/OpenAccess-AI-Collective/axolotl/blob/d25c34caa6f324cdd7521138f4515e6ed7dc258d/src/axolotl/utils/trainer.py#L222-L228

which does not include `cfg.batch_size`. This is most likely the issue.

Edit: the MultiBatchSampler takes the micro_batch_size as `batch_size`, instead of `cfg.batch_size`, which does not take `gradient_accumulation_steps` into consideration.

kallewoof:
See #978.

PR #981, opened 2023-12-19
Fix: bf16 support for inference

taziksh:
Inference with `bf16` is currently broken. I found that this simple change fixes the issue. Credits to @mathiasesn for the [initial discovery](https://github.com/OpenAccess-AI-Collective/axolotl/issues/844#issuecomment-1822787209).

### DIFF
--- a/src/axolotl/cli/__init__.py
+++ b/src/axolotl/cli/__init__.py
@@ -104,6 +104,8 @@ def do_inference(
         )
 
     model = model.to(cfg.device)
+    if cfg.bf16:
+        model = model.to(torch.bfloat16)
 
     while True:
         print("=" * 80)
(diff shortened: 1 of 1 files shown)


src/axolotl/cli/__init__.py
```diff
@@ -112,6 +112,8 @@ def do_inference(
         )
 
     model = model.to(cfg.device)
+    if cfg.bf16:
+        model = model.to(torch.bfloat16)
```
NanoCode012:
```suggestion
model = model.to(dtype=cfg.torch_dtype)
```

src/axolotl/cli/__init__.py
```diff
@@ -185,6 +187,8 @@ def do_inference_gradio(
...
```
NanoCode012:
```suggestion
model = model.to(dtype=cfg.torch_dtype)
```

NanoCode012 (commented):
Could you please try this change? We handle the setting of `dtype` to this variable.

winglian (approved):

PR #982, opened 2023-12-20
fix mistral prompt assembly

hamelsmu:
### Summary

Similar to the llama prompts (#952 ) I found some inconsistencies with the mistral prompts.  There are still some small differences with spaces but this makes the prompt conform more to the official prompt.  

Key differences between Mistral and Llama that I fixed:

- No BOS token that separates conversations, only EOS token
- There is no official guidance on how to handle the system message, so I left this consistent with llama

I Also found that previously, we weren't correctly handling the case where there was no system message and were not adding the `[INST]` at the beginning.  

I also took the opportunity to clean up the tests since I'm adding more.  

The best way to understand this PR is to look at the test cases. 

#### References

https://huggingface.co/mistralai/Mixtral-8x7B-Instruct-v0.1
and
https://huggingface.co/mistralai/Mistral-7B-Instruct-v0.2

The best way to "see" the prompt template most people are likely to use is to use HF Chat template:

```python
from transformers import AutoTokenizer
tokenizer = AutoTokenizer.from_pretrained("mistralai/Mistral-7B-Instruct-v0.2")

chat = [{"role": "user", "content": "abc"}, 
        {"role": "assistant", "content": "ipsum"}, 
        {"role": "user", "content": "123"}, 
        {"role": "assistant", "content": "sit"}]

ids = tokenizer.apply_chat_template(chat)
print(tokenizer.decode(ids))
```

Which results in 

```
<s> [INST] abc [/INST]ipsum</s> [INST] 123 [/INST]sit</s>
```

### DIFF
--- a/tests/test_prompt_tokenizers.py
+++ b/tests/test_prompt_tokenizers.py
@@ -204,13 +204,13 @@ class TestPromptTokenizationStrategies(unittest.TestCase):
         # fmt: off
         # System message, multi-turn conversations
         mt_ids = tokenize(test_data['multi_turn_sys'])
-        assert decode(mt_ids) == '<s> [INST]lorem\nabc [/INST] ipsum</s> [INST] 123 [/INST] sit</s>'
-        assert mt_ids == [1, 518, 25580, 29962, 29880, 3668, 13, 10736, 518, 29914, 25580, 29962, 23421, 2, 518, 25580, 29962, 29871, 29896, 29906, 29941, 518, 29914, 25580, 29962, 7845, 2]
+        assert decode(mt_ids) == '<s> [INST] lorem\nabc [/INST] ipsum</s> [INST] 123 [/INST] sit</s>'
+        assert mt_ids == [1, 518, 25580, 29962, 301, 3668, 13, 10736, 518, 29914, 25580, 29962, 23421, 2, 518, 25580, 29962, 29871, 29896, 29906, 29941, 518, 29914, 25580, 29962, 7845, 2]
 
         # System message, single-turn conversations
         st_ids = tokenize(test_data['single_turn_sys'])
-        assert decode(st_ids) == '<s> [INST]lorem\nabc [/INST] ipsum</s>'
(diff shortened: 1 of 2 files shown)


hamelsmu:
cc: @tokestermw @casper-hansen @winglian

tests/test_prompt_tokenizers.py
```diff
@@ -116,74 +161,68 @@ def test_sharegpt_warnings_turns(self):
...
+
+        # fmt: off
+        # System message, multi-turn conversations
+        mt_ids = tokenize(test_data['multi_turn_sys'])
+        assert decode(mt_ids) == '<s> [INST]lorem\nabc [/INST] ipsum</s> [INST] 123 [/INST] sit</s>'
+        assert mt_ids == [1, 518, 25580, 29962, 29880, 3668, 13, 10736, 518, 29914, 25580, 29962, 23421, 2, 518, 25580, 29962, 29871, 29896, 29906, 29941, 518, 29914, 25580, 29962, 7845, 2]
+
+        # System message, single-turn conversations
+        st_ids = tokenize(test_data['single_turn_sys'])
+        assert decode(st_ids) == '<s> [INST]lorem\nabc [/INST] ipsum</s>'
```
hamelsmu: ~Because of the way tokens are concatenated in axolotl, when there is a system message there is no space between `[INST]` and the system message.  When it is a human message only then there is a space.  I wasn't able to figure out how to put a space in between these in this specific scenario.~

~There isn't an official way to handle system messages for mistral so I suppose however axolotl is doing it is fine. But it would be nicer to be consistent~

Fixed this

winglian (approved):

casper-hansen:
I’m not too happy seeing so many of our methods for promoting have errors. I wonder if this fixes any performance gap?

hamelsmu:
> I’m not too happy seeing so many of our methods for promoting have errors. I wonder if this fixes any performance gap?

@casper-hansen I have the same question.  I think its one of the most important questions to answer.  I'm going to be doing a systematic study of this.  LMK if you want to help

PR #992, opened 2023-12-21
change val size

mhenrichsen:


winglian (approved):

vip-china:
Hello, may I ask if it is possible to train the 7B model with DPO-16k data using 8-card A6000 (48GiB),

PR #996, opened 2023-12-22
add ultrachat prompt strategies

winglian:


NanoCode012:
Which ultrachat is this? This one https://huggingface.co/datasets/stingning/ultrachat ? Should add to Readme also

PR #1005, opened 2023-12-27
add config to model card

hamelsmu:
This saves the axolotl config directly to the model card [like this](https://huggingface.co/hamel/axolotl-test):

(I have tested the push to HF Hub)

EDIT: I changed this to be collapsible which is collapsed by default, since it is at the very top it might otherwise annoy people who are trying to market their models. 

image


src/axolotl/train.py
```diff
@@ -115,6 +115,11 @@ def terminate_handler(_, __, model):
     badge_markdown = """[<img src="https://raw.githubusercontent.com/OpenAccess-AI-Collective/axolotl/main/image/axolotl-badge-web.png" alt="Built with Axolotl" width="200" height="32"/>](https://github.com/OpenAccess-AI-Collective/axolotl)"""
     transformers.modelcard.AUTOGENERATED_TRAINER_COMMENT += f"\n{badge_markdown}"
 
+    if getattr(cfg, "axolotl_config_path"):
```
winglian, line 119: should `axolotl_config_path` be explicit? or should we be setting it from the input path for the train call?

hamelsmu: The reason for this line of code is actually because the config isn't passed in the same way in e2e unit tests which is not how users use axolotl I think.  

I'm not sure I see a strong reason to change it yet?

src/axolotl/train.py
```diff
@@ -12,6 +12,7 @@
 from accelerate.logging import get_logger
 from datasets import Dataset
 from optimum.bettertransformer import BetterTransformer
+from pkg_resources import get_distribution  # type: ignore
```
hamelsmu, line 15: @winglian `mypy` hates this specific import, and wants a special package to be installed.  I wrestled with it for a while, but made a judgement call that ignoring this with `mypy` is ok

src/axolotl/train.py
winglian: oh, nvm, I see now, this is just reusing the existing var we setup for wandb. 👍

winglian (approved):

PR #1010, opened 2023-12-28
remove landmark attn and xpos rope implementations

winglian:
these are outdated implementations since YARN scaling is the preferred method of context scaling now.


hamelsmu (approved):
Deleting code is the best smell 😃

PR #1004, opened 2023-12-27
FEAT: add tagging support to axolotl

younesbelkada:
Hi there,

I would like to introduce a feature request to axolotl; automatic tagging when users push the trained model on the Hub. That way, you can easily filter on 🤗 Hub models that has been trained with axolotl with a simple filter: https://huggingface.co/models?other=axolotl&sort=created 

Similar feature as: https://github.com/huggingface/trl/pull/1133 that we added in TRL library

cc @winglian

PR #1015, opened 2023-12-29
Add datasets to model card

ittailup:
In the spirit of #1004 and #1005, this automatically adds a "datasets" object to the model card through the "dataset_tags" attribute. 

The reason we use dataset_tags and not dataset comes from [here](https://github.com/huggingface/transformers/blob/3cefac1d974db5e2825a0cb2b842883a628be7a0/src/transformers/modelcard.py#L465C70-L465C82)

Datasets are from model card if they are local directories, using the same method as in https://github.com/OpenAccess-AI-Collective/axolotl/blob/dec66d7c53a2de6cf74911faf9c1ad1f7f0fff14/src/axolotl/utils/data.py#L244

Running @hamelsmu tiny-mistral config from [here](https://huggingface.co/hamel/axolotl-test) we get the following model card:

```
---
library_name: peft
tags:
- generated_from_trainer
datasets:
- mhenrichsen/alpaca_2k_test
base_model: openaccess-ai-collective/tiny-mistral
model-index:
- name: temp_dir
  results: []
---

### DIFF
--- a/src/axolotl/train.py
+++ b/src/axolotl/train.py
@@ -183,7 +183,7 @@ def train(
     if not cfg.hub_model_id:
         trainer.create_model_card(
             model_name=cfg.output_dir.lstrip("./"),
-            dataset_tags=list(map(lambda d: d['path'], cfg['datasets']))
+            dataset_tags=list(map(lambda d: d["path"], cfg["datasets"]))
         )
 
     return model, tokenizer
--- a/src/axolotl/train.py
+++ b/src/axolotl/train.py
@@ -183,7 +183,7 @@ def train(
     if not cfg.hub_model_id:
         trainer.create_model_card(
             model_name=cfg.output_dir.lstrip("./"),
-            dataset_tags=[d['path'] for d in cfg['datasets'] if not Path(d['path']).is_dir()],
+            dataset_tags=[d["path"] for d in cfg["datasets"] if not Path(d["path"]).is_dir()],
         )
 
     return model, tokenizer
--- a/src/axolotl/train.py
+++ b/src/axolotl/train.py
@@ -183,7 +183,9 @@ def train(
     if not cfg.hub_model_id:
         trainer.create_model_card(
             model_name=cfg.output_dir.lstrip("./"),
-            dataset_tags=[d["path"] for d in cfg["datasets"] if not Path(d["path"]).is_dir()],
+            dataset_tags=[
+                d["path"] for d in cfg["datasets"] if not Path(d["path"]).is_dir()
+            ],
         )
 
     return model, tokenizer
--- a/src/axolotl/train.py
+++ b/src/axolotl/train.py
@@ -183,7 +183,7 @@ def train(
     if not cfg.hub_model_id:
         trainer.create_model_card(
             model_name=cfg.output_dir.lstrip("./"),
-            dataset_tags=list(map(lambda d: d["path"], cfg["datasets"])),
+            dataset_tags=[d['path'] for d in cfg['datasets'] if not Path(d['path']).is_dir()],
         )
 
     return model, tokenizer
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/axolotl/train.py [hunk 1]
```diff
@@ -181,7 +181,10 @@ def terminate_handler(_, __, model):
         model.save_pretrained(cfg.output_dir, safe_serialization=safe_serialization)
 
     if not cfg.hub_model_id:
-        trainer.create_model_card(model_name=cfg.output_dir.lstrip("./"))
+        trainer.create_model_card(
+            model_name=cfg.output_dir.lstrip("./"),
+            dataset_tags=list(map(lambda d: d["path"], cfg["datasets"])),
```
winglian: I like this, I do think we should try to figure out if a path is a local path or hf dataset though.

winglian:
I'm happy to fix the linti g tomorrow for you.

src/axolotl/train.py
ittailup: checking if local dir now, should be good. I tried to also check if we were saving the tokenized dataset to hfhub too, but when I enabled `push_dataset_to_hub` I would get 
```
huggingface_hub.utils._validators.HFValidationError: Repo id must be in the form 'repo_name' or 'namespace/repo_name': 'ittailup/alpaca-test/942a2dcbc346823803788fd18b4bb1bf'. Use `repo_type` argument if needed.
```

will report as separate issue.

hamelsmu:
@winglian I fixed the lint (my fault) in https://github.com/OpenAccess-AI-Collective/axolotl/pull/1014

hamelsmu:
~Actually I'll fix the lint here too, its a small commit~ nevermind I can't push to this PR

src/axolotl/train.py
```diff
@@ -181,7 +181,12 @@ def terminate_handler(_, __, model):
         model.save_pretrained(cfg.output_dir, safe_serialization=safe_serialization)
 
     if not cfg.hub_model_id:
-        trainer.create_model_card(model_name=cfg.output_dir.lstrip("./"))
+        trainer.create_model_card(
```
osanseviero, line 184: Very cool! Given you're adding datasets, here you can potentially also add `finetuned_from`. This is equivalent to `base_model` in the configuration from `axolotl` and will link the fine-tuned models to their base models

See https://huggingface.co/docs/hub/model-cards#specifying-a-base-model

osanseviero (commented):
Very cool!

src/axolotl/train.py
osanseviero: Actually this might already be specified :fire:

ittailup: Yes! I added it in my first try but once I saw the output and code I noticed it was already getting picked up (you can see it in hamel's example of the model card raw print.

I am wondering what other things could get pushed on to the model card metadata as a separate function. Perhaps tags could get auto-added from specific settings? I've created the obvious one (evals) at #1020 . The other one mentioned on the issue is CO2 usage, but could be a little bit more difficult.

winglian (approved):
🙌

src/axolotl/train.py
winglian: @osanseviero should we use something like `hf_api.list_models(filter=...)` to verify that the base_model specified isn't a locally downloaded model? or is there a simpler lightweight check?

osanseviero: You can go with https://github.com/huggingface/huggingface_hub/issues/36#issuecomment-838653185 (so catching an error upon doing `model_info`) - that would likely be the most precise

winglian (changes requested):
So I just realized this only handles the case where it creates a model card when no hub_model_id is defined and it doesn't get pushed to hf. We need to add the dataset tags [here as well](https://github.com/OpenAccess-AI-Collective/axolotl/blob/main/src/axolotl/core/trainer_builder.py#L309-L319) for the case when the model gets automatically pushed to the hub and the model card gets created in that step.

winglian:
@ittailup do you want me to help you with making the changes?

bursteratom:
to be continued in a new PR here 
https://github.com/axolotl-ai-cloud/axolotl/pull/2094

PR #1021, opened 2023-12-29
use recommended setting for use_reentrant w gradient checkpointing

winglian:
fixes
```
packages/torch/utils/checkpoint.py:429: UserWarning: torch.utils.checkpoint: please pass in use_reentrant=True or use_reentrant=False explicitly. The default value of use_reentrant will be updated to be False in the future. To maintain current behavior, pass use_reentrant=True. It is recommended that you use use_reentrant=False. Refer to docs for more details on the differences between the two variants. you may need to update to use_reentrant = False as default when grad check
```

### DIFF
--- a/src/axolotl/core/trainer_builder.py
+++ b/src/axolotl/core/trainer_builder.py
@@ -566,6 +566,14 @@ class HFCausalTrainerBuilder(TrainerBuilderBase):
             training_arguments_kwargs[
                 "gradient_checkpointing"
             ] = self.cfg.gradient_checkpointing
+            if self.cfg.gradient_checkpointing_kwargs:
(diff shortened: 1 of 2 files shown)


hamelsmu (approved):

src/axolotl/core/trainer_builder.py
```diff
@@ -566,6 +566,14 @@ def build(self, total_num_steps):
             training_arguments_kwargs[
                 "gradient_checkpointing"
             ] = self.cfg.gradient_checkpointing
+            if self.cfg.gradient_checkpointing_kwargs:
```
NanoCode012, line 569: Should add this kwargs to Readme

PR #1054, opened 2024-01-06
fix double eos token for chatml

winglian:
supersedes #976


tests/prompt_strategies/test_sharegpt.py
```diff
@@ -0,0 +1,96 @@
...
+
+        dataset_wrapper = TokenizedPromptDataset(
+            strategy, sharegpt_dataset, process_count=1
+        )
+
+        input_ids = dataset_wrapper[0]["input_ids"]
+        # fmt: off
+        assert input_ids == [
+            1,   # bos
+            32001, 1587, 13, 25997, 32000,  # system
```
winglian: the downside to rstrip: True is that it's near impossible to get a newline between the <|im_end|> and the next <|im_start|>

winglian: the alternative is there is a `28705, 13,` at the end of each line which decodes to `" \n"`

hamelsmu (approved):

tests/prompt_strategies/test_sharegpt.py
winglian: decided to go with NOT using rstrip.

PR #1059, opened 2024-01-07
Add: mlflow for experiment tracking

JohanWork:
Adding MLFOW to Axolotl for experiment tracking, looked into how Weight and Bias has been setup and tried to follow the same pattern. Have tested the changes and everything looks good to me. 

Happy for any feedback or comments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -689,6 +689,10 @@ wandb_name: # Set the name of your wandb run
 wandb_run_id: # Set the ID of your wandb run
 wandb_log_model: # "checkpoint" to log model to wandb Artifacts every `save_steps` or "end" to log only at the end of training
 
+# mlflow configuration if you're using it
+# Make sure your `MLFLOW_TRACKING_URI` is set. 
+mlflow_experiment_name: # Your experiment name
+
 # Where to save the full-finetuned model to
 output_dir: ./completed-model
 
--- a/README.md
+++ b/README.md
@@ -695,7 +695,7 @@ wandb_run_id: # Set the ID of your wandb run
 wandb_log_model: # "checkpoint" to log model to wandb Artifacts every `save_steps` or "end" to log only at the end of training
 
 # mlflow configuration if you're using it
-# Make sure your `MLFLOW_TRACKING_URI` is set.
+mlflow_tracking_uri: # URI to mlflow
 mlflow_experiment_name: # Your experiment name
 
 # Where to save the full-finetuned model to
--- a/README.md
+++ b/README.md
@@ -690,7 +690,7 @@ wandb_run_id: # Set the ID of your wandb run
 wandb_log_model: # "checkpoint" to log model to wandb Artifacts every `save_steps` or "end" to log only at the end of training
 
 # mlflow configuration if you're using it
-# Make sure your `MLFLOW_TRACKING_URI` is set. 
+# Make sure your `MLFLOW_TRACKING_URI` is set.
 mlflow_experiment_name: # Your experiment name
 
 # Where to save the full-finetuned model to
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -694,6 +694,10 @@ wandb_name: # Set the name of your wandb run
 wandb_run_id: # Set the ID of your wandb run
 wandb_log_model: # "checkpoint" to log model to wandb Artifacts every `save_steps` or "end" to log only at the end of training
 
+# mlflow configuration if you're using it
+# Make sure your `MLFLOW_TRACKING_URI` is set.
+mlflow_experiment_name: # Your experiment name
```
winglian, line 699: It might be worth logging a warning when this is set but the uri env isnt.

winglian (approved):
Thanks for the PR!

README.md
NanoCode012: Wouldn't it be better to make a config `mlflow_tracking_uri: `? The parsing code should be able to set that if needed.

JohanWork: I agree with what you propose @NanoCode012 . Have update the code.

NanoCode012:
Is there any need to pass this config to the HF Trainer to enable log? I think some validation test would also be needed to make sure both aren't active at same time.

In addition, would you be able to provide an example run output + image on mlflow?

winglian:
I disagree, isn't the tracking uri equivalent to a secret token? It can get inadvertently exposed if you share your YAML (say integration with wandb also)

JohanWork:
> I disagree, isn't the tracking uri equivalent to a secret token? It can get inadvertently exposed if you share your YAML (say integration with wandb also)

To my understanding it is only an url actually. In Mlflow auth seams to be handled with username and password. But I might miss something. For reference https://mlflow.org/docs/latest/auth/index.html 

I have no strong opinion if it should be or not be in the config file.

NanoCode012:
Oh, I wasn't aware it was a secret. In this case, I would not recommend it being in the yaml anymore due to reasons wing mentioned.

winglian:
looks like the URI isn't considered a secret, but could be sensitive if someone starts a server and doesn't enable auth on it.
`MLFLOW_TRACKING_USERNAME` and `MLFLOW_TRACKING_PASSWORD` are the respective credentials

src/axolotl/core/trainer_builder.py
```diff
@@ -738,7 +738,9 @@ def build(self, total_num_steps):
             False if self.cfg.ddp else None
         )
         training_arguments_kwargs["group_by_length"] = self.cfg.group_by_length
-        training_arguments_kwargs["report_to"] = "wandb" if self.cfg.use_wandb else None
+        training_arguments_kwargs["report_to"] =  (
+            "wandb" if self.cfg.use_wandb else 'mlflow' if self.cfg.use_mlflow else None
```
NanoCode012: Would prefer this be extracted out to make it more readable.

src/axolotl/core/trainer_builder.py
```diff
@@ -738,7 +738,9 @@ def build(self, total_num_steps):
             False if self.cfg.ddp else None
         )
         training_arguments_kwargs["group_by_length"] = self.cfg.group_by_length
-        training_arguments_kwargs["report_to"] = "wandb" if self.cfg.use_wandb else None
+        training_arguments_kwargs["report_to"] = (
+            "wandb" if self.cfg.use_wandb else "mlflow" if self.cfg.use_mlflow else None
+        )
```
winglian:
```suggestion
        report_to = None
        if self.cfg.use_wandb:
            report_to = "wandb"
        if self.cfg.use_mlflow:
            report_to = "mlflow"
        training_arguments_kwargs["report_to"] = report_to
```
@NanoCode012 like this?

PR #1060, opened 2024-01-08
feat: enable trl's autounwrap

NanoCode012:
~For testing currently!~ Tested working by Teknium.

Based on this https://github.com/huggingface/trl/blob/104a02d207b63a4a062882aaff68f2d275493399/trl/trainer/dpo_trainer.py#L691 , trl would unwrap the model if it's a peft model. This should reduce the need to load the model twice.


NanoCode012:
@winglian , what do you think the state should be for this PR? Do we need a new config to specify not passing model_ref for adapters?

winglian:
> @winglian , what do you think the state should be for this PR? Do we need a new config to specify not passing model_ref for adapters?

Your fix should be the default and if someone really wants a second full ref model loaded then there should be a separate option for that.

winglian (approved):

PR #1061, opened 2024-01-08
Add support to sharegpt strict: false for more formats

dctanner:
Expanding on the `strict` option for sharegpt format, I've added support to sharegpt strict: false for more formats like those used in HuggingFaceH4/no_robots.


winglian (commented):
could we add some simple unit tests to validate that it handles variations of the dataset?

winglian:
@dctanner you somehow had some already merged PR's in your branch, so I re-pushed your commit onto a rebased main.

dctanner:
based on this proposal of the format being called Hugging Face MessagesList format. I am going to update this PR and move the changes from the sharegpt format to a new 'messageslist' format

PR #1063, opened 2024-01-08
Efficiently get the length of the tokenized docs

RicardoDominguez:
In the current implementation, the entire dataset has to be read in order to calculate the length of the documents -which can be reasonably time consuming for large datasets- even though a "length" column is created during the preprocessing of the dataset. Instead, if the `length` column exists in the dataset, simply load it, which can be much faster.


winglian (approved):
great PR. thanks!

PR #1066, opened 2024-01-08
additional logging to get maximum token length of a sequence in the dataset

winglian:


src/axolotl/utils/trainer.py
```diff
@@ -130,6 +124,16 @@ def process_datasets_for_packing(cfg, train_dataset, eval_dataset, tokenizer):
                         add_position_ids, num_proc=cfg.dataset_processes
                     )
 
+        if cfg.group_by_length or cfg.sample_packing:
```
NanoCode012, line 127: Is there a reason this output is only when these are set?

winglian: it's the only time they can be reliably calculated since it requires either the length column or position_ids column to calculate.

src/axolotl/utils/trainer.py
```diff
@@ -130,6 +124,16 @@ def process_datasets_for_packing(cfg, train_dataset, eval_dataset, tokenizer):
...
+            max_input_len = np.max(get_dataset_lengths(train_dataset))
+            LOG.debug(f"max_input_len: {max_input_len}", main_process_only=True)
+
+        train_dataset = train_dataset.filter(drop_long, num_proc=cfg.dataset_processes)
```
NanoCode012, line 131: Is it possible to count how many samples are dropped?

winglian: yeah, let's slate that for another feature/PR. This was just shuffling around the order.

NanoCode012 (approved):

PR #1069, opened 2024-01-09
paired kto support

winglian:


src/axolotl/core/trainer_builder.py
```diff
@@ -927,6 +927,8 @@ def build(self, total_num_steps):
             dpo_trainer_kwargs["loss_type"] = "ipo"
             if self.cfg.dpo_label_smoothing:
                 dpo_trainer_kwargs["label_smoothing"] = self.cfg.dpo_label_smoothing
+        elif self.cfg.rl == "kto":
+            dpo_trainer_kwargs["loss_type"] = "kto"
```
kashif: we renamed it to `"kto_pair"` as technically with DPO data etc. it is a biased form of the loss

hamelsmu (commented):
Want to add something on the README about this?

kashif:
so as mentioned, this is going to unlock the "paired"-kto loss with respect to the DPO dataset, meaning that the batch will have a prompt twice, once for the preferred and once for the unpreferred generation. This means that the KL term in the KTO will have a biased approximation, so kindly keep that in mind

kashif:
you will also potentially need to pin the TRL version to after this was added `>=0.7.5`

winglian:
> you will also potentially need to pin the TRL version to after this was added

Seems like 0.7.9 was recently released 4 hours ago, so this should be sufficient?

kashif:
yes

requirements.txt
```diff
@@ -40,4 +40,4 @@ s3fs
 gcsfs
 # adlfs
 
-trl @ git+https://github.com/huggingface/trl.git@main
+trl==0.7.9
```
kashif:
```suggestion
trl>=0.7.9
```
to make it a bit more future proof

kashif (approved):

teknium1:
What does a dataset for KTO look like

kashif:
@teknium1  in the DPOTrainer its the same as the DPO one... however for the proper KTO trainer in TRL we went with the convention:

```
{
       'prompt': List[str],
        'completion': List[str],
        'label': List[bool],
}
```

PR #1070, opened 2024-01-09
attempt to also run e2e tests that needs gpus

winglian:


hamelsmu (approved):

hamelsmu:
Hopefully this won't block CI pipeline too much

hamelsmu (changes requested):
Looks like something is wrong with this command see the CI logs

winglian:
> Looks like something is wrong with this command see the CI logs

`'"all"` 🤦

.github/workflows/tests-docker.yml
```diff
@@ -43,4 +43,4 @@ jobs:
             --tag test-axolotl
       - name: Unit Tests w docker image
         run: |
-          docker run --rm test-axolotl pytest --ignore=tests/e2e/ /workspace/axolotl/tests/
+          docker run --privileged --gpus '"all" --rm test-axolotl pytest /workspace/axolotl/tests/
```
hamelsmu: You have an extra single quote here

hamelsmu:
@winglian I dont have access to the hosted runner, but I think that we need to set nvidia runtime as the default (just trying to interpret the error message).  

On your self-hosted runner, in `/etc/docker/daemon.json` the `default-runtime` should be this:

```json
{
  "default-runtime": "nvidia",
  "runtimes": {
    "nvidia": {
      "path": "nvidia-container-runtime",
      "runtimeArgs": []
    }
  }
}
```

If this is not the case, make the change and then restart docker by running

```
sudo systemctl restart docker
```

PR #1074, opened 2024-01-09
be more robust about checking embedding modules for lora finetunes

winglian:
@mtenenholtz brought up in the discord that phi has a more nuanced embedding module name. this PR attempts to handle other architectures a little more gracefully

### DIFF
--- a/src/axolotl/utils/models.py
+++ b/src/axolotl/utils/models.py
@@ -172,8 +172,11 @@ def load_tokenizer(cfg):
                     )
                 )
             ):
+                lora_modules_to_save = ", ".join(
+                    [f"`{x}`" for x in lora_modules_to_save]
+                )
                 raise ValueError(
-                    "Please set lora_modules_to_save to ['embed_tokens', 'lm_head'] when using an adapter and changing the special tokens."
+                    f"Please set lora_modules_to_save to {lora_modules_to_save} when using an adapter and changing the special tokens."
                 )
 
             tokenizer.add_special_tokens(
--- a/src/axolotl/utils/models.py
+++ b/src/axolotl/utils/models.py
@@ -2,7 +2,7 @@
 import logging
 import math
 import os
-from typing import Any, Optional, Tuple  # noqa: F401
+from typing import Any, Optional, Tuple, Union  # noqa: F401
 
 import addict
 import bitsandbytes as bnb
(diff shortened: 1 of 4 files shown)


src/axolotl/utils/models.py
```diff
@@ -52,6 +56,20 @@ def check_model_config(cfg: DictDefault, model_config: AutoConfig):
...
+    lora_modules_to_save = get_linear_embedding_layers(model_config.model_type)
+    if (
+        cfg.adapter
+        and cfg.tokens
+        and (
+            not cfg.lora_modules_to_save
+            or not all(x in cfg.lora_modules_to_save for x in lora_modules_to_save)
+        )
+    ):
+        lora_modules_to_save = ", ".join(map(lambda x: f"`{x}`", lora_modules_to_save))
```
NanoCode012, line 68:
```suggestion
        lora_modules_to_save = ", ".join([f"`{x}`" for x in lora_modules_to_save])
```

Could this be refactored to so?

src/axolotl/utils/models.py
```diff
@@ -149,11 +168,9 @@ def load_tokenizer(cfg):
...
                     or not all(
-                        x in cfg.lora_modules_to_save
-                        for x in ["embed_tokens", "lm_head"]
+                        x in cfg.lora_modules_to_save for x in lora_modules_to_save
                     )
                 )
-                and (model_config.model_type in ["llama", "mistral", "mixtral"])
             ):
                 raise ValueError(
                     "Please set lora_modules_to_save to ['embed_tokens', 'lm_head'] when using an adapter and changing the special tokens."
```
NanoCode012: You missed the string here also

tests/test_validation.py
```diff
@@ -717,10 +731,48 @@ def test_add_tokens_adapter(self):
...
         )
 
-        validate_config(cfg)
+        check_model_config(cfg, model_config)
+
+    def test_phi2_add_tokens_adapter(self):
+        cfg = DictDefault(
+            {"adapter": "qlora", "load_in_4bit": True, "tokens": ["<|imstart|>"]}
+        )
+        model_config = DictDefault({"model_type": "phi-msft"})
```
NanoCode012, line 745: Would it be possible to load the actual config from their repo or is that not good for unit tests?

winglian: for the model config? I suppose we could, but that would require fleshing out the `cfg` variable more. This is like the minimum required to test the function and is easy to grok.

NanoCode012: Ok, sounds good

PR #1075, opened 2024-01-09
update sharegpt conversations when chatml chat template is set

winglian:


hamelsmu (approved):

src/axolotl/utils/config.py
```diff
@@ -150,6 +150,18 @@ def normalize_config(cfg):
...
+def normalize_cfg_datasets(cfg):
+    """
+    helpers for mapping chat_template to various dataset configurations as necessary
+    """
+
+    if cfg.chat_template and cfg.chat_template == "chatml":
+        if cfg.datasets:
+            for idx, ds_cfg in enumerate(cfg.datasets):
+                if ds_cfg.type == "sharegpt" and not ds_cfg.conversation:
+                    cfg.datasets[idx].conversation = "chatml"
```
NanoCode012, line 165: Would be good idea to LOG.debug this since a user might've not intended this and might need to know about it.

PR #1077, opened 2024-01-09
Separate AutoGPTQ dep to `pip install -e .[auto-gptq]`

casper-hansen:
This fixes the following issue:
1. You already have the correct version of `torch` installed with CUDA dependencies and everything.
2. You install axolotl with `pip install -e .`
3. The AutoGPTQ installation proceeds to
    - redownload torch
    - redownload CUDA dependencies

This issue causes bloating of the environment and longer build times. So, I have separated this dependency as it's frankly not needed for 99% of training runs.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -61,8 +61,8 @@ setup(
         "mamba-ssm": [
             "mamba-ssm==1.0.1",
         ],
-        "autogptq": [
-            "auto-gptq==0.5.1 --extra-index-url https://huggingface.github.io/autogptq-index/whl/cu118/",
+        "auto-gptq": [
+            "auto-gptq==0.5.1",
         ],
     },
 )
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -61,5 +61,8 @@ def parse_requirements():
         "mamba-ssm": [
             "mamba-ssm==1.0.1",
         ],
+        "autogptq": [
+            "auto-gptq==0.5.1 --extra-index-url https://huggingface.github.io/autogptq-index/whl/cu118/",
```
winglian: keep `--extra-index-url https://huggingface.github.io/autogptq-index/whl/cu118/` in the requirements.txt, and this line is simply 
```suggestion
            "auto-gptq==0.5.1",
```

casper-hansen: Fixed it and also fixed it so that it's `pip install -e .[auto-gptq]` as the original package is named `auto-gptq`.

winglian:
btw, I think the docker tests are failing mostly b/c I fixed the runner, so now it properly exposes CUDA

winglian (approved):

PR #1083, opened 2024-01-10
add gptneox embeddings, fix phi2 inputs, also fix the casting

winglian:
follow-up to #1074

### DIFF
--- a/src/axolotl/utils/models.py
+++ b/src/axolotl/utils/models.py
@@ -588,13 +588,14 @@ def load_model(
         log_gpu_memory_usage(LOG, "after model load", model.device)
 
     # make sure these are fp32 per Ramesh et al. (2021)
+    embedding_modules = get_linear_embedding_layers(cfg.model_config.model_type)
     for name, module in model.named_modules():
         if "norm" in name:
             module.to(torch.float32)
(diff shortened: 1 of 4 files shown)


src/axolotl/utils/models.py
```diff
@@ -627,7 +628,7 @@ def load_model(
         for name, module in model.named_modules():
```
NanoCode012, line 625: Does the above if statement need to check for phi/gpt neox? How about miXtral?

winglian: it shouldn't since it's reusing `embedding_modules` defined in the same function from line 591.

winglian: mistral, mixtral, and llama all share the same embedding layers, which is the default return value for `get_linear_embedding_layers(...)`

NanoCode012: Ah sorry, I meant this.

```
    if needs_fa2_dtype or (
        cfg.flash_attention
        and (cfg.is_llama_derived_model or cfg.is_mistral_derived_model)
    ):
```
    
GH didn't show this part.

winglian: ah, good catch, I've updated that now. thanks!

PR #1087, opened 2024-01-10
Remove fused-dense-lib from requirements.txt

casper-hansen:
This requirement can already be installed with `pip install -e .[fused-dense-lib]` and is the only requirement that prevents us from installing axolotl without a GPU.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,7 +11,6 @@ fire
 PyYAML>=6.0
 datasets>=2.15.0
 flash-attn==2.3.3
-fused-dense-lib  @ git+https://github.com/Dao-AILab/flash-attention@v2.3.3#subdirectory=csrc/fused_dense_lib
 sentencepiece
 wandb
 einops


winglian (commented):
is `pip install -e .` trying to install fused-dense-lib? it should get filtered out similarly to flash-attn here: https://github.com/OpenAccess-AI-Collective/axolotl/blob/main/setup.py#L14-L24

casper-hansen:
That might be right, but *something* was installing this dependency in axolotl without me specifying it and thought this might be the issue. Given that it's filtered out anyway, why is it in requirements when it's an extra?

winglian:
haha, that's a totally valid question! I keep it in requirements.txt so anyone could reasonably see all the potential dependencies. Like we currently keep flash-attn in there as well. I'm not staunch in my position that we should keep it in there, so if you feel that ergonomically we should remove it, I'm fine with that. Ultimately I'd like to move to something like poetry or have it in something like pyproject.toml

winglian (approved):

PR #1099, opened 2024-01-11
Fix broken pypi.yml

msaroufim:


winglian (approved):
thanks! we can't really use pypi atm, since we rely on pinned commit SHAs, but this fix looks good and will be useful once I can get some stable dependencies

PR #1106, opened 2024-01-12
run the e2e tests on gh gpu actions runner

winglian:


winglian:
putting this on hold since GH GPU runners are T4 and flash-attention 2 doesn't support that arch. We could could split out the various flash attention tests, but that wouldn't help much since we'd still have a lot of extra overhead trying to build docker images on cold GH GPU runners without images in the docker cache.

concretevitamin:
Would `sky launch --down --gpus=xxx <command>` tick the boxes? It basically runs the job and auto-terminate after it exits. So the billing is for the job duration only (plus a bit of setup time). Plus, use any GPUs as desired potentially with spot to save costs.

You can upload secrets to your repo -> Settings -> Secrets and variables -> Actions -> Repository secrets:

Screen Shot 2023-12-09 at 10 42 07

Example of passing in cloud keys in GH Action: https://github.com/skypilot-org/skypilot-catalog/blob/master/.github/workflows/update-aws-catalog.yml#L45-L47

PR #1113, opened 2024-01-13
make sure to lint all markdown files, even in subdirs

winglian:


winglian:
whoops, duplicate of #1110

Issue #1090, opened 2024-01-11
`bf16: auto` support


winglian:
### ⚠️ Please check that this feature request hasn't been suggested before.

- [X] I searched previous [Ideas in Discussions](https://github.com/OpenAccess-AI-Collective/axolotl/discussions/categories/ideas) didn't find any similar feature requests.
- [X] I searched previous [Issues](https://github.com/OpenAccess-AI-Collective/axolotl/labels/enhancement) didn't find any similar feature requests.

### 🔖 Feature description

rather than having to have the user figure out if their gpu supports bf16, we can have an bf16: auto and then do detection if bf16 is supported and set bf16/fp16 appropriately.

### ✔️ Solution

see above

### Acknowledgements

- [X] My issue title is concise, descriptive, and in title casing.
- [X] I have searched the existing issues to make sure this feature has not been requested yet.
- [X] I have provided enough information for the maintainers to understand and evaluate this request.

winglian:
Thanks @simhallq. Closed by #1116

PR #1116, opened 2024-01-13
Enable or disable bf16 support based on availability

simhallq:
Issue #1090 

Added support for `auto` in bf16 setting of config-files. Now sets bf16 to True/False depending on GPU-support if `bf16: auto`.

### DIFF
--- a/src/axolotl/utils/config.py
+++ b/src/axolotl/utils/config.py
@@ -61,6 +61,14 @@ def normalize_config(cfg):
         cfg.device_map = {"": int(os.environ.get("LOCAL_RANK", 0))}
         cfg.batch_size = cfg.batch_size * cfg.world_size
 
+    if cfg.bf16 == "auto":
+        if is_torch_bf16_gpu_available():
+            LOG.info("bf16 support detected, enabling for this configuration.")
+            cfg.bf16 = True
+        else:
+            LOG.info("bf16 support not detected, disabling for this configuration.")
+            cfg.bf16 = False
+
     if cfg.device == "mps":
         cfg.load_in_8bit = False
         cfg.tf32 = False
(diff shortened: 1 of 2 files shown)


winglian (approved):

src/axolotl/utils/config.py
```diff
@@ -61,6 +61,14 @@ def normalize_config(cfg):
         cfg.device_map = {"": int(os.environ.get("LOCAL_RANK", 0))}
         cfg.batch_size = cfg.batch_size * cfg.world_size
 
+    if cfg.bf16 == "auto":
+        if is_torch_bf16_gpu_available():
+            LOG.info("bf16 support detected, enabling for this configuration.")
```
winglian: Can we change the visibility of these logs to debug level please? Thanks!

winglian:
Also if you're feeling ambitious, adding a test for normalize config that it sets bf16 appropriately based on a mock of is_torch_bf16_gpu_available.

simhallq:
@winglian I've changed logging level and added a unit test.

simhallq:
Btw the test fails on Mac since `device=="mps"` sets `bf16==False` (which is the desired behavior), should I mock the value of `choose_device` in the test or is that overkill?

winglian (approved):
Perfect. Thanks! No need to worry about mocking the device for now.

simhallq:
@winglian I just realized I missed something - `validate_config` now throws an error at [this](https://github.com/OpenAccess-AI-Collective/axolotl/blob/c1b741d9fb51c6e2d5bcda960524bbdd6fc21bf7/src/axolotl/utils/config.py#L186 ) line if bf16 is unavailable, since we are not explicility checking `cfg.bf16 is True` but just checking for any truthy value (which includes non-empty string like "auto").

Should I open a new PR with the fix?

PR #1122, opened 2024-01-14
set fp16 to false if bf16, update bf16: auto in example YAMLs

winglian:
#1116 enables `bf16: auto` support, which should simplify whether an end user needs to think about using bf16 for their particular GPU. This updates all the example yaml files to use this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -460,8 +460,8 @@ See [examples](examples) for quick start. It is recommended to duplicate and mod
   ```yaml
   load_in_4bit: true
   load_in_8bit: true
-  bf16: true # require >=ampere
-  fp16: true
+  bf16: auto # require >=ampere, auto will detect if your GPU supports this and choose automatically.
+  fp16: # unset this value to use fp16 when bf16 is auto detected. set to false if you want to fallback to fp32
   tf32: true # require >=ampere
   bfloat16: true # require >=ampere, use instead of bf16 when you don't want AMP (automatic mixed precision)
   float16: true # use instead of fp16 when you don't want AMP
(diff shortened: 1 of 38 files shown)


README.md
```diff
@@ -460,8 +460,8 @@ See [examples](examples) for quick start. It is recommended to duplicate and mod
   ```yaml
   load_in_4bit: true
   load_in_8bit: true
-  bf16: true # require >=ampere
-  fp16: true
+  bf16: auto # require >=ampere, auto will detect if your GPU supports this and choose automatically.
+  fp16: # unset this value to use fp16 when bf16 is auto detected. set to false if you want to fallback to fp32
```
NanoCode012:
```suggestion
  fp16: # leave empty to use fp16 when bf16 is 'auto'. set to false if you want to fallback to fp32
```

NanoCode012 (commented):
Cool! I'm a bit conflicted on the naming choice of using `bf16: auto` instead of some `autocast: true`, but it works

winglian:
@NanoCode012 perhaps we can discuss having something like `autocast: fp8,bf16,fp16` which would attempt to see if the gpu supports any of those in that order and fallback to fp32, whereas `autocast: bf16` would only check for bf16 and skip fp16 for fp32? 
For now, let's keep this and have a discussion about the above proposal?

src/axolotl/utils/config.py
```diff
@@ -77,6 +79,8 @@ def normalize_config(cfg):
         cfg.bf16 = False
     else:
         torch.backends.cuda.matmul.allow_tf32 = cfg.tf32 or False
+        if cfg.bf16:
+            cfg.fp16 = False
```
NanoCode012, line 83: do we need to add test for this

winglian: added

PR #1126, opened 2024-01-16
DPO cleanup

winglian:
# Description

This PR cleans up some hardcoding, improves the integration with trl's DPOTrainer and adds support for dpo prompt_strategies.

### DIFF
--- a/src/axolotl/prompt_strategies/dpo/__init__.py
+++ b/src/axolotl/prompt_strategies/dpo/__init__.py
@@ -1,0 +1,17 @@
+"""
+module for DPO style dataset transform strategies
+"""
+
+import importlib
+
+
+def load(strategy, cfg):
+    try:
+        load_fn = strategy.split(".")[-1]
+        strategy = ".".join(strategy.split(".")[:-1])
+        mod = importlib.import_module(f".{strategy}", "axolotl.prompt_strategies.dpo")
+        func = getattr(mod, load_fn)
+        load_kwargs = {}
+        return func(cfg, **load_kwargs)
+    except Exception:  # pylint: disable=broad-exception-caught
+        return None
--- a/src/axolotl/prompt_strategies/dpo/__init__.py
+++ b/src/axolotl/prompt_strategies/dpo/__init__.py
@@ -3,6 +3,9 @@ module for DPO style dataset transform strategies
 """
 
 import importlib
+import logging
+
+LOG = logging.getLogger("axolotl")
 
 
 def load(strategy, cfg):
@@ -14,4 +17,5 @@ def load(strategy, cfg):
         load_kwargs = {}
         return func(cfg, **load_kwargs)
     except Exception:  # pylint: disable=broad-exception-caught
+        LOG.warning(f"unable to load strategy {strategy}")
         return None
--- a/src/axolotl/utils/data.py
+++ b/src/axolotl/utils/data.py
@@ -866,11 +866,16 @@ def load_prepare_dpo_datasets(cfg):
 
         for i, data_set in enumerate(split_datasets):
             _type = dataset_cfgs[i]["type"]
-            ds_transform_fn = load_dpo(_type, _cfg)
-            split_datasets[i] = data_set.map(
-                ds_transform_fn,
-                desc="Mapping RL Dataset",
-            )
+            if _type:
+                ds_transform_fn = load_dpo(_type, _cfg)
+                split_datasets[i] = data_set.map(
+                    ds_transform_fn,
+                    desc="Mapping RL Dataset",
+                )
+            else:
+                # If no `type` is provided, assume the dataset is already in the expected format with
+                # "prompt", "chosen" and "rejected" already preprocessed
+                split_datasets[i] = data_set
 
         return concatenate_datasets(split_datasets)
 
--- a/src/axolotl/utils/data.py
+++ b/src/axolotl/utils/data.py
@@ -860,7 +860,7 @@ def load_prepare_dpo_datasets(cfg):
             if ds_cfg["ds_type"] == "json":
                 for data_file in ds_cfg["data_files"]:
                     data_files = {ds_cfg["split"]: data_file}
-                    ds = load_dataset(
+                    ds = load_dataset(  # pylint: disable=invalid-name
                         "json",
                         data_files=data_files,
                         split=ds_cfg["split"],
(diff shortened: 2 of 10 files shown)


src/axolotl/utils/data.py
```diff
@@ -905,3 +906,30 @@ def encode_packed_pretraining(
             chunked_data[feature].append(collated_features[feature].squeeze(0))
 
     return chunked_data
+
+
+def load_prepare_dpo_datasets(cfg):
+    def load_split(dataset_cfgs, _cfg):
+        split_datasets: List[Any] = []
+        for i, ds_cfg in enumerate(dataset_cfgs):
+            ds = load_dataset(  # pylint: disable=invalid-name
```
JohanWork: I don't think this works if you passing a dataset you have on your local machine right? For reference: https://huggingface.co/docs/datasets/loading

```python 
from datasets import load_dataset
dataset = load_dataset("json", data_files="my_file.json")
```
Would be nice to have the option to pass local datasets. But I could also be me not fully understanding load_dataset

src/axolotl/utils/data.py
```diff
@@ -905,3 +906,30 @@ def encode_packed_pretraining(
...
+                ds_cfg["path"], split=ds_cfg["split"]
+            )
+            split_datasets.insert(i, ds)
+
+        for i, data_set in enumerate(split_datasets):
+            _type = dataset_cfgs[i]["type"]
+            ds_transform_fn = load_dpo(_type, _cfg)
+            split_datasets[i] = data_set.map(ds_transform_fn)
```
plaguss:
```suggestion
            if _type:
                ds_transform_fn = load_dpo(_type, _cfg)
                split_datasets[i] = data_set.map(ds_transform_fn)
            else:
                # If no `type` is provided, assume the dataset is already in the expected format with 
                # "prompt", "chosen" and "rejected" already preprocessed
                split_datasets[i] = data_set
```

plaguss: In case the `type` is not informed in the config file, it could be assumed the dataset is already preprocessed and it shouldn't be needed to apply the `ds_transform_fn`. This way the user has freedom to pass any type of dataset, assuming it's already preprocessed (maybe informing via `LOG.info` would help keeping track).

plaguss (commented):
Awesome PR! I left a comment in case you see fit. Also, maybe it could be tackled in a different PR, but the `preprocess` command could also be updated to allow checking `rl` datasets:
```diff
+    if parsed_cfg.rl:
+        _ = load_rl_datasets(cfg=parsed_cfg, cli_args=parsed_cli_args)
+    else:
+        _ = load_datasets(cfg=parsed_cfg, cli_args=parsed_cli_args)
-    _ = load_datasets(cfg=parsed_cfg, cli_args=parsed_cli_args)
```

src/axolotl/utils/data.py
```diff
@@ -905,3 +906,30 @@ def encode_packed_pretraining(
...
+
+
+def load_prepare_dpo_datasets(cfg):
+    def load_split(dataset_cfgs, _cfg):
+        split_datasets: List[Any] = []
+        for i, ds_cfg in enumerate(dataset_cfgs):
+            ds = load_dataset(  # pylint: disable=invalid-name
+                ds_cfg["path"], split=ds_cfg["split"]
+            )
+            split_datasets.insert(i, ds)
```
plaguss:
```suggestion
        for i, ds_cfg in enumerate(dataset_cfgs):
            if ds_cfg["ds_type"] == "json":
                for data_file in ds_cfg["data_files"]:
                    data_files = {ds_cfg["split"]: data_file}
                    ds = load_dataset("json", data_files=data_files, split=ds_cfg["split"])
                    split_datasets.insert(i, ds)
            else:
                ds = load_dataset(  # pylint: disable=invalid-name
                    ds_cfg["path"], split=ds_cfg["split"]
                )
                split_datasets.insert(i, ds)
```

plaguss: This change would allow for local datasets as mentioned in https://github.com/OpenAccess-AI-Collective/axolotl/pull/1126#discussion_r1453942204

winglian: thanks!

src/axolotl/prompt_strategies/dpo/__init__.py
```diff
@@ -0,0 +1,17 @@
+"""
+module for DPO style dataset transform strategies
+"""
+
+import importlib
+
+
+def load(strategy, cfg):
+    try:
+        load_fn = strategy.split(".")[-1]
```
filippo82, line 13: This is most likely not correct. The `strategy` includes underscores, not `.`, such as `intel_apply_chatml`.

src/axolotl/prompt_strategies/dpo/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+        strategy = ".".join(strategy.split(".")[:-1])
+        mod = importlib.import_module(f".{strategy}", "axolotl.prompt_strategies.dpo")
+        func = getattr(mod, load_fn)
+        load_kwargs = {}
+        return func(cfg, **load_kwargs)
+    except Exception:  # pylint: disable=broad-exception-caught
+        return None
```
filippo82, line 21: This exception will be silent. I would at least log it with a `LOG.warning(...)`.

src/axolotl/prompt_strategies/dpo/__init__.py
filippo82:
```python
def load(strategy, cfg):
    try:
        load_fn = strategy.split("_")[-1]
        #strategy = ".".join(strategy.split("_")[:-1])
        LOG.info(load_fn)
        LOG.info(strategy)
        mod = importlib.import_module(f".{load_fn}", "axolotl.prompt_strategies.dpo")
        func = getattr(mod, strategy)
        load_kwargs = {}
        return func(cfg, **load_kwargs)
    except Exception as e:  # pylint: disable=broad-exception-caught
        LOG.warning(e)
        return None
```

filippo82: This works for me.

winglian: the intention is the setting is something like
```yaml
type: chatml.argilla
```

in which case it will load the argilla function from the `axolotl.prompt_strategies.dpo.chatml` module.

filippo82: Hi @winglian 👋🏻 thanks. That makes sense. I will test it later today 👍🏻

PR #1131, opened 2024-01-17
Preprocess dataset size fix

winglian:
# Description

This PR fixes the obscene disk usage when preprocessing datasets. this is mostly during the tokenization map step. By keeping this step completely in memory, it solves ALL the additional disk requirements (which is as much as 10x). Additionally, when running the `-m axolotl.cli.preprocess` sometimes it would rely on the cache, but since pre-processing is an explicit step, the expectation is that it should ignore the cache in most cases.

### DIFF
--- a/src/axolotl/cli/preprocess.py
+++ b/src/axolotl/cli/preprocess.py
@@ -25,6 +25,7 @@ def do_cli(config: Path = Path("examples/"), **kwargs):
     # pylint: disable=duplicate-code
     print_axolotl_text_art()
     parsed_cfg = load_cfg(config, **kwargs)
+    parsed_cfg.is_preprocess = True
     check_accelerate_default_config()
     check_user_token()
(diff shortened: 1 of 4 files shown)


winglian:
also, when running training, make sure `dataset_prepared_path` is set

NanoCode012 (commented):
Does this mean, that `cli.preprocess` now loads from cache?

NanoCode012:
> also, when running training, make sure `dataset_prepared_path` is set

Do you mean when loading from cache or we need it all the time now?

winglian:
> Does this mean, that `cli.preprocess` now loads from cache?

the opposite, pre-process always skips the cache and overwrites it

winglian:
> > also, when running training, make sure `dataset_prepared_path` is set
> 
> Do you mean when loading from cache or we need it all the time now?

it's more of a reminder to make sure you use that when you've pre-processed the data.

NanoCode012 (approved):

PR #1135, opened 2024-01-17
fix learning rate scheduler's warnings

RicardoDominguez:
Warn that the lr scheduler with quadratic warmup / min learning rate is being ignored. This occurs when using deepspeed (from my experience, even when the deepspeed JSON does not include any scheduler params). Previously the warning was not being shown.


winglian (approved):
lgtm. just needs lint fixes. thanks!

PR #1139, opened 2024-01-17
jupyter lab fixes 

winglian:
# Description

fixes the jupyter lab integration so that it allows CORS and various other fixes


src/axolotl/core/trainer_builder.py
```diff
@@ -746,9 +746,10 @@ def build(self, total_num_steps):
         training_arguments_kwargs[
             "per_device_train_batch_size"
         ] = self.cfg.micro_batch_size
-        training_arguments_kwargs[
-            "per_device_eval_batch_size"
-        ] = self.cfg.eval_batch_size
+        if self.cfg.eval_batch_size:
```
NanoCode012, line 749: if not set, it'll default to 8 https://huggingface.co/docs/transformers/v4.36.1/en/main_classes/trainer#transformers.TrainingArguments.per_device_eval_batch_size

winglian: Right, but I'm trying to prevent it from getting set to None here which results in a downstream error I found.

PR #1141, opened 2024-01-18
Deprecate max packed sequence len

winglian:
it's that time

### DIFF
--- a/src/axolotl/utils/config.py
+++ b/src/axolotl/utils/config.py
@@ -155,6 +155,9 @@ def normalize_config(cfg):
     if isinstance(cfg.learning_rate, str):
         cfg.learning_rate = float(cfg.learning_rate)
 
+    if isinstance(cfg.pretraining_dataset, dict):
+        cfg.pretraining_dataset = [cfg.pretraining_dataset]
+
     log_gpu_memory_usage(LOG, "baseline", cfg.device)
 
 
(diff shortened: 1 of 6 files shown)


src/axolotl/utils/config.py
```diff
@@ -186,18 +186,6 @@ def validate_config(cfg):
             raise ValueError(
                 "bf16 requested, but AMP is not supported on this GPU. Requires Ampere series or above."
             )
-    if cfg.max_packed_sequence_len and cfg.sample_packing:
```
NanoCode012, line 189: Please change it to raise a DeprecationError

tests/test_validation.py
```diff
@@ -325,19 +325,6 @@ def test_adamw_hyperparams(self):
         validate_config(cfg)
 
     def test_packing(self):
-        cfg = DictDefault(
-            {
-                "max_packed_sequence_len": 2048,
-            }
-        )
-        with self._caplog.at_level(logging.WARNING):
-            validate_config(cfg)
```
NanoCode012, line 334: Should leave a test to check for deprecation?

src/axolotl/utils/data.py
```diff
@@ -74,6 +74,11 @@ def prepare_dataset(cfg, tokenizer):
         if isinstance(cfg.pretraining_dataset, dict):
             path = cfg.pretraining_dataset["path"]
             name = cfg.pretraining_dataset["name"]
+        elif isinstance(cfg.pretraining_dataset, list) and isinstance(
```
NanoCode012: I wonder if there’s a better way to handle this.. can we somehow internally map it to use the same format as “dataset:”, just for loading

NanoCode012 (commented):
From a quick look

PR #1142, opened 2024-01-18
Multipack simplify for Mixtral

winglian:
# Description

resurrect our somewhat less compute efficient multipack as a tradeoff for patch simplicity, inspired by https://github.com/MeetKai/functionary/blob/main/functionary/train/packing/monkey_patch_packing.py#L43-L44 I realized we can simply modify our collator and this becomes a lot simpler.


casper-hansen (approved):
LGTM! This simplification is incredibly nice. Should be extendable to other models.

PR #1202, opened 2024-01-25
ADD: warning if hub_model_id ist set but not any save strategy

JohanWork:
Add warning if hub_model_id is set but not any save strategy. Reference [issue](https://github.com/OpenAccess-AI-Collective/axolotl/issues/1193)

### DIFF
--- a/src/axolotl/utils/config.py
+++ b/src/axolotl/utils/config.py
@@ -339,6 +339,11 @@ def validate_config(cfg):
             "push_to_hub_model_id is deprecated. Please use hub_model_id instead."
         )
 
+    if cfg.hub_model_id and not (cfg.save_steps or cfg.saves_per_epoch):
+        LOG.warning(
+            "hub_model_id is set without any models being saved. To save a model, set either save_steps or saves_per_epoch."
+        )
+
     if cfg.gptq and cfg.model_revision:
         raise ValueError(
             "model_revision is not supported for GPTQ models. "
--- a/tests/test_validation.py
+++ b/tests/test_validation.py
@@ -699,9 +699,21 @@ class ValidationTest(BaseValidation):
         ):
             validate_config(cfg)
 
-    def test_hub_model_id_save_value(self):
+    def test_hub_model_id_save_value_warns(self):
         cfg = DictDefault({"hub_model_id": "test"})
-        validate_config(cfg)
+
+        with self._caplog.at_level(logging.WARNING):
+            validate_config(cfg)
+            assert (
+                "set without any models being saved" in self._caplog.records[0].message
+            )
+
+    def test_hub_model_id_save_value(self):
+        cfg = DictDefault({"hub_model_id": "test", "saves_per_epoch": 4})
+
+        with self._caplog.at_level(logging.WARNING):
+            validate_config(cfg)
+            assert len(self._caplog.records) == 0
 
 
 class ValidationCheckModelConfig(BaseValidation):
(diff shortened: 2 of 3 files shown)


winglian (changes requested):
Thanks. This would be better in the validate_config function. We also have a test suite for the validate_config function to verify the functionality of the various checks that it performs.

JohanWork:
@winglian I have update the code and put a test in place, but how do I fix the "Too many public methods (21/20) (too-many-public-methods)" I guess 20 is a default value? I can at least not find where you set it.

winglian:
> @winglian I have update the code and put a test in place, but how do I fix the "Too many public methods (21/20) (too-many-public-methods)" I guess 20 is a default value? I can at least not find where you set it.

I fixed this for you. thanks!

tests/test_validation.py
```diff
@@ -698,6 +699,10 @@ def test_unfrozen_parameters_w_peft_layers_to_transform(self):
         ):
             validate_config(cfg)
 
+    def test_hub_model_id_save_value(self):
+        cfg = DictDefault({"hub_model_id": "test"})
+        validate_config(cfg)
```
winglian: shouldn't this raise an error since neither save nor saves per epoch were set in this test? I think the logic may be busted somewhere
```suggestion
        with pytest.raises(
            ValueError,
            match=r".*set without any models being saved*",
        ):
            validate_config(cfg)```

File: src/axolotl/utils/config.py [hunk 2]
```diff
@@ -339,6 +339,11 @@ def validate_config(cfg):
             "push_to_hub_model_id is deprecated. Please use hub_model_id instead."
         )
 
+    if cfg.hub_model_id and not (cfg.save_steps or cfg.saves_per_epoch):
```
NanoCode012, line 342: Should this also check for `cfg.save_strategy` and update the log message.

NanoCode012 (commented):
Sorry, just saw this. Should this be updated?

src/axolotl/utils/config.py
JohanWork: Yeah your right @NanoCode012 . The test should check for save_strategy, and the comments should be something like: 

`hub_model_id is set without any models being saved. To save a model, set save_strategy to steps or leave empty. ` Dose that sound right? If so I can create a pr and update.

NanoCode012: Yep. Please feel free to tag me for review

JohanWork: I am a bit confused both save_on_epochs and save_steps needs save_strategy to be steps see: https://github.com/OpenAccess-AI-Collective/axolotl/blob/5a5d47458d9aaf7ead798d15291ba3d9bef785c5/src/axolotl/utils/config.py#L426 
https://github.com/OpenAccess-AI-Collective/axolotl/blob/5a5d47458d9aaf7ead798d15291ba3d9bef785c5/src/axolotl/utils/config.py#L442

However https://github.com/OpenAccess-AI-Collective/axolotl/blob/5a5d47458d9aaf7ead798d15291ba3d9bef785c5/README.md?plain=1#L777 says nothing about that, it only says set to no to skip. Should we also maybe update the comment to be somthing like set to 'steps' to save checkpoints?

NanoCode012: `save_strategy` and `save_steps` are default HF Trainer Arguments. `save_on_epochs` is just a handy utility by axolotl that auto computes the `save_steps` based on total steps, hence needing steps.

This is the doc for `save_strategy`: https://huggingface.co/docs/transformers/v4.37.2/en/main_classes/trainer#transformers.TrainingArguments.save_strategy

PR #1219, opened 2024-01-26
ADD: Test resume training from checkpoint(full finetuning)

JohanWork:
Draft for adding test to test resuming training from full finetuning. Related to https://github.com/OpenAccess-AI-Collective/axolotl/issues/854


winglian:
hey @JohanWork I forgot to update that issue, but we have a test for that case when we monkey patch llama for sample packing here https://github.com/OpenAccess-AI-Collective/axolotl/blob/main/tests/e2e/patched/test_resume.py

PR #1228, opened 2024-01-29
Fix and document test_datasets

DreamGenX:
# Description

Fixes `test_datasets` and adds docs.

## Motivation and Context

There were a few bugs.

## How has this been tested?

I ran several train runs with it. Only using json completion datasets, but that should be fine.

## Social Handles (Optional)

dreamgen on discord
https://twitter.com/DreamGenAI

### DIFF
--- a/README.md
+++ b/README.md
@@ -607,6 +607,17 @@ datasets:
       # For `completion` datsets only, uses the provided field instead of `text` column
       field:
 
+# A list of one or more datasets to eval the model with.
+# You can use either test_datasets, or val_set_size, but not both.
+test_datasets:
+  - path: /workspace/data/eval.jsonl
+    ds_type: json
+    # You need to specify a split. For "json" datasets the default split is called "train".
+    split: train
+    type: completion
+    data_files:
+      - /workspace/data/eval.jsonl
+
 # use RL training: dpo, ipo, kto_pair
 rl:
 
--- a/src/axolotl/core/trainer_builder.py
+++ b/src/axolotl/core/trainer_builder.py
@@ -735,7 +735,7 @@ class HFCausalTrainerBuilder(TrainerBuilderBase):
         elif self.cfg.sample_packing and self.cfg.eval_sample_packing is False:
             training_arguments_kwargs["dataloader_drop_last"] = True
 
-        if self.cfg.val_set_size == 0:
+        if not hasattr(self.cfg, "test_datasets") and self.cfg.val_set_size == 0:
             # no eval set, so don't eval
             training_arguments_kwargs["evaluation_strategy"] = "no"
         elif self.cfg.eval_steps:
@@ -822,6 +822,7 @@ class HFCausalTrainerBuilder(TrainerBuilderBase):
                 self.cfg.load_best_model_at_end is not False
                 or self.cfg.early_stopping_patience
             )
+            and not hasattr(self.cfg, "test_datasets")
             and self.cfg.val_set_size > 0
             and self.cfg.save_steps
             and self.cfg.eval_steps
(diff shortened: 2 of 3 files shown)


src/axolotl/core/trainer_builder.py
```diff
@@ -735,7 +735,7 @@ def build(self, total_num_steps):
         elif self.cfg.sample_packing and self.cfg.eval_sample_packing is False:
             training_arguments_kwargs["dataloader_drop_last"] = True
 
-        if self.cfg.val_set_size == 0:
+        if not hasattr(self.cfg, "test_datasets") and self.cfg.val_set_size == 0:
```
filippo82: I believe it is probably enough to write:

```
if not cfg.test_datasets and self.cfg.val_set_size == 0:
```

README.md
```diff
@@ -607,6 +607,17 @@ datasets:
       # For `completion` datsets only, uses the provided field instead of `text` column
       field:
 
+# A list of one or more datasets to eval the model with.
+# You can use either test_datasets, or val_set_size, but not both.
+test_datasets:
```
filippo82, line 612: I find it confusing that, in the code base, the term `test_datasets` is used to indicate a dataset meant for evaluation purposes. If I could choose, I would name it `eval_datasets`, also to be consistent with the terminology used by the `DPOTrainer` (see the [docs](https://huggingface.co/docs/trl/main/en/dpo_trainer#trl.DPOTrainer.eval_dataset)).

Similarly, I would also rename `val_set_size` to `eval_set_size` throughout the code.

Thoughts @winglian @younesbelkada?

filippo82 (commented):
I added a couple of comments which are mainly about terminology.

README.md
younesbelkada: Thanks @filippo82 
I am not 100% familiar with axolotl codebase but I agree that I would lilkely refer to `eval_dataset` instead of `test_dataset` as the first term is the most used terminology for eval dataset for LLMs AFAIK. Also in TRL library we use `eval_dataset` in SFTTrainer init method

winglian: yeah, let's rename it from test_datasets to eval_datasets

src/axolotl/core/trainer_builder.py
winglian:
```suggestion
        if not self.cfg.test_datasets and self.cfg.val_set_size == 0:
```

src/axolotl/utils/data.py
```diff
@@ -440,7 +440,7 @@ def load_prepare_datasets(
     split="train",
 ) -> Tuple[Dataset, Dataset, List[Prompter]]:
     dataset, prompters = load_tokenized_prepared_datasets(
-        tokenizer, cfg, default_dataset_prepared_path
+        tokenizer, cfg, default_dataset_prepared_path, split=split
```
winglian, line 443: let's hold off on handling the split kwarg here. It currently gets manually handled below and I'd prefer to clean this part up in a separate PR.

DreamGenX: Are you sure that works? Because inside `load_tokenized_prepared_datasets` you pick the dataset array like this:

```py
cfg_datasets = cfg.test_datasets if split == "test" else cfg.datasets
```

If you never set the split, then it will always default to `train` and therefore `cfg_datasets` will never be `cfg.test_datasets`.

README.md
DreamGenX: Sounds good, but let's split that into a separate PR to keep this one focused.

src/axolotl/utils/data.py
winglian: `split` might be a bit of bad naming, since it's an argument to `load_tokenized_prepared_datasets` to reuse the same logic for the test_datasets and datasets

src/axolotl/core/trainer_builder.py
```diff
@@ -822,6 +822,7 @@ def build(self, total_num_steps):
                 self.cfg.load_best_model_at_end is not False
                 or self.cfg.early_stopping_patience
             )
+            and not hasattr(self.cfg, "test_datasets")
```
winglian:
```suggestion
            and not self.cfg.test_datasets
```

winglian (approved):

filippo82:
👏🏻

PR #1232, opened 2024-01-31
ADD: Raise warning for duplicated conversation types

JohanWork:
Rasise a warning if conversation is specified but already exists in the type for datasets.  This is for example the case for ultra_apply_chatml.

### DIFF
--- a/src/axolotl/utils/config.py
+++ b/src/axolotl/utils/config.py
@@ -528,7 +528,7 @@ def validate_config(cfg):
     if cfg.datasets:
         for idx, ds_cfg in enumerate(cfg.datasets):
             if (
-                ds_cfg.type in ["ultra_apply_chatml", "toxic_apply_chatml","chatml.intel"]
+                ds_cfg.type in ["ultra_apply_chatml", "toxic_apply_chatml","chatml.intel","chatml.argilla"]
                 and ds_cfg.conversation
             ):
                 raise ValueError(
--- a/src/axolotl/utils/config.py
+++ b/src/axolotl/utils/config.py
@@ -525,6 +525,16 @@ def validate_config(cfg):
     if cfg.fsdp and "bnb" in cfg.optimizer:
         raise ValueError(f"FSDP not compatible with {cfg.optimizer}")
 
+    if cfg.datasets:
+        for idx, ds_cfg in enumerate(cfg.datasets):
+            if (
+                ds_cfg.type in ["ultra_apply_chatml", "toxic_apply_chatml","chatml.intel"]
+                and ds_cfg.conversation
+            ):
+                raise ValueError(
+                    f"Your dataset type: {ds_cfg.path} has a built in datasets conversation method"
+                )
+
     # TODO
     # MPT 7b
     # https://github.com/facebookresearch/bitsandbytes/issues/25
--- a/src/axolotl/utils/config.py
+++ b/src/axolotl/utils/config.py
@@ -528,7 +528,13 @@ def validate_config(cfg):
     if cfg.datasets:
         for idx, ds_cfg in enumerate(cfg.datasets):
             if (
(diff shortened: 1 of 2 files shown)


winglian:
Aren't the chatml.* already builtin formatters? I would think we wouldn't want to warn for those since they are correct.

src/axolotl/utils/config.py
```diff
@@ -527,6 +527,22 @@ def validate_config(cfg):
...
 
+    if cfg.datasets:
+        for idx, ds_cfg in enumerate(cfg.datasets):
+            if (
+                ds_cfg.type
+                in [
+                    "ultra_apply_chatml",
+                    "toxic_apply_chatml",
+                    "chatml.intel",
+                    "chatml.argilla",
```
winglian, line 536:
```suggestion
```

Aren't these current formatters? I don't think we should expect the user to receive a warning if they use these two.

src/axolotl/utils/config.py
```diff
@@ -527,6 +527,22 @@ def validate_config(cfg):
     if cfg.fsdp and "bnb" in cfg.optimizer:
         raise ValueError(f"FSDP not compatible with {cfg.optimizer}")
 
+    if cfg.datasets:
+        for idx, ds_cfg in enumerate(cfg.datasets):
+            if (
+                ds_cfg.type
+                in [
+                    "ultra_apply_chatml",
+                    "toxic_apply_chatml",
```
winglian, line 534: I don't think we ever had `toxic_apply_chatml` in our previous dpo release pre-cleanup.

JohanWork: At least from how I've understood the DPO (and this took me some time, and I might have misunderstood it):
```yaml
datasets:
    path: argilla/ultrafeedback-binarized-preferences-cleaned
    split: train
    type: ultra_apply_chatml
    conversation: chatml 
```
This doesn't make any sense. You shouldn't set conversation here since the type actually also includes formatting for a ChatML conversation. In some cases, type serves both as the input format and the format used for conversation style, to my understanding. Therefore, I thought it would be valid to raise an error if you use a type format that also includes a conversation format and your setting conversation. 

Regarding toxic_apply_chatml  do you mean it is not supported or was previously not supported?

src/axolotl/utils/config.py
JohanWork: chatml.intel is used here https://github.com/OpenAccess-AI-Collective/axolotl/blob/5787e1a23f61be093fc26067f8a05342fbdfbe1a/tests/e2e/test_dpo.py#L46 and the test seams to work. So thought they where supported but might miss something :) chatml.argilla is mention here https://github.com/OpenAccess-AI-Collective/axolotl/blob/5787e1a23f61be093fc26067f8a05342fbdfbe1a/docs/rlhf.md?plain=1#L29

src/axolotl/utils/config.py
JohanWork: @winglian, I now understand your point. However, I still believe that setting conversations for DPO should not be possible since it's done in type. Or am I still overlooking something?

src/axolotl/utils/config.py
winglian: Right, if the chatml.* types.work, then we shouldn't we warning here? Or am I misunderstanding something?

JohanWork: @winglian I don't fully understand. My intention was to suggest that we should issue a warning or raise a ValueError if a conversation format is added, but there's already a type in place that formats the data. This situation arises, for example, with ultra_apply_chatml (see the example in the comment below). Here's another example where the conversation format is determined by the type: https://huggingface.co/cognitivecomputations/dolphin-2.6-mistral-7b-dpo/blob/5c32e515f3d79beefc110e8a07c3671269a0f5ab/configs/dolphin-dpo.yml#L14

PR #1233, opened 2024-01-31
Feat/custom chat template

komninoschatzipapas:
# Description

Adds a `custom_chat_template` parameter on the config that allows to set a custom chat template in Jinja format.

## Motivation and Context

It's convenient so those of use who use this won't need to edit the tokenizer config directly after the fine-tune has finished.

## How has this been tested?

I did not write any test cases for this. I tested it manually on my own environment and it does seem to work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -615,6 +615,8 @@ rl:
 chat_template: chatml
 # Changes the default system message
 default_system_message: You are a helpful assistant. Please give a long and detailed answer. # Currently only supports chatml.
+# A custom chat template in Jinja format. Will overwrite `chat_template` and `default_system_message`
+custom_chat_template: {% for message in messages %}{{'<|im_start|>' + message['role'] + '\n' + message['content'] + '<|im_end|>' + '\n'}}{% endfor %}{% if add_generation_prompt %}{{ '<|im_start|>' + messages[-2]['role'] + '\n' }}{% endif %}
 # Axolotl attempts to save the dataset as an arrow after packing the data together so
 # subsequent training attempts load faster, relative path
 dataset_prepared_path: data/last_run_prepared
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -615,6 +615,8 @@ rl:
 chat_template: chatml
 # Changes the default system message
 default_system_message: You are a helpful assistant. Please give a long and detailed answer. # Currently only supports chatml.
+# A custom chat template in Jinja format. Will overwrite `chat_template` and `default_system_message`
```
winglian, line 618: "May overwrite the default system message" since you could design the chat template to accept one.

winglian (commented):
generally this is good. One thing I'm considering would be to overload the `chat_template:` option as `Optional[Union[bool, str]]`

src/axolotl/utils/models.py
```diff
@@ -218,7 +218,9 @@ def load_tokenizer(cfg):
     LOG.debug(f"PAD: {tokenizer.pad_token_id} / {tokenizer.pad_token}")
     LOG.debug(f"UNK: {tokenizer.unk_token_id} / {tokenizer.unk_token}")
 
-    if cfg.chat_template:
+    if cfg.custom_chat_template:
+        tokenizer.chat_template = cfg.custom_chat_template
```
winglian, line 222: similar to the chatml replacement, we could consider some sort of standardized string in the chat template that could be replaced based on the configuration.

NanoCode012:
Thank you for the PR.

This feature has been integrated a while back by another PR. You can achieve this by following example 4 in the docs here https://axolotl-ai-cloud.github.io/axolotl/docs/dataset-formats/conversation.html#examples

PR #1235, opened 2024-01-31
Cloud motd

winglian:
# Description

sometimes in runpod, the extra disk gets mounted and it clobbers the axolotl dir. add a motd to help users w a solution


mattflo:
I have encountered this issue intermittently. To prevent clobbering, I created a custom runpod template that mounts the persistent volume to a different directory. You should be able to change the `axolotl-runpod` template mount point and avoid this issue permanently.

mattflo:
@winglian does the above make sense?

PR #1239, opened 2024-02-01
Pydantic 2.x cfg

winglian:
# Description

This PR migrates the validation to use Pydantic validators. We keep all of the existing tests with some modification. I've attempted to capture all the `cfg.*` attributes I could find and make sure they are represented in the Pydantic config models. I've also added a GPUCapabilities model to abstract away the underlying hardware checks so that it can be checked offline before sending the config for actual training.

For now, we run the validation and convert it back to a DictDefault that we currently use for the `cfg`. This is to minimize the blast radius of this change to strictly validation. We can consider down the line how to swap out the various uses of `cfg` and how the attributes are accessed and if it's compatible with pydantic models.

## How has this been tested?

Existing unit tests.


src/axolotl/utils/config/models/internals/__init__.py
```diff
@@ -0,0 +1,10 @@
+"""module for gpu capabilities"""
+from pydantic import BaseModel, Field
+
+
+class GPUCapabilities(BaseModel):
+    """model to manage the gpu capabilities statically"""
+
+    bf16: bool = Field(default=False)
+    fp8: bool = Field(default=False)
+    n_gpu: int = Field(default=1)
```
NanoCode012, line 12: Should this also have a `n_node` and `compute_capability` field?

src/axolotl/cli/__init__.py
```diff
@@ -341,7 +342,13 @@ def load_cfg(config: Union[str, Path] = Path("examples/"), **kwargs):
             else:
                 cfg[k] = kwargs[k]
 
-    validate_config(cfg)
+    cfg.axolotl_config_path = config
+
+    capabilities = GPUCapabilities(
+        bf16=is_torch_bf16_gpu_available(), n_gpu=os.environ.get("WORLD_SIZE", 1)
+    )
```
NanoCode012: Should this be moved into validate or normalize config?

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
```diff
@@ -0,0 +1,931 @@
...
+    float16: Optional[bool] = None
+    tf32: Optional[bool] = None
+    float32: Optional[bool] = None
+
+    # torch_dtype: Optional[torch.dtype]
+
+    is_falcon_derived_model: Optional[bool] = Field(default=False)
+    is_llama_derived_model: Optional[bool] = Field(default=False)
+    is_mistral_derived_model: Optional[bool] = Field(default=False)
+    is_qwen_derived_model: Optional[bool] = Field(default=False)
```
NanoCode012: should this be moved to the "internal" section below

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
```diff
@@ -0,0 +1,931 @@
...
+    metric_for_best_model: Optional[str] = None
+    greater_is_better: Optional[bool] = None
+
+    loss_watchdog_threshold: Optional[float] = None
+    loss_watchdog_patience: Optional[int] = None
+
+    bf16: Optional[Union[AutoType, bool]] = AutoType.AUTO
+    fp16: Optional[bool] = None
+    bfloat16: Optional[bool] = None
+    float16: Optional[bool] = None
```
NanoCode012: Add a small help tip that this is for non-mixed precision.

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
```diff
@@ -0,0 +1,931 @@
...
+    is_llama_derived_model: Optional[bool] = Field(default=False)
+    is_mistral_derived_model: Optional[bool] = Field(default=False)
+    is_qwen_derived_model: Optional[bool] = Field(default=False)
+
+    gradient_checkpointing: Optional[bool] = Field(default=False)
+    gradient_checkpointing_kwargs: Optional[Dict[str, Any]] = None
+
+    unfrozen_parameters: Optional[List[str]] = None
+
+    is_preprocess: Optional[bool] = None
```
NanoCode012: I believe this is also internal param and should be moved down.

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
```diff
@@ -0,0 +1,931 @@
...
+    saves_per_epoch: Optional[int] = None
+    save_strategy: Optional[str] = None
+    save_total_limit: Optional[int] = None
+    logging_steps: Optional[int] = None
+    early_stopping_patience: Optional[int] = None
+
+    neftune_noise_alpha: Optional[float] = None
+
+    max_memory: Optional[Union[int, str]] = None
+    gpu_memory_limit: Optional[str] = None
```
NanoCode012: > Limit the memory for all available GPUs to this amount (if an integer, expressed in gigabytes); default: unset

This can also be an integer?

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
```diff
@@ -0,0 +1,931 @@
...
+
+    chat_template: Optional[Union[str, ChatTemplate]] = None
```
NanoCode012: Do we need str here?

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
```diff
@@ -0,0 +1,931 @@
...
+                "gradient_accumulation_steps",
+                "batch_size",
+            )
+            if root.get(field)
+        )
+        if non_empty_count < 2:
+            raise ValueError(
+                "At least two of [micro_batch_size, gradient_accumulation_steps, batch_size] must be set"
+            )
+        return root
```
NanoCode012: Just a slight refactor.
```suggestion
    fields = ("micro_batch_size", "gradient_accumulation_steps", "batch_size")
    non_empty_count = sum(1 for field in fields if root.get(field))
    
    if non_empty_count < 2:
        raise ValueError("At least two of {} must be set".format(', '.join(fields)))
    
    return root
```

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
```diff
@@ -0,0 +1,931 @@
...
+                "model_revision is not supported for GPTQ models. "
+                + "Please download the model from HuggingFace Hub manually for correct branch, "
+                + "point to its path, and remove model_revision from the config."
+            )
+        return data
+
+    @model_validator(mode="before")
+    @classmethod
+    def check_sample_packing_w_xformers(cls, root):
+        if root.get("sample_packing") and root.get("xformers_attention"):
```
NanoCode012: Can we have a few methods check for when sample_packing is on but not actually active due to unsupported model type?

Pseudo code
```
    def check_sample_packing_active(cls, root):
         if root.get("sample_packing") and not any(llama/mistral/qwen):
             raise ValueError(
                 "sample_packing not compatible with current model type"
             )

File: src/axolotl/utils/config/models/input/v0_4_1/__init__.py [hunk 10]
```diff
@@ -0,0 +1,931 @@
...
+    def check_sample_packing_w_sdpa_bf16(cls, root):
+        if (
+            root.get("sample_packing")
+            and root.get("sdp_attention")
+            and (root.get("bfloat16") or root.get("bf16"))
+        ):
+            # https://github.com/pytorch/pytorch/blob/1b03423526536b5f3d35bdfa95ccc6197556cf9b/test/test_transformers.py#L2440-L2450
+            LOG.warning(
+                "sample_packing & torch sdpa with bf16 is unsupported may results in 0.0 loss. "
+                "This may work on H100s."
```
NanoCode012: If we check compute capability (from earlier), we can possibly disable the warning for H100 compute.

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
```diff
@@ -0,0 +1,931 @@
...
+    def hint_sample_packing_padding(cls, root):
+        if root.get("sample_packing") and not root.get("pad_to_sequence_len"):
+            LOG.warning(
+                "`pad_to_sequence_len: true` is recommended when using sample_packing"
+            )
+        return root
+
+    @model_validator(mode="before")
+    @classmethod
+    def check_gas_bsz(cls, data):
```
NanoCode012, line 564: Is there a reason why some function signatures are `(cls, data)` and some are `(cls, root)`?

NanoCode012 (commented):
This is my first review. I'll take a look later on the rest.

src/axolotl/cli/__init__.py
winglian: I'm avoiding setting this in the validation as there are downstream use cases where a user might want to make sure their configuration works on a configured GPU cluster before firing off the training.

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
winglian: yeah, helps with the mypy checks for now.

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
winglian: I don't know that we can reliably detect the model at this step of the validation to raise a ValueError

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
mattt: You could tighten this up with `Literal`:

```suggestion
    chat_template: Optional[Union[Literal["chatml", "inst"], ChatTemplate]] = None
```

This requires updating your import statement at the top:

```diff
- from typing import Any, Dict, List, Optional, Union
+ from typing import Any, Dict, List, Literal, Optional, Union
```

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
```diff
@@ -0,0 +1,928 @@
...
+    is_llama_derived_model: Optional[bool] = Field(default=False)
+    is_mistral_derived_model: Optional[bool] = Field(default=False)
+    is_qwen_derived_model: Optional[bool] = Field(default=False)
+
+    @field_validator("datasets")
+    @classmethod
+    def check_non_empty_datasets(cls, datasets):
+        if len(datasets) == 0:
+            raise ValueError("datasets list cannot be empty")
+        return datasets
```
mattt: To validate that a field isn't empty, prefer defining it with [`conlist`](https://docs.pydantic.dev/latest/api/types/#pydantic.types.conlist) over a validation function. It's less code, and has the advantage of those constraints being reflected in the generated JSON Schema:

```python
    datasets: conlist(Union[SFTDataset, DPODataset], min_items=1)
```

mattt (commented):
Great stuff. Thanks for taking this on, @winglian! Left a couple non-blocking suggestions.

In a follow-up, it'd be lovely to annotate each field with the documentation from "All yaml options (click me)" in the [README](https://github.com/OpenAccess-AI-Collective/axolotl#config).

README.md
```diff
@@ -543,7 +543,7 @@ is_mistral_derived_model:
 is_qwen_derived_model:
 
 # optional overrides to the base model configuration
-model_config:
+model_config_overrides:
```
NanoCode012, line 546: I just noticed this, but we would need to deprecate the old name (add valueerror)

winglian: we can't actually deprecate it with Pydantic because `model_config` is an internal variable name for pydantic models

PR #1241, opened 2024-02-01
fix(model): apply gate fp32 only for mixtral

NanoCode012:
Gate was accidentally moved to fp32 for non-mixtral models.

This has been discovered by Nafnlaus on discord: https://discord.com/channels/1104757954588196865/1104757955204743201/1202445059438542899

# Description

Details on negative effect: https://discord.com/channels/1104757954588196865/1104757955204743201/1202287752486604853

Worse eval loss.

## Social Handles (Optional)

His discord handle is `@nafnlaus00`. Github: @enn-nafnlaus

### DIFF
--- a/src/axolotl/utils/models.py
+++ b/src/axolotl/utils/models.py
@@ -676,7 +676,7 @@ def load_model(
     if not cfg.fsdp:
         # FSDP doesn't like mixed Float and BFloat16
         for name, module in model.named_modules():
-            if "norm" in name or (model_config.model_type == "mixtral" and "gate" in name):
+            if "norm" in name or name == "gate":
                 module.to(torch.float32)
             if model_config.model_type == "btlm":
                 # don't upcast lm_head for btlm
--- a/src/axolotl/utils/models.py
+++ b/src/axolotl/utils/models.py
@@ -676,7 +676,7 @@ def load_model(
     if not cfg.fsdp:
         # FSDP doesn't like mixed Float and BFloat16
         for name, module in model.named_modules():
-            if any(m in name for m in ["norm", "gate"]):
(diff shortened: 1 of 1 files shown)


src/axolotl/utils/models.py
```diff
@@ -671,7 +671,7 @@ def load_model(
     if not cfg.fsdp:
         # FSDP doesn't like mixed Float and BFloat16
         for name, module in model.named_modules():
-            if any(m in name for m in ["norm", "gate"]):
+            if "norm" in name or (model_config.model_type == "mixtral" and "gate" in name):
```
```suggestion
            if "norm" in name or name == "gate":
```
there are other MoE's out there too. also, `"gate" in name` will still still catch the `gate_proj` module in mixtral.

NanoCode012: do all MoE's strictly use the same name `gate`? Should we specifically check all those model types?

winglian: Maybe gate in name and proj not in name?

winglian: See https://huggingface.co/mlabonne/phixtral-2x2_8/tree/main

NanoCode012: I think `name == "gate"` would fail as:
```python
    if "gate" in name:
        print(name)

# model.layers.0.block_sparse_moe.gate
# model.layers.1.block_sparse_moe.gate
..
```

I have also checked, and I could not find `gate_proj` in either the modeling code or after initiating the model.

NanoCode012: Unresolving back as I think you resolved this while I was midtyping

NanoCode012: The other suggestion of yours may work: `"gate" in name and "proj" not in name`.

NanoCode012:
yep. that works!

PR #1254, opened 2024-02-04
add mps support

maximegmd:
# Description

Supports basic training on Mac M series.

## Motivation and Context

It partially solves Mac support.

## How has this been tested?

Ran a train job with lora-mps.yml from start to finish.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,8 @@
 """setup.py for axolotl"""
 
 from importlib.metadata import PackageNotFoundError, version
-
+from packaging.version import Version, parse
+import platform
 from setuptools import find_packages, setup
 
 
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -1,7 +1,9 @@
 """setup.py for axolotl"""
 
+import platform
 from importlib.metadata import PackageNotFoundError, version
 
+from packaging.version import Version, parse
```
winglian, line 6: If we're going to require packaging, we probably have to update the docs so people install this before axolotl. So we'll have to update the ci to install packaging too.

winglian:
@maximegmd I can help with this PR, but because your PR is from your `main` branch, it doesn't play nicely with github's maintainer tools. Could you submit this PR from a different branch from your fork please? This PR will be a great addition. Thanks!

setup.py
maximegmd: We already do say to install packaging first in the readme

maximegmd:
Replaced by #1264

Issue #1253, opened 2024-02-03
More save strategies for DPO training.


PhilipMay:
### ⚠️ Please check that this feature request hasn't been suggested before.

- [X] I searched previous [Ideas in Discussions](https://github.com/OpenAccess-AI-Collective/axolotl/discussions/categories/ideas) didn't find any similar feature requests.
- [X] I searched previous [Issues](https://github.com/OpenAccess-AI-Collective/axolotl/labels/enhancement) didn't find any similar feature requests.

### 🔖 Feature description

When I do DPO training it only allows steps save strategy.
This is because it is hard wired here:

https://github.com/OpenAccess-AI-Collective/axolotl/blob/2d65f470d54986f2df542cb8332e31113c04112d/src/axolotl/core/trainer_builder.py#L1095

The consequence is that it is not possible to specify "epoch" strategy and it is almost
impossible to save after the last step.

### ✔️ Solution

Make save_strategy configurable like in `HFCausalTrainerBuilder`.

### Acknowledgements

- [X] My issue title is concise, descriptive, and in title casing.
- [X] I have searched the existing issues to make sure this feature has not been requested yet.
- [X] I have provided enough information for the maintainers to understand and evaluate this request.

filippo82:
Hi @PhilipMay, are you willing to start working on a PR to improve this issue?

PhilipMay:
> Hi @PhilipMay, are you willing to start working on a PR to improve this issue?

I can do.

PhilipMay:
@filippo82 see #1255 - can you approve the workflows please?

filippo82:
Hi @PhilipMay, I'll review this as soon as I can.

filippo82:
@PhilipMay the corresponding PR was merged. Hence, you can close this issue.

PhilipMay:
> @PhilipMay the corresponding PR was merged. Hence, you can close this issue.

thanks

PR #1255, opened 2024-02-04
Add more save strategies for DPO training.

PhilipMay:
see #1253


PhilipMay:
Fixed: `TypeError: transformers.training_args.TrainingArguments() got multiple values for keyword argument 'save_steps'`

PhilipMay:
Hmm - did I break a regression test? 
@filippo82

filippo82:
I suspect that the tests are broken for other reasons ... 😅 

I had a look at your PR. I want to first check something in the [validate_config](https://github.com/OpenAccess-AI-Collective/axolotl/blob/main/src/axolotl/utils/config.py#L188) method.

Issue #1258, opened 2024-02-05
Seems `model_kwargs` is broken, resulting in failures


zacbrannelly:
### Please check that this issue hasn't been reported before.

- [X] I searched previous [Bug Reports](https://github.com/OpenAccess-AI-Collective/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

Trying to run a merge_lora operation using the axolotl Docker image on Azure cloud.

Config with:
```
model_kwargs:
  use_safetensors: true
```
should allow specifying a path to a model that is represented as .safetensors files.

### Current behaviour

Fails with:
```
File "/root/miniconda3/envs/py3.11/lib/python3.11/site-packages/transformers/modeling_utils.py", line 3120, in from_pretrained
    raise EnvironmentError(
OSError: Error no file named pytorch_model.bin, tf_model.h5, model.ckpt.index or flax_model.msgpack found in directory ...
```

### Steps to reproduce

Use the config in the expected behaviour.

### Config yaml

```
model_kwargs:
  use_safetensors: true
```

### Possible solution

Seems to be a typo in here:
https://github.com/OpenAccess-AI-Collective/axolotl/blob/2d65f470d54986f2df542cb8332e31113c04112d/src/axolotl/utils/models.py#L394-L398

Needs to be:
```
model_kwargs: Dict[str, Any] = {}

    if cfg.model_kwargs:
        for key, val in cfg.model_kwargs.items():
            model_kwargs[key] = val
```

### Which Operating Systems are you using?

- [X] Linux
- [ ] macOS
- [ ] Windows

### Python Version

3.9.5

### axolotl branch-commit

main/2d65f47

### Acknowledgements

- [X] My issue title is concise, descriptive, and in title casing.
- [X] I have searched the existing issues to make sure this bug has not been reported yet.
- [X] I am using the latest version of axolotl.
- [X] I have provided enough information for the maintainers to reproduce and diagnose the issue.

NanoCode012:
Thank you for catching this. Would you be able to submit a PR for this?

zacbrannelly:
Done :) #1262

PR #1262, opened 2024-02-06
Fix typo preventing `model_kwargs` being injected

zacbrannelly:
# Description

* Fix issue #1258 

## Motivation and Context

I need to be able to inject `use_safetensors` into the model kwargs.

## How has this been tested?

Haven't tested the code, but don't see why it would break.
If required will find some time to test it.


NanoCode012 (approved):

winglian (approved):
great catch. thanks!

PR #1277, opened 2024-02-08
add support for https remote yamls

hamelsmu:
@winglian I tested this extensively.  I added additional error handling to dummy proof it against myself because I broke it several times by passing links to yaml files on GitHub without using the "raw" link.   I have also seen people passing around json as configs but wanted to throw a warning as most of the time that is a user error (but may not be).  

Finally I added some docs.  Thanks for starting this, this feature is pretty useful for quickly testing configs from other people!  Big Quality of Life improvement.


winglian (approved):
Thanks for the extra sanity checks and docs!

PR #1278, opened 2024-02-08
allow remote data paths

hamelsmu:
This ended up being pretty simple.  Added support for remote URL paths per discussion with @winglian - HF Datasets natively allow remote paths

### DIFF
--- a/src/axolotl/utils/data.py
+++ b/src/axolotl/utils/data.py
@@ -336,7 +336,7 @@ def load_tokenized_prepared_datasets(
                         split=None,
                         storage_options=storage_options,
                     )
-            elif config_dataset.path.startswith("http"):
+            elif config_dataset.path.startswith("https://"):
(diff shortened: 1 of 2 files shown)


src/axolotl/utils/data.py
```diff
@@ -336,6 +336,16 @@ def for_d_in_datasets(dataset_configs):
                         split=None,
                         storage_options=storage_options,
                     )
+            elif config_dataset.path.startswith("http"):
```
winglian: do we want to allow loading from non-ssl urls?

winglian (approved):
lgtm, only consideration might be the security of allowing unsecured http connections.

src/axolotl/utils/data.py
hamelsmu: ok I'll add the `s`!

PR #1289, opened 2024-02-12
run tests again on Modal

winglian:
# Description

follow up to #1217, run on modal runners to kick off jobs to modal


NanoCode012:
Seems like you hit timeout of 45mins:

```
FunctionTimeoutError: Task's current input in-wrd5Ktt6HmXMgu9T6pyhAj hit its 
timeout of 2700s
```

Is that timeout per task or for the entire run?

PR #1297, opened 2024-02-17
fix(examples): remove is_*_derived as it's parsed automatically

NanoCode012:
Remove `is_*_derived` from configs. This is particularly problematic for qwen configs.

# Description


winglian (approved):

winglian:
I'll probably need to move these to the DeprecatedConfig for #1239

PR #1298, opened 2024-02-17
fix(dataset): normalize tokenizer config and change hash from tokenizer class to tokenizer path

NanoCode012:
Fix hash across tokenizers. We should not base on tokenizer class name as they could have different vocabs.

# Description


NanoCode012:
Test failing due to not normalizing config, I believe.

winglian (approved):

PR #1303, opened 2024-02-20
[bug-report template] Use yaml codeblock for config.yaml field

kallewoof:
The default bug report does not triple-backtick-yaml the config file, resulting in weird markdown formatting that is hard to read.


winglian (approved):
thank you!

PR #1316, opened 2024-02-22
fix steps check for anneal on first cycle

winglian:
# Description

Fixes relora anneal step on first restart

### DIFF
--- a/src/axolotl/monkeypatch/relora.py
+++ b/src/axolotl/monkeypatch/relora.py
@@ -267,7 +267,7 @@ class ReLoRAScheduler(LRScheduler):
         original = self.inner_schedule.get_lr()
         step = self.last_epoch
 
-        if step < self.relora_steps:
+        if step < self.relora_steps - self.warmup_steps:
             scale = 1
         else:
             per_relora_progress = step % self.relora_steps


winglian:
@JohanWork any ideas why mlflow isn't playing nicely with the latest transformers update? It seems to cause a new error here https://github.com/OpenAccess-AI-Collective/axolotl/actions/runs/7998732389/job/21847085356#step:7:1

JohanWork:
@winglian I have no idea, to be honest, but I can try to debug it later today or tomorrow

src/axolotl/monkeypatch/relora.py
```diff
@@ -267,7 +267,7 @@ def get_lr(self) -> float:
         original = self.inner_schedule.get_lr()
         step = self.last_epoch
 
-        if step < self.relora_steps:
+        if step < self.relora_steps - self.warmup_steps:
```
NanoCode012, line 270: If `warmup_steps` is greater than `relora_steps`, wouldn't this never trigger?

winglian: yeah, then it would only anneal, but I don't think there is a good solution in that case, that's just bad hyperparams.

PR #1329, opened 2024-02-25
Introducing auto_install.sh

monk1337:
**Objective**: 

This PR introduces an automated setup script (auto_install.sh) designed to streamline the installation process of Axolotl. It addresses common challenges faced by users not utilizing Docker, particularly in configuring the correct CUDA, PyTorch versions, and related dependencies. This enhancement is motivated by [community feedback](https://www.reddit.com/r/LocalLLaMA/comments/18pk6wm/how_to_qlora_fine_tune_using_axolotl_zero_to/) highlighting difficulties in manual setup processes.

**Features and Implementation:**

- Conda Environment Setup: Automatically checks for an existing Conda installation and proceeds to install Conda if it's not found, ensuring a controlled setup environment.
- CUDA Version Management: Installs a specific CUDA version within the Conda environment tailored for compatibility, addressing common version mismatch issues.
- PyTorch Ecosystem Compatibility: Ensures the installation of PyTorch, TorchAudio, and TorchVision that are compatible with the selected CUDA version, mitigating dependency conflicts.
- Axolotl Dependencies: Automatically installs all required dependencies for Axolotl, facilitating a smoother setup experience.

**Planned Enhancements:**

- OS-Specific Installation: Currently it supports Linux only however, future iterations will aim to automatically identify the operating system and adjust the CUDA installation and other dependencies accordingly.
- Docker Support Argument: Incorporate an option to install Docker if required by the user, enhancing flexibility for users preferring containerized environments.
- Installation Validation: Introduce a simple test script capable of running a small model on a dataset of 500 samples to verify the correctness of the installation and operational integrity of Axolotl post-setup.

**Note:** This script presents an additional option for users seeking an easier installation route without mandating its use. Users retain the freedom to choose their preferred installation method. The primary goal of this script is to reduce setup time and complexity, offering an alternative to manual configuration.

### DIFF
--- a/examples/mistral/mixtral_22.yml
+++ b/examples/mistral/mixtral_22.yml
@@ -1,0 +1,59 @@
+base_model: mistral-community/Mixtral-8x22B-v0.1
+model_type: AutoModelForCausalLM
+tokenizer_type: LlamaTokenizer
+trust_remote_code: true
+
+load_in_8bit: false
+load_in_4bit: false
+strict: false
+
+unfrozen_parameters:
+  - ^lm_head.weight$
+  - ^model.embed_tokens.weight$
+  - model.layers.4[4-9]+.block_sparse_moe.gate
+  - model.layers.4[4-9]+.block_sparse_moe.experts
+  - model.layers.5[0-5]+.block_sparse_moe.gate
+  - model.layers.5[0-5]+.block_sparse_moe.experts
+  
+model_config:
+  output_router_logits: true
+
+DATA_STUFF_HERE
+output_dir: ./out
+
+sequence_len: 8000
+sample_packing: true
+pad_to_sequence_len: true
+
+gradient_accumulation_steps: 1
+micro_batch_size: 1
+num_epochs: 3
+optimizer: adamw_bnb_8bit
+lr_scheduler: cosine
+learning_rate: 0.0001
+
+train_on_inputs: false
+group_by_length: false
+bf16: auto
+fp16:
+tf32: false
+
+gradient_checkpointing: true
+early_stopping_patience:
+resume_from_checkpoint:
+local_rank:
+logging_steps: 1
+xformers_attention:
+flash_attention: true
+
+save_total_limit: 1
+save_steps:
+debug:
+deepspeed: deepspeed_configs/zero3_bf16_cpuoffload_all.json
+weight_decay: 0.0
+fsdp:
+fsdp_config:
+special_tokens:
+  eos_token: "<|im_end|>"
+tokens:
+  - "<|im_start|>"


maximegmd:
I am a bit opposed to the auto install being at the root, while it covers the "common" use case, there is a growing variety of setups (ROCm, MPS, XPU...)

I also think that there is no need for such a script as the docker images cover the difficult part of setting it up. So I am not sure who this helps, power users already have a conda install and most likely the correct pytorch version. End users should just pull the docker image.

monk1337:
> I am a bit opposed to the auto install being at the root, while it covers the "common" use case, there is a growing variety of setups (ROCm, MPS, XPU...)
> 
> I also think that there is no need for such a script as the docker images cover the difficult part of setting it up. So I am not sure who this helps, power users already have a conda install and most likely the correct pytorch version. End users should just pull the docker image.

You make a fair case that the Docker images likely cover much of the complexity for many users. However, as you rightly mentioned, not all users currently utilize Docker. I recently came across several incidents on Discord and elsewhere where users faced issues installing Axolotl, for example this popular  Reddit post [Reddit post](https://www.reddit.com/r/LocalLLaMA/comments/18pk6wm/how_to_qlora_fine_tune_using_axolotl_zero_to/), where a user faced frustration while installing Axolotl. This inspired me to consider simplifying the initial setup process for those who could benefit. 

It can also support different platforms. I just created a very initial version; there is more to go. Moreover, it would just serve as an additional option for users - they could follow other methods if they prefer.

mhenrichsen:
Gemma config change and auto setup should be two different PR, so they don't go in together.

Also, please add some documentation that this autoconfig is only for linux.

monk1337:
> Gemma config change and auto setup should be two different PR, so they don't go in together.
> 
> Also, please add some documentation that this autoconfig is only for linux.

Done, Thank you!

winglian:
Thanks @monk1337, I do find value in having this script as having docker adds another layer of complexity over a native install. I do worry that having a script adds another point of maintenance/testing/support. Currently our docker images are well tested as we typically run tests against the built images themselves. 

I think a good compromise would be to indicate that the install script itself is "beta" and have a date somewhere where it was last updated/tested, just for visibility to end users.

auto_install.sh
```diff
@@ -0,0 +1,57 @@
...
+conda init bash
+
+# Activate the newly created Conda environment
+echo "Activating the 'axolotl' environment..."
+source activate axolotl || source $HOME/miniconda3/bin/activate axolotl
+
+# Install CUDA and PyTorch with dependencies
+echo "Installing CUDA and PyTorch with dependencies..."
+conda install -y -c "nvidia/label/cuda-12.1.1" cuda --yes
+conda install pytorch==2.1.2 torchvision==0.16.2 torchaudio==2.1.2 pytorch-cuda=12.1 -c pytorch -c nvidia --yes
```
NanoCode012: Some systems do not support 12.1, so this may be an issue? One alternative could be like ooba where they allow option to use 11.8?

auto_install.sh
```diff
@@ -0,0 +1,57 @@
...
+# Change directory to axolotl
+cd axolotl
+
+# Install additional Python packages
+echo "Installing additional Python packages..."
+pip3 install packaging
+pip3 install -e '.[flash-attn,deepspeed]'
+
+pip3 uninstall deepspeed -y
+pip3 install deepspeed==0.13.1
```
NanoCode012:
```suggestion
```

This will not be necessary anymore. There's a separate PR which addresses this.

README.md
```diff
@@ -260,6 +260,17 @@ HF_TOKEN=xx sky launch axolotl.yaml --env HF_TOKEN
...
+
+To use the automated installation script, follow these steps in your terminal:
+
+```bash
+# Grant execution permissions to the script
+chmod +x auto_install.sh
+
+# Execute the script to start the installation
+./auto_install.sh
+```
```
NanoCode012: This section may be better integrated with the pip/conda section?

NanoCode012 (commented):
If this is the start of a one-stop installer (like how text-webui supports several options), this may be a good starting point. Otherwise, I believe it would be better to improve the current pip/conda installation instructions.

Edit: I also agree with suggestion above where this has to be stated for Nvidia-only.

PR #1362, opened 2024-03-05
Update tinyllama lora.yml to fix eval packing issue

rasbt:
Disables evaluation set sample packing for the TinyLlama Lora config. Without this setting, it raises a 

```
ValueError: eval dataset split is too small for sample_packing. You should set `eval_sample_packing: False`. 
```

# Description

Adds:

`eval_sample_packing: false`

## Motivation and Context

Fixes an error with the default config

## How has this been tested?

Ran it with the setting, and it then worked fine (see screenshot below)

Screenshot 2024-03-05 at 10 47 53 AM

## Social Handles (Optional)

@rasbt (on Twitter)

### DIFF
--- a/examples/tiny-llama/lora.yml
+++ b/examples/tiny-llama/lora.yml
@@ -15,6 +15,7 @@ output_dir: ./lora-out
 
 sequence_len: 4096
 sample_packing: true
+eval_sample_packing: false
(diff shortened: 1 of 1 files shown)


winglian (approved):
thanks!

PR #1364, opened 2024-03-06
Fix supported python versions in README, as python 3.9 was recently deprecated

nirogu:
Removes python3.9 from the README as the minimum required version, as it was [recently deprecated](https://github.com/OpenAccess-AI-Collective/axolotl/commit/6d4bbb877f7317bb8ee2ff59f08efc0df5f99bcb). Now, the README correctly states that python>=3.10 should be used.


winglian (approved):
Thanks for fixing this!

PR #1373, opened 2024-03-07
re-enable phi for tests in modal ci

winglian:
# Description


NanoCode012:
It looks like disk writing failed in CI. Could this have been due lack of storage?

thundergolfer:
@winglian as said in Slack, I think this will pass now 🙏

PR #1390, opened 2024-03-11
Train parameters exclusively in specific ranges

seungduk-yanolja:
Train parameters exclusively in specific ranges

# Description

This PR enables the freezing configuration to specify parameter ranges so that we can ignore gradients for specific tokens.
For more details, please take a look at this technical report.
https://arxiv.org/abs/2402.14714

## Motivation and Context

When some tokens are added to the vocab, we may want to train only those specific tokens. For example, while fine-tuning, imagine you want to train specific control tokens only.
This technique can be also used when expanding the vocabulary for non-English languages.

## How has this been tested?

I wrote unit tests for it.

N/A

## Types of changes

New feature (non-breaking change which adds functionality)

### DIFF
--- a/examples/mistral/mixtral.yml
+++ b/examples/mistral/mixtral.yml
@@ -16,12 +16,12 @@ output_dir: ./qlora-out
 
 ## You can optionally freeze the entire model and unfreeze a subset of parameters
 unfrozen_parameters:
-#  - lm_head.*
-#  - model.embed_tokens.*
-#  - model.layers.2[0-9]+.block_sparse_moe.gate.*
-#  - model.layers.2[0-9]+.block_sparse_moe.experts.*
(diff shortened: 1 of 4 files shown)


examples/mistral/mixtral.yml
```diff
@@ -16,12 +16,12 @@ output_dir: ./qlora-out
 
 ## You can optionally freeze the entire model and unfreeze a subset of parameters
 unfrozen_parameters:
-#  - lm_head.*
```
seungduk-yanolja, line 19: The previous example was technically wrong (but worked) because this becomes the following:
`lm_head\.*` and checks if the dot repeats 0+ times

winglian (approved):
thanks!

PR #1396, opened 2024-03-12
Update ChatTemplate enum to include alpaca and gemma

chiragjn:
Was going through internals and trying gemma format but pydantic validation wouldn't let it pass
Not sure if we want to expose `alpaca`


winglian (approved):
lgtm, thanks!

PR #1370, opened 2024-03-07
WIP: galore optimizer

maximegmd:
Adds support for Galore optimizers

Still a WIP, untested.

PR #1409, opened 2024-03-16
support galore once upstreamed into transformers

winglian:
Supersedes #1370, pending merging of Galore upstream in transformers. We'll also need to bump transformers in our requirements once that is merged upstream.

https://github.com/huggingface/transformers/pull/29588

### DIFF
--- a/README.md
+++ b/README.md
@@ -883,7 +883,26 @@ lr_div_factor: # Learning rate div factor
 # - paged_adamw_8bit
 # - paged_lion_32bit
 # - paged_lion_8bit
+# - galore_adamw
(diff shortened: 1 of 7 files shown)


fakerybakery:
Hi, super excited to try this out! Is there an example config file?

winglian:
> Hi, super excited to try this out! Is there an example config file?

```yaml
optimizer: galore_adamw | galore_adamw_8bit | galore_adafactor
optim_args:
  rank: 128
  update_proj_gap: 200
  scale: 0.25
  proj_type: std
optim_target_modules:
  - mlp
  - attn
```

PR #1412, opened 2024-03-16
fix(config): passing gradient_checkpoint_kwargs

NanoCode012:
According to https://github.com/huggingface/transformers/issues/28339 , setting it to `False` increases VRAM. My quick testing shows ~1GB increase at lowest settings.

Furthermore, the default in transformers and torch is going to be `True` https://github.com/huggingface/transformers/issues/29638#issuecomment-1995034098

Finally, removing the default to False in trainer_building to clean old configs. I see that this kwarg is now set in https://github.com/OpenAccess-AI-Collective/axolotl/blob/a914cb37dc455a3fd0368e3a0898867f25b3a6c9/src/axolotl/utils/config/__init__.py#L170-L176

# Description

### DIFF
--- a/README.md
+++ b/README.md
@@ -843,7 +843,7 @@ group_by_length: false
 gradient_checkpointing: false
 # additional kwargs to pass to the trainer for gradient checkpointing
 # gradient_checkpointing_kwargs:
-#   use_reentrant: false
+#   use_reentrant: true
 
 # Stop training after this many evaluation losses have increased in a row
 # https://huggingface.co/transformers/v4.2.2/_modules/transformers/trainer_callback.html#EarlyStoppingCallback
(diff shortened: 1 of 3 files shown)


winglian:
In my experience, `use_reentrant: False` used less VRAM prevented OOM in some situations where `use_reentrant: True` would OOM, hence the current default.

winglian:
see my comment in #1167

NanoCode012:
In that code block above, it seems that the `validate_config` which runs first, sets it to True by default already.

Then, the `trainer_builder` which runs second, redundantly checks for it again and tries to set it to False.

Either way, I think one of the defaults should be removed to prevent future confusion.

Edit: your linked PR sets it to true, despite the comment saying it's false.

winglian (approved):

winglian:
> In that code block above, it seems that the `validate_config` which runs first, sets it to True by default already.
> 
> Then, the `trainer_builder` which runs second, redundantly checks for it again and tries to set it to False.
> 
> Either way, I think one of the defaults should be removed to prevent future confusion.
> 
> Edit: your linked PR sets it to true, despite the comment saying it's false.

Thanks for digging into this. Good to go!

PR #1413, opened 2024-03-16
fix for accelerate env var for auto bf16, add new base image and expand torch_cuda_arch_list support

winglian:


src/axolotl/utils/trainer.py
```diff
@@ -319,6 +320,11 @@ def prepare_optim_env(cfg):
         os.environ["ACCELERATE_USE_DEEPSPEED"] = "true"
         os.environ["ACCELERATE_DEEPSPEED_CONFIG_FILE"] = cfg.deepspeed
 
+    if (cfg.bf16 == "auto" and is_torch_bf16_gpu_available()) or cfg.bf16 is True:
```
NanoCode012, line 323: If `cfg.bf16` is true, shouldn't we also check for whether it's available?

I think there should be an assertion check elsewhere like in validate_config.

```py
if cfg.bf16 and not is_torch_bf16_gpu_available(): raise ValueError
```

winglian: we already check this against the capabilities of the gpu in the `check_bf16` method

PR #1414, opened 2024-03-16
fix optimizer reset for relora sft

winglian:
the relora optimizer reset wasn't pruning anything, this fixes that. 

what seemed to be wrong with the original implementation is that iterating over the optimizer states isn't sufficient to determine the optimizer keys.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -2,4 +2,3 @@ pre-commit
 black
 mypy
 types-requests
-tbparse
--- a/requirements-tests.txt
+++ b/requirements-tests.txt
@@ -2,3 +2,4 @@ pytest
 pytest-xdist
 pytest-retry
 pytest-sugar
+tbparse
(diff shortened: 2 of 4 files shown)


winglian:
hmm, still doesn't seem to be correct,

```
[2024-11-21 21:17:08,843] [INFO] [axolotl.relora.reset_optimizer:72] [PID:2840] [RANK:0] Percent of optimizer states zeroed: 0.00                                                                                                                         
[2024-11-21 21:17:08,843] [INFO] [axolotl.relora.reset_optimizer:73] [PID:2840] [RANK:0] absolute n of optimizer states zeroed: 0                                                                                                                         
[2024-11-21 21:17:09,011] [INFO] [axolotl.relora.reset_optimizer:72] [PID:2841] [RANK:1] Percent of optimizer states zeroed: 0.00                                                                                                                         
[2024-11-21 21:17:09,011] [INFO] [axolotl.relora.reset_optimizer:73] [PID:2841] [RANK:1] absolute n of optimizer states zeroed: 0   
```
also, ooms after the first relora restart

winglian:
reference implementation for Relora is here https://github.com/Guitaricet/relora/blob/main/peft_pretraining/training_utils.py#L267-L364

PR #1425, opened 2024-03-20
HF / FEAT: Optimize HF tags

younesbelkada:
# What does this PR do?

Currently the `axolotl` tag is pushed only if a user calls `trainer.push_to_hub()`, this PR makes sure whenever a user calls `model.push_to_hub()` the tag `axolotl` gets properly pushed on the Hub

cc @winglian


winglian (approved):
this is great. thanks!

younesbelkada:
Thanks so much @winglian !

PR #1429, opened 2024-03-22
Bootstrap Hosted Axolotl Docs w/Quarto

hamelsmu:
I discussed this with @winglian on Discord.  I might merge to test if the machinery works.

**Update** The docs are here: https://openaccess-ai-collective.github.io/axolotl/

I tried not to disrupt the directory structure too much so that people do not get lost.  We can change this over time.  This is why all the docs are still in the `docs/` directory, and some files are now in the root of this repo.  I had to put `index.qmd`, `styles.css` and `_quarto.yml` in the root of this repo to avoid moving files. 

Also, I'm programmatically sourcing contents from the README for the home page, and also **a separate page that just has the configs - this is personally the section of the README that I visit the most often so I duplicated it into its own page**


hamelsmu:
@winglian I'll go ahead and merge this, but you will have to update your repo settings (I don't have access) to enable GitHub pages.  Also perhaps you want to add a `CNAME` to the root of this repo to point at a specific domain?

Issue #1430, opened 2024-03-22
Loading local Dataset saved via `save_to_disk` fails


fozziethebeat:
### Please check that this issue hasn't been reported before.

- [X] I searched previous [Bug Reports](https://github.com/OpenAccess-AI-Collective/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

I create some local dataset that I saved via calling `ds.save_to_disk('some_path')`

If I then configure Axolotl to load a dataset from `save_path` such as

```yaml
datasets: 
- path: tmp_dataset
...
```

I would expect it to load much like loading from a remote huggingface repo.

### Current behaviour

The loaded dataset includes the contents of `state.json` rather than the actual dataset and everything fails.

### Steps to reproduce

1. Save some dataset to disk via `save_to_disk`
2. Configure Axolotl to read from that dataset
3. See Axolotl crash

### Config yaml

```yaml
datasets:
- path: local_dataset_saved_via_save_to_disk
  type:
    field_instruction: prompts
    field_output: responses
    field_system: system
    format: '[INST] {instruction} [/INST]'
    no_input_format: '[INST] {instruction} [/INST]'
    system_prompt: ''
```

### Possible solution

In `src/axolotl/utils/data.py` I see that for local directory paths, datasets are loaded via `load_dataset`.  `load_from_disk` seems to be the new preferred solution (this is also a comment).  I manually hacked this in and everything worked without a problem.

Would switching that to `load_from_disk` break anything? It seems like a reasonable quick fix (which I can do)

### Which Operating Systems are you using?

- [X] Linux
- [ ] macOS
- [ ] Windows

### Python Version

3.10

### axolotl branch-commit

main/4e69aa4

### Acknowledgements

- [X] My issue title is concise, descriptive, and in title casing.
- [X] I have searched the existing issues to make sure this bug has not been reported yet.
- [X] I am using the latest version of axolotl.
- [X] I have provided enough information for the maintainers to reproduce and diagnose the issue.

anttttti:
I've had to work around this for a while. You'll get:
`ValueError: You are trying to load a dataset that was saved using `save_to_disk`. Please use `load_from_disk` instead.`

The workaround is to manually delete "state.json" from the dataset directory, and axolotl will then read it. But other HF-compatible tools expect "state.json" so this is a hassle

fozziethebeat:
[...]

PR #1432, opened 2024-03-22
Support loading datasets saved via save_to_disk

fozziethebeat:
# Description

Fixes  

Uses `load_from_disk` when no data files are listed in a local directory dataset.  This is to ensure datasets saved via `save_to_disk` are properly loaded.

## Motivation and Context

Datasets saved via `save_to_disk` currently can't be loaded without explicitly listing all files.
#1430

## How has this been tested?

Created a local dataset saved via `save_to_disk` and trained a model on it.

## Types of changes

Small code change

## Social Handles (Optional)

fozziethebeat on all the platforms


winglian (approved):
lgtm. @NanoCode012 ?

winglian:
Looks like there are some issues with this change and the current integration tests we have, especially when loading remote datasets.

fozziethebeat:
I'll look through the docker tests that failed and add unittests for what's failing.

fozziethebeat:
I think this should fix the tests.  The updates to main require `tokenizer_config` to be set in the unittests given that its not using the config utils that auto-set the field.

fozziethebeat:
One E2E test suite failed with this error:

```
Writing manifest to image destination
Unpacking OCI image
   • unpacking rootfs ...
Runner terminated.
Stopping app - uncaught exception raised locally: RemoteError('Image build for im-RlXPRxFbKAnNrpADzxyalb terminated due to external shut-down. Please try again.').
```

This seems like it needs a re-run.  The other E2E suite completed successfully

fozziethebeat:
yay~

PR #1438, opened 2024-03-25
Turn on sample_packing for Gemma training

satpalsr:
In last commit, sample_packing was entirely turned off #1405 
However, only `eval_sample_packing` must be kept false here.


winglian (approved):
good catch. thanks!

PR #1441, opened 2024-03-26
Fix falcon tokenization step

pharaouk:
Running example falcon 7b throws error:
ValueError: Column name ['token_type_ids'] not in the dataset. Current columns in the dataset: ['input_ids', 'attention_mask', 'labels']

I've added a check for whether token_type_ids even exists which cleared this bug.


winglian (approved):
Thanks. There is probably a linting issue, but I can fix that later this morning.

PR #1445, opened 2024-03-27
support layer replication for peft and fix rslora integration

winglian:
adds support for https://github.com/huggingface/peft/pull/1368


fakerybakery:
Hi, is there an example config for this?

ehartford:
How to set layer_replication in axolotl config?

xzuyn:
> How to set layer_replication in axolotl config?

If you wanted to do an [11B stack like Undi's recipe](https://huggingface.co/Undi95/Mistral-11B-v0.1), you would do `peft_layer_replication: [[0,24], [8,32]]`

PR #1463, opened 2024-03-30
fix pretraining_ on odd datasets

mapmeld:
# Description

I'm interested in using axolotl to pretrain and then finetune on a dataset which I have on HF.

1. Fixes an issue I experienced because the training script expects pretraining to have a split named "train" and single column named "text". I propose adding `split` and `name` to the `PretrainingDataset` parser and a new `text_column` field.

2. The `load_dataset(path, streaming=True)` used in pretraining currently returns `features: None` or `features: Unknown`, so I propose retrieving the first record to get column names (this is later passed to `remove_columns`).

As an alternative config change, I could see using `SFTDataset` and removing the separate `PretrainingDataset` type

## How has this been tested?

Output YAML, ran training script in CoLab notebook

Sample dataset with one column named "Comment"
```yaml
pretraining_dataset:
  - path: urkopa/comments-lv
    split: train
    text_column: Comment
    type: pretrain
```

My dataset where each split has a different schema so I need to specify split and name
```yaml
pretraining_dataset:
  - path: monsoon-nlp/greenbeing-proteins
    split: pretraining
    text_column: sequence
    name: pretraining
    type: pretrain
```

### DIFF
--- a/src/axolotl/prompt_strategies/pretrain.py
+++ b/src/axolotl/prompt_strategies/pretrain.py
@@ -20,10 +20,11 @@ class PretrainTokenizationStrategy(PromptTokenizingStrategy):
     def supports_batched(self):
         return True
 
-    def __init__(self, *args, max_length=None, **kwargs):
+    def __init__(self, *args, max_length=None, text_column="text", **kwargs):
         super().__init__(*args, **kwargs)
         if max_length:
             self.max_length = max_length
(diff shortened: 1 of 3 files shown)


NanoCode012:
Hm, is it possible to add unit tests for this?

winglian (approved):
this all lgtm. thanks!

PR #1465, opened 2024-03-30
feat: validate sample packing requires flash_attention

NanoCode012:
# Description

Solves https://github.com/OpenAccess-AI-Collective/axolotl/issues/449

We can also remove a few if checks because of this.

### DIFF
--- a/src/axolotl/utils/config/models/input/v0_4_1/__init__.py
+++ b/src/axolotl/utils/config/models/input/v0_4_1/__init__.py
@@ -659,9 +659,13 @@ class AxolotlInputConfig(
     @model_validator(mode="before")
     @classmethod
     def check_sample_packing_wo_flash(cls, data):
-        if data.get("sample_packing") and not data.get("flash_attention"):
+        if (
+            data.get("sample_packing")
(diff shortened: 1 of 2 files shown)


NanoCode012:
Hm, I found that we support sample_packing with sdpa, so this PR and relevant Issue should be closed?

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
```diff
@@ -624,6 +625,16 @@ def check_sample_packing_w_xformers(cls, data):
 
         return data
 
+    @model_validator(mode="before")
+    @classmethod
+    def check_sample_packing_wo_flash(cls, data):
+        if data.get("sample_packing") and not data.get("flash_attention"):
```
winglian:
```suggestion
        if data.get("sample_packing") and not data.get("flash_attention") and not data.get("sup_attention"):
```
we also support sample packing for sdpa.

NanoCode012: fixed. thanks

winglian (approved):

PR #1466, opened 2024-03-30
feat: add deepspeed 3 with cpuoffload

NanoCode012:
# Description

Solves https://github.com/OpenAccess-AI-Collective/axolotl/issues/1216


winglian:
with zero3, if you don't explicitly specify bf16: true, it implicitly loads in fp32 since it needs to know at model load time. Also, I added a param only offload since the optimizer offload can blow up CPU ram on multrigpu environments.

winglian (approved):

NanoCode012:
That sounds good!

PR #1468, opened 2024-03-31
Reorganize Docs

hamelsmu:
This is to start cleaning up the README and organizing docs so it is more navigable and searchable

1. Move dataset construction out of README
2. Move all config options out of README

I also fix a few bugs in the docs site. 

@winglian preview of rendered site is [here](https://rococo-pasca-245a07.netlify.app/)


winglian (approved):
Much cleaner. thanks!

PR #1470, opened 2024-03-31
Pretrain multipack v2

winglian:
This PR enables more traditional pretraining, concatenating datasets, but not handling cross attention between the packed sequences. This helps with filling the context length when a sample doesn't come close to using the configured context length.

the multipack buffer size is how far ahead to look to try to pack samples together for pretraining/streaming datasets. What happens is that we pull in N samples and attempt to optimally pack those since we can't pack the entire dataset b/c they are usually too large to pre-process.


ehartford:
Thank you for the attention to pretraining

PR #1471, opened 2024-03-31
improved Jamba deepspeed z3 compat

tmm1:
# Description

implements z3 leaf for jamba

## Motivation and Context

as requested on https://discord.com/channels/1104757954588196865/1104758010959634503/1224077312648024235

### DIFF
--- a/src/axolotl/utils/models.py
+++ b/src/axolotl/utils/models.py
@@ -69,7 +69,7 @@ def get_module_class_from_name(module, name):
         if module_class is not None:
             return module_class
 
-    raise ValueError(f"module class {name} not found in {module.__class__.__name__}")
+    return None
 
 
 def check_model_config(cfg: DictDefault, model_config: Union[AutoConfig, DictDefault]):
--- a/src/axolotl/utils/models.py
+++ b/src/axolotl/utils/models.py
@@ -49,6 +49,29 @@ from axolotl.utils.lora_embeddings import get_linear_embedding_layers
 LOG = logging.getLogger("axolotl")
 
 
+# copied from accelerator.FullyShardedDataParallelPlugin
+def get_module_class_from_name(module, name):
+    """
+    Gets a class from a module by its name.
+    Args:
+        module (`torch.nn.Module`): The module to get the class from.
+        name (`str`): The name of the class.
+    """
+    modules_children = list(module.children())
+    if module.__class__.__name__ == name:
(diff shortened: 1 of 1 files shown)


bjoernpl:
Getting `ValueError: module class JambaSparseMoeBlock not found in JambaMambaMixer`. Probably because it's going down the JambaMambaMixer recursive path first, which has children but none of them are the required module -> it raises an exception.

src/axolotl/utils/models.py
```diff
@@ -48,6 +48,29 @@
...
+
+    if len(modules_children) == 0:
+        return None
+
+    for child_module in modules_children:
+        module_class = get_module_class_from_name(child_module, name)
+        if module_class is not None:
+            return module_class
+
+    raise ValueError(f"module class {name} not found in {module.__class__.__name__}")
```
```suggestion
```

I added this when copying it over, but it's probably incorrect since it recurses above and shouldn't raise if not found in a child.

winglian:
> Getting `ValueError: module class JambaSparseMoeBlock not found in JambaMambaMixer`. Probably because it's going down the JambaMambaMixer recursive path first, which has children but none of them are the required module -> it raises an exception.

@bjoernpl I updated so this should work now.

winglian:
maybe we don't need this for the Jamba MoE? 

```
raise RuntimeError(
RuntimeError: tracing error at step 916: 
module id: 1565, training: True
expected the next 1 parameters in the parameter fetch queue to be ({'id': 'name=base_model.model.model.layers.31.mamba.dt_proj.lora_A.default.weight id=3012', 'status': 'AVAILABLE', 'numel': 2048, 'ds
_numel': 2048, 'shape': (8, 256), 'ds_shape': (8, 256), 'requires_grad': True, 'grad_shape': None, 'persist': True, 'active_sub_modules': {1560, 1565}, 'ds_tensor.shape': torch.Size([256])},) 
but got 
 ({'id': 'name=base_model.model.model.layers.31.mamba.dt_proj.base_layer.bias id=3011', 'status': 'AVAILABLE', 'numel': 8192, 'ds_numel': 8192, 'shape': (8192,), 'ds_shape': (8192,), 'requires_grad': 
False, 'grad_shape': None, 'persist': True, 'active_sub_modules': {1537}, 'ds_tensor.shape': torch.Size([1024])},).
```

NanoCode012:
Is this still needed?

PR #1476, opened 2024-04-04
refactor utils.data module for line count linter

winglian:
The CI was complaining about too many lines in utils/data.py, so refactoring it into more specific modules to workaround that.

### DIFF
--- a/src/axolotl/utils/data/dpo.py
+++ b/src/axolotl/utils/data/dpo.py
@@ -1,0 +1,114 @@
+"""data handling specific to DPO"""
+
+import logging
+from pathlib import Path
+from typing import Any, List
+
+import yaml
+from datasets import concatenate_datasets, load_dataset, load_from_disk
+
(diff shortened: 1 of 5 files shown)


src/axolotl/utils/data/dpo.py
```diff
@@ -0,0 +1,114 @@
...
+from axolotl.common.const import DEFAULT_DATASET_PREPARED_PATH
+from axolotl.prompt_strategies.dpo import load as load_dpo
+from axolotl.utils.data.utils import md5
+from axolotl.utils.dict import DictDefault
+from axolotl.utils.distributed import is_main_process, zero_first
+
+LOG = logging.getLogger("axolotl")
+
+
+def _get_path(ds_hash, cfg):
```
NanoCode012, line 19: I think these 3 utility functions make more sense at `./data/utils`?

NanoCode012 (approved):
I think it's okay. We can maybe start removing some pylint ignores in a separate PR.

PR #1477, opened 2024-04-04
Guide For Multi-Node Distributed Finetuning

shahdivax:
Title: Distributed Finetuning for Multi-Node Setup Guide

Description:
This PR introduces a comprehensive guide for setting up a distributed finetuning environment using Axolotl and Accelerate. The guide covers the following steps:

1. Configuring SSH for passwordless access across multiple nodes
2. Generating and exchanging public keys for secure communication
3. Configuring Axolotl with shared settings and host files
4. Configuring Accelerate for multi-node training with Deepspeed
5. Running distributed finetuning using Accelerate


NanoCode012:
Thanks for sharing. I think we converted to a `qmd`. Also, we have an existing `multi-node` doc. Maybe you could edit that one instead?

The previous one uses fsdp whereas yours use deepspeed, so you can create a heading for fsdp which is the current doc, then one for deepspeed (or you can combine them to make it clearer).

shahdivax:
> Thanks for sharing. I think we converted to a `qmd`. Also, we have an existing `multi-node` doc. Maybe you could edit that one instead?
> 
> The previous one uses fsdp whereas yours use deepspeed, so you can create a heading for fsdp which is the current doc, then one for deepspeed (or you can combine them to make it clearer).

Sure, will edit the existing one. As you mentioned.

shahdivax:
> Thanks for sharing. I think we converted to a `qmd`. Also, we have an existing `multi-node` doc. Maybe you could edit that one instead?
> 
> The previous one uses fsdp whereas yours use deepspeed, so you can create a heading for fsdp which is the current doc, then one for deepspeed (or you can combine them to make it clearer).

Hello, I have created another pull request for the same https://github.com/OpenAccess-AI-Collective/axolotl/pull/1688 
please review it thanks

PR #1483, opened 2024-04-05
Update SaveAxolotlConfigtoWandBCallback to use artifact instead of save

tcapelle:
Removing the use `wandb.save` in favor of `wandb.artifact`

### DIFF
--- a/src/axolotl/utils/callbacks/__init__.py
+++ b/src/axolotl/utils/callbacks/__init__.py
@@ -748,7 +748,9 @@ class SaveAxolotlConfigtoWandBCallback(TrainerCallback):
                     mode="w", delete=False, suffix=".yml", prefix="axolotl_config_"
                 ) as temp_file:
                     copyfile(self.axolotl_config_path, temp_file.name)
-                    wandb.save(temp_file.name)
+                    at = wandb.Artifact(f"config-{wandb.run.id}", type="axolotl-config")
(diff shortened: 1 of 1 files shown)


winglian:
@tcapelle So we had originally implemented it this way here: https://github.com/OpenAccess-AI-Collective/axolotl/pull/716

but, people (I agree too) found it annoying having to find it in the artifacts so it was changed to it's current state here. https://github.com/OpenAccess-AI-Collective/axolotl/pull/1014

Let's discuss what is sane. perhaps some sort of configurability for it?

src/axolotl/utils/callbacks/__init__.py
```diff
@@ -748,6 +748,9 @@ def on_train_begin(
                     mode="w", delete=False, suffix=".yml", prefix="axolotl_config_"
                 ) as temp_file:
                     copyfile(self.axolotl_config_path, temp_file.name)
+                    at = wandb.Artifact(f"config-{wandb.run.id}", type="axolotl-config")
+                    at.add_file(temp_file.name)
+                    wandb.log_artifact(at)
```
winglian:
```suggestion
                    artifact = wandb.Artifact(f"config-{wandb.run.id}", type="axolotl-config")
                    artifact.add_file(temp_file.name)
                    wandb.log_artifact(artifact)
```
the linter doesn't like `at`

PR #1496, opened 2024-04-08
Print versions

winglian:
print out the various major dependencies and their versions for troubleshooting


NanoCode012:
Would it also be good to add axolotl git commit?

winglian:
> Would it also be good to add axolotl git commit?

that's going to be a bit more complicated and need something like https://setuptools-git-versioning.readthedocs.io/en/stable/runtime_version.html which we can tackle as another PR

PR #1501, opened 2024-04-09
Fix the wrong adapter in qwen2-moe-qlora example

maziyarpanahi:
It should be `qlora` instead of `lora`

# Description

## Motivation and Context

The example is for QLoRA, so in a copy/paste the `adapter` causes an exception


winglian (approved):
good catch. thanks!

PR #1512, opened 2024-04-10
add tests for merging lora and validating the dtype

winglian:


tests/e2e/test_lora_llama.py
```diff
@@ -65,3 +64,67 @@ def test_lora(self, temp_dir):
...
+
+        train(cfg=cfg, cli_args=cli_args, dataset_meta=dataset_meta)
+        assert (Path(temp_dir) / "adapter_model.bin").exists()
+
+        cfg.lora_model_dir = cfg.output_dir
+        cfg.load_in_4bit = False
+        cfg.load_in_8bit = False
+        cfg.flash_attention = False
+        cfg.deepspeed = None
+        cfg.fsdp = None
```
NanoCode012, line 112: This can be excluded as the modify_cfg_for_merge should've set it?
```suggestion
```

tests/e2e/test_lora_llama.py
```diff
@@ -65,3 +64,67 @@ def test_lora(self, temp_dir):
...
+
+        cfg = modify_cfg_for_merge(cfg)
+        cfg.merge_lora = True
```
NanoCode012, line 115: Let's move this setting inside the modify_cfg function as well.

src/axolotl/cli/merge_lora.py
```diff
@@ -27,21 +28,26 @@ def do_cli(config: Path = Path("examples/"), **kwargs):
         flash_attention=False,
```
NanoCode012, line 28: If the above section already sets these properties, is it necessary to set it again below?

tests/e2e/test_lora_llama.py
```diff
@@ -65,3 +64,67 @@ def test_lora(self, temp_dir):
 
         train(cfg=cfg, cli_args=cli_args, dataset_meta=dataset_meta)
         assert (Path(temp_dir) / "adapter_model.bin").exists()
+
+    @with_temp_dir
+    def test_lora_merge(self, temp_dir):
+        # pylint: disable=duplicate-code
+        cfg = DictDefault(
+            {
+                "base_model": "JackFram/llama-68m",
```
NanoCode012, line 73: Also, sometimes, this issue can occur for different model types. For ex, previous llama merge was fine, but mistral was not. Do we need to test this for other arch?

tests/e2e/test_lora_llama.py
```diff
@@ -65,3 +64,67 @@ def test_lora(self, temp_dir):
...
+                "lr_scheduler": "cosine",
+                "max_steps": 10,
+                "bf16": "auto",
+            }
+        )
+        normalize_config(cfg)
+        cli_args = TrainerCliArgs()
+        dataset_meta = load_datasets(cfg=cfg, cli_args=cli_args)
+
+        train(cfg=cfg, cli_args=cli_args, dataset_meta=dataset_meta)
```
NanoCode012, line 104: I don't think you need to train a model, maybe a tiny adapter can be uploaded to HF which we use for merge?

PR #1516, opened 2024-04-11
Switch to parallel FFD bin packing algorithm (closes #1492)

### LINKED ISSUE #1492: More efficient sample packing

### 🔖 Feature description

Move packing to the preprocess step and add a slower but more efficient algorithm.

### ✔️ Solution

[ReEvo](https://github.com/ai4co/LLM-as-HH) is a way of using LLMs for generating heuristics to tackle optimization problems (similar to FunSearch.) They have an online bin packing example, which I ran using GPT-4 to see if we could get something better suited for the distribution seen in sample packing.

Here's a script to try it:
https://gist.github.com/dsesclei/4fcf3763f07feaf67b4141429afb3fb8

It does very well compared to first-fit-decreasing from `multipack.py`. For `teknium/OpenHermes-v2` with context size 4096, FFD gives 109138 bins while this returns 8.3% fewer at 100076 (the minimum possible being 100061.) This could have been cut down even further by having GPT-4 generate a heuristic specifically for that dataset, though it doesn't seem worth it at that point.

Packing this way is much slower - OpenHermes took about five hours - so it would need to be moved to the preprocess step.

If this sounds good, I'm happy to implement - just wanted to open a discussion first.

### ❓ Alternatives

We can try generating a faster heuristic, although we'd probably want to put anything short of near-instant into preprocessing anyway.

### 📝 Additional Context

_No response_

### Acknowledgements

dsesclei:
# Description

Replace the existing sample packing algorithm with a parallel implementation of first-fit-decreasing.

## Motivation and Context

I noticed [recently](https://github.com/OpenAccess-AI-Collective/axolotl/issues/1492) that we could get denser sample packing with a different algorithm. Looking into it more, FFD performs just as well and is much faster than the heuristic I had 😅.

We can run FFD in parallel without losing much performance by packing samples in groups rather than all at once. On an i9-14900k, it takes 2.2s to pack 1M samples with 99.7% efficiency (current `multipack.py` is 91.7% in 0.32s.)

I removed the length estimates around packing in favor of just counting the batches, but let me know if I should add that back in. Two new config options are added: `sample_packing_group_size` controls the the number of samples packed by each process, and `sample_packing_bin_size` sets the number of samples that can be placed in one pack (may need to be increased for large context lengths.)

## How has this been tested?

Tests have been updated to verify that packing is correct. Training appears to run the same, just with fewer steps.

It seems reasonable that sorting the items in FFD would interfere with shuffling between epochs, but I haven't been able to find any evidence of that being the case. Testing against a few [similarity metrics](https://gist.github.com/dsesclei/1b63e782045fed52c93c1dc55657dfd2) shows that even when we do the packing at once in one group, shuffling still generates a mostly new set of packs.

## Screenshots

Some performance checks below for 1M items. 

group_size_vs_excess
bin_size_vs_excess

### DIFF
--- a/src/axolotl/utils/distributed.py
+++ b/src/axolotl/utils/distributed.py
@@ -4,25 +4,27 @@ utility helpers for distributed checks
 import os
 import pickle  # nosec
 from contextlib import contextmanager
-from datetime import timedelta
 
 import torch
 import torch.distributed as dist
-from accelerate import PartialState
+from accelerate import Accelerator
 
-distributed_state = None  # pylint: disable=invalid-name
+accelerate = None  # pylint: disable=invalid-name
+
+
+def load_accelerate():
+    global accelerate  # pylint: disable=global-statement
+    accelerate = Accelerator()
 
 
 def is_distributed():
     """
     Check if distributed training is initialized.
     """
-    global distributed_state  # pylint: disable=global-statement
-    if not distributed_state:
-        timeout = int(os.environ.get("AXOLOTL_NCCL_TIMEOUT", 1800))
-        distributed_state = PartialState(timeout=timedelta(seconds=timeout))
-
-    return distributed_state.use_distributed and distributed_state.initialized
+    global accelerate  # pylint: disable=global-statement
+    if not accelerate:
+        accelerate = Accelerator()
+    return dist.is_available() and dist.is_initialized()
 
 
 def barrier():
@@ -36,27 +38,16 @@ def barrier():
 
 def is_main_process():
     """
-    Return whether the current process is on the main rank.
+    Check if the current process is the main process.
+    If not in distributed mode, always return True.
     """
-    return get_rank() == 0
+    if not is_distributed():
+        return True
+    return dist.get_rank() == 0
 
 
 def get_world_size():
-    """
-    Return world size (1 if not distributed.)
-    """
-    if not is_distributed():
-        return 1
     return int(os.getenv("WORLD_SIZE", "1"))
-
-
-def get_rank():
-    """
-    Return rank (0 if not distributed.)
-    """
-    if not is_distributed():
-        return 0
-    return dist.get_rank()
 
 
 @contextmanager
--- a/src/axolotl/utils/distributed.py
+++ b/src/axolotl/utils/distributed.py
@@ -36,16 +36,27 @@ def barrier():
 
 def is_main_process():
     """
-    Check if the current process is the main process.
-    If not in distributed mode, always return True.
+    Return whether the current process is on the main rank.
     """
-    if not is_distributed():
-        return True
-    return dist.get_rank() == 0
+    return get_rank() == 0
 
 
 def get_world_size():
+    """
+    Return world size (1 if not distributed.)
+    """
+    if not is_distributed():
+        return 1
     return int(os.getenv("WORLD_SIZE", "1"))
+
+
+def get_rank():
+    """
+    Return rank (0 if not distributed.)
+    """
+    if not is_distributed():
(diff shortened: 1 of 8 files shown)


winglian:
> I removed the length estimates around packing in favor of just counting the batches, but let me know if I should add that back in. 

I need to do some checking, but the estimates exist due to different processes getting different splits of data, so the actual count of packed samples can vary from process to process. When this happens, you get one process thinking it needs to run another step, but another process thinking it's done and they get out of sync. The estimate was the most sane way I could come up with having each process come up with a deterministic length. I'm open to other ideas to working around this.

src/axolotl/utils/samplers/multipack.py
```diff
@@ -1,202 +1,119 @@
...
-        batch_max_len: int,
-        lengths: np.ndarray,
-        packing_efficiency_estimate: float = 1.0,
+        sampler,
+        lengths,
+        batch_max_len,
+        batch_size,
+        group_size,
+        bin_size,
+        shuffle=False,
```
winglian: I don't believe this is necessary, as the sampler should handle any shuffling and would be more deterministic from run to run.

dsesclei:
Could we generate all the packs, and then evenly split those up (like in the updated multipack.py)? I think each rank should then get an exact number of batches and stay in sync.

winglian:
> Could we generate all the packs, and then evenly split those up (like in the updated multipack.py)? I think each rank should then get an exact number of batches and stay in sync.

Perhaps we could do something like dispatch_batches=True to only run the packing on rank 0. I'm not 100% certain of the implications though

NanoCode012:
Hey, this is very interesting. Should there be some full run comparisons to make sure that there is no loss in performance?

dsesclei:
> Perhaps we could do something like dispatch_batches=True to only run the packing on rank 0. I'm not 100% certain of the implications though

Gotcha, for now I'll keep this PR simple by leaving the packing estimates in. Ready for another look.

> Hey, this is very interesting. Should there be some full run comparisons to make sure that there is no loss in performance?

Yeah definitely, once the code is greenlit/finalized I'll rent an instance to test it in a distributed setup.

src/axolotl/utils/distributed.py
```diff
@@ -4,25 +4,27 @@
...
     Check if distributed training is initialized.
     """
-    global distributed_state  # pylint: disable=global-statement
-    if not distributed_state:
-        timeout = int(os.environ.get("AXOLOTL_NCCL_TIMEOUT", 1800))
-        distributed_state = PartialState(timeout=timedelta(seconds=timeout))
-
-    return distributed_state.use_distributed and distributed_state.initialized
+    global accelerate  # pylint: disable=global-statement
+    if not accelerate:
```
winglian, line 25: Hey @dsesclei, Sorry for the delay in getting back to this PR. Is there a particular reason Accelerator was added back rather than using PartialState. It's best to not explicitly load up Accelerator until the last possible moment, and I believe we can get everything about the distributed state from PartialState.

winglian:
Hey @dsesclei we cherry picked and merged your fixes in #1619. Thanks! Would love to give you a shoutout if you're on twitter or discord and could share your handle. thanks!

dsesclei:
Thanks for getting this in Wing! No handles to give, but I appreciate it

winglian:
Thanks @dsesclei, I ended up having to revert the change b/c the loss was off by an order of magnitude. I need to dig into what the multipack sampler is outputting another time to see if there is something obvious that it is doing differently

dsesclei:
Oh gotcha, I'll look into it

PR #1525, opened 2024-04-16
Add data streaming support through `mosaic-streaming`

fmv1992:
# Description

This PR adds support for (non-volatile) memory efficient training through [StreamingDataset](https://github.com/mosaicml/streaming/blob/3ba93010ce86cba1dc4d51ab9977c8cbbbbfb2c9/streaming/base/dataset.py#L165).

## Motivation and Context

Context: https://github.com/OpenAccess-AI-Collective/axolotl/issues/585 .

## How has this been tested?

I have tested this through `docker` on a VM.

I'm open to ideas as to how this should be added. Does the repo support an `s3` bucket for instance?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -31,6 +31,7 @@ art
 fschat==0.2.36
 gradio==3.50.2
 tensorboard
+mosaicml-streaming
 
 mamba-ssm==1.2.0.post1
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -31,7 +31,7 @@ art
 fschat==0.2.36
 gradio==3.50.2
 tensorboard
-mosaicml-streaming
+mosaicml-streaming==0.7.5
 
 mamba-ssm==1.2.0.post1
 
--- a/setup.py
+++ b/setup.py
@@ -92,8 +92,5 @@ setup(
         "galore": [
             "galore_torch",
         ],
-        "mosaicml-streaming": [
-            "mosaicml-streaming",
-        ],
     },
 )
--- a/setup.py
+++ b/setup.py
@@ -92,5 +92,8 @@ setup(
         "galore": [
             "galore_torch",
         ],
+        "mosaicml-streaming": [
+            "mosaicml-streaming",
+        ],
     },
 )
--- a/src/axolotl/utils/data/sft.py
+++ b/src/axolotl/utils/data/sft.py
@@ -66,9 +66,10 @@ def load_streaming_dataset(config_dataset):
     - ds (datasets.Dataset): A Hugging Face dataset object that streams data from the specified remote location.
     """
     # These imports are local due to the optionality of `mosaicml-streaming`.
-    from streaming import StreamingDataset
-    from datasets import Features, Value, Dataset
     from functools import partial
+
+    from datasets import Features, Value
+    from streaming import StreamingDataset
 
     # Initialize the `StreamingDataset` with configurations.
     streaming_dataset = StreamingDataset(
@@ -86,7 +87,7 @@ def load_streaming_dataset(config_dataset):
     #
     # This is necessary because downstream functions use a different interface
     # than `StreamingDataset` (e.g. the `features` attribute).
-    ds = Dataset.from_generator(
+    ds = Dataset.from_generator(  # pylint: disable=invalid-name
         generator=partial(generator_from_streaming_dataset, streaming_dataset),
         features=features,
     )
--- a/src/axolotl/utils/data/sft.py
+++ b/src/axolotl/utils/data/sft.py
@@ -51,6 +51,49 @@ from axolotl.utils.trainer import (
 LOG = logging.getLogger("axolotl")
 
 
+def load_streaming_dataset(config_dataset):
+    """
+    Load a streaming dataset from a remote storage.
+
+    This function initializes a streaming dataset from a remote S3 bucket,
+    wraps the data into a generator, and then converts it into a Hugging Face
+    dataset for compatibility purposes.
+
+    Parameters:
+    - config_dataset (dict): Configuration dictionary that may contain settings necessary for initializing the dataset.
+
+    Returns:
+    - ds (datasets.Dataset): A Hugging Face dataset object that streams data from the specified remote location.
+    """
+    # These imports are local due to the optionality of `mosaicml-streaming`.
+    from streaming import StreamingDataset
+    from datasets import Features, Value, Dataset
+    from functools import partial
+
+    # Initialize the `StreamingDataset` with configurations.
+    streaming_dataset = StreamingDataset(
+        local=None, remote=config_dataset.path, shuffle=True, batch_size=4
+    )
+
+    # Define dataset features according to the axolotl structure.
+    features = Features({"text": Value("string")})
+
+    # Shim between `StreamingDataset` and `Dataset`.
+    def generator_from_streaming_dataset(streaming_dataset):
+        yield from streaming_dataset
+
+    # Create a Hugging Face dataset from the generator.
+    #
+    # This is necessary because downstream functions use a different interface
+    # than `StreamingDataset` (e.g. the `features` attribute).
+    ds = Dataset.from_generator(
+        generator=partial(generator_from_streaming_dataset, streaming_dataset),
+        features=features,
+    )
+
+    return ds
+
+
 def prepare_dataset(cfg, tokenizer):
     prompters = []
     if not cfg.pretraining_dataset:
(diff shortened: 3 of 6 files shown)


requirements.txt
```diff
@@ -31,6 +31,7 @@ art
 fschat==0.2.36
 gradio==3.50.2
 tensorboard
+mosaicml-streaming
```
fmv1992: [`question`](https://conventionalcomments.org/): `mosaicml-streaming` should be an optional dependency. Is this the right way of adding it in this capacity?

setup.py
```diff
@@ -92,5 +92,8 @@ def parse_requirements():
         "galore": [
             "galore_torch",
         ],
+        "mosaicml-streaming": [
```
fmv1992: [`question`](https://conventionalcomments.org/): `mosaicml-streaming` should be an optional dependency. Is this the right way of adding it in this capacity?

src/axolotl/utils/data/sft.py
```diff
@@ -51,6 +51,49 @@
...
+
+    # Shim between `StreamingDataset` and `Dataset`.
+    def generator_from_streaming_dataset(streaming_dataset):
+        yield from streaming_dataset
+
+    # Create a Hugging Face dataset from the generator.
+    #
+    # This is necessary because downstream functions use a different interface
+    # than `StreamingDataset` (e.g. the `features` attribute).
+    ds = Dataset.from_generator(
```
winglian: This becomes an `IterableDataset`, right?

requirements.txt
winglian: I'm okay with it being a required dependency. If it causes issues down the line we can make it optional then. Hoping to keep things simpler.

setup.py
```diff
@@ -92,5 +92,8 @@ def parse_requirements():
...
+            "mosaicml-streaming",
+        ],
```
winglian:
```suggestion
```
we can remove this since we're okay with this being a required dependency

winglian (approved):
good to go. thank you @fmv1992 !

fmv1992:
Thanks, much appreciated; I'm just checking a few more things before merging.

The experience of contributing to this repo has been very positive.

src/axolotl/utils/data/sft.py
fmv1992: @winglian  , 

Sorry for the delay here.

No, that was something that I wanted to verify but it looks like it goes to [def process](https://github.com/fmv1992/axolotl/blob/main/src/axolotl/datasets.py#L47) and everything is evaluated eagerly.

I started a draft like:

```
    def process(self, dataset):
        map_kwargs = {}
        if self.prompt_tokenizer.supports_batched:
            map_kwargs["batched"] = True
            map_kwargs["batch_size"] = 100
        map_kwargs["desc"] = "Tokenizing Prompts"

        if isinstance(dataset, IterableDataset):
            dataset_wrapper = dataset.map(
                self.prompt_tokenizer.tokenize_prompt,
                remove_columns=features,
                keep_in_memory=self.keep_in_memory,
                **map_kwargs,
            )
        else:
            num_proc = min(
                64, self.process_count if self.process_count else os.cpu_count()
            )

            return dataset.map(
                self.prompt_tokenizer.tokenize_prompt,
                num_proc=num_proc,
                remove_columns=features,
                keep_in_memory=self.keep_in_memory,
                desc="Tokenizing Prompts",
                **map_kwargs,
            )
```

But I don't know whether that's a good idea. The `.map` API is different between `Dataset` ([here](https://github.com/huggingface/datasets/blob/a3bc89d8bfd47c2a175c3ce16d92b7307cdeafd6/src/datasets/arrow_dataset.py#L2918)) and `IterableDataset` ([here](https://github.com/huggingface/datasets/blob/a3bc89d8bfd47c2a175c3ce16d92b7307cdeafd6/src/datasets/iterable_dataset.py#L1581)).

Feel free to remove the "ready to merge" tag from this.

requirements.txt
NanoCode012: Could this version be locked?

NanoCode012:
Hey, thanks for the PR. I just wanted to clarify something I asked previously. This would require user's to preprocess their dataset to Mosaic's format first right? If so, I would prefer this to be documented somewhere near the cloud loading section. For ex, add `stream: true` to load a Mosaic streaming dataset.

You should also add this parameter to this https://github.com/OpenAccess-AI-Collective/axolotl/blob/main/docs/config.qmd

https://github.com/mosaicml/streaming?tab=readme-ov-file#quick-start

Kesta-bos:
I think it need additional 'StreamingDataset' support for pretraining dataset (completion) in addition to Finetuning dataset.

ehartford:
We can pretrain with Axolotl streaming a data mix from s3?

winglian:
> Hey, thanks for the PR. I just wanted to clarify something I asked previously. This would require user's to preprocess their dataset to Mosaic's format first right? If so, I would prefer this to be documented somewhere near the cloud loading section. For ex, add `stream: true` to load a Mosaic streaming dataset.
> 
> You should also add this parameter to this https://github.com/OpenAccess-AI-Collective/axolotl/blob/main/docs/config.qmd
> 
> https://github.com/mosaicml/streaming?tab=readme-ov-file#quick-start

JSONL should be fine for streaming. see https://github.com/mosaicml/streaming?tab=readme-ov-file#1-prepare-your-data

requirements.txt
fmv1992: Addressed by `976bc13`.

fmv1992:
> We can pretrain with Axolotl streaming a data mix from s3?

We can, but I prefer if we include this in a second PR. Right now I would rather see this smaller change working and merged. Expanding on it should be easier later.

fmv1992:
> We can pretrain with Axolotl streaming a data mix from s3?

We can, but I prefer if we include this in a second PR. Right now I would rather see this smaller change working and merged. Expanding on it should be easier later.

> Hey, thanks for the PR. I just wanted to clarify something I asked previously. This would require user's to preprocess their dataset to Mosaic's format first right? If so, I would prefer this to be documented somewhere near the cloud loading section. For ex, add `stream: true` to load a Mosaic streaming dataset.
> 
> You should also add this parameter to this https://github.com/OpenAccess-AI-Collective/axolotl/blob/main/docs/config.qmd
> 
> https://github.com/mosaicml/streaming?tab=readme-ov-file#quick-start

Addressed by ba86339 . Let me know if that addresses all your points.

fmv1992:
As per [this comment](https://github.com/OpenAccess-AI-Collective/axolotl/pull/1525#discussion_r1570661103) this is not ready for merging. Maybe we want to remove that tag.

I posted a draft of the changes there, but the issue is that the tokenization should happen as we download the data, and right now I'm almost certain it does everything in a batch: it downloads everything, then tokenizes everything, then proceeds to do the fine tuning.

NanoCode012:
> but the issue is that the tokenization should happen as we download the data, and right now I'm almost certain it does everything in a batch: it downloads everything, then tokenizes everything, then proceeds to do the fine tuning.

@fmv1992 , this is correct. I only got to review your code in detail earlier. The section I provided you was incorrect. https://github.com/OpenAccess-AI-Collective/axolotl/blob/60f5ce0569b7f1d522ef81ea986ebfdc98780e6a/src/axolotl/utils/data/sft.py#L121

This function runs the whole dataset, merges it, and perform tokenization at this point here.

https://github.com/OpenAccess-AI-Collective/axolotl/blob/60f5ce0569b7f1d522ef81ea986ebfdc98780e6a/src/axolotl/utils/data/sft.py#L410-L411 

The only part that "skips" tokenization before finetuning is the `pretaining` section that you attempted to modify before.

https://github.com/OpenAccess-AI-Collective/axolotl/blob/60f5ce0569b7f1d522ef81ea986ebfdc98780e6a/src/axolotl/utils/data/sft.py#L70-L102

I have two ideas as of now:
1. discuss a better way to handle data preprocessing between the current `pretraining_dataset` and `dataset` format as the code is currently messy before continuing further.
2. Hack around and support streaming for pretraining datasets first and figure sft later. This is also because, your code expects the data in `completion` aka (`{ "text": ..." }`) format. This is not the case for SFT datasets. https://github.com/OpenAccess-AI-Collective/axolotl/blob/ba863392250539eaa1347217672f0e92881583e1/src/axolotl/utils/data/sft.py#L79-L80

I would also appreciate @winglian 's comments on this.

---

Side note: what should this `batch_size` be set to? Is it hardcoded to 4 on purpose?

https://github.com/OpenAccess-AI-Collective/axolotl/blob/ba863392250539eaa1347217672f0e92881583e1/src/axolotl/utils/data/sft.py#L76

fmv1992:
I'm closing this due to inactivity.

tmm1:
I'm interested in this

winglian:
@NanoCode012 @djsaunde @SalmanMohammadi @mhenrichsen Let's think about how we can resurrect this feature.

PR #1530, opened 2024-04-17
Add support for Gemma chat template

Haoxiang-Wang:
# Description

Supports Gemma chat template for SFT. Currently, when users specify `type=gemma` in their YAML file. e.g.,

```
base_model: google/gemma-2b-it
model_type: AutoModelForCausalLM
tokenizer_type: AutoTokenizer

datasets:
  - path: HuggingFaceH4/ultrachat_200k
    conversation: gemma
    type: sharegpt.load_ultrachat
    split: "train_sft"
    train_on_split: "train_sft"
```

errors will occur, as the current version of axolotl doesn't support Gemma chat template as there is no `if self.sep_style == SeparatorStyle.GEMMA` in [fastchat_conversation_turns.py](https://github.com/OpenAccess-AI-Collective/axolotl/blob/main/src/axolotl/monkeypatch/fastchat_conversation_turns.py)

I added the following lines to fastchat_conversation_turns.py
```
    if self.sep_style == SeparatorStyle.GEMMA:
        if self.system_message:
            raise ValueError("Gemma chat template does not support system messages")
        for i, (role, message) in enumerate(self.messages):
            prefix = "<bos>" if i == 0 else ""
            message_str = message if message else ""
            yield prefix + "<start_of_turn>" + role + "\n", message_str + "<end_of_turn>\n"
        return
```
which will generate texts matching [Gemma-instruct's chat template](https://huggingface.co/google/gemma-1.1-7b-it#chat-template). 

Besides, the FastChat latest version 0.2.36 doesn't support Gemma chat template, so I updated the `requirements.txt` to install the up-to-date FastChat from GitHub. 

## How has this been tested?

I tested it with 

```
base_model: google/gemma-2b-it
model_type: AutoModelForCausalLM
tokenizer_type: AutoTokenizer

datasets:
  - path: HuggingFaceH4/ultrachat_200k
    conversation: gemma
    type: sharegpt.load_ultrachat
    split: "train_sft"
    train_on_split: "train_sft"
```

and it worked well.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -28,7 +28,7 @@ scipy
 scikit-learn==1.2.2
 pynvml
 art
-fschat @ git+https://github.com/lm-sys/FastChat.git@main
+fschat @ git+https://github.com/lm-sys/FastChat.git@5095615810cf613dba7f27dd155f571fcff976d8
 gradio==3.50.2
 tensorboard
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -28,7 +28,7 @@ scipy
 scikit-learn==1.2.2
 pynvml
 art
-fschat==0.2.36
+fschat @ git+https://github.com/lm-sys/FastChat.git@main
 gradio==3.50.2
 tensorboard
 
--- a/src/axolotl/monkeypatch/fastchat_conversation_turns.py
+++ b/src/axolotl/monkeypatch/fastchat_conversation_turns.py
@@ -123,6 +123,14 @@ def get_turns(  # pylint: disable=too-many-return-statements
             else:
                 yield role, ""
         return
+    if self.sep_style == SeparatorStyle.GEMMA:
+        if self.system_message:
+            raise ValueError("Gemma chat template does not support system messages")
+        for i, (role, message) in enumerate(self.messages):
+            prefix = "<bos>" if i == 0 else ""
+            message_str = message if message else ""
+            yield prefix + "<start_of_turn>" + role + "\n", message_str + "<end_of_turn>\n"
+        return
     if self.sep_style == SeparatorStyle.CHATGLM:
         # source: https://huggingface.co/THUDM/chatglm-6b/blob/1d240ba371910e9282298d4592532d7f0f3e9f3e/modeling_chatglm.py#L1302-L1308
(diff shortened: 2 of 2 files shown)


requirements.txt
```diff
@@ -28,7 +28,7 @@ scipy
 scikit-learn==1.2.2
 pynvml
 art
-fschat==0.2.36
+fschat @ git+https://github.com/lm-sys/FastChat.git@main
```
winglian:
```suggestion
fschat @ git+https://github.com/lm-sys/FastChat.git@5095615810cf613dba7f27dd155f571fcff976d8
```

let's pin this to the current HEAD so we don't pick up unexpected breaking changes

winglian (approved):

requirements.txt
```diff
@@ -28,7 +28,7 @@ scipy
 scikit-learn==1.2.2
 pynvml
 art
-fschat==0.2.36
+fschat @ git+https://github.com/lm-sys/FastChat.git@5095615810cf613dba7f27dd155f571fcff976d8
```
NanoCode012, line 31: Would it be possible to use a version instead of commit?

src/axolotl/monkeypatch/fastchat_conversation_turns.py
```diff
@@ -123,6 +123,14 @@ def get_turns(  # pylint: disable=too-many-return-statements
             else:
                 yield role, ""
         return
+    if self.sep_style == SeparatorStyle.GEMMA:
+        if self.system_message:
+            raise ValueError("Gemma chat template does not support system messages")
+        for i, (role, message) in enumerate(self.messages):
+            prefix = "<bos>" if i == 0 else ""
```
NanoCode012, line 130: I am a bit confused about this. I don't think you need to add `bos` manually like this. Axolotl should prepend it.

NanoCode012 (commented):
Could you give me an example output from running `cli.preprocess` with `--debug`? Does the bos repeat?

src/axolotl/monkeypatch/fastchat_conversation_turns.py
winglian: since this is Gemma specific, and this function doesn't have access to the tokenizer, I think this is fine for now. in other formats in this function, we have literal "<s>" strings.

requirements.txt
winglian: Screenshot 2024-04-21 at 4 55 02 PM
unfortunately, fastchat hasn't cut a release in a while too.

PR #1541, opened 2024-04-19
fix broken linting

winglian:
something snuck through that broke the linting

### DIFF
--- a/tests/prompt_strategies/test_sharegpt.py
+++ b/tests/prompt_strategies/test_sharegpt.py
@@ -96,7 +96,9 @@ def fixture_multi_role_dataset():
 
 @pytest.fixture(name="tokenizer")
 def fixture_tokenizer():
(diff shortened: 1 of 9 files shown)


winglian:
updated the workflow regexes to lint when example yaml are updated too

tests/prompt_strategies/test_sharegpt.py
```diff
@@ -96,7 +96,9 @@ def fixture_multi_role_dataset():
 
 @pytest.fixture(name="tokenizer")
 def fixture_tokenizer():
-    tokenizer = AutoTokenizer.from_pretrained("mistralai/Mistral-7B-v0.1")
+    tokenizer = AutoTokenizer.from_pretrained(
+        "casperhansen/mistral-7b-instruct-v0.1-awq"
```
NanoCode012, line 100: Should we use a consistent repo (aka `LoneStriker/Mixtral-8x7B-v0.1-HF` above?

Issue #1544, opened 2024-04-19
Preprocess with debug flag fails.


amitagh:
### Please check that this issue hasn't been reported before.

- [X] I searched previous [Bug Reports](https://github.com/OpenAccess-AI-Collective/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

Preprocess with debug flag should work.
python -m axolotl.cli.preprocess  /content/test_axolotl.yaml --debug

### Current behaviour

Gives error. Have json file with each example in the json file is with {"text": <text_str>}.
I am doing Pretraining with Lora for Non-Eng lang.

[2024-04-19 09:05:02,918] [DEBUG] [axolotl.log:61] [PID:2346] [RANK:0] max_input_len: 600
Dropping Long Sequences (num_proc=2): 100% 17/17 [00:00<00:00, 99.19 examples/s]
Add position_id column (Sample Packing) (num_proc=2): 100% 17/17 [00:00<00:00, 70.88 examples/s]
[2024-04-19 09:05:03,502] [INFO] [axolotl.load_tokenized_prepared_datasets:423] [PID:2346] [RANK:0] Saving merged prepared dataset to disk... /content/d538aae6e42c7df428d20d3ff2685ad0
Traceback (most recent call last):
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/content/src/axolotl/src/axolotl/cli/preprocess.py", line 70, in <module>
    fire.Fire(do_cli)
  File "/usr/local/lib/python3.10/dist-packages/fire/core.py", line 143, in Fire
    component_trace = _Fire(component, args, parsed_flag_args, context, name)
  File "/usr/local/lib/python3.10/dist-packages/fire/core.py", line 477, in _Fire
    component, remaining_args = _CallAndUpdateTrace(
  File "/usr/local/lib/python3.10/dist-packages/fire/core.py", line 693, in _CallAndUpdateTrace
    component = fn(*varargs, **kwargs)
  File "/content/src/axolotl/src/axolotl/cli/preprocess.py", line 60, in do_cli
    load_datasets(cfg=parsed_cfg, cli_args=parsed_cli_args)
  File "/content/src/axolotl/src/axolotl/cli/__init__.py", line 397, in load_datasets
[...]

PR #1545, opened 2024-04-19
fix(packages): lock datasets version

NanoCode012:
# Description

There seems to be a bug with newer datasets version. Ref: https://discord.com/channels/1104757954588196865/1111279858136383509/1230806736127004674

It seems that downgrading to the current locked version fixes it for them.

```bash
File "/usr/local/lib/python3.10/dist-packages/fsspec/core.py", line 316, in _un_chain
    if "::" in path
TypeError: argument of type 'PosixPath' is not iterable
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,7 +11,7 @@ addict
 fire
 PyYAML>=6.0
 requests
-datasets>=2.15.0
+datasets==2.15.0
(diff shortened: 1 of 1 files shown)


Napuh:
Also solves #1544

jorge-tromero:
I need this, to fix #1544

PR #1549, opened 2024-04-19
Update Tokenizer Overrides Handling in models.py

mhenrichsen:
Example:
```
tokenizer_overrides:
  - 28006: <|im_start|>
  - 28007: <|im_end|>
```

Description:

This PR introduces an enhancement to the way we handle tokenizer overrides in our models.py file.

Previously, the code did not account for the possibility of a token override not being found in the tokenizer's special tokens. This could potentially lead to silent failures where an override is skipped without any indication to the user.

The updated code now includes a check for each key in the cfg.tokenizer_overrides dictionary within the tokenizer.all_special_tokens list. If a match is found, the corresponding token's content is updated with the override value.

This change ensures that our tokenizer correctly applies all specified overrides, improving the robustness and reliability of our tokenization process.

### DIFF
--- a/docs/config.qmd
+++ b/docs/config.qmd
@@ -154,8 +154,6 @@ datasets:
       content: value
       # ...
 
-    message_property_mappings:
-
     # Optional[Dict[str, List]]. Roles mapping in the messages. The default is:
     roles:
       user: ["human", "user"]
--- a/docs/config.qmd
+++ b/docs/config.qmd
@@ -546,6 +546,13 @@ special_tokens:
 # Add extra tokens.
 tokens:
 
+# Mapping token_id to new_token_string to override reserved added_tokens in the tokenizer.
+# Only works for tokens that are not part of the base vocab (aka are added_tokens).
+# Can be checked if they exist in tokenizer.json added_tokens.
+tokenizer_overrides:  # Dict[int, str]
+#  128041: "<|im_start|>"
+#  128042: "<|im_end|>"
+
 # FSDP
 fsdp:
 fsdp_config:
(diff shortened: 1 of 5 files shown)


winglian (approved):
thanks!! this was on my todo list too.

NanoCode012:
Could a unit test be added for this?

winglian:
@mhenrichsen played around with this today. The tokenizer seems to be frozen, especially for llama 3 where this would be more useful.

winglian:
here's a script I have on hand that will download and replace tokens in the tokenizer that we could integrate. The way I think it could work would be in `load_tokenizer` it detects if the yaml is requesting token replacements, and build a has based on the tokenizer name and replacements (similar to our dataset caching) and define a path to look for the updated tokenizer. If it isn't found, it can execute the function that replaces the tokens and save it in the cache, and then loads it and returns it to load_tokenizer.

EDIT: script here https://gist.github.com/winglian/b3ee34825b8e5bbfe2b933b4ba37df51

NanoCode012:
Edit: rename the prior config name to the new name `added_tokens_overrides` discussed below.

---

Fixed this PR to allow overriding `added_tokens` within a tokenizer. This does not allow changing base vocab tokens.

I tried the linked gist but couldn't accurately change the strings (either due to unicode or it changes multiple tokens unintentionally due to string matching)

---

The current method works via:
```yaml
added_tokens_overrides:
  28006: <|im_start|>
  28007: <|im_end|>
```

Note: Dict[int, str] instead of List.

The naming of the config is still upto discussion as it only overrides `added_tokens` which should cover most reserved tokens in current tokenizers.

winglian:
@NanoCode012 Do you want to name it something else? `added_tokens_overrides` ?

NanoCode012:
> @NanoCode012 Do you want to name it something else? `added_tokens_overrides` ?

Yeah, that seems suitable. I'll make that change.

NanoCode012:
Updated to use `added_tokens_overrides` instead!

minpeter:
I love this feature, it's exactly what I was trying to implement someday, thank you

PR #1559, opened 2024-04-22
make sure everything stays in the same dtype when using dpo + FSDP

winglian:
trl's DPO Trainer attempts to call prepare_for_kbit_training on the peft model, but this should be avoided when using FSDP, otherwise you end up with the error:
```
ValueError: Must flatten tensors with uniform dtype but got torch.bfloat16 and torch.float32
```

we go behind it after instantiation and fix the dtypes.


maziyarpanahi:
can confirm, this PR fixes the issue for me. (QLoRA+FSDP Llama-3-70B)

PR #1567, opened 2024-04-25
Pose context length ext

winglian:
PoSE paper: https://huggingface.co/papers/2309.10400
Model: https://huggingface.co/winglian/Llama-3-8b-64k-PoSE
YAML: https://huggingface.co/winglian/Llama-3-8b-64k-PoSE/blob/main/axolotl/pose.yaml

Add the PoSE technique for extending context length without needing long context data.

### DIFF
--- a/src/axolotl/utils/trainer.py
+++ b/src/axolotl/utils/trainer.py
@@ -100,7 +100,10 @@ def add_position_ids(sample):
 
 
 def add_pose_position_ids(
-    sample, max_context_len=32768, split_on_token_ids: Optional[List[int]] = None
+    sample,
+    max_context_len=32768,
+    split_on_token_ids: Optional[List[int]] = None,
+    chunks: int = 2,
 ):
     """
     use the PoSE technique to extend the context length by randomly skipping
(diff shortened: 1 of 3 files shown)


src/axolotl/utils/trainer.py
```diff
@@ -98,17 +99,87 @@ def add_position_ids(sample):
...
+
+    if split_on_token_ids is None:
+        split_on_token_ids = []
+
+    if split_on_token_ids:
+        split_indices = [
+            i for i, token_id in enumerate(input_ids) if token_id in split_on_token_ids
+        ]
+    else:
+        split_indices = [sample_len // chunks]
```
arkapal3: I don't think this is going to work for any n_chunks > 2 right?

arkapal3 (commented):
I'm pretty sure your PR doesn't quite work for chunks > 2.

src/axolotl/utils/trainer.py
winglian: ah, you're right

PR #1580, opened 2024-04-30
Adding Phi-3 model

monk1337:
Adding Phi-3 model

`fsdp_transformer_layer_cls_to_wrap: Phi3DecoderLayer`

### DIFF
--- a/examples/phi/phi3-ft-fsdp.yml
+++ b/examples/phi/phi3-ft-fsdp.yml
@@ -1,0 +1,83 @@
+base_model: microsoft/Phi-3-mini-4k-instruct
+model_type: AutoModelForCausalLM
+tokenizer_type: AutoTokenizer
+
+load_in_8bit: false
+load_in_4bit: false
+strict: false
+
+datasets:
+  - path: mhenrichsen/alpaca_2k_test
+    type: alpaca
+
+dataset_prepared_path:
+val_set_size: 0
+output_dir: ./phi-sft-out
+
+sequence_len: 4096
+sample_packing: true
+pad_to_sequence_len: true
+trust_remote_code: true
+
+adapter:
+lora_model_dir:
+lora_r:
+lora_alpha:
+lora_dropout:
+lora_target_linear:
+lora_fan_in_fan_out:
+
+wandb_project: phi3
+wandb_entity:
+wandb_watch:
+wandb_name:
+wandb_log_model:
+
+gradient_accumulation_steps: 2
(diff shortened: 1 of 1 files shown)


examples/phi/phi3-ft.yml
```diff
@@ -0,0 +1,83 @@
...
+  fsdp_limit_all_gathers: true
+  fsdp_sync_module_states: true
+  fsdp_offload_params: true
+  fsdp_use_orig_params: false
+  fsdp_cpu_ram_efficient_loading: true
+  fsdp_auto_wrap_policy: TRANSFORMER_BASED_WRAP
+  fsdp_transformer_layer_cls_to_wrap: Phi3DecoderLayer
+  fsdp_state_dict_type: FULL_STATE_DICT
+  fsdp_sharding_strategy: FULL_SHARD
+resize_token_embeddings_to_32x: true
```
winglian: generally we prefer not to have FSDP or deepspeed specific configs (which require multi-gpu) for smaller models that should fit on single GPU systems.

maziyarpanahi:
Will you also add support for `Phi-3` chat template for fine-tuning? As a reference: https://github.com/unslothai/unsloth/blob/4211cc01409e3ced4f7abebaf68e244193b46e2c/unsloth/chat_templates.py#L269C3-L269C8

winglian:
will merge this after #1582

vinamrabenara:
Was wondering when would this be merged?

monk1337:
> Was wondering when would this be merged?

Too busy to work on this :P

maziyarpanahi:
@winglian I am having some issues with phi-3-small and phi-3-medium models. 

- phi-3-medium with 4k instruct, it constantly fails with `RuntimeError: Cannot re-initialize CUDA in forked subprocess. To use CUDA with multiprocessing, you must use the 'spawn' start method`
- phi-3-small with 8k instruct, fails with:
```
ValueError: For now, we do not support unknown special tokens
In the future, if there is a need for this, we can add special tokens to the tokenizer
starting from rank 100261 - 100263 and then 100266 - 100275.
And finally, we can re-construct the enc object back
```

Is it correct to assume this PR won't solve these, right?

winglian:
@maziyarpanahi yeah, I don't think this PR will resolve that issue.

winglian:
thanks @monk1337 !

maziyarpanahi:
Thank you @winglian and @monk1337

PR #1581, opened 2024-04-30
Pass weakref to model in the SIGINT handler to free up model post train function

chiragjn:
Basically passing model to sigint handler via a normal reference captures it in the handler closure and prevents it from being garbage collected even when train() function has completely finished.

To avoid leaking memory like this, this change passes the model to sigint handler using a weak reference

---

I discovered this because post train() function I am calling gc.collect() and torch.cuda.empty_cache() and noticed the model still sticks around on the gpu.

### DIFF
--- a/src/axolotl/train.py
+++ b/src/axolotl/train.py
@@ -3,6 +3,7 @@
 import os
 import signal
 import sys
+import weakref
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Optional, Tuple, Union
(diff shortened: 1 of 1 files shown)


winglian (approved):
great catch. thanks!

chiragjn:
On a side note, although this sigint handler is nice to have, I think this is not fully accurate
- This does not save the trainer state, so can't resume
- If I am not wrong when a model is using deepspeed / fsdp where weights are being sharded, save should be called from all ranks

winglian:
> On a side note, although this sigint handler is nice to have, I think this is not fully accurate
> 
> * This does not save the trainer state, so can't resume
> * If I am not wrong when a model is using deepspeed / fsdp where weights are being sharded, save should be called from all ranks

Yeah, definitely not perfect atm. We can add this as a feature enhancement to make it work with fsdp/deepspeed and also capturing the trainer states for resuming.

PR #1583, opened 2024-05-02
FIX: TRL trainer preprocessing step was running in one process

ali-mosavian:
# Description

We weren't passing dataset_num_proc to TRL training config, thus the initial data preprocessing steps in the TRL trainer was running in one process only.
 
## Motivation and Context

Speeds up training start time by a lot depending on the number logical cores you have.

## How has this been tested?

Tested it with the ORPO trainer in axolotl.

## Social Handles (Optional)

https://www.linkedin.com/in/ali-mosavian-7a27457/
https://github.com/ali-mosavian


winglian (approved):
thanks!

ali-mosavian:
I made a misstake, i sent it as trainer args in all cases, but it only applies CPO, KTO and ORPO trainer args. For the DPO it needs to be sent directly to the trainer. My latest commit in the PR fixes that.

PR #1584, opened 2024-05-02
FIX: max_length and max_prompt_length was not being sent to ORPOTrainer

ali-mosavian:
# Description

### DIFF
--- a/src/axolotl/core/trainer_builder.py
+++ b/src/axolotl/core/trainer_builder.py
@@ -1522,6 +1522,8 @@ class HFRLTrainerBuilder(TrainerBuilderBase):
         if self.cfg.rl == "orpo":
             training_args_cls = ORPOConfig
             training_args_kwargs["dataset_num_proc"] = self.cfg.dataset_processes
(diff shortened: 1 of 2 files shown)


winglian (approved):

src/axolotl/core/trainer_builder.py
```diff
@@ -1522,6 +1522,8 @@ def build_training_arguments(self, total_num_steps):
         if self.cfg.rl == "orpo":
             training_args_cls = ORPOConfig
             training_args_kwargs["dataset_num_proc"] = self.cfg.dataset_processes
+            training_args_kwargs["max_length"] = self.cfg.sequence_len
+            training_args_kwargs["max_prompt_length"] = self.cfg.sequence_len // 4
```
winglian: I don't know that we want to artificially limit this to 25% of the sequence len. Either it should be equal to max_length or configurable.

ali-mosavian: The reason behind it was that the ORPO trainer would hang in some of my tests. But having it configurable is fine.

PR #1585, opened 2024-05-02
add support for SPPO

winglian:
arxiv: https://arxiv.org/abs/2405.00675

temporarily overriding the `dpo_loss` method for testing until this is merged upstream: https://github.com/huggingface/trl/pull/1612


kallewoof:
Nice. Does this work with the llama-3 8b models as well or only mistral?

PR #1587, opened 2024-05-03
improve tool handling roles

winglian:
fixes an issue with allowing tools with other sharegpt formatted variants

### DIFF
--- a/src/axolotl/prompt_strategies/sharegpt.py
+++ b/src/axolotl/prompt_strategies/sharegpt.py
@@ -79,13 +79,28 @@ def load_ultrachat(tokenizer, cfg, ds_cfg: Optional[Dict[str, Any]] = None):
     return strategy
 
 
-def load_role(tokenizer, cfg):
-    return SimpleRoleShareGPTPromptTokenizingStrategy(
-        ShareGPTPrompterV2(),
+def load_role(tokenizer, cfg, ds_cfg: Optional[Dict[str, Any]] = None):
+    """use when your dataset has a role field instead of a from field in the conversation"""
+    conversation = (
+        ds_cfg["conversation"] if ds_cfg and "conversation" in ds_cfg else None
+    )
+    field_human = ds_cfg["field_human"] if ds_cfg and "field_human" in ds_cfg else None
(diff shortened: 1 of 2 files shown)


winglian:
reported as working via discord https://discord.com/channels/1104757954588196865/1111279858136383509/1235987415907172517

src/axolotl/prompt_strategies/sharegpt.py
```diff
@@ -79,13 +79,28 @@ def load_ultrachat(tokenizer, cfg, ds_cfg: Optional[Dict[str, Any]] = None):
...
+            role_key_model=field_model,
+            role_key_human=field_human,
+            roles=roles,
+        ),
         tokenizer,
         cfg.train_on_inputs,
         cfg.sequence_len,
     )
+    if ds_cfg and "strict" in ds_cfg:
+        strategy.strict = ds_cfg["strict"]
```
NanoCode012: I think this should be extracted to a utility function that accepts a class of type `SimpleShareGPTPromptTokenizingStrategy`. This can then be invoked for this `.load_role` along with the base `.load` and other load examples below.

winglian: done!

NanoCode012 (approved):

hasan9090:
Hi, I am trying to use another variant of the glaive dataset from https://huggingface.co/datasets/lilacai/glaive-function-calling-v2-sharegpt/viewer/default/train?q=functioncall&row=18841. I know I could just use the glaiveai/glaive-function-calling-v2 one and follow this config: https://github.com/axolotl-ai-cloud/axolotl/pull/1365 . 
But I need a way to also load a custom dataset which is exactly formatted as the lilicai glaive one, already in sharegpt format. My problem is I could not get it tokenized correctly with a "tool" role. How do I need to set the config to use the conversation template "chatml_glaive" which seems to add the "tool" role but with the dataset type "sharegpt" instead of "sharegpt.load_glaive" ?

The following config does not work:

datasets:
   -path: lilacai/glaive-function-calling-v2-sharegpt
    type: sharegpt
    conversation: chatml_glaive

PR #1591, opened 2024-05-04
Gradio configuration parameters

marijnfs:
Various parameters of Gradio were hardcoded (e.g. share=True, ip address, port, number of tokens, temperature)
I made them configurable here.

Additionally the default tokens were overwritten into the tokenizer, which seems odd behaviour and broke llama3 for me, so I removed that.

# Description

Added configuration parameters to  ''src/axolotl/utils/config/models/input/v0_4_1/__init__.py', not sure if they should be added there?
The rest is in src/axolotl/cli/__init__.py, leaving the default behaviour from before

## Motivation and Context

Now aware of open issues for this, but it helps using Gradio. Mostly, I didn't want the public link, and wanted to bind to 0.0.0.0 instead of 127.0.0.1 so you can access gradio from your local network.

## How has this been tested?

I tested it locally, ideally someone else tests it as well

## Types of changes

Simple additions to configuration parameters for gradio

### DIFF
--- a/src/axolotl/cli/__init__.py
+++ b/src/axolotl/cli/__init__.py
@@ -233,6 +233,12 @@ def do_inference_gradio(
 
     model, tokenizer = load_model_and_tokenizer(cfg=cfg, cli_args=cli_args)
     prompter = cli_args.prompter
+    default_tokens = {"unk_token": "<unk>", "bos_token": "<s>", "eos_token": "</s>"}
+
+    for token, symbol in default_tokens.items():
+        # If the token isn't already specified in the config, add it
+        if not (cfg.special_tokens and token in cfg.special_tokens):
+            tokenizer.add_special_tokens({token: symbol})
 
     prompter_module = None
     if prompter:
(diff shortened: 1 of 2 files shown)


src/axolotl/cli/__init__.py
```diff
@@ -233,12 +233,6 @@ def do_inference_gradio(
 
     model, tokenizer = load_model_and_tokenizer(cfg=cfg, cli_args=cli_args)
     prompter = cli_args.prompter
-    default_tokens = {"unk_token": "<unk>", "bos_token": "<s>", "eos_token": "</s>"}
-
-    for token, symbol in default_tokens.items():
-        # If the token isn't already specified in the config, add it
-        if not (cfg.special_tokens and token in cfg.special_tokens):
-            tokenizer.add_special_tokens({token: symbol})
```
tmm1: Unrelated change?

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
```diff
@@ -408,7 +408,16 @@ def check_wandb_run(cls, data):
...
         return data
 
-
+class GradioConfig(BaseModel):
+    """Gradio configuration subset"""
+
+    gradio_title: Optional[str] = None
+    gradio_share: Optional[bool] = None
+    gradio_server_name: Optional[str] = None
+    gradio_port: Optional[int] = None
```
winglian: you used `gradio_server_port` above.

src/axolotl/cli/__init__.py
winglian: I went ahead and added that back, thanks

winglian (approved):

PR #1601, opened 2024-05-07
Fused Cross Entropy Loss

winglian:
https://github.com/pytorch/pytorch/issues/124480 served as the impetus for getting this integrated. 

Adapted the code from @mgmalek's https://github.com/mgmalek/efficient_cross_entropy/blob/main/modules.py so that it properly handles causal/next-token prediction.

### DIFF
--- a/.github/workflows/nightlies.yml
+++ b/.github/workflows/nightlies.yml
@@ -109,7 +109,7 @@ jobs:
         with:
           context: .
           build-args: |
-            BASE_TAG=${{ github.ref_name }}-py${{ matrix.python_version }}-cu${{ matrix.cuda }}-${{ matrix.pytorch }}${{ matrix.axolotl_extras != '' && '-' || '' }}${{ matrix.axolotl_extras }}
+            BASE_TAG=${{ github.ref_name }}-{{ date 'YYYYMMDD' }}-py${{ matrix.python_version }}-cu${{ matrix.cuda }}-${{ matrix.pytorch }}${{ matrix.axolotl_extras != '' && '-' || '' }}${{ matrix.axolotl_extras }}
             CUDA=${{ matrix.cuda }}
           file: ./docker/Dockerfile-cloud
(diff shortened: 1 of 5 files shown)


YouJiacheng:
mgmalek's implementation might have numerical issues when using half precision dtype.
The gradient w.r.t. the weight is accumulated through chunks with a half precision accumulator.
Some users have reported loss curve mismatch on Twitter.

winglian:
> mgmalek's implementation might have numerical issues when using half precision dtype.
> The gradient w.r.t. the weight is accumulated through chunks with a half precision accumulator.
> Some users have reported loss curve mismatch on Twitter.

haha, yeah, that was probably me reporting the loss curve differences out on twitter. Do you have any insights on the best way to fix the kernels? standard CEL uses bfloat16 as well iirc.

YouJiacheng:
The problem is the lm_head weight (proj_weight), not CEL. By default bf16 matmul use fp32 accumulator (along the GEMM K axis). Since matmul is performed tile by tile in M&N axis (MxK@KxN matmul), registers are used as accumulator, without the need of a fp32 global memory buffer.

However, here we perform chunking along the GEMM K axis (DxT@TxV matmul)…

A simple fix can be using fp32 grad_proj_weight, but that will incur a significant memory overhead.

Can you check the loss curve with fp32 grad_proj_weight?

winglian:
@YouJiacheng 
Screenshot 2024-05-10 at 11 24 28 PM

gave your recommendation a try, but changing grad_proj_weight to fp32 didn't seem to help much.

YouJiacheng:
Ah Oh, now we need more investigation…

YouJiacheng:
It,seems that the fused version has a different loss & grad_norm from the first step? That looks strange…

YouJiacheng:
It seems that the fused version has a different loss & grad_norm from the first step? That looks strange…

winglian:
closing now that Liger is available

PR #1609, opened 2024-05-11
Unsloth optims for Llama

winglian:
WIP to integrate Unsloth's optimizations into axolotl. 

The manual autograd for MLP, QKV, O only seems to help VRAM by 1% as opposed to the reported 8%. 
The Cross Entropy Loss does help significantly, but only reduced VRAM by 13% as opposed to the reported 17%.

edit: clarification, the cross entropy loss works for both full fine tunes and lora. The MLP, QKV, and O are only for 4-bit qlora with flash attention.


bratao:
The cross_entropy_loss optimization is applicable even in a full fine tune, right?

winglian:
> The cross_entropy_loss optimization is applicable even in a full fine tune, right?

Correct!

linux-leo:
Are these optimizations compatible with flash attention? (Complete Noob here)

bratao:
It is possible to alter to also patch Qwen? It was added to unsloth and all optimizations work for it:
https://github.com/unslothai/unsloth/pull/428

winglian:
> It is possible to alter to also patch Qwen? 

Let's tackle that in a follow up PR

Nero10578:
Does this only work for single GPU? Getting error of "Runtime Error: Model must be 2-D" when enabling unsloth cross entropy loss on 2x3090Ti finetuning Llama 3 8B LORA.

PR #1635, opened 2024-05-17
Fix llama3 chat_template (extra <|eot_id|> on last turn)

lhl:
Pretty straightforward fix, I noticed when printing out my chat on a fine-tune `chat_template: llama3` that the output differed than when using the original Llama 3 Instruct chat_template/tokenizer.

In the loop, an `<|eot_id|>` is already being added to every single line, so the last conditional that adds another `{{eos_token}}` leads to an extra `<|eot_id|>`.

This is a minor fix but I did notice one weird stoppage so better to get that fix in I think.

## How has this been tested?

I've manually tested to make sure the `apply_chat_template()` results look right with both w/ and w/o `add_generation_prompt`.

You can check this by comparing output vs the original chat template:
```
tokenizer = AutoTokenizer.from_pretrained("meta-llama/Meta-Llama-3-8B-Instruct")

messages = [
    {'role': 'system', 'content': 'This is the system prompt.'},
    {'role': 'user', 'content': 'This is the first user input.'},
    {'role': 'assistant', 'content': 'This is the first assistant response.'},
    {'role': 'user', 'content': 'This is the second user input.'},
]

print()
print('Chat Template:')
print(tokenizer.chat_template)
print()
print('---')
print()
print(tokenizer.apply_chat_template(messages, tokenize=False))
```


winglian (approved):
great catch. Thanks!

would be great to include a test in the `TestSharegptLlama3` class, but I can add that in another PR if you don't have the time.

lhl:
sure, lemme look at the test class now and can submit another PR

lhl:
Hmm, am I reading this right, looks like there's already a tokenization test and it should have been catching the repeat? https://github.com/OpenAccess-AI-Collective/axolotl/blob/main/tests/prompt_strategies/test_sharegpt.py#L150

eg, w/ the previous chat_template you'd get a 128009, 128009 (double eot) which should have caused the assert to fail?

PR #1640, opened 2024-05-18
Add KTO support

benredmond:
# Description

This PR:

- Adds support for KTO training using the `trl` `KTOTrainer` https://huggingface.co/docs/trl/main/en/kto_trainer
  - Done using `KTOTrainer` in `trainer_builder.py`
- Adds support for KTO datasets by introducing `KTODataset`

This KTO support differs from the existing KTO support in that it only requires a `label` feature on the data (see linked issue for more motivation)

KTO arxiv: https://arxiv.org/abs/2402.01306

## Motivation and Context

The motivation here is that some recent research has suggested that KTO outperforms other alignment models ([Insights into Alignment: Evaluating DPO and its Variants Across Multiple Tasks](https://arxiv.org/pdf/2404.14723)). 

Additionally, using KTO may or may not require prior SFT, compared to DPO which does require prior SFT.

This addresses issue https://github.com/OpenAccess-AI-Collective/axolotl/issues/1200

## How has this been tested?

TODO still need to test this out in practice, will try it soon

### DIFF
--- a/src/axolotl/core/trainer_builder.py
+++ b/src/axolotl/core/trainer_builder.py
@@ -1543,7 +1543,7 @@ class HFRLTrainerBuilder(TrainerBuilderBase):
         if self.cfg.rl == "kto":
             training_args_cls = KTOConfig
 
-            training_args_kwargs["beta"] = self.cfg.dpo_beta or 0.1
+            training_args_kwargs["beta"] = self.cfg.rl_beta or 0.1
             training_args_kwargs["desirable_weight"] = (
                 self.cfg.kto_desirable_weight or 1.0
             )
@@ -1591,7 +1591,7 @@ class HFRLTrainerBuilder(TrainerBuilderBase):
             ] = self.cfg.precompute_ref_log_probs
         if self.cfg.rl in ["dpo", "ipo", "kto_pair"]:
             trainer_cls = AxolotlDPOTrainer
-            dpo_trainer_kwargs["beta"] = self.cfg.dpo_beta or 0.1
+            dpo_trainer_kwargs["beta"] = self.cfg.rl_beta or 0.1
             trainer_cls_args = [self.model, self.model_ref]
 
             # these aren't used for the ORPO trainer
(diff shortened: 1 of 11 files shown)


src/axolotl/core/trainer_builder.py
```diff
@@ -1532,6 +1539,18 @@ def build_training_arguments(self, total_num_steps):
             if self.cfg.max_prompt_len:
                 training_args_kwargs["max_prompt_length"] = self.cfg.max_prompt_len
 
+        if self.cfg.rl == "kto":
+            training_args_cls = KTOConfig
+
+            training_args_kwargs["beta"] = self.cfg.dpo_beta or 0.1
```
benredmond: IMO `dpo_beta` is slightly confusing here (because this isn't dpo), but it pretty much serves the same purpose. should I introduce a `kto_beta` config value?

tests/e2e/test_dpo.py
```diff
@@ -205,3 +205,66 @@ def test_orpo_lora(self, temp_dir):
...
+                "lora_dropout": 0.1,
+                "lora_target_linear": True,
+                "special_tokens": {},
+                "rl": "kto",
+                "dpo_beta": 0.5,
+                "kto_desirable_weight": 1.0,
+                "kto_undesirable_weight": 1.0,
+                "remove_unused_columns": False,
+                "datasets": [
+                    # {
```
benredmond, line 230: I'm getting some strange schema validation errors when loading the `argilla/kto-mix-15k` dataset, don't think it's related to this change. I might poke into it a bit more but not really sure whats going on.

src/axolotl/core/trainer_builder.py
winglian: let's change it to `rl_beta`. We can add the dpo_beta to [DeprecatedParameters](https://github.com/OpenAccess-AI-Collective/axolotl/blob/main/src/axolotl/utils/config/models/input/v0_4_1/__init__.py#L21) and if it is set, we can map it to the new `rl_beta` similar to https://github.com/OpenAccess-AI-Collective/axolotl/blob/main/src/axolotl/utils/config/models/input/v0_4_1/__init__.py#L863-L868

winglian: I went ahead and changed this and updated the validation.

winglian (approved):
thanks @benredmond!!!

PR #1657, opened 2024-05-25
Fix setting correct repo id when pushing dataset to hub 

chrislee973:
# Description
In the dataset preprocessing step, pass the dataset hash as the [config_name ](https://huggingface.co/docs/datasets/v2.19.0/en/package_reference/main_classes#datasets.Dataset.push_to_hub.config_name)parameter in push_to_hub if `push_dataset_to_hub` is set. The `config_name `parameter is used for handling different versions/configurations of a dataset, which fits our case here.

## Motivation and Context

The `push_dataset_to_hub` config setting takes a value in the form `<user>/<dataset_name>`, but the previous implementation was pushing to hub with the id `f"{cfg.push_dataset_to_hub}/{ds_hash}"`, which was causing an error because `<user>/<dataset_name>/<ds_hash>` is an invalid hf repo id. 

## How has this been tested?

- With `push_dataset_to_hub` set, ran the dataset preprocessing step several times with different dataset settings each time. 
- Confirmed that this successfully pushed different variations of the dataset to hf hub (screenshot)
- On subequent dataset preprocessing runs with the same settings, confirmed that we successfully pulled down the variation of that dataset from hf hub.
image

## Types of changes
- [x] Dataset preprocessing step

### DIFF
--- a/src/axolotl/utils/data/sft.py
+++ b/src/axolotl/utils/data/sft.py
@@ -161,7 +161,8 @@ def load_tokenized_prepared_datasets(
     try:
         if cfg.push_dataset_to_hub:
             dataset = load_dataset(
-                f"{cfg.push_dataset_to_hub}/{ds_hash}",
+                cfg.push_dataset_to_hub,
+                ds_hash,
                 token=use_auth_token,
             )
             dataset = dataset[split]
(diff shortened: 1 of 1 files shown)


winglian:
Thanks @chrislee973 The original intention was to "cache" the pre-processed dataset to its own dataset, but this is much better. Were you able to verify that it loads the subdirectory only when there are 2 or more datasets uploaded and you remove the local dataset cache ?

winglian (approved):
@chrislee973 Thanks for the PR. Just tested. good to go!

PR #1659, opened 2024-05-26
Fix Lora config error for Llama3

oaishi:
# Description

I modified the sample yml file used to train Llama 8b with custom tokens in LORA.

## Motivation and Context

The current yml code throws an error: ValueError: Please set lora_modules_to_save to [`embed_tokens`, `lm_head`] when using an adapter and changing the special tokens.

## How has this been tested?
I successfully fine-tuned a model with the new changes without any error.


winglian (approved):
great catch. thanks!

PR #1619, opened 2024-05-15
Switch to parallel FFD bin packing algorithm.

winglian:
Add support for packing in a distributed context.
Add packing efficiency estimate back.

See #1516 by @dsesclei. Attempting to rebase the original PR onto the latest main wasn't terribly clean. I also reverted the change to the distributed code.

PR #1672, opened 2024-05-29
revert multipack batch sampler changes

winglian:
The loss isn't quite right w/ #1619, off by an order of magnitude.


winglian:
just to make it easier, the previous state was https://github.com/OpenAccess-AI-Collective/axolotl/blob/fe650dd3260d0c24d6c5366b90d3507773872b2c/src/axolotl/utils/samplers/multipack.py

PR #1673, opened 2024-05-29
use L4 vs A10G for modal cicd

winglian:


winglian:
scratch this, A10G performance is much better than L4

PR #1680, opened 2024-06-03
jagged lr restart scheudler

winglian:
generic LR restart scheduler for ReLoRA and other techniques.

## Summary by CodeRabbit

* **New Features**
  * Introduced a new learning rate scheduler option supporting "jagged restarts," enabling periodic resets with customizable warmup and annealing steps during training.
  * Added new training arguments to configure jagged restart frequency, warmup steps, and anneal steps.
  * Integrated jagged restart parameters into training configuration and trainer selection.
  * Updated ReLoRA support to use jagged restart settings for step control.

* **Improvements**
  * Enhanced scheduler configuration to support advanced learning rate scheduling patterns for improved training flexibility.
  * Refined validation logic to require jagged restart parameters when using ReLoRA.

* **Removals**
  * Removed the dedicated ReLoRATrainer class and related custom scheduler, unifying functionality under the jagged restart scheduler settings.
  * Simplified ReLoRA configuration by replacing multiple step-related fields with a single enable flag.


src/axolotl/core/training_args.py
```diff
@@ -76,6 +76,22 @@ class AxolotlTrainingMixins:
...
+        metadata={
+            "help": "how many warmup steps to take after reset for jagged restarts"
+        },
+    )
+    jagged_restarts_anneal_steps: Optional[int] = field(
+        default=None,
+        metadata={
+            "help": "how many anneal steps to take before reset for jagged restarts"
+        },
+    )
```
NanoCode012: Could we add to `config.qmd`?

src/axolotl/core/trainers/base.py
```diff
@@ -143,7 +144,21 @@ def create_scheduler(
...
+                    anneal_steps = (
+                        self.args.jagged_restarts_anneal_steps or 1
+                    )
+                    self.lr_scheduler = JaggedLRRestartScheduler(  # pylint: disable=attribute-defined-outside-init
+                        optimizer,
+                        self.lr_scheduler,
+                        self.args.jagged_restart_steps,
+                        warmup_steps,
+                        anneal_steps,
+                    )
```
NanoCode012: - Do we need to get the return from `super().create_scheduler`?

- Would it make more sense to move this section one tab left as it could apply for other lr schedulers too?

src/axolotl/core/training_args.py
winglian: shouldn't be necessary now with auto-doc

src/axolotl/core/builders/causal.py
```diff
@@ -271,20 +268,34 @@ def build(self, total_num_steps):
...
+                self.cfg.jagged_restart_steps
+            )
+            if self.cfg.jagged_restart_warmup_steps:
+                training_arguments_kwargs["jagged_restart_warmup_steps"] = (
+                    self.cfg.jagged_restart_warmup_steps
+                )
+            if self.cfg.jagged_restart_anneal_steps:
+                training_arguments_kwargs["jagged_restart_anneal_steps"] = (
+                    self.cfg.jagged_restart_anneal_steps
+                )
```
NanoCode012: Confused on the difference between the two. Is jagged the new method? Why do we have both here but remove relora from input schema?

src/axolotl/utils/schemas/peft.py
```diff
@@ -187,18 +187,10 @@ def validate_lora_dropout(cls, data):
...
-        json_schema_extra={"description": "Number of per-restart warmup steps"},
-    )
-    relora_anneal_steps: int | None = Field(
+    relora: bool | None = Field(
         default=None,
         json_schema_extra={
-            "description": "Number of anneal steps for each relora cycle"
+            "description": "Whether to use ReLoRA. Use with jagged_restart_*steps options."
         },
     )
```
NanoCode012, line 195: We need to add deprecation error and migration steps before we remove this from schema?

src/axolotl/core/builders/causal.py
winglian: We only have the user define the jagged, the relora args are for the callback handler

src/axolotl/utils/schemas/peft.py
winglian: I think this is a minor enough feature that we can skip that for this one

NanoCode012 (approved):
I think this looks good

PR #1682, opened 2024-06-03
update to be deprecated evaluation_strategy

winglian:


src/axolotl/utils/config/models/input/v0_4_1/__init__.py
```diff
@@ -99,6 +100,13 @@ def validate_dpo_beta(cls, dpo_beta):
             LOG.warning("dpo_beta is deprecated, use rl_beta instead")
         return dpo_beta
 
+    @field_validator("evaluation_strategy")
+    @classmethod
+    def validate_evaluation_strategy(cls, evaluation_strategy):
+        if evaluation_strategy is not None:
+            LOG.warning("evaluation_strategy is deprecated, use eval_strategy instead")
+        return evaluation_strategy
```
NanoCode012, line 108: Do we need to add one validation where we remap `eval_strategy = evaluation_strategy` if both aren't set. If both are set, we should error out.

src/axolotl/utils/config/__init__.py
```diff
@@ -247,370 +244,3 @@ def validate_config(cfg: DictDefault, capabilities: Optional[dict] = None):
...
-                    f"conflicting optimizer: {cfg.optimizer} used alongside deepspeed optimizer."
-                )
-
-    if cfg.test_datasets and cfg.val_set_size:
-        raise ValueError(
-            "non-zero val_set_size should not be used with test_datasets configuration"
-        )
-
-    if cfg.fsdp and "bnb" in cfg.optimizer:
-        raise ValueError(f"FSDP not compatible with {cfg.optimizer}")
```
NanoCode012, line 590: This does not exist in our config validation check. Is that okay?

src/axolotl/utils/config/__init__.py
```diff
@@ -247,370 +244,3 @@ def validate_config(cfg: DictDefault, capabilities: Optional[dict] = None):
...
-                            is True
-                        )
-                    ):
-                        raise ValueError(
-                            "bf16.enabled or fp16.enabled must be set to true when using ZeRO-3 with flash-attention"
-                        )
-            if "8bit" in cfg.optimizer and deepspeed_cfg.optimizer:
-                LOG.warning(
-                    f"conflicting optimizer: {cfg.optimizer} used alongside deepspeed optimizer."
-                )
```
NanoCode012, line 582: This does not exist in our config validation check. Is that okay?

NanoCode012:
I see, `c4 dataset` being mentioned but don't see change related to it.

winglian:
> I see, `c4 dataset` being mentioned but don't see change related to it.

Ah, this is an old enough PR that the c4 issue was pixed in another PR already

src/axolotl/utils/config/__init__.py
winglian: let's address these in another PR, since these aren't being checked currently

src/axolotl/utils/config/__init__.py
winglian: let's address these in another PR, since these aren't being checked currently

PR #1686, opened 2024-06-06
sanity check ranges in freeze.py

josharian:
# Description

this will catch range-related problems earlier and more clearly.

## Motivation and Context

in my case, it appears that deepspeed doesn't play well with ranges.
i am still investigating.

## How has this been tested?

manually, in my broken setup.
(it fails because the param length is 0.)


winglian (approved):

josharian:
Looks like that test failure is an unrelated flake. No sign of "unfreeze" in the logs for it.

PR #1688, opened 2024-06-07
Update multi-node.qmd

shahdivax:
Title: Distributed Finetuning For Multi-Node with Axolotl and Deepspeed

Description:
This PR introduces a comprehensive guide for setting up a distributed finetuning environment using Axolotl and Accelerate. The guide covers the following steps:

1. Configuring SSH for passwordless access across multiple nodes
2. Generating and exchanging public keys for secure communication
3. Configuring Axolotl with shared settings and host files
4. Configuring Accelerate for multi-node training with Deepspeed
5. Running distributed finetuning using Accelerate

### DIFF
--- a/docs/multi-node.qmd
+++ b/docs/multi-node.qmd
@@ -2,11 +2,12 @@
 title: Multi Node
 description: How to use Axolotl on multiple machines
 ---
+# Distributed Finetuning For Multi-Node with Axolotl and FSDP
+You will need to create a configuration for accelerate, either by using accelerate config and follow the instructions or you can use 
+one of the preset below:
 
-You will need to create a configuration for accelerate, either by using `accelerate config` and follow the instructions or you can use one of the preset below:
-
-~/.cache/huggingface/accelerate/default_config.yaml
 ```yaml
+~/.cache/huggingface/accelerate/default_config.yaml
 compute_environment: LOCAL_MACHINE
 debug: false
 distributed_type: FSDP
@@ -27,6 +28,7 @@ use_cpu: false
 ```
 
 Configure your model to use FSDP with for example:
+
 ```yaml
 fsdp:
   - full_shard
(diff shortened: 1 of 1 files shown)


winglian:
@muellerzr seem right?

casper-hansen:
This seems to assume that you have access to each node before your training starts. However, a lot of cloud systems like AzureML, SLURM, SageMaker does not let you follow guides like this because the assumptions of the guide is that you can modify these variables.

@shahdivax @winglian I would suggest a bit more of an automatic setup if you want this to work well for users.

shahdivax:
> This seems to assume that you have access to each node before your training starts. However, a lot of cloud systems like AzureML, SLURM, SageMaker does not let you follow guides like this because the assumptions of the guide is that you can modify these variables.
> 
> @shahdivax @winglian I would suggest a bit more of an automatic setup if you want this to work well for users.

This assumes that user are using EC2 instances from AWS. 

( I forgot to add that 😓)

Edit: Added in the heading

docs/multi-node.qmd
```diff
@@ -2,11 +2,12 @@
 title: Multi Node
 description: How to use Axolotl on multiple machines
 ---
+# Distributed Finetuning For Multi-Node with Axolotl and FSDP
```
muellerzr, line 5: Let's also change the `title: ` yaml to reflect this

docs/multi-node.qmd
```diff
@@ -2,11 +2,12 @@
...
+You will need to create a configuration for accelerate, either by using accelerate config and follow the instructions or you can use
```
muellerzr:
```suggestion
You will need to create a configuration for accelerate, either by using `accelerate config` and follow the instructions or you can use 
```

docs/multi-node.qmd
```diff
@@ -38,11 +40,114 @@ fsdp_config:
...
+Repeat the above steps for Node 2, but change the machine rank to `1`.
+
+## Step 5: Finetuning
+On Node 1 (server), run the finetuning process using Accelerate:
+
+```bash
+accelerate launch -m axolotl.cli.train examples/llama-2/qlora.yml
+```
+
+This will start the finetuning process across all nodes. You can check the different IP addresses before each step to verify that the training is running on every node.
```
muellerzr, line 153: To my knowledge this is not the case. You need to do `accelerate launch -m` on *every server* else it will sit there and never actually start

shahdivax: When we tasted, we were only starting on single node (server) and it was able to use the resources from other nodes,
As a proof, we were able to see the ip of both the machines on the left, and in the total GPU it were showing all the GPU form all the nodes.

docs/multi-node.qmd
shahdivax: Thanks for the review,
But I couldn't understand with what you what to change, can you please rephrase or mention the exact changes, if possible.

docs/multi-node.qmd
```diff
@@ -2,11 +2,12 @@
 title: Multi Node
```
winglian:
```suggestion
title: Distributed Finetuning For Multi-Node with Axolotl and FSDP
```

docs/multi-node.qmd
winglian: I added the suggestion above that can be committed directly.

docs/multi-node.qmd
winglian: @muellerzr My guess is this is probably specific to deepspeed since the IP addresses are set in a hostfile. We should probably disambiguate this that it only needs to be run on the first node when this is the case. Most other cases like FSDP or plain multinode DDP will likely still need `accelerate launch` to be run on each node.

docs/multi-node.qmd
shahdivax: @winglian Committed The Changes you suggested.

docs/multi-node.qmd
shahdivax: Done, Thanks

shahdivax (commented):
All the changes required are Done , And I think this doc is now good to go.

docs/multi-node.qmd
shahdivax: @winglian, @muellerzr  That might be the case , because for us deepspeed was a good options where we were using multi node for finetuning via EC2, as it provides the public ip , and we used hostfile, it was really easy to connect both machines and run the finetuning on root only, this indeed connected all the other instances. (using all the resources from all the nodes via single node)

PR #1693, opened 2024-06-07
download model weights on preprocess step

winglian:
downloading model weights on multi-gpu during training can leave filesystem in a wonky state. attempt to download during pre-process step


winglian:
@charlesfrye fyi, I don't think this will affect the modal llm-finetuning repo, but something to be aware of on your end.

PR #1703, opened 2024-06-11
Allow "weight: 0" in messages to mask them

DavidFarago:
Allow in message objects the additional key `weight`, which can be set to 0 (or 1) to cause that message to be masked out (or left unmasked) for training (similar to [1]). 

[1]: https://github.com/mistralai/mistral-finetune

# Description

Extend `src/axolotl/prompters.py::_build_result` to return the turns with weights as additional tuple element. Do this in axolotl directly instead of modifying `fastchat.conversation`'s `Conversation`.

Extend `src/axolotl/prompt_tokenizers.py::tokenize_prompt` to mask out tokens when weight is set to 0.

## Motivation and Context

This is helpful for training the model to be robust and capable of error recovery upon a bad assistant message. A missing `weight` key defaults to weight 1, to guarantee downward compatibility.

## How has this been tested?

Extend `tests/prompt_strategies/test_sharegpt.py` to contain messages with `weight` keys.

### DIFF
--- a/tests/prompt_strategies/test_sharegpt.py
+++ b/tests/prompt_strategies/test_sharegpt.py
@@ -52,6 +52,51 @@ def fixture_sharegpt_dataset():
     )
 
 
+@pytest.fixture(name="sharegpt_dataset_with_weights")
+def fixture_sharegpt_dataset_with_weights():
+    return Dataset.from_list(
+        [
+            {
+                "conversations": [
+                    {
+                        "from": "system",
+                        "value": "repeat",
+                    },
+                    {
+                        "from": "human",
+                        "value": "hello",
+                        "weight": 1,
+                    },
+                    {
+                        "from": "gpt",
+                        "value": "hello",
+                        "weight": 0,
+                    },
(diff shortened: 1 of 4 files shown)


tests/prompt_strategies/test_sharegpt.py
winglian, line 1: Rather than modifying the existing test, could you leave that in place to test the prior feature capability and add a new test for the new feature?

winglian (changes requested):

tests/prompt_strategies/test_sharegpt.py
DavidFarago: Good point, @winglian. I added 4 new test cases in the same file because the test cases are closely related.

winglian (approved):
thank you!

DavidFarago:
you are welcome, @winglian -- thank you for reviewing.

Will you merge this PR or give me write access to this repository?

winglian:
merged. thanks @DavidFarago !

PR #1716, opened 2024-06-25
Fix eval_sample_packing in llama-3 lora example

RodriMora:
# Description
Fix eval_sample_packing: False in example file
https://github.com/OpenAccess-AI-Collective/axolotl/issues/1644

## Motivation and Context

The provided example didn't work as is
## How has this been tested?

accelerate launch -m axolotl.cli.train examples/llama-3/lora-8b.yml
## Types of changes

Parameter in .yml file

### DIFF
--- a/examples/llama-3/lora-8b.yml
+++ b/examples/llama-3/lora-8b.yml
@@ -14,7 +14,7 @@ val_set_size: 0.05
 output_dir: ./outputs/lora-out
 
 sequence_len: 4096
-sample_packing: true
+sample_packing: false
 pad_to_sequence_len: true
 
 adapter: lora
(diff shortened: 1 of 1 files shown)


examples/llama-3/lora-8b.yml
```diff
@@ -14,7 +14,7 @@ val_set_size: 0.05
 output_dir: ./outputs/lora-out
 
 sequence_len: 4096
-sample_packing: true
+sample_packing: false
```
winglian:
```suggestion
sample_packing: true
eval_sample_packing: false
```

I think you intended to set eval_sample_packing?

RodriMora: You are totally right. Sorry for the mistake

winglian (approved):
thanks!

PR #1718, opened 2024-06-27
support for gemma2 w sample packing

winglian:
# Description


winglian:
needs https://github.com/huggingface/transformers/pull/31670 for sample packing to work

PR #1731, opened 2024-07-09
Add option to raise error for long seqs + drop seqs with no outputs

chiragjn:
A summary of changes:

1. Removes with zero first handling in sample packing function. This function is already called from `utils.data.sft.prepare_dataset` under a zero first guard. In fact, if you keep it and process the dataset in a multi gpu setting, when you provide a eval set explicitly it causes incorrect synchornization

   i. `utils.data.sft.prepare_dataset` establishes a barrier() on rank0, all other ranks stopped 
   ii. `load_prepare_datasets` > `load_tokenized_prepared_datasets` > `process_datasets_for_packing` is called for train set and again establishes a barrier() on rank 0, all other ranks still stopped. 
  iii. when `process_datasets_for_packing finishes`, barrier() is called on all other ranks which allows the ranks blocked in i. to resume and will lead to other ranks trying to load the validation set before it has been processed by rank 0

2. Adds an option `drop_long_sequences` which when set to `False` allows the user to raise an error if sequences longer than `sequence_len` are found in the dataset. Consequently this also requires truncation to be turned off in prompt strategies. Also adds a sequence lengths histogram printer util

3. Adds dropping of sequences with no learnable outputs. It can happen that with empty completions or messages or with truncation=True + train_on_inputs=False can lead to all -100 labels with nothing to learn - such samples make the loss nan

4. Adds a check and error if the dataset ends up empty after all this processing

---

Things I am not sure about yet and would love some help:

- Turning trucation off in all prompt strategies and its consequences
- Changes for DPO

---

Things Todo:

- Writing a proper test

---

For context, my employer has been running these changes for quite a while now (https://github.com/truefoundry/axolotl/pull/5/files) but we majorly use chat_template prompt strategy and SFT but axololt is quite a large surface area code base so we would need some help 😅

### DIFF
--- a/src/axolotl/utils/trainer.py
+++ b/src/axolotl/utils/trainer.py
@@ -15,8 +15,9 @@ from torch.utils.data import DataLoader, RandomSampler
 from transformers.utils import is_torch_bf16_gpu_available
 
 from axolotl.core.trainer_builder import HFCausalTrainerBuilder, HFRLTrainerBuilder
-from axolotl.utils.distributed import is_main_process, reduce_and_broadcast, zero_first
+from axolotl.utils.distributed import reduce_and_broadcast
 from axolotl.utils.samplers import MultipackBatchSampler, get_dataset_lengths
+from axolotl.utils.samplers.utils import plot_ascii_lengths_histogram
 
 LOG = get_logger("axolotl")
 
@@ -169,6 +170,10 @@ def add_length(sample):
     return sample
 
 
+def drop_samples_with_no_outputs(sample):
+    return any(v != -100 for v in sample["labels"])
+
+
 def drop_long_seq(sample, sequence_len=2048, min_sequence_len=2):
     return (
         len(sample["input_ids"]) <= sequence_len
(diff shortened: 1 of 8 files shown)


src/axolotl/utils/trainer.py
```diff
@@ -169,116 +170,181 @@ def add_length(sample):
...
+        dataset=train_dataset,
+        ds_split_name="Train",
         sequence_len=cfg.sequence_len,
-        min_sequence_len=cfg.min_sample_len or 2,
+        min_sample_len=cfg.min_sample_len or 2,
+        ignore_long_sequences=cfg.drop_long_sequences,
+        drop_no_outputs=True,
+        dataset_processes=cfg.dataset_processes,
+        load_from_cache_file=not cfg.is_preprocess,
+        show_lengths_summary=True,
```
winglian, line 269:
```suggestion
        show_lengths_summary=cfg.is_preprocess,
```
the outputs are likely going to be really wonky if run on multi-process/gpu. since preprocessing is typically single process, it's going to be safer.

src/axolotl/utils/trainer.py
```diff
@@ -169,116 +170,181 @@ def add_length(sample):
...
-
-    train_dataset = train_dataset.filter(
-        drop_long,
-        desc="Dropping Long Sequences",
+    train_dataset = _maybe_drop_sequences(
+        dataset=train_dataset,
+        ds_split_name="Train",
+        sequence_len=sequence_len,
+        ignore_long_sequences=True,  # TODO (chiragjn): Pass via cfg here
+        show_lengths_summary=False,
```
winglian, line 345:
```suggestion
        show_lengths_summary=False,  # we don't plot summary to avoid iterating the entire dataset here
```

src/axolotl/utils/trainer.py
```diff
@@ -169,116 +170,181 @@ def add_length(sample):
...
 
 def process_pretraining_datasets_for_packing(
     train_dataset, sequence_len, skip_position_ids=True
 ):
-    drop_long = partial(drop_long_seq, sequence_len=sequence_len)
-
-    train_dataset = train_dataset.filter(
-        drop_long,
-        desc="Dropping Long Sequences",
+    train_dataset = _maybe_drop_sequences(
```
winglian, line 340: Actually, I don't think we can swap the partial/filter for the function here. in the _maybe fn, there is
```python
dropped_rows = len_pre_drop - len(dataset)
```
which we can't really get from an utterable dataset

chiragjn: Ah yes, makes sense, I'll refactor and think of a better way

src/axolotl/utils/trainer.py
chiragjn: I generally agree, I'll make this change for now, but considering this function is called under with zero first, there is no wonkiness

PR #1732, opened 2024-07-09
Allow using tokenizer's default chat template or pass custom jinja chat template

### LINKED ISSUE #1689: Using native chat_template from tokenizer config in `chat_template` strategy

### 🔖 Feature description

Currently when using `type: chat_template` we have to chose a chat template which are encoded in the axolotl codebase (defaults to chatml)

https://github.com/OpenAccess-AI-Collective/axolotl/blob/a82a7115224b7aef14301387a11ad4729fd6ca52/src/axolotl/prompt_strategies/chat_template.py#L99-L102

https://github.com/OpenAccess-AI-Collective/axolotl/blob/a82a7115224b7aef14301387a11ad4729fd6ca52/src/axolotl/prompt_strategies/chat_template.py#L115-L126

https://github.com/OpenAccess-AI-Collective/axolotl/blob/a82a7115224b7aef14301387a11ad4729fd6ca52/src/axolotl/utils/chat_templates.py#L21-L29

I would like to use the tokenizer `tokenizer.chat_template` as that would help the model start from a familiar place.

### ✔️ Solution

I would like to pass `chat_template` as `None` in which case it should be picked from tokenizer or raise error / fallback to chatml if `tokenizer.chat_template` missing

I can work on this, seems like a small change.

### ❓ Alternatives

_No response_

### 📝 Additional Context

_No response_

### Acknowledgements

Fixed by #1970

chiragjn:
Closes #1689 

Summary of changes:

1. Adds `tokenizer_default` as option for `chat_template` in `chat_template` prompt strategy that allows using the chat template from tokenizer's config.json
2. Allows falling back to chat templates available in axolotl if tokenizer does not have a chat template
3. Allows passing in a custom chat template using a new option `chat_template_jinja` which is only considered when `chat_template` option is set to `jinja`

---

Why?

Many popular models are not trained with chatml format. As a result for the model to correctly learn chatml we have to turn on train_on_inputs which requires more compute and time. If we can use the model's already learned chat template we can just learn the output tokens

---


chiragjn:
On a separate note, I would also like to explore how can we support passing a jinja template entirely via the config itself which might me more flexible

src/axolotl/utils/chat_templates.py
```diff
@@ -2,9 +2,15 @@
...
+import logging
 
+LOG = logging.getLogger("axolotl.utils.chat_templates")
 
-def chat_templates(user_choice: str):
+_DEFAULT_TEMPLATE_CHOICE = "tokenizer_default"
+_DEFAULT_FALLBACK_CHATML_TEMPLATE_CHOICE_PREFIX = "tokenizer_default_fallback_"
+
+
+def chat_templates(user_choice: str, tokenizer=None):
```
winglian: I had initially similarly included the tokenizer as an argument to chat template, but was trying to minimize the interdependency on this module with tokenizers. Seems like this is a reasonable workaround for now. thanks!

winglian:
Thanks! Would you have some time to put together some tests for this?

ganler:
I wonder if it is a good idea to use the `chat_template` available from the trained model's `tokenizer_config.json` as the default template?

chiragjn:
Added a few tests

> I wonder if it is a good idea to use the chat_template available from the trained model's tokenizer_config.json as the default template?

Yes, that is what this PR aims to do, if we all agree, we can change the default in config to `tokenizer_default_fallback_chatml` :)

ganler:
Yes. I am sorry I just read the PR in more detail and found that's exactly what I was hoping for (what a coincidence!) so I removed my previous dumb question. Thanks for the contribution!

ganler:
I have one question. Can this option be used together with the sharegpt conversation format?

winglian:
> I have one question. Can this option be used together with the sharegpt conversation format?

Unfortunately no, they are different templating libraries.

Tostino:
Thank you for working on this! I was needing just this...and was going to either integrate my custom chat template into axolotl (limits me on experimenting) or work on getting this feature going. Glad you beat me to it. Doesn't look like it'll conflict much with the changes I made for https://github.com/axolotl-ai-cloud/axolotl/pull/1756.

chiragjn:
> Thank you for working on this! I was needing just this...and was going to either integrate my custom chat template into axolotl (limits me on experimenting) or work on getting this feature going. Glad you beat me to it. Doesn't look like it'll conflict much with the changes I made for https://github.com/axolotl-ai-cloud/axolotl/pull/1756.

Amazing, would love to see both getting merged

chiragjn:
Added `chat_template_jinja` option to provide a Jinja template as per @fozziethebeat's suggestion

chiragjn:
Bumping for a review here @winglian

nyxkrage:
Regarding point 4, it seems that the jinja template from that repo is very different when compared to how fastchat/axolotl currently formats the prompt, and mistrals templates

FastChat/Axolotl currently places the system message with a single newline inside the first user turn
Mistral 0.1 places the system message with 2 newlines inside the first user turn, and with different spacing between [INST] tokens
the Mistral v3 tokenizer places the system message with 2 newlines inside the last user turn

The one in the linked repo puts the system prompt before the first [INST], I'm unsure if this is an oversight by the creator of that template/repo or deliberately chosen to be different from the official mistral template

chiragjn:
I can get rid of the mistral template now, anyway this PR now allows passing custom jinja template, I had only added it for convenience

> deliberately chosen to be different from the official mistral template

Most likely deliberately as older mistral models did not support system message - so there was no official way for older models

chiragjn:
Bumping this again @winglian

src/axolotl/utils/chat_templates.py
```diff
@@ -17,20 +44,67 @@ def chat_templates(user_choice: str):
...
-    }
-
-    if user_choice in templates:
-        return templates[user_choice]
+        user_choice = user_choice[
+            len(_DEFAULT_FALLBACK_CHATML_TEMPLATE_CHOICE_PREFIX) :
+        ]
+        LOG.warning(
+            f"No chat template found on tokenizer, falling back to {user_choice}. It is recommended to set --train_on_inputs to True for the model to learn this chat template."
```
NanoCode012, line 85: Is this warning accurate? I wasn't aware of it.

src/axolotl/utils/chat_templates.py
```diff
@@ -17,20 +44,67 @@ def chat_templates(user_choice: str):
...
+
+
+def extract_chat_template_args(cfg, ds_cfg: Optional[Dict[str, Any]] = None):
+    if ds_cfg and ds_cfg.get("chat_template"):
+        chat_template_choice = ds_cfg.get("chat_template") or "chatml"
+        chat_template_jinja = ds_cfg.get("chat_template_jinja")
+    else:
+        chat_template_choice = cfg.get("chat_template") or "chatml"
+        chat_template_jinja = cfg.get("chat_template_jinja")
+    return chat_template_choice, chat_template_jinja
```
NanoCode012: Is this default to `chatml` a good default? Wouldn't a proper default be the tokenizer's current template?

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
```diff
@@ -120,12 +146,22 @@ class SFTDataset(BaseModel):
...
 
+    @model_validator(mode="before")
+    @classmethod
+    def check_chat_template_config(cls, data):
+        if data.get("chat_template") == ChatTemplate.jinja and not data.get(
+            "chat_template_jinja"
+        ):
+            raise ValueError(
+                "chat_template_jinja is required when chat_template is set to jinja"
+            )
```
NanoCode012: Same like points above, do we need to set it to `jinja` manually? We can set this internally if required by check for `chat_template_jinja` directly.

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
```diff
@@ -786,6 +812,18 @@ def check_sample_packing_w_xformers(cls, data):
...
+
+        return data
```
NanoCode012: Same as above.

src/axolotl/utils/chat_templates.py
```diff
@@ -17,20 +44,67 @@ def chat_templates(user_choice: str):
...
+
+
+def get_chat_template_from_config(
+    cfg,
+    ds_cfg: Optional[Dict[str, Any]] = None,
+    tokenizer: Optional["PreTrainedTokenizerBase"] = None,
+) -> str:
+    ds_cfg = ds_cfg or {}
```
NanoCode012:
```suggestion
```

May not be needed as your internal function checks if valid.

src/axolotl/utils/chat_templates.py
```diff
@@ -2,9 +2,36 @@
...
+    from transformers import PreTrainedTokenizerBase
 
-def chat_templates(user_choice: str):
+LOG = logging.getLogger("axolotl.utils.chat_templates")
+
+_JINJA_TEMPALTE_CHOICE = "jinja"
+_DEFAULT_TEMPLATE_CHOICE = "tokenizer_default"
+_DEFAULT_FALLBACK_CHATML_TEMPLATE_CHOICE_PREFIX = "tokenizer_default_fallback_"
+
+_TEMPLATES = {
```
NanoCode012: Should this var name be more descriptive? For ex, `_CHAT_TEMPLATES`

src/axolotl/utils/chat_templates.py
```diff
@@ -17,20 +44,67 @@ def chat_templates(user_choice: str):
...
     """
+    if user_choice == _JINJA_TEMPALTE_CHOICE:
+        if not jinja_template:
+            raise ValueError(
+                f"`jinja_template` cannot be None when `chat_template` choice is {_JINJA_TEMPALTE_CHOICE}"
+            )
+        return jinja_template
+
+    if user_choice == _DEFAULT_TEMPLATE_CHOICE:
+        if not tokenizer:
```
NanoCode012, line 62: Would this ever be None? It is to my understanding that the tokenizer must be loaded before all of this.

NanoCode012 (commented):
Hey, there. Thank you for the PR. We will try to get this PR in. Starting off, I have some points below:

- We can assume if `chat_template_jinja` , `chat_template==jinja` without having to set it separately, probably?
- Could these new configs be added to `docs/config`? We may also consider adding into `docs/dataset-formats/conversation` as an example on how to use this chat_template method as it was not documented before.
- Which `chat_template` is being loaded, the root one or within `ds_config`? I wasn't able to clearly tell from the code. Maybe there needs to be improvement on doc to clarify this.

src/axolotl/utils/chat_templates.py
chiragjn: This is assuming if chat template is missing then most likely the model was not trained to chat at all

If the model was indeed trained for chat but chat template is not present in tokenizer config then this warning would be incorrect - but this should be a rare case

src/axolotl/utils/chat_templates.py
chiragjn: We can change the default to tokenizer_default. Kept it same as before till we reached a decision

src/axolotl/utils/chat_templates.py
chiragjn: There were a few call sites that might not pass tokenizer. I'll check if this is no longer the case and get back

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
chiragjn: Sure, can set automatically

src/axolotl/utils/chat_templates.py
chiragjn: There were a few call sites that might not pass tokenizer. I'll check if this is no longer the case and get back

src/axolotl/utils/chat_templates.py
NanoCode012: Was the prior default, chatml? I think the tokenizer's template should be more logical as a default.

src/axolotl/utils/chat_templates.py
NanoCode012: Oh, I wasn't aware of that. Resolved

src/axolotl/utils/chat_templates.py
```diff
@@ -2,9 +2,36 @@
...
 
+if TYPE_CHECKING:
+    from transformers import PreTrainedTokenizerBase
 
-def chat_templates(user_choice: str):
+LOG = logging.getLogger("axolotl.utils.chat_templates")
+
+_JINJA_TEMPALTE_CHOICE = "jinja"
+_DEFAULT_TEMPLATE_CHOICE = "tokenizer_default"
+_DEFAULT_FALLBACK_CHATML_TEMPLATE_CHOICE_PREFIX = "tokenizer_default_fallback_"
```
NanoCode012, line 15: Could you tell me what's the difference between the two? How would one explicitly set the config to use the tokenizer?

```yaml
type: chat_template
```

Is this correct? I think if this part is clear, we can close https://github.com/axolotl-ai-cloud/axolotl/pull/1855/files .

src/axolotl/utils/chat_templates.py
NanoCode012: This makes sense after reading the code in more detail. It's for the condition if setting template with fallback.

NanoCode012:
Please let me know if there's any part you need help with. I think as a summary, the only parts left to be addressed are:
- [x] Documentation into `config.qmd` to list all these new configs
- [x] Documentation into `conversation.qmd` to show a few examples of this and explain the new options
- [x] Address the comments above
- [ ] Test for DPO / ORPO (if not done)

src/axolotl/utils/chat_templates.py
chiragjn: Yes, this is correct, updated docs as per your review comment

chiragjn:
Have addressed comments and updated docs to the best of my knowledge. Will test DPO/ORPO.

One thing I am not clear on - what should be the behavior is some one provides different chat templates across different datasets? It causes ambiguity in selecting which template to finally save in the tokenizer. Ideally, no one should do it this way but there is no validation to protect against this.

NanoCode012:
> One thing I am not clear on - what should be the behavior is some one provides different chat templates across different datasets? It causes ambiguity in selecting which template to finally save in the tokenizer. Ideally, no one should do it this way but there is no validation to protect against this.

That's a good point. I missed this. 

In this case, we may need to adjust the hash to include the chat_template (if exists?). https://github.com/axolotl-ai-cloud/axolotl/blob/17af1d7081414c32614cbabe324e1197ca9f43a7/src/axolotl/utils/data/sft.py#L143

docs/config.qmd
```diff
@@ -141,9 +141,16 @@ test_datasets:
...
+# - alpaca/inst/chatml/gemma/cohere/llama3/phi_3/deepseek_v2/jamba: These chat templates are available in the axolotl codebase at src/axolotl/utils/chat_templates.py
+# - tokenizer_default_fallback_*: where * is the name of the chat template to fallback to. E.g. tokenizer_default_fallback_chatml. This is useful when the chat template is not available in the tokenizer.
+# - jinja: Uses a custom jinja template for the chat template. The custom jinja template should be provided in the chat_template_jinja field.
+# The selected chat template will be saved to the tokenizer_config.json for easier inferencing
+# Note: It is recommended to set train_on_inputs to true when using a chat template that is different from the model's default chat template.
+chat_template: tokenizer_default
+# custom jinja template for chat template. This will be only used if chat_template is set to `jinja` or `null` (in which case chat_template is automatically set to `jinja`). Default is null.
+chat_template_jinja: null
```
NanoCode012, line 196: I believe I may have some confusion about this section

There should be `chat_template` and `chat_template_jinja` under the `datasets: ` level to set for training (as shown in your example doc) under `type: chat_template` only.

The `chat_template` config at root should focus on setting the tokenizer_config.json only?

Please correct me if I'm understanding it incorrectly. Is the root config also accessed during tokenization?

docs/dataset-formats/conversation.qmd
```diff
@@ -69,3 +69,152 @@ creates a chat where bot is asked to tell a joke, then explain why the joke is f
...
+
+
+## chat_template
+
+Chat Template strategy uses a jinja2 template that converts a list of messages into a prompt. Usually this chat template is stored in tokenizer_config.json under the key `chat_template`.
+
+Conversational data would normally look like follows:
+
+```{.json filename="data.jsonl"}
+{"messages": [{"role": "...", "content": "..."}]}
```
NanoCode012:
```suggestion
{"conversations": [{"from": "...", "value": "..."}]}
```

We should use the default values as you already explained it can be customized below.

docs/config.qmd
chiragjn: My understanding was that root level config were to fallback to if the dataset item does not mention them which is why I added it at root. Ideally it does not make sense to even have different chat templates across different datasets because that would make inferencing tricky

Ideally we shouldn't have both chat_template and chat_template_jinja at root level.

levguy:
Hi @chiragjn ,
I'd like to point out a potential issue that might occur when training DPO with prompt strategy `chat_template.default` and using the base model's (or custom) chat template.
In many open source models (e.g. `Phi-3-medium-128k-instruct`), the chat template validates that the first message has the `user` role, and raises an exception otherwise.
Consider the code [here](https://github.com/truefoundry/axolotl/blob/cj_tokenizer_default_prompt_template/src/axolotl/prompt_strategies/dpo/chat_template.py#L65):
```
        result["chosen"] = tokenizer.apply_chat_template(
            [chosen],
            add_generation_prompt=False,
            chat_template=chat_template_string,
            tokenize=False,
        )
```
We pass here `[chosen]`, which is a list containing a single message with `assistant` role. Hence the validation mentioned above will cause raising an exception.

This can be solved by adding a dummy user message, e.g.:
```
        dummy_user_message = {"role": "user", "content": "dummy"}
        result["chosen"] = tokenizer.apply_chat_template(
            [dummy_user_message, chosen],
            add_generation_prompt=False,
            chat_template=chat_template_str,
            tokenize=False,
        )
```
For trimming the dummy message, no additional code is needed, as it will be trimmed by the lines that follow:
```
        chosen_strip_index = result["chosen"].find(chosen["content"])
        result["chosen"] = result["chosen"][chosen_strip_index:].rstrip()
```
These lines remove what comes before the actual message content (typically an `assistant` header / special token).

The same workaround is needed in the `rejected` case as well ([here](https://github.com/truefoundry/axolotl/blob/cj_tokenizer_default_prompt_template/src/axolotl/prompt_strategies/dpo/chat_template.py#L74)).

NanoCode012:
Hello @levguy , thank you for the find. Would it be better to separate that into a different PR to limit the scope of this PR? I think we would want to get this PR in first.

@chiragjn , may I ask if there are any parts blocking to this current PR now? I believe you have addressed the points I lifted. Only the hash of the tokenized dataset remains.

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
```diff
@@ -677,7 +704,11 @@ class Config:
     gpu_memory_limit: Optional[Union[int, str]] = None
     low_cpu_mem_usage: Optional[bool] = None
 
-    chat_template: Optional[ChatTemplate] = None
+    chat_template: Union[
+        ChatTemplate,
+        Annotated[str, StringConstraints(pattern="^tokenizer_default_fallback_")],
+    ] = ChatTemplate.chatml
```
NanoCode012: Does this `chatml` default also need to be changed to `tokenizer_default`?

I ran with the below model that has no chat_template in its config, and it defaulted to processing into `chatml` format.
```
base_model: mlx-community/gemma-2-2b

datasets:
  - path: LDJnr/Puffin
    type: chat_template
```

NanoCode012:
Edit: Resolved in commit below

---

I notice also that the loss masking may have some issues for chat template (likely not due to this PR, but an old issue).

For ex, it sometimes does not detect the assistant portion correctly.

```
# user masked correctly
(-100, 235322) |(-100, 235371) im(-100, 571) _(-100, 235298) start(-100, 2997) |>(-100, 73786) user(-100, 1645) 
(-100, 108) Each(-100, 7639)  problem(-100, 3210)  consists(-100, 13569)  of(-100, 576)  three(-100, 2149)
...
# assistant also masked
 <(-100, 235322) |(-100, 235371) im(-100, 571) _(-100, 235298) start(-100, 2997) |>(-100, 73786) assistant(-100, 105776) 
(-100, 108) If(-100, 2495)  the(-100, 573)  third(-100, 4906)  statement(-100, 6218)  is(-100, 603)  true(-100, 1382)  ((-100, 591) Bananas(-100, 217061)  cost(-100, 2458) 
```

Config to recreate:
```yaml
base_model: mlx-community/gemma-2-2b

datasets:
  - path: LDJnr/Puffin
    type: chat_template
    chat_template: chatml
    shards: 10
chat_template: chatml
```

chiragjn:
Sorry, I did not mean to block this PR
It is just that I was running busy at my main job, I wanted to address the DPO issues before merging, I'll make some time for this

fozziethebeat:
I'm excited to see this land!

NanoCode012:
Hey @chiragjn , I made some changes but did not have write permission to the PR. Could you take a look at the one I made to yours and merge it if it's okay?

I merged main to the branch, fixed some config, updated docs, and added some more tests.

@levguy , I added a test for the case you talked about, and it seemed like phi3's chat template did not have that issue https://github.com/truefoundry/axolotl/blob/dd87d8c438998b60dceaa5f5093ba619c3bca683/tests/prompt_strategies/test_dpo_chat_templates.py#L167-L190

chiragjn:
Apologies, life just got too busy to get back to this.

@NanoCode012 thanks for carrying it 🙌 , I have forwarded all your commits. Not sure why it was not editable by maintainers in first place.

Changes seem good to me

levguy:
> Hey @chiragjn , I made some changes but did not have write permission to the PR. Could you take a look at the one I made to yours and merge it if it's okay?
> 
> I merged main to the branch, fixed some config, updated docs, and added some more tests.
> 
> @levguy , I added a test for the case you talked about, and it seemed like phi3's chat template did not have that issue https://github.com/truefoundry/axolotl/blob/dd87d8c438998b60dceaa5f5093ba619c3bca683/tests/prompt_strategies/test_dpo_chat_templates.py#L167-L190

Hey @NanoCode012 ,
Thank you for adding a test for that case!
1. Your test uses a conversation without a system message. You need to add system message for testing this.
2. The chat template of Phi-3 was updated by model owners and it supports system message now, so it can't be used for testing this issue. You can use `google/gemma-2b-it` instead. [Here](https://huggingface.co/google/gemma-2b-it/blob/main/tokenizer_config.json) we see that its chat template raises an exception if system message is passed:
```
"chat_template": "{{ bos_token }}{% if messages[0]['role'] == 'system' %}{{ raise_exception('System role not supported') }}{% endif %}{% ...
````

**Comment by levguy** [comment]
> levguy

Hey @NanoCode012 ,
Thank you for adding a test for that case!
The chat template of Phi-3 was updated by model owners and the relevant validity check was removed. You can use `google/gemma-2b-it` instead. You can take a look at its chat template [here](https://huggingface.co/google/gemma-2b-it/blob/main/tokenizer_config.json#L1507). There is this part:
```
{% if (message['role'] == 'user') != (loop.index0 % 2 == 0) %}{{ raise_exception('Conversation roles must alternate user/assistant/user/assistant/...') }}
```
Which will cause an exception raised in case the first message has `assistant` role.

**Comment by levguy** [comment]
> Hey @chiragjn , I made some changes but did not have write permission to the PR. Could you take a look at the one I made to yours and merge it if it's okay?
> 
> I merged main to the branch, fixed some config, updated docs, and added some more tests.
> 
> @levguy , I added a test for the case you talked about, and it seemed like phi3's chat template did not have that issue https://github.com/truefoundry/axolotl/blob/dd87d8c438998b60dceaa5f5093ba619c3bca683/tests/prompt_strategies/test_dpo_chat_templates.py#L167-L190

Hey @NanoCode012 ,
Thank you for adding a test for that case!
The chat template of Phi-3 was updated by model owners and the relevant validity check was removed. You can use `google/gemma-2b-it` instead. You can take a look at its chat template [here](https://huggingface.co/google/gemma-2b-it/blob/main/tokenizer_config.json#L1507). There is this part:
```
{% if (message['role'] == 'user') != (loop.index0 % 2 == 0) %}{{ raise_exception('Conversation roles must alternate user/assistant/user/assistant/...') }}
```
Which will cause an exception raised in case the first message has `assistant` role.

NanoCode012:
Hey @levguy , thank you for clarifying this. I have added a commit to address this along with a test case https://github.com/truefoundry/axolotl/pull/8/commits/d101cfc125ad6b879209ef32eb5422f27539c9b5

@chiragjn I made a PR to your branch. Could you please take a look and merge when you can? This should be the last changes required, but if you can enable contributors to have write access, that’ll make things more convenient.

levguy:
> Hey @levguy , thank you for clarifying this. I have added a commit to address this along with a test case [truefoundry@d101cfc](https://github.com/truefoundry/axolotl/commit/d101cfc125ad6b879209ef32eb5422f27539c9b5)
> 
> @chiragjn I made a PR to your branch. Could you please take a look and merge when you can? This should be the last changes required, but if you can enable contributors to have write access, that’ll make things more convenient.

Hey @NanoCode012 , great, thank you!
As a final remark, maybe it would be better to change (in `dummy_user_message`) the "dummy" content to a very unlikely string such as "[[[dummy]]]" or something like this. This would reduce the chance of having a problem when calling `find(chosen["content"])` and `find(rejected["content"])` - a wrong behaviour will occur in case `chosen["content"]` and/or `rejected["content"]` are identical to the dummy string.

chiragjn:
> @chiragjn I made a PR to your branch. Could you please take a look and merge when you can? This should be the last changes required, but if you can enable contributors to have write access, that’ll make things more convenient.

I think I am not able to enable direct write access because I started the PR from my employer's fork 🤦 
If it helps I can re-create this PR from my personal account fork.
Or NanoCode012, you can also close this PR and start a PR directly from your fork, I don't mind, either way commit history would be preserved

chiragjn:
Superseded by #1970

PR #1740, opened 2024-07-12
bump xformers to 0.0.27

akshaylive:
# Description
Update `xformers` version.

## Motivation and Context

Preserve compatibility with torch 2.3.1. [Reference](https://github.com/facebookresearch/xformers/issues/1052)

## How has this been tested?
```
 pip3 install -e '.[flash-attn,deepspeed]'
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,7 +16,7 @@ flash-attn==2.5.8
 sentencepiece
 wandb
 einops
-xformers==0.0.26.post1
+xformers==0.0.27
 optimum==1.16.2
 hf_transfer
 colorama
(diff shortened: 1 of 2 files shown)


winglian:
Screenshot 2024-07-12 at 9 14 56 PM
so we need to cross check the wheels they are building for the new version and make sure they didn't drop pytorch/cuda version for some wheels.  I think this is likely fine, but when we update xformers in requirements, we parse it in [setup.py](https://github.com/axolotl-ai-cloud/axolotl/blob/main/setup.py#L51-L58)  so we can replace older versions based on the installed version of python in case there isn't a wheel for it.

winglian:
Thanks @akshaylive!

PR #1756, opened 2024-07-15
Add flexible configuration options for `chat_template` dataset training

Tostino:
- Introduce roles_to_train parameter to set training labels by role
- Add train_on_eos option to configure training on end-of-sequence tokens
- Implement per-message training configuration in dataset
- Allow fine-grained control over training specific portions of messages
- Add message_field_training and message_field_training_detail settings
- Implement mapping between dataset character offsets and tokenized prompt
- Enhance test suite to cover new functionality

## Motivation and Context

This change is required to provide more granular control over the training process, particularly for fine-tuning models on specific roles and message components. It solves the problem of overly broad training that may not focus on the most relevant parts of the dataset.

These enhancements will enable researchers and developers to create more targeted and efficient training workflows, potentially leading to better model performance on specific tasks or domains.

## How has this been tested?

The changes have been tested locally using a comprehensive test suite. The enhanced test suite now covers all new functionality, including:

1. Unit tests for the new parameters
2. Validation of per-message training configurations
3. Tests for fine-grained control over message portion training
4. Accuracy checks for the mapping between dataset character offsets and tokenized prompts

## Types of changes

[x] New feature (non-breaking change which adds functionality)
[o] Breaking change (fix or feature that would cause existing functionality to not work as expected)
[o] Documentation update
[o] Performance enhancement
[o] Code cleanup or refactor
[o] Dependency update

### DIFF
--- a/src/axolotl/prompt_strategies/chat_template.py
+++ b/src/axolotl/prompt_strategies/chat_template.py
@@ -6,14 +6,16 @@ import logging
 from typing import Any, Dict, List, Optional
 
 from axolotl.prompt_tokenizers import PromptTokenizingStrategy
-from axolotl.prompters import Prompter
+from axolotl.prompters import IGNORE_TOKEN_ID, Prompter
 from axolotl.utils.chat_templates import chat_templates
 
+# Configure the logger
+logging.basicConfig(level=logging.DEBUG)
 LOG = logging.getLogger("axolotl")
 
 
 class ChatTemplatePrompter(Prompter):
-    """prompter for HF chat templates"""
+    """Prompter for HF chat templates"""
 
     def __init__(
         self,
@@ -22,6 +24,8 @@ class ChatTemplatePrompter(Prompter):
         max_length=2048,
         message_field_role: str = "from",
         message_field_content: str = "value",
+        message_field_training: str = "train",
+        message_field_training_detail: str = "train_detail",
         roles: Optional[Dict[str, List[str]]] = None,
         drop_system_message: bool = False,
     ):
@@ -37,6 +41,8 @@ class ChatTemplatePrompter(Prompter):
             }
         self.message_field_role = message_field_role
         self.message_field_content = message_field_content
+        self.message_field_training = message_field_training
+        self.message_field_training_detail = message_field_training_detail
         self.tokenizer = tokenizer
         self.chat_template = chat_template
         self.max_length = max_length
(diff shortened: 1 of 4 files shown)


src/axolotl/prompt_strategies/chat_template.py
```diff
@@ -79,62 +201,169 @@ def messages(self, messages):
...
-            tokenizer,
-            chat_templates(chat_template),
-            message_field_role=message_field_role,
-            message_field_content=message_field_content,
-            roles=roles,
-            drop_system_message=drop_system_message,
+    prompter_params = {
+        "tokenizer": tokenizer,
+        "chat_template": chat_templates(ds_cfg.get("chat_template", "chatml")),
+        "message_field_role": ds_cfg.get("message_field_role", "from"),
```
winglian, line 345:
```suggestion
        "max_length": cfg.sequence_len,,
        "message_field_role": ds_cfg.get("message_field_role", "from"),
```
I realized I had originally forgotten to include this. without it, everything gets truncated to the default of 2k.

Tostino:
@winglian Just FYI, in the tests, I changed the model because `NousResearch/Meta-Llama-3-8B-Instruct` has bad tokenizer configurations.

Using that model will cause some of my tests to fail because of an incorrectly configured EOS token.

I mentioned it to them in Discord, so hopefully they fix it and tests can start working.

src/axolotl/prompt_strategies/chat_template.py
Tostino: So there has already been some confusion about this, because there was already the additional `max_length` config option added (which as you said, apparently wasn't passed in).

Should we just be using global sequence_len throughout, and get rid of the dataset configuration for max_length?

winglian: This is really to control the length of the outputs of the tokenizer. Here's something I have in mind that might be more flexible and give users more control over whether samples should be dropped or truncated if they are too long.

Screenshot 2024-07-23 at 10 48 58 PM

hammoudhasan:
@Tostino I tried this code it works alright however one thing to fix imo is the fact that no error was thrown out when I had the key:

message_field_training: training

but didn't have it in the JSON file. It still tokenized and all.

winglian (approved):
thanks @Tostino !

Tostino:
Just for reference later...
@hammoudhasan that is entirely intentional. That field is meant to override whatever other settings you already have. So it let's you set the row to train when it otherwise wouldn't, or not train when it otherwise would...e.g. a conversation where the assistant is learning to correct from errors it has made, and you don't want to train on the known errors. 

So it's not expected to be on every row of training data. Same with message_field_training_detail.

PR #1771, opened 2024-07-19
Fix untrained tokens

winglian:
mistral 12b has a bunch of untrained reserved tokens that are useful for using as chatml tokens, this will allow us to use them without having `inf` as the grad norm.


ehartford:
Weird
I'm training with no inf/nan, and I started before this went into main

PR #1779, opened 2024-07-23
fix sharegpt tool support

winglian:


hasan9090:
@winglian When trying to preprocess glaive sharegpt dataset I came across this issue as well and observed that the tool role parameter in the prompters and strategy would never get used. What is the reason for taking back and closing this commit, since I think it is very much needed? 
My actual issue though is that in /src/axolotl/prompt_strategies/__init__.py  it would always in line 28 after returning the func, run into "ModuleNotFoundError" which I could not debug. Any ideas what could be happening?

PR #1782, opened 2024-07-24
better handling of llama-3 tool role

winglian:
supersedes #1779

### DIFF
--- a/src/axolotl/prompters.py
+++ b/src/axolotl/prompters.py
@@ -350,9 +350,12 @@ class ShareGPTPrompter(Prompter):  # pylint: disable=too-few-public-methods
                         "Please help us by creating an Issue to add support for this conversation type."
                     )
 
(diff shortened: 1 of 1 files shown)


src/axolotl/prompters.py
```diff
@@ -350,9 +350,12 @@ def _build_result(self, source):
                         "Please help us by creating an Issue to add support for this conversation type."
                     )
 
-                role = CONVERSATION_ROLE_FORMAT[self._conversation.name].format(
-                    ROLE=from_role
-                )
+                if self._conversation.name in ["llama3"]:
+                    role = from_role
```
seungduk-yanolja, line 354: Why not update this instead?
```
"llama3": "{ROLE}",
```

winglian: @teknium1 said he tried changing that 
Screenshot 2024-07-23 at 11 10 53 PM

PR #1789, opened 2024-07-26
Update instruct-lora-8b.yml

monk1337:
Config is giving an error if not using the end of the token as the `pad_to_sequence_len` is true.


winglian (approved):
thanks!

Issue #1749, opened 2024-07-13
Potential max_input_len Issue/Inconsistency?


williambarberjr:
### Please check that this issue hasn't been reported before.

- [X] I searched previous [Bug Reports](https://github.com/axolotl-ai-cloud/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

I expect to see input/output that combines for a token length of less than 8192 after the chat template is applied to be retained in full as valid training data when examining the arrow files in the /last_run_prepared/ data after running `python -m axolotl.cli.preprocess instruct-lora-8b.yml`

### Current behaviour

Training data is being cut off at a max length of 2048.

### Steps to reproduce

My yml sets `sequence_len: 8192` but the logs keep printing `max_input_len` as having been set to 2048. Even when I alter the source code in `src/axolotl/utils/trainer.py` to hard code `max_input_len: 7192` and `def drop_long_seq(sample, sequence_len=2048, min_sequence_len=2)` to `def drop_long_seq(sample, sequence_len=7192, min_sequence_len=2)`, and the log print out confirms at least that `max_input_len` has been set to 7192 after rebuilding/reinstalling the axolotl, the training data still gets cut off at a length of 2048 tokens. The issue persistently occurs, when using these `datasets`/`chat_template` settings:
```
chat_template: llama3
datasets:
  - path: williambarberjr/L3_8B_Instruct_MarkdownToSummaryConvert
    type: chat_template
    chat_template: llama3
    field_messages: messages
    message_field_role: role
    message_field_content: content
    roles:
      user:
        - user
      assistant:
        - assistant
      system:
        - system
```

However, when I set my own custom chat template like this:
```
datasets:
  - path: ./cleanAxolotlTrainingDataL3_8B_Instruct.jsonl
    type: input_output
```

And have the jsonl data already prepped with all the correct beginning, ending etc. chat template tokens. It doesn't cut the length at a max of 2048. Here's how I'm printing out the prepared data to check if the template looks correct. First I run `python -m axolotl.cli.preprocess instruct-lora-8b.yml` in the command line then I run this python code:

```python
import json, yaml
from transformers import AutoTokenizer
from datasets import load_from_disk
import os

#find the yml file in the current directory
ymlFile = next(f for f in os.listdir('.') if f.endswith('.yml'))
with open(ymlFile, 'r') as f:
    cfg = yaml.safe_load(f)
model_id = cfg['base_model']
[...]

PR #1818, opened 2024-08-08
fix the incorrect `max_length` for chat template

chiwanpark:
# Description

The preprocessed data was being cut off at a maximum length of 2048 tokens, even when the user set `sequence_len` to a number greater than 2048. This PR fixes the bug by passing `sequence_len` from the configuration to `max_length` in `ChatTemplatePrompter`.

## Motivation and Context

There is an issue reporting the same cut off problem. (#1749)


winglian (approved):
thanks @chiwanpark !

PR #1821, opened 2024-08-09
update sklearn versrion, torch compile env vars, don't worry about failure on preprocess load model

winglian:
# Description

### DIFF
--- a/src/axolotl/utils/trainer.py
+++ b/src/axolotl/utils/trainer.py
@@ -390,6 +390,14 @@ def calculate_total_num_steps(cfg, train_dataset, update=True):
     return total_num_steps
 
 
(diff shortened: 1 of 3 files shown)


src/axolotl/utils/trainer.py
```diff
@@ -434,6 +442,9 @@ def prepare_optim_env(cfg):
             stage = deepspeed_config.get("zero_optimization", {}).get("stage", None)
         setup_deepspeed_env(cfg, stage=stage)
 
+    if cfg.torch_compile:
+        setup_torch_compile_env(cfg)
```
NanoCode012:
```suggestion
    setup_torch_compile_env(cfg)
```

There is already a condition check within the function. This outer one is not necessary

PR #1834, opened 2024-08-19
fix so inference can be run against quantized models without adapters

winglian:


src/axolotl/utils/config/models/input/v0_4_1/__init__.py
```diff
@@ -322,8 +322,10 @@ class LoraConfig(BaseModel):
...
     def validate_adapter(cls, data):
-        if not data.get("adapter") and (
-            data.get("load_in_8bit") or data.get("load_in_4bit")
+        if (
+            not data.get("adapter")
+            and not data.get("inference")
+            and (data.get("load_in_8bit") or data.get("load_in_4bit"))
         ):
             raise ValueError(
                 "load_in_8bit and load_in_4bit are not supported without setting an adapter."
```
NanoCode012:
```suggestion
                "load_in_8bit and load_in_4bit are not supported without setting an adapter for training."
```

PR #1845, opened 2024-08-21
fix: prompt phi

JohanWork:
# Description

The current phi templet is not following the phi template [here](https://huggingface.co/microsoft/Phi-3-mini-4k-instruct#chat-format). It is missing some <|end|> and som \n

## How has this been tested?

I have added test to the prompt template to make sure it is correct.


winglian (approved):
thanks @JohanWork !

PR #1855, opened 2024-08-22
Enable using the tokenizer default chat template of base model

levguy:
# Description

Added the option to use the tokenizer default chat template of the base model (the one loaded from tokenizer_config.json of the base model).
This was done by adding an entry to the [templates](https://github.com/axolotl-ai-cloud/axolotl/blob/b33dc07a7757819f68f6ceb4bbd2deb4c59c105b/src/axolotl/utils/chat_templates.py#L21) dict, where the key is `"tokenizer"` and the value is `None`. When `tokenizer.apply_chat_template` is called with `chat_template=None`, the default chat template is used, i.e. the one loaded from tokenizer_config.json.

## Motivation and Context

This update solves the following problems:
* Chat templates of only few models are defined in the [templates](https://github.com/axolotl-ai-cloud/axolotl/blob/b33dc07a7757819f68f6ceb4bbd2deb4c59c105b/src/axolotl/utils/chat_templates.py#L21) dict. Whenever we want to support an additional model, it is needed to update this dict.
* Models in the Phi-3 family, for example, have different chat templates. In the [templates](https://github.com/axolotl-ai-cloud/axolotl/blob/b33dc07a7757819f68f6ceb4bbd2deb4c59c105b/src/axolotl/utils/chat_templates.py#L21) dict, however, there is a single `phi_3` entry. Such cases may lead to confusion and errors.
* There are many cases where model owners modify the chat template of their model at Huggingface hub. In such cases, the corresponding chat template in the [templates](https://github.com/axolotl-ai-cloud/axolotl/blob/b33dc07a7757819f68f6ceb4bbd2deb4c59c105b/src/axolotl/utils/chat_templates.py#L21) dict becomes outdated, and it's hard to track that.

To summarize, the ability to use the default chat template of base model provides flexibility, prevents errors and saves future code updates.

## How has this been tested?

I tested SFT using the following dataset configuration:
```
datasets:
- path: cgato/SlimOrcaDedupCleaned
  type: chat_template
  chat_template: tokenizer
```

I tested DPO using the following dataset configuration:
```
chat_template: tokenizer
rl: dpo
datasets:
  - path: fozziethebeat/alpaca_messages_2k_dpo_test
    type: chat_template.default
    chat_template: tokenizer
    field_messages: conversation
    field_chosen: chosen
    field_rejected: rejected
    message_field_role: role
    roles:
      system:
        - system
      user:
        - user
      assistant:
        - assistant
```


winglian:
Thanks for opening this! What do you think about having the value be _tokenizer_ instead to indicate it is a special value?

chiragjn:
#1732

levguy:
> Thanks for opening this! What do you think about having the value be _tokenizer_ instead to indicate it is a special value?

Thank you for reviewing @winglian.
I chose the value of `None` because this way you seamlessly pass `chat_template=None` when calling `tokenizer.apply_chat_template`, which leads to using the default chat template as desirable.
I mean, what we currently do is:
```
    chat_template_str = chat_templates(cfg.chat_template)
    ...
    result = tokenizer.apply_chat_template(
        ...
        chat_template=chat_template_str,
        ...
    )
```
If `chat_templates` returns `None`, then we can keep the code above as is.
If `chat_templates` returns `"tokenizer"`, then we will have to add something like:
```
chat_template_str = None if chat_template_str == "tokenizer" else chat_template_str
```
in all functions that call `tokenizer.apply_chat_template`.

levguy:
@winglian would it be clearer if i change `"tokenizer"` to `"base_model"`?
As this option triggers the usage of the base model's chat template.

bursteratom:
This change makes sense to me. I like using `"base_model"` instead of `"tokenizer"`. Not only is it clearer, it also informs newcomers that if tokenizer receives `None` for `chat_template` it will default to `base_model` 's chat template.

bursteratom (changes requested):
Hi Guy, thank you for your input! It would definitely make more sense to name the key `base_model` instead of `tokenizer` when imbuing it with `None` value.

It also occurred to me that users some times might intend NOT to use any chat template at all and want to make clear of that in the configuration file. Perhaps we can as well slightly modify line 287 of 

> src/axolotl/utils/models.py

into the following: 
`if cfg.chat_template AND cfg.chat_template.lower() != "none":`

What do you think?

chiragjn:
I would like to consolidate changes from this PR and https://github.com/axolotl-ai-cloud/axolotl/pull/1732 to avoid conflicts since they are targetting same things 😅 

Seeking advice from @bursteratom and @NanoCode012

NanoCode012:
Hey, after reading the PR contents for this and https://github.com/axolotl-ai-cloud/axolotl/pull/1732 . I believe, the latter solves this issue (via `chat_template: tokenizer_default`) as well as supports more features ( custom jinja and fallback templates). Therefore, I think it would be better to focus on #1732. 

Thank you for lifting up this PR! I will close this for now, but if there are any changes, we can re-open.

levguy:
Okay guys, makes sense to consolidate this PR and #1732.
Thank you for reviewing! Looking forward to having #1732 merged :)
@chiragjn thanks for implementing a wider solution. It would be nice if you can adopt the suggestion discussed above and change `tokenizer_default` to `base_model` for better clarity.

NanoCode012:
>  It would be nice if you can adopt the suggestion discussed above and change `tokenizer_default` to `base_model` for better clarity.

Hm, this might be semantics, but, since it's possible (albeit rare) to specify a different tokenizer than the base model's within axolotl yaml, this naming change might be misunderstood. 

A use case would be to point to a similar tokenizer config with some custom changes than were not there in the base_model's repo. I think I found this helpful during llama2 where the tokenizer config in original repo was wrong at one point.

Should the naming be kept then?

PR #1857, opened 2024-08-22
drop valueerror as this was from when 4bit required gptq

winglian:


abedkhooli:
I still get: `ValueError: `.to` is not supported for `4-bit` or `8-bit` bitsandbytes models.`  
Error is triggered from `axolotl/cli/merge_lora.py` L48 to `axolotl/cli/__init__.py` L147  -> accelerate -> transformers/modeling_utils.py#L2861

ved1beta:
close as stale

PR #1861, opened 2024-08-23
Liger Kernel integration

winglian:


src/axolotl/integrations/liger/__init__.py
```diff
@@ -0,0 +1,104 @@
...
+            if cfg.liger_rope:
+                modeling_gemma.apply_rotary_pos_emb = liger_rotary_pos_emb
+            if cfg.liger_rms_norm:
+                modeling_gemma.GemmaRMSNorm = LigerRMSNorm
+            if cfg.liger_swiglu:
+                modeling_gemma.GemmaMLP = LigerGEGLUMLP
+            if cfg.liger_cross_entropy:
+                modeling_gemma.CrossEntropyLoss = LigerCrossEntropyLoss
+            if cfg.liger_fused_linear_cross_entropy:
+                logging.warning(
```
ByronHsu, line 86: Add a TODO?

ByronHsu (approved):
Looks awesome to me!!

PhilipMay:
Hi @winglian .
Thanks for adding this.

I have 2 questions:

1. would it be possible to add some documentation on how to activate and use this? On a top level?

2. I am wondering why you do not just provide code like this which would do some monkey patching:

```python
    if cfg.apply_liger_kernel_to_llama:
        from liger_kernel.transformers import apply_liger_kernel_to_llama
        apply_liger_kernel_to_llama()
```

Issue #1340, opened 2024-02-27
Kindly add DeepSeek family for training


ajinkya123-robo:
### ⚠️ Please check that this feature request hasn't been suggested before.

- [X] I searched previous [Ideas in Discussions](https://github.com/OpenAccess-AI-Collective/axolotl/discussions/categories/ideas) didn't find any similar feature requests.
- [X] I searched previous [Issues](https://github.com/OpenAccess-AI-Collective/axolotl/labels/enhancement) didn't find any similar feature requests.

### 🔖 Feature description

Axolotl for DeepSeek model finetuning. Thanks

### ✔️ Solution

Sft, Lora and qLora will suffice. It will be great to have their models in Axolotl training platform.

### Acknowledgements

- [X] My issue title is concise, descriptive, and in title casing.
- [X] I have searched the existing issues to make sure this feature has not been requested yet.
- [X] I have provided enough information for the maintainers to understand and evaluate this request.

NanoCode012:
Hey, as said here https://github.com/OpenAccess-AI-Collective/axolotl/discussions/1171, it's llama-based, so you can use the llama configs :)

ajinkya123-robo:
Very cool, Thank you!

ehartford:
It doesn't look so:

https://huggingface.co/deepseek-ai/deepseek-moe-16b-base/blob/main/config.json

```
{
--
  | "architectures": [
  | "DeepseekForCausalLM"
  | ],
  | "attention_bias": false,
  | "attention_dropout": 0.0,
  | "auto_map": {
  | "AutoConfig": "configuration_deepseek.DeepseekConfig",
  | "AutoModel": "modeling_deepseek.DeepseekModel",
  | "AutoModelForCausalLM": "modeling_deepseek.DeepseekForCausalLM"
  | },
  | "bos_token_id": 100000,
  | "eos_token_id": 100001,
```

If it's a Llama then it will say "architectures": ["LlamaForCausalLM"] right?

NanoCode012:
Oh, I wasn’t aware of that model. I thought they were referencing models such as https://huggingface.co/deepseek-ai/deepseek-coder-6.7b-instruct/tree/main which is llama based

ajinkya123-robo:
Thanks @ehartford for spotting it & reopening the issue. I hope Axolotl team will come up with the solution.

NanoCode012:
[...]

PR #1874, opened 2024-08-27
reapply multipack patches after model load for trust_remote_code compat

tmm1:
cc #1873 #1872 https://github.com/axolotl-ai-cloud/axolotl/issues/1340#issuecomment-2301160001

note that this was a regression in https://github.com/huggingface/transformers/pull/30370 after which transformers will reload modules causing our patched function to get replaced

### DIFF
--- a/src/axolotl/utils/models.py
+++ b/src/axolotl/utils/models.py
@@ -5,6 +5,7 @@ import gc
 import logging
 import math
 import os
+import sys
 import types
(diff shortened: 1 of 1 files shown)


tmm1:
i think a better solution might be fix transformers so it doesn't wipe away monkeypatches between model loads

otherwise we'll be chasing down and trying to restore a bunch of patches all over. for example, i ran into this again when trying to patch deepseekv2 for liger support

PR #1885, opened 2024-09-01
support for auto_find_batch_size when packing

winglian:
added support for

```yaml
auto_find_batch_size: true
```

simply crank up the micro_batch_size to the top of the search range. keep in mind that transformers has no support for changing the lr/gradient_accumulation_Steps to compensate for the different batch size.


winglian:
need https://github.com/huggingface/transformers/pull/33238 for this to really work for multigpu

winglian:
the loss for `multipack_real_batches` is off by 2x 😭 
EDIT: not an issue, fixed since FA requires flattening and we recompute dataloader len now

PR #1886, opened 2024-09-01
Fix RMSNorm monkey patch for Gemma models

chiwanpark:
# Description

Gemma 1/2 use the different offset and dtype casting method comparing to that of LLaMA. Liger Kernel already updated these differences (linkedin/Liger-Kernel#85, linkedin/Liger-Kernel#92) in 0.2.1, but axolotl uses still buggy implementation. This PR fixes the bug.


winglian (approved):
Thanks!

PR #1894, opened 2024-09-04
Fix documentation for pre-tokenized dataset

alpayariyak:
It's currently asking to not add BOS and EOS, stating that Axolotl adds them, but this is not true. Fixed documentation to instead ask to add BOS and EOS, as well as to emphasize making sure that EOS is trained on by not setting its label to -100.


NanoCode012 (commented):
I believe this correction is accurate. Axolotl should not do anything for bring-your-own-datasets.

@winglian , do you know if the process changed?

I followed the PR that introduced it and noticed that it mentioned that the information is from me. I believe there was some miscommunication on that end.

winglian (approved):
yeah, the updated docs should reflect the correct behavior, as we shouldn't call the tokenizer or any chat templating at this point.

PR #1915, opened 2024-09-13
Trigger the original tokenization behavior when no advanced turn settings are provided

fozziethebeat:
# Description

#1756 introduces some very advanced features for turn level masking which regresses on the simplified and default behavior for training on the last assistant turn in a conversation history.

This fixes that for the default simplified case

## Motivation and Context

This sets `None` for all the new per turn config features so that the default tokenization strategy prior to #1756 can be done as is without any complicated logic.  Now, the advanced logic is *only* triggered when any of the advanced features are introduced

## How has this been tested?

Added a new unittest to the prompt strategies chat template test suite that tests against Phi-3.5 (where the current behavior most clearly breaks things)

## Types of changes

Simplifying the chat template prompt strategy and undoing some (new) default values

## Social Handles (Optional)

- @fozziethebeat everywhere


fozziethebeat:
Pylint doesn't seem to like the unittest now that it's gotten bigger.  I tried splitting it into two separate smaller unittests for sanity sake but now pylint is also upset that some lines are duplicated.  Open to suggestions on test cleanliness preferences

winglian:
thanks @fozziethebeat, I moved the shared fixtures out, which should help with the linting.

fozziethebeat:
That test cleanup strategy works for me!

PR #1917, opened 2024-09-17
Grokfast support

winglian:
Adds support for @ironjr's grokfast to axolotl see. https://github.com/ironjr/grokfast


NanoCode012 (approved):
Looks good. I did small tests on single and multi gpu to check it ran fine.

PR #1919, opened 2024-09-18
shampoo optim support

winglian:


requirements.txt
```diff
@@ -35,6 +35,7 @@ python-dotenv==1.0.1
 autoawq>=0.2.5
 triton>=2.3.0
 liger-kernel==0.2.1
+distributed_shampoo @ git+https://github.com/facebookresearch/optimizers.git@main
```
NanoCode012, line 38: If we're aiming for pypi release, I think this will be tricky as we don't want to install from source. However, their last release was Feb.

src/axolotl/core/trainer_builder.py
```diff
@@ -465,6 +476,102 @@ def create_optimizer(self):
...
+                            param_to_metadata=compile_fsdp_parameter_metadata(
+                                self.model_wrapped
+                            )
+                        )
+                    else:
+                        distributed_config = DDPShampooConfig(
+                            communication_dtype=CommunicationDType.BF16,
+                            num_trainers_per_group=self.args.world_size,
+                            communicate_params=False,
+                        )
```
NanoCode012, line 554: Does this support deepspeed?

PR #1927, opened 2024-09-25
Fixing/Adding Mistral Templates

pandora-s-git:
# Description

This PR has come objective fixing the templates and making them match as closesly as possible the ground truth ([Mistral Common](https://github.com/mistralai/mistral-common)), a document delves into details this issue [here](https://github.com/pandora-s-git/cookbook/blob/main/concept-deep-dive/tokenization/chat_templates.md).

## Motivation and Context

This PR should make the templates match properly the corresponding tokenizer of mistral models, it does not tackle function calling but the basic instruction template. It is possible to use the hugging face templates if function calling is supported by Axolotl.
The previous template was not correct, and only matched partially the expected format, this format that will be used assumes that the BOS is not added by default, since im new to Axolotl feel free to confirm this is the proper way to do it!

## How has this been tested?

I did not test heavily since I am aware this templates are correct, the only possible issue would depend on how Axolotl implements it, and I may lack the insights confirm it will work out of the box and match perfectly, my main 2 concerns are:
 - does Axolotl add a BOS by default? If yes I can remove the BOS from the template.
 - does Axolotl also do generation/inference? If yes its important to note the model with this template will always start with a token with a leading white space for the V1, V2 and V3 tokenizers (not the Tekken).

The current templates will format it as such:

**V1**
```
<s> [INST] Hello [/INST] Hi there!</s> [INST] How are you? [/INST] I am good, thanks!</s>
```
**V2**
```
<s>[INST] Hello[/INST] Hi there!</s>[INST] How are you?[/INST] I am good, thanks!</s>
```
**V3-Tekken**
```
<s>[INST]Hello[/INST]Hi there!</s>[INST]How are you?[/INST]I am good, thanks!</s>
```

## Types of changes

Simple Jinja templates.


winglian (approved):
thanks @pandora-s-git !

guillermo-gabrielli-fer:
If I use inst it will fail with ValueError: Template 'inst' not found.
```
  File "axolotl/src/axolotl/utils/chat_templates.py", line 40, in chat_templates
    raise ValueError(f"Template '{user_choice}' not found.")
```
If I use the new ones it will fail with :
```
pydantic_core._pydantic_core.ValidationError: 1 validation error for AxolotlConfigWCapabilities
chat_template
  Input should be 'alpaca', 'chatml', 'inst', 'gemma', 'cohere', 'llama3', 'llama3_2_vision', 'phi_3', 'phi_35', 'deepseek_v2' or 'jamba' [type=enum, input_value='mistral_v2v3', input_type=str]
   
```

PR #1930, opened 2024-09-26
WIP Multimodal support

winglian:


src/axolotl/utils/models.py
```diff
@@ -298,10 +305,28 @@ def load_tokenizer(cfg):
         )
     return tokenizer
 
+def load_processor(cfg: DictDefault, tokenizer: PreTrainedTokenizerBase):
+    #model_config = load_model_config(cfg)
+    processor_kwargs = {} #do we actually need this?
+
+    processor_cls = AutoProcessor
+    if cfg.processor_type:
```
winglian, line 312: need to add `processor_type` to https://github.com/axolotl-ai-cloud/axolotl/blob/main/src/axolotl/utils/config/models/input/v0_4_1/__init__.py

src/axolotl/train.py
```diff
@@ -69,6 +69,9 @@ def train(
         main_process_only=True,
     )
     tokenizer = load_tokenizer(cfg)
+    processor = None
+    if cfg.is_multimodal:
```
winglian, line 75: also need to include `is_multimodal` arg to the config class

winglian, line 75:
```suggestion
    if cfg.is_multimodal or cfg.processor_type:
```

src/axolotl/utils/models.py
```diff
@@ -461,6 +491,14 @@ def load_model(
     max_memory = cfg.max_memory
     device_map = cfg.device_map
 
+    if cfg.is_multimodal:
+        if model_config.model_type == "llava":
+            Auto = LlavaForConditionalGeneration
+        elif model_config.model_type == "mllama":
+            Auto = MllamaForConditionalGeneration
```
winglian: we can use `AutoModelForVision2Seq` instead here.

bursteratom:
this is too outdated. Gonna start a new one.

PR #1939, opened 2024-10-02
Comet integration

Lothiraldan:
# Description

Hi, I'm Boris working on Integrations for Comet. This PR add the possibility to use the Comet Transformers integrations to log fine-tuning metrics and parameters.

## Motivation and Context

We got some Comet users that are using Axolotl for fine-tuning and they want to log as many data to Comet for ensuring their fine-tuning is reproducible.

## How has this been tested?

I've added a unit test for some logic and ran some basic example with it to validate that the data were correctly logged to Comet.

## Screenshots

Screenshot 2024-10-02 at 15-15-50 lothiraldan_comet-example-axolotl-llama2

## Types of changes

This add a new experiment tracking integration

## Social Handles

Here is our Twitter handle: https://x.com/Cometml


winglian (approved):
thanks @Lothiraldan !

winglian:
@Lothiraldan looks like the tests are failing since comet_ml isn't installed by default. Can you look at the mlflow integration and how we conditionally import and use functions like `is_mlflow_available` ?

Lothiraldan:
@winglian I think I have updated all of the code to make the Comet integration optional, thanks for the pointers to MLFlow. I have also updated the configuration page of the documentation, please let me know if there is another portion of the doc to update.

PR #1957, opened 2024-10-10
add warning that sharegpt will be deprecated

winglian:


src/axolotl/prompt_strategies/sharegpt.py
```diff
@@ -61,6 +61,9 @@ def build_loader(
     default_conversation: Optional[str] = None,
 ):
     def _load(tokenizer, cfg, ds_cfg: Optional[Dict[str, Any]] = None):
+        LOG.warning(
+            "sharegpt type support will be deprecated in the next release of Axolotl",
```
NanoCode012: Could we add more info to inform what they should migrate to?

We may also consider updating all yaml configs to stop using this (if any)

winglian: none of our examples actually use sharegpt

NanoCode012:
```suggestion
            "sharegpt type support will be deprecated in the next release of Axolotl. Please use chat_template instead.",
```

PR #1958, opened 2024-10-11
Handle image input as string paths for MMLMs

afrizalhasbi:
### Description

- Accept datasets with the `image` path as a string
- The original code would throw a `ValueError` because the trainer processor expects an image directly
- Images are loaded online during training in batches

### Motivation and Context

- Some datasets store the image rows as string paths so it can be handled easier and faster

### How has this been tested?

- `ValueError` is gone and training runs smoothly


winglian:
thanks! One thing is this requires the images to be downloaded to a local path, right? https://pillow.readthedocs.io/en/stable/reference/Image.html#PIL.Image.open

winglian (approved):

afrizalhasbi:
Thanks for the approval!

Yeah, this only works for local paths.

PR #1963, opened 2024-10-11
only install torchao for torch versions >= 2.4.0

winglian:


winglian:
pending https://github.com/axolotl-ai-cloud/axolotl/actions/runs/11305510026

PR #1970, opened 2024-10-14
Feat: Add support for tokenizer’s or custom jinja chat_template

NanoCode012:
Continues #1732 due to no write access to PR (with prior author consent)

# Description

## Social Handles (Optional)

@chiragjn


src/axolotl/prompt_strategies/dpo/chat_template.py
```diff
@@ -53,7 +53,7 @@ def transform_fn(sample, tokenizer=None):
             "role": role_map[sample[field_rejected][field_message_role]],
             "content": sample[field_rejected][field_message_content],
         }
-        dummy_user_message = {"role": "user", "content": "dummy"}
+        dummy_user_message = {"role": "user", "content": "[[dummy_message]]"}
```
winglian, line 56: @NanoCode012  why does this need to be a string with double "[[ ... ]]" ?

NanoCode012: I chose this arbitrary symbol following suggestions from levy from the earlier PR.

It's just to prevent any matching with the line below (`result["chosen"].find(chosen["content"])` on the off chance someone actually has a chat saying "dummy").

winglian: why do we need a dummy user message here? shouldn't this be from the DPO dataset?

NanoCode012: There is an edge case where if the `chat_template` asserts that the first message must be the `user` like in Phi3's case, the line below would fail as the *first* role would be `assistant`. We add the dummy user message, so that the chat template would run fine. Afterwards, the dummy message would be stripped when `find` is used.

```py
        result["chosen"] = tokenizer.apply_chat_template(


             [chosen], # without the dummy user message
             add_generation_prompt=False,
             chat_template=chat_template_string,
             tokenize=False,
         )

        chosen_strip_index = result["chosen"].find(chosen["content"])
```

I added a test to also check for this here: https://github.com/axolotl-ai-cloud/axolotl/blob/207e7627f9d5e00dde5257d4f09ce1d81c1ef2d2/tests/prompt_strategies/test_dpo_chat_templates.py#L169-L197

If you were to remove the dummy message, the test would fail.

Ref: https://github.com/axolotl-ai-cloud/axolotl/pull/1732#issuecomment-2408981853

winglian (approved):

NanoCode012:
Will do another CI check once base images are updated to ensure it doesn't break anything

NanoCode012:
New CI passed https://github.com/axolotl-ai-cloud/axolotl/actions/runs/11433735390. Adding `ready-to-merge` tag.

chiragjn:
Bumping this up @winglian 😅

Issue #1938, opened 2024-10-02
ML Flow Checkpointing 


wanderingweights:
### Please check that this issue hasn't been reported before.

- [X] I searched previous [Bug Reports](https://github.com/axolotl-ai-cloud/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

I have the below config and I was hoping to have the model checkpoints saving as artifacts but I only get the metrics and config saving. 

Is this expected to work or am I missing something? 

Thanks for looking over! 

### Current behaviour

No model checkpoints. 

### Steps to reproduce

I use the docker image:

"winglian/axolotl:main-latest"

hf login 

then 

accelerate launch -m axolotl.cli.train theconfig.yml

### Config yaml

```yaml
base_model: NousResearch/Meta-Llama-3-8B-Instruct
model_type: LlamaForCausalLM
tokenizer_type: AutoTokenizer

load_in_8bit: true
load_in_4bit: false
strict: false

chat_template: llama3
datasets:
  - path: base.yml
    type: alpaca
    ds_type: json

dataset_prepared_path:
val_set_size: 0.05
output_dir: attempt_1

sequence_len: 4096
sample_packing: false
pad_to_sequence_len: true

adapter: lora
lora_model_dir:
lora_r: 32
lora_alpha: 16
lora_dropout: 0.05
lora_target_linear: true
lora_fan_in_fan_out:

mlflow_tracking_uri: ml_flow_url
mlflow_experiment_name: test_1
hf_mlflow_log_artifacts: true

gradient_accumulation_steps: 4
micro_batch_size: 2
num_epochs: 1
optimizer: adamw_bnb_8bit
lr_scheduler: cosine
learning_rate: 0.0002

train_on_inputs: false
group_by_length: false
bf16: auto
fp16:
tf32: false

gradient_checkpointing: true
early_stopping_patience:
resume_from_checkpoint:
local_rank:
logging_steps: 1
xformers_attention:
flash_attention: true
s2_attention:

hub_model_id: some_repo
hub_strategy: checkpoint
hub_token: 

  
save_steps: 5
save_strategy: steps
warmup_steps: 10
evals_per_epoch: 4
eval_table_size:
eval_max_new_tokens: 128
max_steps: 10
debug:
deepspeed:
weight_decay: 0.0
fsdp:
fsdp_config:
special_tokens:
   pad_token: " "
```

### Which Operating Systems are you using?

- [X] Linux
- [ ] macOS
[...]

PR #1976, opened 2024-10-17
Log checkpoints as mlflow artifacts

awhazell:
# Description

Currently checkpoints/models aren't logged as artifacts by mlflow even if the `hf_mlflow_log_artifacts` config var is set to true.

This PR ensures that the config var is propagated to MLflow as an env variable, and also adds the transformers MLflowCallback when MLflow is in use
## Motivation and Context

It's easier to organise your results if they're logged to mlflow as artifacts- it also automates syncing them with cloud storage.

https://github.com/axolotl-ai-cloud/axolotl/issues/1938

## How has this been tested?

Added a couple of MLflow tests to `test_validation.py`, also tested on a couple of example configs and can confirm each save was logged as an mlflow artifact

### DIFF
--- a/src/axolotl/utils/mlflow_.py
+++ b/src/axolotl/utils/mlflow_.py
@@ -16,3 +16,7 @@ def setup_mlflow_env_vars(cfg: DictDefault):
     # Enable mlflow if experiment name is present
     if cfg.mlflow_experiment_name and len(cfg.mlflow_experiment_name) > 0:
         cfg.use_mlflow = True
+
+    # Enable logging hf artifacts in mlflow if value is truthy
+    if cfg.hf_mlflow_log_artifacts is True:
+        os.environ["HF_MLFLOW_LOG_ARTIFACTS"] = "true"
--- a/src/axolotl/core/trainer_builder.py
+++ b/src/axolotl/core/trainer_builder.py
@@ -1119,12 +1119,17 @@ class TrainerBuilderBase(abc.ABC):
                 SaveAxolotlConfigtoWandBCallback(self.cfg.axolotl_config_path)
             )
         if self.cfg.use_mlflow and is_mlflow_available():
+            from transformers.integrations.integration_utils import MLflowCallback
+
             from axolotl.utils.callbacks.mlflow_ import (
                 SaveAxolotlConfigtoMlflowCallback,
             )
 
-            callbacks.append(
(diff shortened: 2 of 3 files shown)


src/axolotl/utils/mlflow_.py
```diff
@@ -16,3 +16,7 @@ def setup_mlflow_env_vars(cfg: DictDefault):
     # Enable mlflow if experiment name is present
     if cfg.mlflow_experiment_name and len(cfg.mlflow_experiment_name) > 0:
         cfg.use_mlflow = True
+
+    # Enable logging hf artifacts in mlflow if value is truthy
+    if cfg.hf_mlflow_log_artifacts is True:
+        os.environ["HF_MLFLOW_LOG_ARTIFACTS"] = "true"
```
awhazell, line 22: The variable wasn't being set in the code above before because it's parsed as a bool- I saw similar code to this elsewhere but if there's a preferable way to set the env var please lmk

winglian (approved):

src/axolotl/utils/mlflow_.py
winglian: I assume this means that mlflow prefers to pick up this from the env var rather than being passed the setting somewhere else more explicitly?

NanoCode012: @winglian , yes, this callback depends on env: https://huggingface.co/docs/transformers/v4.45.2/en/main_classes/callback#transformers.integrations.MLflowCallback .

NanoCode012:
Would you be able to do a few tests run to ensure
1) Trains successfully
2) Checkpoints are logged on_save (and end)
3) There is no duplicate logs for each time step?

awhazell:
Thanks both for the review, I've been using this branch and am seeing that

1. The models are training successfully as before
2. Checkpoints are logged according to the save setting in the config (e.g. `saves_per_epoch = 2` and `num_epochs = 2` is logging 4 models, the last one being the end model)
3. Less sure about this one but I'm not seeing any duplicate logs, the additional logging from the `MLflowCallback` is specific to it's actions- so the logs added in this PR are things like the [location of the logged artifacts](https://github.com/huggingface/transformers/blob/53fad641cfdb5105e2470bcf3ef17ea8e25cc300/src/transformers/integrations/integration_utils.py#L1322).

NanoCode012 (approved):
@awhazell , thanks for confirming those!

For point 3, I had an incorrect assumption. Never mind about that.

PR #1978, opened 2024-10-17
add soap optimizer support

winglian:
# Description


winglian:
I believe there are some edge cases where the training hangs on the precondition frequency step

PR #1979, opened 2024-10-18
add pytorch 2.5.0 base images

winglian:
# Description

### DIFF
--- a/src/axolotl/cli/__init__.py
+++ b/src/axolotl/cli/__init__.py
@@ -466,7 +466,7 @@ def load_datasets(
         cli_args.debug
         or cfg.debug
         or cli_args.debug_text_only
-        or cli_args.debug_num_examples
+        or int(cli_args.debug_num_examples) > 0
     ):
         LOG.info("check_dataset_labels...")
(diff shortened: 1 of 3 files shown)


winglian:
let's make sure https://github.com/axolotl-ai-cloud/axolotl/actions/runs/11398189114 passes

EDIT: looks like these all based.

src/axolotl/cli/__init__.py
```diff
@@ -466,7 +466,7 @@ def load_datasets(
         cli_args.debug
         or cfg.debug
         or cli_args.debug_text_only
-        or cli_args.debug_num_examples
+        or int(cli_args.debug_num_examples) > 0
```
NanoCode012, line 469:
```suggestion
        or (cli_args.debug_num_examples and int(cli_args.debug_num_examples) > 0)
```

Do you perhaps need to adjust to check for None/str?

winglian: it will always get cast to an int. ` debug_num_examples: int = field(default=0)`

PR #1980, opened 2024-10-18
Fix: Gradient Accumulation issue

NanoCode012:
Tests the upstream changes in `transformers` for GA fix and compare results. 

# Description

TODO:

- [ ] Check if Mamba Trainer needs to update loss
- [ ] Check if Orpo Trainer needs to update loss

Wait for:

- [x] New transformers release

## How has this been tested?

Testing and comparing loss in wandb.

### DIFF
--- a/src/axolotl/core/trainer_builder.py
+++ b/src/axolotl/core/trainer_builder.py
@@ -666,7 +666,9 @@ class AxolotlTrainer(SchedulerMixin, Trainer):
         return DataLoader(bench_dataset, **dataloader_params)
         # return self.accelerator.prepare(DataLoader(bench_dataset, **dataloader_params))
 
-    def compute_loss(self, model, inputs, return_outputs=False):
+    def compute_loss(
+        self, model, inputs, return_outputs=False, num_items_in_batch=None
+    ):
         # use one's weighted cross entropy loss calc
         # if self.args.sample_packing:
         #     labels = inputs.pop("labels")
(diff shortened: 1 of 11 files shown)


winglian:
also ran a manual check against ~https://github.com/axolotl-ai-cloud/axolotl/actions/runs/11418424220~
 https://github.com/axolotl-ai-cloud/axolotl/actions/runs/11443067437

winglian:
@NanoCode012 I think we need to do some test runs and also verify the unsloth cross entropy loss since we had to swap that

winglian:
I upgraded trl to the latest versions too for this, but seems like there are some changes that are breaking DPO

src/axolotl/core/trainer_builder.py
```diff
@@ -1910,7 +1924,7 @@ def build(self, total_num_steps):
             dpo_trainer_kwargs["max_length"] = self.cfg.sequence_len
             dpo_trainer_kwargs["max_target_length"] = None
             dpo_trainer_kwargs["max_prompt_length"] = self.cfg.sequence_len
-            dpo_trainer_kwargs["generate_during_eval"] = True
+            dpo_trainer_kwargs["generate_during_eval"] = self.cfg.use_wandb
```
NanoCode012, line 1958: Is there a reason we specify to wandb only and not the others as well?

winglian:
@NanoCode012 something isn't quite right with this PR, the loss for 8b with sample packing starts @ ~10, even without packing, the loss is ~10 for 8B

NanoCode012:
I'll do some tests and report back. I'll move this to Draft in meantime.

winglian:
@NanoCode012 upgrading to the transformers release that was pushed out today fixes the issue. I added a test too for a tiny model that also validates the loss

winglian:
waiting on mulitgpu tests too here https://github.com/axolotl-ai-cloud/axolotl/actions/runs/11520374143

PR #1982, opened 2024-10-18
first pass at pytorch 2.5.0 support

winglian:


winglian:
multigpu tests also running here https://github.com/axolotl-ai-cloud/axolotl/actions/runs/11408412866/job/31746518624

setup.py
```diff
@@ -50,7 +50,9 @@ def parse_requirements():
             else:
                 raise ValueError("Invalid version format")
 
-            if (major, minor) >= (2, 4):
+            if (major, minor) >= (2, 5):
+                _install_requires.pop(_install_requires.index(xformers_version))
```
NanoCode012, line 54: Would it be better to just remove and make it an optional dependency like mamba?

winglian: making it an optional dependency makes it harder to change the version on the fly though. unless we want to do something crazy like https://github.com/unslothai/unsloth/blob/main/pyproject.toml#L53-L103

PR #1995, opened 2024-10-25
feat: add Exaone3 chat_template

shing100:
# Description
Added exaone template to support new upcoming exaone models


chiwanpark:
IMHO, there is a patch to load the chat template from model (#1970); thus, we do not need to add the chat template for a specific model.

shing100:
It's very good.
But is it possible to apply Exaone template to the llama base model without this PR?

NanoCode012:
@shing100 , yep! You can see this example: https://github.com/axolotl-ai-cloud/axolotl/blob/28e7e444ee673d15057f2b2838d835fdde0e9c60/docs/dataset-formats/conversation.qmd#L150-L160

This way, we don't need to manually merge/add every model's template.

We're currently trying to merge a few PRs soon, and I hope we can get that one in too!

shing100:
@NanoCode012 Thanks!

NanoCode012:
@shing100 , I merged #1970 . Please give it a try! I will close this PR for now.

winglian (approved):
thanks @shing100 !

shing100:
@NanoCode012 @winglian I just saw the message. I reset it again. Thank you!

PR #1999, opened 2024-10-28
Fix: modelloader handling of model_kwargs load_in*bit

NanoCode012:
# Description

I missed one important case where `model_kwargs["load_in_8bit"]` and `model_kwargs["load_in_4bit"]` are deleted at the end of `self.set_quantization_config()` which would break all further dependency on the above kwarg. This PR fixes that and refactors the handling.


MengqingCao:
This looks much better than checking if the key-value pair exists before each check. :-)

NanoCode012:
To validate the new e2e test, I ran it on current `main` which threw an error.

`main`
image

`this branch`
image

winglian (approved):

PR #2003, opened 2024-10-30
Fix get_chat_template call for trainer builder

chiragjn:
# Description

Code crashes when choosing `chat_template: tokenizer_default`
We missed a few places in previous PRs

## Social Handles (Optional)

@NanoCode012

### DIFF
--- a/src/axolotl/cli/__init__.py
+++ b/src/axolotl/cli/__init__.py
@@ -272,7 +272,7 @@ def do_inference_gradio(
             importlib.import_module("axolotl.prompters"), prompter
         )
     elif cfg.chat_template:
-        chat_template_str = get_chat_template(cfg.chat_template)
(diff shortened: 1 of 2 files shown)


src/axolotl/cli/__init__.py
```diff
@@ -272,7 +272,7 @@ def do_inference_gradio(
             importlib.import_module("axolotl.prompters"), prompter
         )
     elif cfg.chat_template:
-        chat_template_str = get_chat_template(cfg.chat_template)
+        chat_template_str = get_chat_template(cfg.chat_template, tokenizer=tokenizer)
```
NanoCode012, line 275: Nice catch. While testing this, I got this error but figured it was due to my tokenizer chosen not having a template in its config.

NanoCode012 (approved):

PR #2006, opened 2024-10-30
Add plugin manager's callback hooks to training flow

chiragjn:
# Description

Thanks a lot for creating the plugins system! It has really helped me remove a lot of monkey patching to add my custom callbacks to the Trainer. Now that it exists, all this PR is doing is invoking the callback adding hooks during the training flow allowing plugins to add their own callbacks

I have also made the plugins attribute a OrderedDict so that a plugin is only registered once no matter how many times `.register(...)` is called, preventing errors

## How has this been tested?

Using it to run my own plugin for in house Experiment Tracking framework (E.g. https://github.com/truefoundry/llm-finetune/blob/cj_axolotl_update_oct2024/plugins/axolotl_truefoundry/axolotl_truefoundry/__init__.py)

## Future Work

Add other methods of the plugin manager in the flow


src/axolotl/integrations/base.py
```diff
@@ -237,7 +244,7 @@ def __new__(cls):
         """
         if cls._instance is None:
             cls._instance = super(PluginManager, cls).__new__(cls)
-            cls._instance.plugins: List[BasePlugin] = []
+            cls._instance.plugins = collections.OrderedDict()
```
winglian, line 247: I'm curious about why this change is needed. thanks!

chiragjn: So basically, in my flow `.register` was being called more than once, that lead to plugins being registered twice which fails `merge_args` dynamic code eval with duplicate class in inheritance chain

E.g. generated code

```
class AxolotlInputConfig(AxolotlInputConfigBase, MyPlugin, MyPlugin): pass
```

Making it dict, prevents the plugins from duplicating

winglian (approved):

src/axolotl/integrations/base.py
```diff
@@ -277,7 +284,7 @@ def get_input_args(self):
         list[str]: A list of Pydantic classes for all registered plugins' input arguments.'
         """
         input_args = []
-        for plugin in self.plugins:
+        for _, plugin in self.plugins.items():
```
NanoCode012: Just curious, why not `.values()`?

chiragjn: Let me do this then, initially when I was writing code I was doing some login with the plugin names for debugging, hence the `.items()` stuck around 😅

PR #2016, opened 2024-11-05
Add weighted optimisation support for trl DPO trainer integration

bursteratom:
# Description

Added support for DPO weighted optimisation in [trl-v0.12.0](https://github.com/huggingface/trl/releases/tag/v0.12.0) 

To enable this option, simply add the following line in your configuration yaml file.

```
dpo_use_weighting: true
```

## How has this been tested?

pytest tests/e2e.test_dpo.py

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -43,8 +43,7 @@ s3fs>=2024.5.0
 gcsfs>=2024.5.0
 # adlfs
 
-trl @ git+https://github.com/huggingface/trl.git@31d02cfb795284591a084416b9dcb7bef5d08924
-# trl==0.12.0
+trl==0.12.0
 zstandard==0.22.0
 fastcore
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -43,7 +43,8 @@ s3fs>=2024.5.0
 gcsfs>=2024.5.0
 # adlfs
 
-trl @ git+https://github.com/huggingface/trl.git@31d02cfb795284591a084416b9dcb7bef5d08924
+# trl @ git+https://github.com/huggingface/trl.git@31d02cfb795284591a084416b9dcb7bef5d08924
+trl==0.12.0
 zstandard==0.22.0
 fastcore
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -44,7 +44,7 @@ gcsfs>=2024.5.0
 # adlfs
 
 # trl @ git+https://github.com/huggingface/trl.git@31d02cfb795284591a084416b9dcb7bef5d08924
-trl==0.12.0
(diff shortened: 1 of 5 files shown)


src/axolotl/core/trainer_builder.py
```diff
@@ -1890,17 +1890,18 @@ def build_training_arguments(self, total_num_steps):
...
 
+        training_args_kwargs["dataset_num_proc"] = self.cfg.dataset_processes
+
         if self.cfg.rl_beta:
             training_args_kwargs["beta"] = self.cfg.rl_beta
         if self.cfg.orpo_alpha:
             # trl does some odd mapping of alpha to beta to reuse the beta parameter ???
             training_args_kwargs["beta"] = self.cfg.orpo_alpha
 
-        training_args_kwargs["dataset_num_proc"] = self.cfg.dataset_processes
```
winglian, line 1899: is this no longer used?

requirements.txt
```diff
@@ -44,6 +44,7 @@ gcsfs>=2024.5.0
 # adlfs
 
 trl @ git+https://github.com/huggingface/trl.git@31d02cfb795284591a084416b9dcb7bef5d08924
+# trl==0.12.0
```
winglian: doesn't seem to actually be using `trl==0.12.0`

winglian (changes requested):

src/axolotl/core/trainer_builder.py
bursteratom: `dataset_num_proc` is still in use. I just moved it from line 1899 to line 1893

requirements.txt
bursteratom: I tested `trl==0.12.0` and 
`trl @ git+https://github.com/huggingface/trl.git@31d02cfb795284591a084416b9dcb7bef5d08924` 

and both work cuz the latter is also on the 0.12.0 code base (pip lists it as `0.12.0.dev0`) 

We know it works because when tested with an older version `trl==0.11.4` you get the following error
`TypeError: AxolotlDPOConfig.__init__() got an unexpected keyword argument 'use_weighting'` Since `AxolotlDPOConfig` inherits directly from `trl`'s `DPOConfig`, we know `use_weighting` has been passed into `trl` `DPOConfig` correctly.

bursteratom: I think I simply forgot to revert my comment post testing, just fixed it. But yeah `trl==0.12.0` is correctly integrated.

bursteratom:
@winglian I've added the stable version of trl in the `requirements.txt`. Let me know if this PR is good

winglian (approved):

PR #2018, opened 2024-11-07
fix(vlm): handle legacy conversation data format and check image in data

NanoCode012:
# Description

The current vision data processing expects in the OAI format and with images. This PR allows passing text-only dataset to training vision models and converts the old sharegpt-like format datasets.

The caveat is that, it still does not allow mixing text-only and text+image data points per batch yet. We would need to patch upstream for this.

## Motivation and Context

Could not pass normal sharegpt dataset.

## How has this been tested?

Not yet after refactor.

## Social Handles (Optional)

@hjc-puro

### DIFF
--- a/src/axolotl/utils/collators/mm_chat.py
+++ b/src/axolotl/utils/collators/mm_chat.py
@@ -4,7 +4,7 @@ Collators for multi-modal chat messages and packing
 
 from copy import deepcopy
 from dataclasses import dataclass
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Optional, Union
 
 from PIL import Image
 from transformers import PreTrainedTokenizerBase, ProcessorMixin
@@ -32,8 +32,8 @@ class MultiModalChatDataCollator(DataCollatorMixin):
             raise ValueError("Packing is currently not supported.")
 
     def torch_call(
-        self, examples: List[Union[List[int], Any, Dict[str, Any]]]
-    ) -> Dict[str, Any]:
+        self, examples: list[Union[list[int], Any, dict[str, Any]]]
+    ) -> dict[str, Any]:
         # Handle dict or lists with proper padding and conversion to tensor.
 
         return self.__class__.process_rows(
@@ -48,7 +48,7 @@ class MultiModalChatDataCollator(DataCollatorMixin):
         # *** This is COPIED from the trl example sft_vlm.py code ***
         # use this as a starting point
 
-        def _preprocess(examples: list[dict]) -> dict:
+        def _preprocess(examples: list[dict]) -> list[dict]:
             """
             Preprocess conversation examples to ensure consistent format.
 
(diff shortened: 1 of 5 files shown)


winglian:
would love to have a unit test for this in place.

tests/e2e/test_llama_vision.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+    Test case for Llama Vision models
+    """
+
+    @with_temp_dir
+    def test_llama_vision_text_only_dataset(self, temp_dir):
+        # pylint: disable=duplicate-code
+        cfg = DictDefault(
+            {
+                "base_model": "alpindale/Llama-3.2-11B-Vision-Instruct",
```
winglian: is there a smaller model we can use?

tests/e2e/test_llama_vision.py
```diff
@@ -0,0 +1,111 @@
...
+                "lora_target_modules": r"language_model.model.layers.[\d]+.(mlp|cross_attn|self_attn).(up|down|gate|q|k|v|o)_proj",
+                "val_set_size": 0,
+                "chat_template": "llama3_2_vision",
+                "datasets": [
+                    {
+                        "path": "LDJnr/Puffin",
+                        "type": "chat_template",
+                    },
+                ],
+                "num_epochs": 1,
```
winglian, line 54: let's also add `max_steps: 5` since this is likely to timeout the tests with such a large model

tests/e2e/test_llama_vision.py
```diff
@@ -0,0 +1,111 @@
...
+
+        train(cfg=cfg, cli_args=cli_args, dataset_meta=dataset_meta)
+        assert (Path(temp_dir) / "model-00001-of-00005.safetensors").exists()
+
+    @with_temp_dir
+    def test_llama_vision_multimodal_dataset(self, temp_dir):
+        # pylint: disable=duplicate-code
+        cfg = DictDefault(
+            {
+                "base_model": "alpindale/Llama-3.2-11B-Vision-Instruct",
```
winglian: same as above, smaller vlm model available?

tests/e2e/test_llama_vision.py
```diff
@@ -0,0 +1,111 @@
...
+                "chat_template": "llama3_2_vision",
+                "datasets": [
+                    {
+                        "path": "HuggingFaceH4/llava-instruct-mix-vsft",
+                        "type": "chat_template",
+                        "split": "train[:1%]",
+                        "field_messages": "messages",
+                    },
+                ],
+                "num_epochs": 1,
```
winglian, line 99: same, max_steps: 5

src/axolotl/utils/collators/mm_chat.py
```diff
@@ -46,22 +48,129 @@ def process_rows(examples, processor, chat_template, max_images, length_only=Fal
...
+                dict in OpenAI format with 'messages' key
+
+            Raises:
+                ValueError: If the conversation format is not supported
+            """
+            role_mapping = {
+                "human": "user",
+                "gpt": "assistant",
+            }
+
```
winglian, line 73: what about data that's already in user/assistant? won't this mapping fail with KeyError?

NanoCode012: Ah no, we have the following function which fallbacks to the current role if that role is not in dict L74-76.

```py
            def normalize_role(role: str) -> str:
                """Normalize role names to OpenAI format. Default to original role if not found."""
                return role_mapping.get(role, role)
```

winglian (approved):

PR #2028, opened 2024-11-10
move deprecated kwargs from trainer to trainingargs

winglian:
# Description

This PR should help resolve some of the upstream changes in main that will break tests when we upgrade.


winglian:
additional tests against upstream main here ~https://github.com/axolotl-ai-cloud/axolotl/actions/runs/11761420503~ https://github.com/axolotl-ai-cloud/axolotl/actions/runs/11764652384

PR #2029, opened 2024-11-11
Add example YAML file for training Mistral using DPO

olivermolenschot:
# Description

I provided a YAML file that can be used as an example to perform DPO on a mistral model, using open source datasets and open source models. 

## Motivation and Context

This is required as the chat template used originally in a Mistral 7B model does not allow for us to regularly perform DPO on the model. The original chat template enforces iterating between user and assistant. However in the chosen and rejected columns of DPO, we only have an assistant turn. For reference, the original chat template can be found [in this file](https://huggingface.co/mistralai/Mistral-7B-Instruct-v0.2/blob/main/tokenizer_config.json). I gave an example where I use the chatml template instead and introduced the respective special tokens, which allows for training.

## How has this been tested?

I ran the yaml file locally and trained the model. I merged the model and then tested if the model produced outputs and was successfully using the new special tokens. The whole testing environment is a regular conda environment with axolotl and its dependencies. My changes do not affect other areas of the code since it is a yaml file.

## Types of changes

None


examples/mistral/mistral-dpo.yml
```diff
@@ -0,0 +1,103 @@
...
+saves_per_epoch: 1
+debug:
+deepspeed:
+weight_decay: 0.0
+fsdp:
+fsdp_config:
+special_tokens:
+  bos_token: "<|im_start|>"
+  eos_token: "<|im_end|>"
+device_map: "cuda:0"
```
NanoCode012:
```suggestion
```

examples/mistral/mistral-dpo.yml
```diff
@@ -0,0 +1,103 @@
+#Note that we are switching from the regular chat template to chatml.
+#If you experience problems with the special tokens, training for more epochs can help.
+#After trainig, merge the model before inference otherwise you might
```
NanoCode012:
```suggestion
#After training, merge the model before inference otherwise you might
```

examples/mistral/mistral-dpo.yml
```diff
@@ -0,0 +1,103 @@
...
+
+dataset_prepared_path:
+val_set_size: 0.05
+output_dir: ./outputs/dpo-lora-out-5
+
+sequence_len: 64
+sample_packing: false
+pad_to_sequence_len: true
+
+adapter: lora
```
NanoCode012: I notice you used `load_in_4bit: True` but did `adapter: lora`. Did you perhaps mean `qlora` or to use `load_in_8bit: True`?

examples/mistral/mistral-dpo.yml
```diff
@@ -0,0 +1,103 @@
...
+      user:
+        - user
+      assistant:
+        - assistant
+
+dataset_prepared_path:
+val_set_size: 0.05
+output_dir: ./outputs/dpo-lora-out-5
+
+sequence_len: 64
```
NanoCode012: This is quite a low default? I suggest a bit higher, but feel free to adjust.
```suggestion
sequence_len: 2048
```

examples/mistral/mistral-dpo.yml
```diff
@@ -0,0 +1,103 @@
...
+      system:
+        - system
+      user:
+        - user
+      assistant:
+        - assistant
+
+dataset_prepared_path:
+val_set_size: 0.05
+output_dir: ./outputs/dpo-lora-out-5
```
NanoCode012:
```suggestion
output_dir: ./outputs/dpo-lora
```

examples/mistral/mistral-dpo.yml
```diff
@@ -0,0 +1,103 @@
...
+    field_rejected: rejected
+    message_field_role: role
+    message_field_content: content
+    roles:
+      system:
+        - system
+      user:
+        - user
+      assistant:
+        - assistant
```
NanoCode012:
```suggestion
    roles:
      system:
        - system
      user:
        - user
      assistant:
        - assistant
```

I don't think this is necessary here. If it doesn't work without this, please let me know!

examples/mistral/mistral-dpo.yml
```diff
@@ -0,0 +1,103 @@
...
+micro_batch_size: 16
+num_epochs: 6
+optimizer: adamw_bnb_8bit
+lr_scheduler: cosine
+learning_rate: 0.0001
+
+train_on_inputs: false
+group_by_length: false
+bf16: false
+fp16: true
```
NanoCode012:
```suggestion
bf16: auto
fp16:
```

We recommend this to allow proper fallback.

NanoCode012 (changes requested):
Hey, thanks for the PR. I listed a few small points in this yaml that stood out

examples/mistral/mistral-dpo.yml
olivermolenschot: Yeah it is very low. I did this mainly to speed up training as my GPU only has 12GB of memory.

examples/mistral/mistral-dpo.yml
olivermolenschot: Do you mean to remove the roles? Because your suggestion both removed and added the same thing. I am guessing it is a typo.

examples/mistral/mistral-dpo.yml
olivermolenschot: I meant to minimize the model size for testing. But I can push the example as a `qlora` and `load_in_8bit: True` so that it can be easier for people using this example as later on.

olivermolenschot: Actually is the `load_in_8bit` for the model or the lora adapter? and if it is for the lora adapter how is setting `qlora `achieving anything different?

examples/mistral/mistral-dpo.yml
olivermolenschot: I removed the roles and the model still trains. So I made the changes in the file and committed.

examples/mistral/mistral-dpo.yml
NanoCode012: Hey! We would use `load_in_4bit: True` with `adapter: qlora` and 8bit with lora. 

This is for the model quantized, not the adapter. If you use lora/qlora, could you perhaps rename the file to add `-lora.yml` or `-qlora.yml` at the end, so it's easier to distinguish?

examples/mistral/mistral-dpo.yml
NanoCode012: Would this not drop data points or are the sequences in the referred dataset really short?

examples/mistral/mistral-dpo.yml
NanoCode012: Sorry for that incorrect suggestion. Yes, I meant to remove that. Thank you for doing so.

examples/mistral/mistral-dpo.yml
olivermolenschot: Done!

examples/mistral/mistral-dpo.yml
olivermolenschot: The datapoints are quite short by design, so it is not dropping many rows. However, I tested it with max sequence length of 2048 and it also trained well. So I updated it and I think you are right, this would be a better example for others. 

I also made all the other changes you suggested and trained the model successfully.

olivermolenschot:
I think I addressed all your concerns @NanoCode012. Let me know if there is anything else :)

examples/mistral/mistral-dpo-qlora.yml
```diff
@@ -0,0 +1,95 @@
...
+lora_target_modules:
+  - gate_proj
+  - down_proj
+  - up_proj
+  - q_proj
+  - v_proj
+  - k_proj
+  - o_proj
+  - embed_tokens
+  - lm_head
```
NanoCode012:
```suggestion
```

Adding it in `lora_modules_to_save` internally adds it to be targeted

examples/mistral/mistral-dpo-qlora.yml
```diff
@@ -0,0 +1,95 @@
...
+    type: chat_template.default
+    field_messages: conversation
+    field_chosen: chosen
+    field_rejected: rejected
+    message_field_role: role
+    message_field_content: content
+
+dataset_prepared_path:
+val_set_size: 0.05
+output_dir: ./outputs/dpo-lora
```
NanoCode012:
```suggestion
output_dir: ./outputs/dpo-qlora
```

NanoCode012 (commented):
Just two more tiny nitpicks, and it looks good to go!

olivermolenschot:
Cool, I commited your comments @NanoCode012

NanoCode012 (approved):

NanoCode012:
Thank you very much for the PR!

olivermolenschot:
You're welcome! Glad to help.

PR #2031, opened 2024-11-11
add axolotlai docker hub org to publish list

winglian:
# Description

changes to publish docker images under new axolotlai docker hub org

### DIFF
--- a/.github/workflows/nightlies.yml
+++ b/.github/workflows/nightlies.yml
@@ -41,7 +41,9 @@ jobs:
         id: metadata
         uses: docker/metadata-action@v5
         with:
-          images: winglian/axolotl
+          images: |
+            winglian/axolotl
+            axolotlai/axolotl
           tags: |
             type=raw,value={{ branch }}-{{ date 'YYYYMMDD' }}
       - name: Set up Docker Buildx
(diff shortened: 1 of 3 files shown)


.github/workflows/nightlies.yml
```diff
@@ -41,7 +41,9 @@ jobs:
         id: metadata
         uses: docker/metadata-action@v5
         with:
-          images: winglian/axolotl
+          images: |
+            winglian/axolotl
+            axolotlai/axolotl
```
NanoCode012, line 46: On L108, `images: winglian/axolotl-cloud` should be changed to `images: axolotlai/axolotl-cloud` (once an image is initially pushed)

NanoCode012 (commented):
I believe we also need to change the Dockerfiles (including `Dockerfile.jinja`) from `FROM winglian/axolotl-base:$BASE_TAG` to `FROM axolotlai/axolotl-base:$BASE_TAG`.

Another one would be the references in `README.md` and `debugging.qmd`.

I did not make the commits as I wasn't sure if you would need to set/change secrets in Settings for it to work and not break main setup.

winglian:
there will be a second PR after this to swap the base images after we get images published from our automation

PR #2032, opened 2024-11-11
ADOPT optimizer integration

bursteratom:
# Description

Modification to incorporate [ADOPT optimizer](https://github.com/iShohei220/adopt) 

## How has this been tested?

Ran it on tinyllama, works so far.


adopt.py
winglian, line 1: let's move this to an importable module. see https://github.com/axolotl-ai-cloud/axolotl/pull/1978/files for some hints

winglian: oh, maybe this should be deleted

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
```diff
@@ -293,7 +293,7 @@ class LoraConfig(BaseModel):
     gptq: Optional[bool] = None
     bnb_config_kwargs: Optional[Dict[str, Any]] = None
 
-    loraplus_lr_ratio: Optional[float] = Field(
+    loraplus__ratio: Optional[float] = Field(
```
winglian: unintentional change?

src/axolotl/core/optimizers/adopt.py
```diff
@@ -0,0 +1,510 @@
+# mypy: allow-untyped-decorators
```
NanoCode012: Could you please add citation https://github.com/iShohei220/adopt/tree/main?tab=readme-ov-file#citation

bursteratom: What is the proper way to add citation?

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
bursteratom: ah yes, unintentional indeed. Thank you for catching it

examples/llama-3/fft-8b-adopt.yml
winglian, line 1: Let's remove this file since it's simply just an optimizer change

src/axolotl/utils/optimizers/adopt.py
```diff
@@ -0,0 +1,516 @@
+"""
+@inproceedings{taniguchi2024adopt,
```
winglian:
```suggestion
"""
Copied from https://github.com/iShohei220/adopt
@inproceedings{taniguchi2024adopt,
```

winglian:
@bursteratom I think we need to wrap the test case in a decorator that requires torch 2.5.1

bursteratom:
> @bursteratom I think we need to wrap the test case in a decorator that requires torch 2.5.1

just added that and it passed the e2e

tests/e2e/utils.py
```diff
@@ -49,6 +49,18 @@ def is_min_2_3_1():
...
 
 
+def require_torch_2_5_1(test_case):
+    """
+    Decorator marking a test that requires torch >= 2.3.1
+    """
+
+    def is_min_2_5_1():
+        torch_version = version("torch")
+        return torch_version >= "2.5.1"
```
winglian: using string checks for version isn't really safe. You should use something like `from packaging import version` with `version.parse(...)`

winglian: see https://github.com/huggingface/transformers/blob/main/src/transformers/pytorch_utils.py#L20-L39

docs/config.qmd
```diff
@@ -405,6 +405,7 @@ lr_div_factor: # Learning rate div factor
 # - adamw_torch_fused
 # - adamw_torch_xla
 # - adamw_apex_fused
+# - adopt_adamw
```
winglian: we should probably hint in the documentation that this is only for PyTorch version >= 2.5.1

tests/e2e/utils.py
bursteratom: fixed

docs/config.qmd
bursteratom: fixed

winglian (approved):
🙌

Issue #2030, opened 2024-11-11
custom prompt format (custom dataset) doesn't work


mindcrunch4u:
### Please check that this issue hasn't been reported before.

- [X] I searched previous [Bug Reports](https://github.com/axolotl-ai-cloud/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

Expected `accelerate` to train the model using a custom dataset: [AI_Storyteller_Dataset](https://huggingface.co/datasets/jaydenccc/AI_Storyteller_Dataset)

### Current behaviour

`accelerate` throws errors.

```
Traceback (most recent call last):
  File "/dataset/test/condaenv/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/dataset/test/condaenv/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/dataset/test/axolotl/src/axolotl/cli/train.py", line 58, in <module>
    fire.Fire(do_cli)
  File "/dataset/test/condaenv/lib/python3.10/site-packages/fire/core.py", line 135, in Fire
    component_trace = _Fire(component, args, parsed_flag_args, context, name)
  File "/dataset/test/condaenv/lib/python3.10/site-packages/fire/core.py", line 468, in _Fire
    component, remaining_args = _CallAndUpdateTrace(
  File "/dataset/test/condaenv/lib/python3.10/site-packages/fire/core.py", line 684, in _CallAndUpdateTrace
    component = fn(*varargs, **kwargs)
  File "/dataset/test/axolotl/src/axolotl/cli/train.py", line 29, in do_cli
    parsed_cfg = load_cfg(config, **kwargs)
  File "/dataset/test/axolotl/src/axolotl/cli/__init__.py", line 425, in load_cfg
    cfg = validate_config(
  File "/dataset/test/axolotl/src/axolotl/utils/config/__init__.py", line 242, in validate_config
    AxolotlConfigWCapabilities(
  File "/dataset/test/condaenv/lib/python3.10/site-packages/pydantic/main.py", line 171, in __init__
    self.__pydantic_validator__.validate_python(data, self_instance=self)
  File "/dataset/test/axolotl/src/axolotl/utils/config/models/input/v0_4_1/__init__.py", line 792, in deprecate_sharegpt_datasets
    if ds_cfg["type"].startswith("sharegpt"):
```

### Steps to reproduce

[...]

PR #2035, opened 2024-11-11
fix: handle sharegpt dataset missing

NanoCode012:
# Description

Fix the first problem in #2030 

## Motivation and Context

I forgot to consider that `type: ` can be non-str when setting the deprecation error.


winglian (approved):

winglian:
@NanoCode012 should probably add a test case for this scenario

NanoCode012:
Added the test!

PR #2072, opened 2024-11-18
Add Exact Deduplication Feature to Preprocessing Pipeline

### LINKED ISSUE #1946: Feature Request: Adding dataset deduplication process

- [X] I searched previous [Ideas in Discussions](https://github.com/axolotl-ai-cloud/axolotl/discussions/categories/ideas) didn't find any similar feature requests.
- [X] I searched previous [Issues](https://github.com/axolotl-ai-cloud/axolotl/labels/enhancement) didn't find any similar feature requests.

### 🔖 Feature description

A dataset deduplication progress feature could be useful for Axolotl. Especially since many users input their datasets in various formats and configurations, having a deduplication process at the end when all these datasets are merged would be very beneficial for developers fine-tuning models.

### ✔️ Solution

In my use case, adding a 'dedup_datasets_in_end' (this variable name is only a example) variable and the necessary parameters for the deduplication process would be very beneficial.

### ❓ Alternatives

There are many algorithms, GitHub repositories, and tools for dataset deduplication. For example, the main algorithm that comes to mind is MinHash. Incorporating such algorithms over time would be very beneficial.

### 📝 Additional Context

_No response_

### Acknowledgements

olivermolenschot:
# Description

I created a feature that will deduplicate datasets (exact matches) and maintain the same order. This feature is activated by setting `exact_deduplication = true` in the yaml configuration file. The feature works for both RL tasks and non-RL tasks. However, it will not be triggered for `pretraining_dataset`. The feature is triggered only via the preprocess command and not the train command. 

## Motivation and Context
Users often have datasets from different sources, which may contain duplicate rows. This feature automates the deduplication process, reducing the manual effort previously required by users to remove duplicate entries.  
Resolves #1946

## How has this been tested?
- **Unit Tests:** Provided and verified for all deduplication functions, and functions that directly call deduplication functions.
- **Manual Testing:** 
  - Tested with two YAML configurations (`exact_deduplication = true/false/None`) by running `python -m axolotl.cli.preprocess`
- **Environment:** Python 3.10.15, Torch 2.3.1+cu121, Conda environment.

N/A

## Types of changes


olivermolenschot:
@NanoCode012 the pre-commit is failing on code that I did not change. Any insights on this?

NanoCode012:
> @NanoCode012 the pre-commit is failing on code that I did not change. Any insights on this?

I think it's detecting duplicates as test configs are similar across. It's fine to ignore this case in tests.

You can add a `# pylint: disable=duplicate-code` at the top of the duplicated section. For ex: https://github.com/axolotl-ai-cloud/axolotl/blob/0c8b1d824aeaa41f27b4ed4e73e8906751180322/tests/test_packed_dataset.py#L20-L28

You can find the section from the CI / locally running pre-commit.

olivermolenschot:
@NanoCode012 Your siggestion worked. The pre-commit now passes, thanks. 

For the pytests, they are passing locally (I skipped the e2e folder as done here too), but they fail on github. It seems that it has to do with loading the dataset from huggingface. A test file that I did not create/modify is also failing:

`FAILED tests/test_datasets.py::TestDatasetPreparation::test_load_hub - datasets.data_files.EmptyDatasetError: The directory at /home/runner/work/axolotl/axolotl/mhenrichsen/alpaca_2k_test doesn't contain any data files`

I think it might have to do with loading the dataset from huggingface. Did you experience this or have any insights on quick fixes?

NanoCode012:
I ran your test locally, and it does pass.

Huggingface sometimes bugs in our CI esp for dataset loading.

winglian:
We recently added some pytest fixtures to pre-download the popular datasets and models used in the tests, but it was mostly for the e2e tests. We could consider moving those up to the top level tests. https://github.com/axolotl-ai-cloud/axolotl/blob/main/tests/e2e/conftest.py

olivermolenschot:
It looks like many of the previous failed tests now pass, @winglian. Only one test failed, because it exceeded the maximum allotted time:
`
The job running on runner GitHub Actions 230 has exceeded the maximum execution time of 20 minutes.
`
I'm guessing it has something to do with pre-storing the datasets and models.

olivermolenschot:
On Attempt 4, the only test that failed is the e2e-tests. I thought that folder needed to be skipped, @NanoCode012. Can we skip it or is it needed?

NanoCode012:
Hey @olivermolenschot , sorry for this. I'm not sure why the tests are acting up for this PR. I restarted yours again, but it is rare for us to have to do this more than once.

> Can we skip it or is it needed?

E2E test are quite important to run. We run it last after our other quick tests as these are slower and more expensive.

src/axolotl/utils/data/sft.py
```diff
@@ -178,7 +183,7 @@ def load_tokenized_prepared_datasets(
                 + "|".join(
                     sorted(
                         [
-                            f"{d.path}:{d.type}:{d.shards}:{d.conversation}{d.split}"
+                            f"{d.path}: {d.type}: {d.shards}: {d.conversation}{d.split}"
```
winglian, line 182: The spaces here are probably not necessary for consistency since this gets hashed.

src/axolotl/utils/data/utils.py
```diff
@@ -1,10 +1,81 @@
...
+) -> tuple[Dataset, Dataset]:  # type: ignore
+    """
+    Deduplicates train and eval datasets if provided, logging original and new sizes.
+
+    Returns:
+        tuple: Deduplicated train and eval datasets.
+    """
+
+    if train_dataset is not None:
+        logging.info(
```
winglian: Would be preferable to use the existing logging pattern we have in other modules.

tests/test_exact_deduplication.py
```diff
@@ -0,0 +1,402 @@
+"""
+Test suite for functions in the axolotl.utils.data.utils module, focusing on the deduplicate_and_log_datasets function.
+
+Additionally, this test suite includes tests for functions that indirectly call deduplicate_and_log_datasets during the execution of the preprocess command.
+"""
+# pylint: disable=duplicate-code
+# import logging
```
winglian: Can be removed

tests/test_exact_deduplication.py
```diff
@@ -0,0 +1,402 @@
...
+from datasets import Dataset
+from transformers import AutoTokenizer
+
+from axolotl.utils.data import prepare_dataset
+from axolotl.utils.data.rl import load_prepare_dpo_datasets
+from axolotl.utils.data.utils import deduplicate_and_log_datasets
+from axolotl.utils.dict import DictDefault
+from axolotl.utils.models import load_processor, load_tokenizer
+
+# logging.basicConfig(level=logging.INFO)
```
winglian: Can be removed

winglian (commented):
Just a few minor nits, but this is a great addition!

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
```diff
@@ -606,6 +606,7 @@ class Config:
         default=None, metadata={"help": {"streaming dataset to use for pretraining"}}
     )
     dataset_processes: Optional[int] = Field(default=os.cpu_count())
+    exact_deduplication: Optional[bool] = None
```
NanoCode012: Hm, thoughts on labelling this, `dataset_exact_deduplication`?

src/axolotl/utils/data/utils.py
```diff
@@ -1,10 +1,81 @@
...
         return hashlib.md5(to_hash.encode(encoding)).hexdigest()  # nosec
+
+
+def collect_unique_rows(*, dataset: Dataset) -> list:
+    """Converts each row to a tuple, keeping only unique rows."""
+    # Convert dict_values to list for JSON serialization
+    unique_row_tuples = list(
+        dict.fromkeys(tuple(json.dumps(list(row.values())) for row in dataset))
+    )
+    return unique_row_tuples
```
NanoCode012: Is this the best way for getting the unique rows? It seems very complex and expensive.

tests/test_exact_deduplication.py
```diff
@@ -0,0 +1,402 @@
...
+        self.assertEqual(
+            len(train_dataset),
+            2000,
+            "Train dataset should have 1800 samples after deduplication.",
+        )
+
+    def test_prepare_dataset_with_deduplication_eval(self):
+        """Verify that prepare_dataset function processes the dataset correctly with deduplication."""
+        self.cfg_1.exact_deduplication = True
+        self.cfg_1.val_set_size = 3998
```
NanoCode012: Interesting number. Why was this chosen?

src/axolotl/utils/data/utils.py
NanoCode012: What if we use their built-in `.unique()`? (Please test if it flattens to the data or not) https://huggingface.co/docs/datasets/v3.1.0/en/package_reference/main_classes#datasets.Dataset.unique

Another concern is memory usage from having to load the entire data in memory like in your method.

tests/test_exact_deduplication.py
winglian: See inline comment below on line 366

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
winglian: I agree, that would be clearer

src/axolotl/utils/data/utils.py
olivermolenschot: From what I saw online, there is no inbuilt method in huggingface _datasets_ to get the unique rows of a dataset (see [this](https://stackoverflow.com/questions/73145394/how-can-i-take-the-unique-rows-of-a-huggingface-dataset). Their inbuilt `.unique()` function takes in a column and provides a sorted list of unique values for that row. 

Regarding whether there is an easier way, if there is no inbuilt function I think this might be the easiest way. 

Regarding the cost of memory, this method only gets activated in areas of the code where the data is already fully loaded (for `pretraining_datasets` it is not activated). Since we are creating a list of unique rows, we are using at most roughly 2x the dataset size as memory (this includes the space taken by the original dataset). 

Regarding how expensive/fast the method is, one thing I could do to make it faster is to directly use a hashing algorithm rather than converting every row to a JSON. Originally I was converting a row to a JSON because it makes it hashable, but I could simply use a hashing algorithm which should remove the overhead of converting to a JSON and back. 

Please let me know what you think so I can start implementing those changes.

NanoCode012: > Their inbuilt .unique() function takes in a column and provides a sorted list of unique values for that row.

Hm I see. I misunderstood the docs and thought if given multiple columns, it would adjust.

> Regarding how expensive/fast the method is, one thing I could do to make it faster is to directly use a hashing algorithm rather than converting every row to a JSON. Originally I was converting a row to a JSON because it makes it hashable, but I could simply use a hashing algorithm which should remove the overhead of converting to a JSON and back.

By expensive, I mean more on computation cost as it gets converted between multiple data types.

How about using pandas with their `df.drop_duplicates`? You should be able to convert between `Dataset` and `DataFrame`.

olivermolenschot: Yeah that can also work. I'll try it out and commit the changes to the PR soon.

olivermolenschot: I tried using `df.drop_duplicates`. Initially it fails when the cell content is a list, because it is not hashable. From my research the most reasonable way to proceed would be to cast the whole dataframe as string and then proceed. Something like this:

```
    unique_indices = df.astype(str).drop_duplicates().index
    deduplicated_df = df.loc[unique_indices]
    deduplicated_dataset = Dataset.from_pandas(deduplicated_df, preserve_index=False)
```

But this would be fairly expensive because apparently `df.astype(str)` will create a duplicate of the dataset. Instead, we can use hashing, which would be more memory efficient for our task, especially for large datasets. Here we would only need to store one hash per row, and one integer:

```
seen_hashes = set()
unique_indices = []

for idx, row in enumerate(dataset):
    row_hash = hash(str(row))
    if row_hash not in seen_hashes:
        seen_hashes.add(row_hash)
        unique_indices.append(idx)

deduplicated_dataset = dataset.select(unique_indices)
```

src/axolotl/utils/data/utils.py
olivermolenschot: Done. I used `LOG.info()` instead

tests/test_exact_deduplication.py
olivermolenschot: Done. I removed `#pylint` and created a `constants.py` file to avoid the errors mentioned by pylint (as some test cases in different files contained the same constants). This should make it more maintainable for the future too.

src/axolotl/utils/data/sft.py
olivermolenschot: flake8 fails when I try to commit because of these spaces, that's why I included them. What should I do instead?

src/axolotl/utils/data/constants.py
olivermolenschot, line 1: Note that this file avoids pylint issues between other files, as there were repeating constants between some tests. @winglian

src/axolotl/utils/data/utils.py
olivermolenschot: I changed it to the method where I hash, using sha256 to have enough expressiveness for diverse dataset sizes. Using this method also provides us with the benefit of being able to remove duplicates between `train_dataset `and `eval_dataset `because we can use the same set for both. This is what I did in the last commit (7-8).

winglian:
FAILED tests/e2e/test_packing_loss.py::TestPackedLlama::test_loss_packed - Un...
FAILED tests/e2e/test_relora_llama.py::TestReLoraLlama::test_relora - Unbound...
FAILED tests/e2e/test_reward_model_llama.py::TestRewardModelLoraLlama::test_rm_fft
```
		else:
			if cfg.dataset_exact_deduplication:
				train_dataset, _, _ = deduplicate_and_log_datasets(train_dataset=dataset)
>           eval_dataset = None
E           UnboundLocalError: cannot access local variable 'train_dataset' where it is not associated with a value

src/axolotl/utils/data/sft.py:607: UnboundLocalError
```

src/axolotl/utils/data/sft.py
```diff
@@ -583,12 +585,13 @@ def load_prepare_datasets(
         train_dataset = dataset["train"]
         eval_dataset = dataset["test"]
     elif split == "test":
+        if cfg.dataset_exact_deduplication:
+            _, eval_dataset, _ = deduplicate_and_log_datasets(eval_dataset=dataset)
         train_dataset = None
```
winglian: should this handle the original case when there is no deduplication?

```suggestion
        if cfg.dataset_exact_deduplication:
            _, eval_dataset, _ = deduplicate_and_log_datasets(eval_dataset=dataset)
        else:
            eval_dataset = dataset
        train_dataset = None
```

src/axolotl/utils/data/sft.py
```diff
@@ -583,12 +585,13 @@ def load_prepare_datasets(
...
-        eval_dataset = dataset
     else:
-        train_dataset = dataset
+        if cfg.dataset_exact_deduplication:
+            train_dataset, _, _ = deduplicate_and_log_datasets(train_dataset=dataset)
         eval_dataset = None
```
winglian: similarly here
```suggestion
        if cfg.dataset_exact_deduplication:
            train_dataset, _, _ = deduplicate_and_log_datasets(train_dataset=dataset)
        else:
            train_dataset = dataset
        eval_dataset = None
```

winglian:
~~Hmmm. Doesn't make sense why installing from sdist results in a test failure of~~

~~FAILED tests/test_exact_deduplication.py::TestDeduplicateNonRL::test_prepare_dataset_without_deduplication - ValueError: data_files must be either a string or list of strings~~

flaky test it seems, rerunning seems to pass now

olivermolenschot:
Any idea why the pytest fails @winglian? It works locally.

winglian:
Looks like it passed on some of the other versions of Python. Usually this indicates it's a flaky test. Best guess is that it failed to download dataset directly from HF leading to ending up in that else clause. I'll trigger it to run again shortly 

EDIT: but the failure was the same as the previous one:
```
FAILED tests/test_exact_deduplication.py::TestDeduplicateNonRL::test_prepare_dataset_with_deduplication_eval - ValueError: data_files must be either a string or list of strings
```

src/axolotl/utils/data/constants.py
winglian, line 1: This should probably get moved to the tests directory since it's test specific.

winglian (commented):
One minor file move and this is good to go. Thanks!

src/axolotl/utils/data/constants.py
olivermolenschot: Done!

olivermolenschot:
@winglian Great! I moved the file. Let me know if there's anything else needed. I had to rename a file so that might create some complications.

PR #2079, opened 2024-11-19
bump transformers for fsdp-grad-accum fix, remove patch

winglian:
# Description


winglian:
multigpu tests: https://github.com/axolotl-ai-cloud/axolotl/actions/runs/11907036373

PR #2087, opened 2024-11-19
fix None-type not iterable error when deepspeed is left blank w/ use_…

bursteratom:
If you use qlora and set `use_reentrant: false` while leaving deepspeed blank `deepspeed:  ` in the config yml, you get:

 `File "/workspace/axolotl/src/axolotl/utils/config/models/input/v0_4_1/__init__.py", line 1317, in warn_qlora_zero3_w_use_reentrant
    and "zero3" in data.get("deepspeed", "")
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: argument of type 'NoneType' is not iterable`

This PR fixes it.

# Description

### DIFF
--- a/src/axolotl/utils/config/models/input/v0_4_1/__init__.py
+++ b/src/axolotl/utils/config/models/input/v0_4_1/__init__.py
@@ -1314,6 +1314,7 @@ class AxolotlInputConfig(
             and data.get("gradient_checkpointing_kwargs", {})
             and data.get("gradient_checkpointing_kwargs", {}).get("use_reentrant")
             is False
+            and data.get("deepspeed", "") is not None
             and "zero3" in data.get("deepspeed", "")
         ):
             # may result in:
--- a/tests/test_validation.py
+++ b/tests/test_validation.py
@@ -86,7 +86,7 @@ class TestValidation(BaseValidation):
                 "qlora + zero3 with use_reentrant: false may result in a CheckpointError about recomputed values"
                 in self._caplog.records[0].message
             )
-    
+
     def test_deepspeed_empty(self, minimal_cfg):
         test_cfg = DictDefault(
             {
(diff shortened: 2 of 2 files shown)


src/axolotl/utils/config/models/input/v0_4_1/__init__.py
```diff
@@ -1314,6 +1314,7 @@ def warn_qlora_zero3_w_use_reentrant(cls, data):
             and data.get("gradient_checkpointing_kwargs", {})
             and data.get("gradient_checkpointing_kwargs", {}).get("use_reentrant")
             is False
+            and data.get("deepspeed", "") is not None
             and "zero3" in data.get("deepspeed", "")
```
winglian, line 1318: isn't this redundant with the next line? if it's empty, doesn't the next line `data.get("deepspeed", "")` default it to an empty string?

bursteratom: when deepspeed is not specified, `data.get("deepspeed", "")` returns `None` and I get `TypeError: argument of type 'NoneType' is not iterable`

winglian:
let's add a unit test for this. Doesn't need to be an e2e test, just a basic config validation for the two cases where deepspeed  is empty vs not set in the config.

winglian:
also, you can skip the E2E tests since they are not needed for this change. simply append `[skip e2e]` to your commit messages.

bursteratom:
@winglian just added unit tests tho I'm not sure if I did it correctly as there doesn't seem to be room to generalise beyond the very specific combination `adapter: qlora` + `deepspeed: ` + `use_reentrant: false` . How should I modify the tests?

winglian:
Look further down in that test validation file, like around line 200 in this pr. There is an example of checking the warning message. But also good to have the checks where it passes too.

winglian (approved):

PR #2090, opened 2024-11-19
Fix duplication of plugin callbacks

chiragjn:
# Description

https://github.com/axolotl-ai-cloud/axolotl/pull/1917 Added code to call the plugin manager to get the callbacks at the end of `HFCausalTrainerBuilder.get_post_trainer_create_callbacks`, however the plugins were already being added via a `super().get_post_trainer_create_callbacks` call. This has caused the plugin callbacks to be duplicated. This PR fixes it by moving the super call at the end

## How has this been tested?

Manually

## Types of changes

N/A

## Social Handles (Optional)

N/A


winglian:
Thanks. I think this was likely a merge issue when I rebased that PR and we were both simultaneously adding that functionality. Great catch.

winglian (approved):

PR #2091, opened 2024-11-20
feat: add cut_cross_entropy

NanoCode012:
# Description

Add https://github.com/apple/ml-cross-entropy

Note: `torch>=2.4`

## How has this been tested?

Run test comparison:
- [x] llama
- [x] mistral
- [x] gemma2
- [x] phi3

Misc:
- [ ] Add guards check conflict against other patches


src/axolotl/integrations/cut_cross_entropy/__init__.py
```diff
@@ -0,0 +1,97 @@
...
+
+        cce_spec_transformers = importlib.util.find_spec(
+            "cut_cross_entropy.transformers"
+        )
+        if cce_spec_transformers is None:
+            raise ImportError(_CCE_INSTALL_MESSAGE)
+
+    def pre_model_load(self, cfg):
+        """Apply cut cross entropy before model loading if enabled."""
+        if cfg.cut_cross_entropy:
```
NanoCode012, line 72: Design question: Do we need arg to be true or just loading the plugin is enough to load?

winglian:
The cce e2e tests don't actually work since we don't install cce anywhere. We probably actually want an error when cce is configured but not installed/available.

src/axolotl/integrations/cut_cross_entropy/__init__.py
winglian: I would like to have this be explicitly set to true.

tests/e2e/integrations/cut_cross_entropy.py
```diff
@@ -0,0 +1,62 @@
...
+    """
+    e2e tests for cut_cross_entropy integration with Axolotl
+    """
+
+    @with_temp_dir
+    def test_llama_w_cce(self, temp_dir):
+        cfg = DictDefault(
+            {
+                "base_model": "JackFram/llama-68m",
+                "tokenizer_type": "LlamaTokenizer",
```
winglian:
```suggestion
                "base_model": "HuggingFaceTB/SmolLM2-135M",
```

tests/e2e/integrations/cut_cross_entropy.py
```diff
@@ -0,0 +1,62 @@
...
+                    "axolotl.integrations.cut_cross_entropy.CutCrossEntropyPlugin",
+                ],
+                "cut_cross_entropy": True,
+                "sequence_len": 1024,
+                "val_set_size": 0.1,
+                "special_tokens": {
+                    "unk_token": "<unk>",
+                    "bos_token": "<s>",
+                    "eos_token": "</s>",
+                },
```
winglian:
```suggestion
                "special_tokens": {
                    "pad_token": "<|endoftext|>",
                },
```

tests/e2e/integrations/cut_cross_entropy.py
```diff
@@ -0,0 +1,62 @@
...
+
+from ..utils import with_temp_dir
+
+
+class CutCrossEntropyIntegrationTestCase(unittest.TestCase):
+    """
+    e2e tests for cut_cross_entropy integration with Axolotl
+    """
+
+    @with_temp_dir
```
winglian: this decorator can be removed. `temp_dir` should be an automatic pytest fixture.

NanoCode012:
```
FAILED tests/e2e/integrations/cut_cross_entropy.py::CutCrossEntropyIntegrationTestCase::test_llama_w_cce - TypeError: CutCrossEntropyIntegrationTestCase.test_llama_w_cce() missing 1 required positional argument: 'temp_dir'
```

It would error out if I remove the decorator.

NanoCode012:
There is a bug that plugins aren't loaded despite in the configs for our integration e2e tests. This has been fixed in https://github.com/axolotl-ai-cloud/axolotl/pull/2091/commits/a983c3bc73982c0aae428cb36b9b2b34893d40b3

NanoCode012:
> The cce e2e tests don't actually work since we don't install cce anywhere. We probably actually want an error when cce is configured but not installed/available.

There were a few checks for installation. However, the plugin wasn't loaded leading to them not being checked. I have just fixed this.

tests/e2e/integrations/cut_cross_entropy.py
winglian: change it so the test class doesn't extend unittest.TestCase, i.e. `class CutCrossEntropyIntegrationTestCase:`

winglian (approved):

winglian:
I rebased this PR as there were some conflicts.

winglian:
There's an integration issue since unsloth also requires cut-cross-entropy in it's pyproject.toml

PR #2093, opened 2024-11-20
add e2e tests for Unsloth qlora and test the builds

winglian:


.github/workflows/tests.yml
```diff
@@ -67,6 +67,7 @@ jobs:
         run: |
           pip3 show torch
           pip3 install -U -e .
+          python scripts/unsloth_install.py | sh
```
NanoCode012, line 70: Should this be added to `pytest-sdist` and to the nightly workflow too?

src/axolotl/utils/models.py
```diff
@@ -451,27 +453,34 @@ def patch_attention(self) -> None:
 
                 replace_stablelm_attn_with_flash_attn(self.cfg.base_model)
 
-    def patch_loss(self) -> None:
+    @cached_property
+    def flash_attn(self) -> bool:
```
```suggestion
    def has_flash_attn(self) -> bool:
```

Perhaps this name is more clear?

src/axolotl/utils/models.py
```diff
@@ -451,27 +453,34 @@ def patch_attention(self) -> None:
...
+                patch_llama_cross_entropy,
+                patch_llama_rms_norm,
+            )
 
-        if self.cfg.flash_attn_cross_entropy:
+        if self.cfg.flash_attn_cross_entropy and self.flash_attn:
             patch_llama_cross_entropy()
-        if self.cfg.flash_attn_rms_norm:
+        elif self.cfg.unsloth_cross_entropy_loss:
+            from axolotl.monkeypatch.unsloth_ import integrate_cross_entropy_loss_patch
```
NanoCode012, line 474: I didn't get to ask but why does the name have an extra underscore in it `unsloth_`?

src/axolotl/utils/models.py
```diff
@@ -451,27 +453,34 @@ def patch_attention(self) -> None:
...
-        )
+        if self.flash_attn:
+            from axolotl.monkeypatch.llama_attn_hijack_flash import (
+                patch_llama_cross_entropy,
+                patch_llama_rms_norm,
+            )
 
-        if self.cfg.flash_attn_cross_entropy:
+        if self.cfg.flash_attn_cross_entropy and self.flash_attn:
             patch_llama_cross_entropy()
```
NanoCode012: I looked at this patch and it sets 
```py
    transformers.models.llama.modeling_llama.CrossEntropyLoss = partial(
        CrossEntropyLoss, inplace_backward=True
    )
```
However, the loss fn has been changed and should be patched like
```py
transformers.loss.loss_utils.ForCausalLMLoss = ...
```
This patch may have been silently ignored.

tests/e2e/patched/test_unsloth_qlora.py
```diff
@@ -0,0 +1,188 @@
...
+    )
+    def test_unsloth_llama_qlora_unpacked_no_fa2_fp16(self, temp_dir, sdp_attention):
+        # pylint: disable=duplicate-code
+        cfg = DictDefault(
+            {
+                "base_model": "HuggingFaceTB/SmolLM2-135M",
+                "sequence_len": 1024,
+                "sample_packing": False,
+                "load_in_4bit": True,
+                "adapter": "qlora",
```
NanoCode012: Is `sdp_attention: False` just the same as `test_unsloth_llama_qlora_unpacked_no_fa2`?

tests/e2e/patched/test_unsloth_qlora.py
```diff
@@ -0,0 +1,188 @@
...
+        assert (Path(temp_dir) / "adapter_model.bin").exists()
+
+        tb_log_path = most_recent_subdir(temp_dir + "/runs")
+        event_file = os.path.join(tb_log_path, sorted(os.listdir(tb_log_path))[0])
+        reader = SummaryReader(event_file)
+        df = reader.scalars  # pylint: disable=invalid-name
+        df = df[(df.tag == "train/train_loss")]  # pylint: disable=invalid-name
+        assert df.value.values[-1] < 2.0, "Loss is too high"
+
+    def test_unsloth_llama_qlora_unpacked_no_fa2(self, temp_dir):
```
NanoCode012:
```suggestion
    def test_unsloth_llama_qlora_unpacked(self, temp_dir):
```

tests/e2e/patched/test_unsloth_qlora.py
```diff
@@ -0,0 +1,188 @@
...
+    """
+    Test class for Unsloth QLoRA Llama models
+    """
+
+    @pytest.mark.parametrize(
+        "sample_packing",
+        [True, False],
+    )
+    def test_unsloth_llama_qlora_fa2(self, temp_dir, sample_packing):
+        # pylint: disable=duplicate-code
```
NanoCode012: You could perhaps move this pylint disable to above the`class TestUnslothQLoRA:` and delete the rest.

tests/e2e/patched/test_unsloth_qlora.py
```diff
@@ -0,0 +1,188 @@
...
+from tbparse import SummaryReader
+
+from axolotl.cli import load_datasets
+from axolotl.common.cli import TrainerCliArgs
+from axolotl.train import train
+from axolotl.utils.config import normalize_config
+from axolotl.utils.dict import DictDefault
+
+LOG = logging.getLogger("axolotl.tests.e2e")
+os.environ["WANDB_DISABLED"] = "true"
```
NanoCode012, line 19: I don't see wandb configs being set below. Is there a need to set this?

src/axolotl/utils/models.py
winglian: if you don't it clashes with the actual `unsloth` package

tests/e2e/patched/test_unsloth_qlora.py
winglian: it is, but this is also an fp16 test, that one is bf16

tests/e2e/patched/test_unsloth_qlora.py
winglian: why the proposed change?

tests/e2e/patched/test_unsloth_qlora.py
winglian: this is in all our e2e tests. if you don't explicilty disable wandb, it will pause the trainer when you start and ask if you want to use wandb

tests/e2e/patched/test_unsloth_qlora.py
NanoCode012: I think it’s just a regular run. We don’t need to mention it doesn’t have FA specifically?

src/axolotl/utils/models.py
winglian: fixed the implementation and updated the test to validate the loss

PR #2094, opened 2024-11-20
Add ds model card

bursteratom:
Continuation of PR#1015 with rebasing

# Description

Add dataset tagging to model card for both local and push to huggingface hub.

https://github.com/huggingface/hub-docs/blob/main/modelcard.md


bursteratom:
@winglian this is a temporary WIP patch, the proper way will probably be actually parsing dataset_tags into the trainer for proper kwarg sanitising, I think. Do you think it's necessary?

winglian:
Why skip for dpo? Let's make sure to properly handle that case too.

bursteratom:
@winglian trl trainer's `def create_model_card()` has different arguments that don't seem to directly support dataset tags (it only allows you to parse the name of a single dataset),
https://github.com/huggingface/trl/blob/main/trl/trainer/dpo_trainer.py
so thinking of how to get around that properly

.github/workflows/pypi.yml
winglian, line 1: This file should have no edits, you have some errors in your merge commit. I personally prefer to rebase branches rather than merge commits to avoid issues like this.

.gitignore
winglian, line 1: saem as above.

docs/unsloth.qmd
winglian, line 1: saem as above.

examples/colab-notebooks/colab-axolotl-example.ipynb
winglian, line 1: saem as above.

requirements.txt
winglian, line 1: saem as above.

scripts/unsloth_install.py
winglian, line 1: saem as above.

winglian (changes requested):
there are too many unrelated changes in this PR from doing a merge commit. Please resolve and re-request a review.

bursteratom:
to be continued with rebasing here: https://github.com/axolotl-ai-cloud/axolotl/pull/2101

PR #2101, opened 2024-11-24
Add ds model card, rebased

bursteratom:
# Description

Continuation of PR#1015,#2094 with rebasing

### DIFF
--- a/src/axolotl/core/trainer_builder.py
+++ b/src/axolotl/core/trainer_builder.py
@@ -108,8 +108,6 @@ def _sanitize_kwargs_for_tagging(tag_names, kwargs=None):
 
 
 def _sanitize_kwargs_for_ds_tagging(dataset_tag_names, kwargs=None):
-    if isinstance(dataset_tag_names, type(None)):
-        return kwargs
     if isinstance(dataset_tag_names, str):
         dataset_tag_names = [dataset_tag_names]
 
--- a/src/axolotl/core/trainer_builder.py
+++ b/src/axolotl/core/trainer_builder.py
@@ -1239,11 +1239,17 @@ class TrainerBuilderBase(abc.ABC):
         Callbacks added after the trainer is created, usually b/c these need access to the trainer
         """
         callbacks = []
-
-        plugin_manager = PluginManager.get_instance()
-        callbacks.extend(
-            plugin_manager.add_callbacks_post_trainer(cfg=self.cfg, trainer=trainer)
-        )
+        if self.cfg.plugins:
+            plugin_manager = PluginManager.get_instance()
+            callbacks.extend(
+                [
+                    cb
+                    for cb in plugin_manager.add_callbacks_post_trainer(
+                        self.cfg, trainer
+                    )
+                    if cb
+                ]
+            )
         return callbacks
 
     def hook_pre_create_training_args(self, training_arguments_kwargs):
@@ -1290,7 +1296,7 @@ class HFCausalTrainerBuilder(TrainerBuilderBase):
         return callbacks
 
     def get_post_trainer_create_callbacks(self, trainer):
-        callbacks = super().get_post_trainer_create_callbacks(trainer=trainer)
+        callbacks = []
         if self.cfg.use_wandb and self.cfg.eval_table_size > 0:
             LogPredictionCallback = log_prediction_callback_factory(
                 trainer, self.tokenizer, "wandb"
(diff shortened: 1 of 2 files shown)


src/axolotl/core/trainer_builder.py
```diff
@@ -107,6 +107,24 @@ def _sanitize_kwargs_for_tagging(tag_names, kwargs=None):
     return kwargs
 
 
+def _sanitize_kwargs_for_ds_tagging(dataset_tag_names, kwargs=None):
+    if isinstance(dataset_tag_names, type(None)):
```
NanoCode012: Is this a check for None?

```suggestion
    if dataset_tag_names is None:
```

src/axolotl/core/trainer_builder.py
```diff
@@ -1755,6 +1782,10 @@ def build(self, total_num_steps):
         else:
             trainer_kwargs["tokenizer"] = self.tokenizer
 
+        if (trainer_cls is not AxolotlRewardTrainer) and self.cfg.datasets is not None:
+            trainer_kwargs["dataset_tag_names"] = [
+                d["path"] for d in self.cfg.datasets if not Path(d["path"]).is_dir()
```
NanoCode012, line 1785: Hm, this check may miss some cases.

- `path: json` for local data type with `data_files`
- `path: file.jsonl` for local data path with `ds_type`

Is this mainly for linking with existing HF paths? If so, we should check it is an HF dataset instead?

src/axolotl/train.py
```diff
@@ -259,11 +259,27 @@ def terminate_handler(_, __, model_weakref):
...
         try:
-            trainer.create_model_card(
-                model_name=cfg.output_dir.lstrip("./").encode("utf-8").decode("utf-8")
-            )
-        except (AttributeError, UnicodeDecodeError):
+            # Check to make sure the base model is from HuggingFace not a local directory
+            hf_api = HfApi()
+            hf_api.model_info(cfg.base_model)
+
+            model_card_kwarg = {"model_name": cfg.output_dir.lstrip("./")}
```
NanoCode012: Saw this part removed `.encode("utf-8").decode("utf-8")` from model_name. Is this intentional? I'm not sure what it's for though.

src/axolotl/train.py
```diff
@@ -259,11 +259,27 @@ def terminate_handler(_, __, model_weakref):
...
+            if cfg.datasets is not None:
+                if cfg.rl is not None or cfg.reward_model:
+                    model_card_kwarg["dataset_name"] = [
+                        d["path"] for d in cfg.datasets if not Path(d["path"]).is_dir()
+                    ]
+                else:
+                    model_card_kwarg["dataset_tags"] = [
+                        d["path"] for d in cfg.datasets if not Path(d["path"]).is_dir()
+                    ]
```
NanoCode012, line 283: What's difference between `dataset_tags` and `dataset_name`

src/axolotl/core/trainer_builder.py
bursteratom: Yes, but I'm gonna delete that line cuz it's redundant.

src/axolotl/train.py
bursteratom: oh you are right, I did unintentionally remove that part, will restore them. Thanks for picking it up!

src/axolotl/train.py
bursteratom: They are supposed to be the same, but some how trl trainer named it differently than the regular transformers trainer, so we will have to patch that here.

src/axolotl/core/trainer_builder.py
bursteratom: yeah it's for existing HF paths only. I don't think we need to check if it is an HF dataset cuz it's been done when the model prepares for training.

src/axolotl/core/trainer_builder.py
```diff
@@ -410,10 +426,12 @@ def __init__(
         *_args,
         bench_data_collator=None,
         eval_data_collator=None,
+        dataset_tag_names=None,
```
winglian: `dataset_tag_names` isn't used for the base Trainer is it? Only for DPOTrainer?

bursteratom: it is used for both AxolotlTrainer and AxolotlDPOTrainer within `def push_to_hub` via `_sanitize_kwargs_for_ds_tagging`

winglian: I'm not fond of introducing arguments that don't match the original Trainer signature for this case. The call to `_sanitize_kwargs_for_ds_tagging` should handle the case where the attribute isn't available.

bursteratom: Thanks for the advice. While trl trainer has an override for `def create_model_card` over the base transformers trainer, it uses the same `def push_to_hub` inherited from the base transformers trainer. I changed `dataset_tag_names` to `dataset_tags` to match `def push_to_hub`

bursteratom: @winglian Pls let me know what you think. Thanks!

winglian (approved):

winglian (approved):
lgtm

PR #2102, opened 2024-11-25
fix: ds3 and fsdp lmbench eval

### LINKED ISSUE #2058: Axolotl hanging on bench evals with fsdp

- [X] I searched previous [Bug Reports](https://github.com/axolotl-ai-cloud/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

I m finetuning an LLAMA, and after an ep[och we need to launch the validation normally, but it is hanging

### Current behaviour

Axolotl is hanging for too long after the first epoch. I m finetuning a custom dataset on 2 RTX4090 using Qlora FSDP

`wandb: WARNING Saving files without folders. If you want to preserve subdirectories pass base_path to wandb.save, i.e. wandb.save("/mnt/folder/file.h5", base_path="/mnt")
[2024-11-14 13:13:49,394] [INFO] [axolotl.callbacks.on_train_begin:794] [PID:3439] [RANK:0] The Axolotl config has been saved to the WandB run under files.
  0%|                                                                                                                                         | 0/2792 [00:00<?, ?it/s]You're using a PreTrainedTokenizerFast tokenizer. Please note that with a fast tokenizer, using the `__call__` method is faster than using a method to encode the text followed by a call to the `pad` method to get a padded encoding.
{'loss': 0.9007, 'grad_norm': 18.620527267456055, 'learning_rate': 1.0000000000000002e-06, 'epoch': 0.0}                                                               
{'eval_loss': nan, 'eval_runtime': 68.9564, 'eval_samples_per_second': 8.527, 'eval_steps_per_second': 2.132, 'epoch': 0.0}                                            
  0%|                                                                                                                               | 1/2792 [01:16<5:39:51,  7.31s/it/root/miniconda3/envs/py3.11/lib/python3.11/site-packages/transformers/generation/utils.py:2097: UserWarning: You are calling .generate() with the `input_ids` being on a device type different than your model's device. `input_ids` is on cuda, whereas the model is on cpu. You may experience unexpected behaviors or slower generation. Please make sure that you have put `input_ids` to the correct device by calling for example input_ids = input_ids.to('cpu') before running `.generate()`.
  warnings.warn(
`

### Steps to reproduce

1. allocate 2 rtx 4090
2. launch the train using the config file
3. it will hang after first epoch

### Config yaml

```yaml
base_model: ./Llama-3.2-1B-Instruct
model_type: LlamaForCausalLM
tokenizer_type: AutoTokenizer  
load_in_8bit: false
load_in_4bit: true
strict: false

seed: 42

datasets:
  - path: formated_partial_context.jsonl
    type: chat_template

    field_messages: conversations
    message_field_role: role

val_set_size: 0.05
output_dir: ./outputs/out/qlora-Llama-3.2-1B-Instruct-seq4096-cuad-qa-partial

adapter: qlora
lora_model_dir:

sequence_len: 512
sample_packing: false
pad_to_sequence_len: true


evaluation_strategy : steps #epoch
eval_steps: 100
#evals_per_epoch: 1

eval_table_size: 3
eval_max_new_tokens: 512

do_causal_lm_eval: true
eval_causal_lm_metrics:
  - "context_matching"  # custom metric name
  - "perplexity" 


lora_r: 8
lora_alpha: 16
lora_dropout: 0.05
lora_target_modules:
lora_target_linear: true
lora_fan_in_fan_out:

wandb_project: llama3.2-1B-seq512-fine-tune-cuad-qa-partial
wandb_entity:
wandb_watch:
wandb_name:
wandb_log_model: checkpoint

micro_batch_size: 1
num_epochs: 4
optimizer: adamw_torch
lr_scheduler: cosine
learning_rate: 0.00001

train_on_inputs: false
group_by_length: false
bf16: auto
fp16:
tf32: false

early_stopping_patience:
resume_from_checkpoint:
local_rank:
logging_steps: 1
xformers_attention:
flash_attention: true

warmup_steps: 10
saves_per_epoch: 1
debug:
deepspeed:
weight_decay: 0.0
fsdp:
  - full_shard
  - auto_wrap
fsdp_config:
  fsdp_limit_all_gathers: true
  fsdp_sync_module_states: true
  fsdp_offload_params: true
  fsdp_use_orig_params: false
  fsdp_cpu_ram_efficient_loading: true
  fsdp_auto_wrap_policy: TRANSFORMER_BASED_WRAP
  fsdp_sharding_strategy: FULL_SHARD
  activation_checkpointing: true
  pin_memory: false
special_tokens:
  pad_token: <|end_of_text|>
```

### Possible solution

_No response_

- [X] Linux
- [ ] macOS
- [ ] Windows

### Python Version

3.11

### axolotl branch-commit

official version 0.5.0

### Acknowledgements

- [X] I have searched the existing issues to make sure this bug has not been reported yet.
- [X] I am using the latest version of axolotl.
- [X] I have provided enough information for the maintainers to reproduce and diagnose the issue.

NanoCode012:
# Description

Fixes #2058 

We properly utilize multi-gpu for lm bench evals and pass `trainer.model_wrapped` to Perplexity's `compute()` instead of during `init`.

## How has this been tested?

Fft:
- [x] DS1
- [x] DS2
- [x] DS3
- [x] DS3 (w offload)
- [x] FSDP  (with and without act checkpointing)
- [x] FSDP (w offload)

Note: I find FSDP to take much longer than DS in this.

qlora:
- [x] DS1
- [x] DS2
- [x] DS3
- [x] DS3 (w offload)
- [x] FSDP  (with and without act checkpointing)
- [x] FSDP (w offload)

### DIFF
--- a/src/axolotl/utils/callbacks/perplexity.py
+++ b/src/axolotl/utils/callbacks/perplexity.py
@@ -8,6 +8,8 @@ from transformers.modeling_outputs import CausalLMOutput
 from transformers.modeling_utils import PreTrainedModel
 from transformers.tokenization_utils import PreTrainedTokenizer
 
+from axolotl.utils.distributed import is_main_process
+
 
 class Perplexity:
     """
@@ -17,16 +19,13 @@ class Perplexity:
 
     def __init__(
         self,
-        model: PreTrainedModel,
         tokenizer: PreTrainedTokenizer,
         max_seq_len: int,
         stride: int = 512,
     ) -> None:
         self.max_seq_len = max_seq_len
         self.stride = stride
-        self.model = model
         self.tokenizer = tokenizer
-        self.device = model.device
         self.name = "perplexity"
 
     def _feature_names(self) -> List[str]:
@@ -34,6 +33,7 @@ class Perplexity:
 
     def compute(
         self,
+        model: PreTrainedModel,
         references: Optional[List[str]] = None,
     ) -> Dict[str, float]:
         """
@@ -41,17 +41,21 @@ class Perplexity:
         """
         assert references is not None, "Missing parameter: references"
 
+        model.eval()
+
         references_tokenized = self.tokenizer(
             references, return_tensors="pt", padding=True, truncation=True
         )
         input_ids: Tensor = references_tokenized["input_ids"]  # type: ignore
-        input_ids = input_ids.to(self.device)
+        input_ids = input_ids.to(model.device)
 
         sequence_length = input_ids.size(1)
 
         losses = []
         prev_end_loc = 0
-        for begin_loc in tqdm(range(0, sequence_length, self.stride)):
+        for begin_loc in tqdm(
+            range(0, sequence_length, self.stride), disable=not is_main_process()
+        ):
             end_loc = min(begin_loc + self.max_seq_len, sequence_length)
             trg_len = end_loc - prev_end_loc
             input_ids_slice = input_ids[:, begin_loc:end_loc]
@@ -59,7 +63,7 @@ class Perplexity:
             labels_slice[:, :-trg_len] = -100
 
             with torch.no_grad():
-                outputs: CausalLMOutput = self.model(
+                outputs: CausalLMOutput = model(
                     input_ids=input_ids_slice, labels=labels_slice
                 )
 
--- a/src/axolotl/utils/callbacks/__init__.py
+++ b/src/axolotl/utils/callbacks/__init__.py
@@ -28,6 +28,7 @@ from transformers import (
     TrainingArguments,
 )
 from transformers.trainer_utils import PREFIX_CHECKPOINT_DIR, IntervalStrategy
+from trl.models import unwrap_model_for_generation
 
 from axolotl.utils import is_comet_available, is_mlflow_available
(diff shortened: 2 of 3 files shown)


src/axolotl/utils/callbacks/perplexity.py
```diff
@@ -17,49 +19,51 @@ class Perplexity:
...
         self,
+        model: PreTrainedModel,
         references: Optional[List[str]] = None,
     ) -> Dict[str, float]:
         """
         Compute perplexity in a fixed length sliding window across the sequence.
         """
         assert references is not None, "Missing parameter: references"
 
+        model.eval()
```
winglian, line 44: if we set model.eval() here, do we need to switch it back to model.train at the end?

NanoCode012: Should be no. This plugin  is only run `on_evaluate` from the `CausalLMBenchEvalCallback`. In fact, we already have one at the start too, so this line may not be needed.

If you're worried, we can check its state and set it back to it after. I would actually be cautious as that would make any runs after while `on_evaluate` to have a potential model in training mode.

winglian (approved):

PR #2104, opened 2024-11-25
Check torch version for ADOPT optimizer + integrating new ADOPT updates

bursteratom:
# Description

Make sure the torch version is compatible when ADOPT optimizer is used.
Incorporated latest changes to ADOPT optimizer made by original author. https://github.com/iShohei220/adopt

### DIFF
--- a/src/axolotl/cli/__init__.py
+++ b/src/axolotl/cli/__init__.py
@@ -22,6 +22,7 @@ from accelerate.commands.config import config_args
 from art import text2art
 from huggingface_hub import HfApi
 from huggingface_hub.utils import LocalTokenNotFoundError
+from packaging import version
 from transformers import GenerationConfig, TextIteratorStreamer, TextStreamer
 from transformers.utils import is_torch_bf16_gpu_available
 from transformers.utils.import_utils import _is_package_available
@@ -100,8 +101,8 @@ def print_dep_versions():
         print("*" * 40)
         print("**** Axolotl Dependency Versions *****")
         for pkg in packages:
-            version = _is_package_available(pkg, return_version=True)
-            print(f"{pkg: >{max_len}}: {version[1]: <15}")
+            pkg_version = _is_package_available(pkg, return_version=True)
+            print(f"{pkg: >{max_len}}: {pkg_version[1]: <15}")
         print("*" * 40)
 
 
@@ -444,6 +445,7 @@ def load_cfg(config: Union[str, Path] = Path("examples/"), **kwargs):
             "n_gpu": int(os.environ.get("WORLD_SIZE", 1)),
             "compute_capability": gpu_version,
         },
+        env_capabilities={"torch_version": version.parse(torch.__version__)},
     )
 
     prepare_optim_env(cfg)
--- a/src/axolotl/cli/__init__.py
+++ b/src/axolotl/cli/__init__.py
@@ -22,7 +22,6 @@ from accelerate.commands.config import config_args
 from art import text2art
 from huggingface_hub import HfApi
 from huggingface_hub.utils import LocalTokenNotFoundError
-from packaging import version
 from transformers import GenerationConfig, TextIteratorStreamer, TextStreamer
 from transformers.utils import is_torch_bf16_gpu_available
 from transformers.utils.import_utils import _is_package_available
@@ -445,7 +444,9 @@ def load_cfg(config: Union[str, Path] = Path("examples/"), **kwargs):
             "n_gpu": int(os.environ.get("WORLD_SIZE", 1)),
             "compute_capability": gpu_version,
         },
-        env_capabilities={"torch_version": version.parse(torch.__version__)},
+        env_capabilities={
+            "torch_version": str(torch.__version__).split("+", maxsplit=1)[0]
+        },
     )
 
     prepare_optim_env(cfg)
--- a/src/axolotl/utils/__init__.py
+++ b/src/axolotl/utils/__init__.py
@@ -3,6 +3,9 @@ Basic utils for Axolotl
 """
 import importlib.util
 
+import torch
+from packaging import version
+
 
 def is_mlflow_available():
     return importlib.util.find_spec("mlflow") is not None
@@ -10,3 +13,8 @@ def is_mlflow_available():
 
 def is_comet_available():
     return importlib.util.find_spec("comet_ml") is not None
+
+
+def is_torch_min(version_string: str):
+    torch_version = version.parse(torch.__version__)
+    return torch_version >= version.parse(version_string)
--- a/src/axolotl/utils/__init__.py
+++ b/src/axolotl/utils/__init__.py
@@ -3,9 +3,6 @@ Basic utils for Axolotl
 """
 import importlib.util
 
-import torch
-from packaging import version
-
 
 def is_mlflow_available():
     return importlib.util.find_spec("mlflow") is not None
@@ -13,8 +10,3 @@ def is_mlflow_available():
 
 def is_comet_available():
     return importlib.util.find_spec("comet_ml") is not None
-
-
-def is_torch_min(version_string: str):
-    torch_version = version.parse(torch.__version__)
-    return torch_version >= version.parse(version_string)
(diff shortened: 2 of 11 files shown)


src/axolotl/utils/config/models/internals/__init__.py
```diff
@@ -12,3 +12,9 @@ class GPUCapabilities(BaseModel):
     n_gpu: int = Field(default=1)
     n_node: int = Field(default=1)
     compute_capability: Optional[str] = Field(default=None)
+
+
+class ENVCapabilities(BaseModel):
```
winglian: `EnvCapabilities` would be preferable here since env isn't an acronym.

bursteratom: yup, changed that

src/axolotl/utils/config/models/internals/__init__.py
```diff
@@ -12,3 +12,9 @@ class GPUCapabilities(BaseModel):
     n_gpu: int = Field(default=1)
     n_node: int = Field(default=1)
     compute_capability: Optional[str] = Field(default=None)
+
+
+class EnvCapabilities(BaseModel):
+    """model to manage the environment capabilities statically"""
+
+    torch_version: Optional[str] = Field(default=None)
```
NanoCode012, line 20: @winglian , this could be an alternative to the `def flash_attn` you created in the Unsloth PR and allow re-use of import checks across modules. This can also include bnb / other stuff later to check before we import to allow optional packages in future.

src/axolotl/utils/config/__init__.py
```diff
@@ -240,13 +244,20 @@ def validate_config(cfg: DictDefault, capabilities: Optional[dict] = None):
         ) = merge_input_args()
 
     if capabilities:
+        if env_capabilities is None:
+            env_capabilities = {
+                "torch_version": str(torch.__version__).split("+", maxsplit=1)[0]
+            }
```
NanoCode012: Is this necessary as you did set it earlier in `load_cfg`? Perhaps in tests I guess?

If so, should we remove the one in `load_cfg`?

src/axolotl/utils/config/models/internals/__init__.py
winglian: good call-out. probably out of scope for this PR, but we can upgrade the others in follow-up PRs.

src/axolotl/utils/config/__init__.py
bursteratom: actually I'll just remove the one in `validate_cfg`

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
```diff
@@ -1548,3 +1550,15 @@ def check_multigpu_unsloth(cls, data):
                     "unsloth_lora_mlp, unsloth_lora_qkv, and unsloth_lora_o are not compatible with multi-GPU training."
                 )
         return data
+
+    @model_validator(mode="before")
+    @classmethod
+    def check_adopt_torch_version(cls, data):
+        if (data.get("optimizer") is not None) and ("adopt" in data.get("optimizer")):
+            env_capabilities = data.get("env_capabilities")
+            torch_version = env_capabilities.get("torch_version")
```
winglian: if env_capabilities is `Optional` this will fail with a KeyError or AttributeError I believe

bursteratom: in load_cfg I've set `env_capabilities` to include the torch_version, so during regular usage it will never be `None`. I previously set `env_capabilities` to be `Optional` so that it passes some of the tests. For here I think I can make it such that if `env_capabilities` is `None` it will perform a check for torch version

winglian: Right, but if in regular usage it should not be optional, we should remove the optional typehint. Otherwise you need to add guards here for when it is passed as none otherwise it will fail spectacularly and someone will ultimately end up filing a bug report

bursteratom: I see, should we instead change the pytest files?

winglian:
Don't forget to write a couple of tests where you mock the PyTorch version and validate the config

bursteratom:
> Don't forget to write a couple of tests where you mock the PyTorch version and validate the config

on it

src/axolotl/utils/config/__init__.py
NanoCode012: One caveat is that, in tests, we don't use `load_cfg`, we usually use `validate_cfg`, so this env may not be set. Just a thought for now

src/axolotl/utils/__init__.py
```diff
@@ -3,10 +3,18 @@
...
 def is_mlflow_available():
     return importlib.util.find_spec("mlflow") is not None
 
 
 def is_comet_available():
     return importlib.util.find_spec("comet_ml") is not None
+
+
+def is_torch_min(version_string: str):
+    torch_version = version.parse(torch.__version__)
```
NanoCode012: I found that `torch.__version__` can equal `2.3.1+cu121`. Does this parser handle it?

bursteratom: yes

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
```diff
@@ -1548,3 +1550,21 @@ def check_multigpu_unsloth(cls, data):
...
+
+            if torch_version is None:
+                import torch
+
+                torch_version = str(torch.__version__).split("+", maxsplit=1)[0]
+
+            if version.parse(torch_version) < version.parse("2.5.1"):
+                raise ValueError(
+                    "ADOPT optimizer is incompatible with torch version < 2.5.1"
+                )
```
winglian, line 1572: above in `utils` you created a new `is_torch_min` helper function, but never used it. I assumed you intended to use it here?

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
```diff
@@ -1548,3 +1550,21 @@ def check_multigpu_unsloth(cls, data):
                     "unsloth_lora_mlp, unsloth_lora_qkv, and unsloth_lora_o are not compatible with multi-GPU training."
                 )
         return data
+
+    @model_validator(mode="before")
+    @classmethod
+    def check_adopt_torch_version(cls, data):
+        if (data.get("optimizer") is not None) and ("adopt" in data.get("optimizer")):
+            env_capabilities = data.get("env_capabilities")
```
winglian:
```suggestion
            env_capabilities = data.get("env_capabilities", {})
```

this should help with the case where env_capabilities is None

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
bursteratom: that was my original intention but since we are using env_capabilities `is_torch_min` becomes unnecessary, tho I left it in `utils` cuz I thought it might be useful for some other functionalities in the future. Should I remove it from `utils`?

winglian: yes, let's remove `is_torch_min` then please as it's unused code then.

bursteratom: Got it, I just removed it.

src/axolotl/cli/__init__.py
```diff
@@ -444,6 +444,9 @@ def load_cfg(config: Union[str, Path] = Path("examples/"), **kwargs):
             "n_gpu": int(os.environ.get("WORLD_SIZE", 1)),
             "compute_capability": gpu_version,
         },
+        env_capabilities={
+            "torch_version": str(torch.__version__).split("+", maxsplit=1)[0]
```
NanoCode012, line 448: Would this ever throw some weird error? How about using the `parse` fn you used elsewhere?

winglian: I think this is fine. `version.parse(...)` returns a `Version` object, which isn't what we want to be passing through and having to know about on the other end. A `str` here should be good.

winglian (approved):
I just rebased this PR against `origin/main`, once tests pass, this should be good to go

PR #2107, opened 2024-11-27
CLI Implementation with Click

djsaunde:
# Description

This PR implements a simple CLI using [`click`](https://click.palletsprojects.com/en/stable/). Note the following entrypoint:

```python
    entry_points={
        "console_scripts": [
            "axolotl=axolotl.cli.main:main",
        ],
    },
```

This means, once `pip install`ed, `axolotl`'s CLI can be used as:

```python
axolotl [command] [args]
```

It support the following commands:
- `axolotl preprocess`
- `axolotl train`
- `axolotl inference`
- `axolotl shard`
- `axolotl merge_sharded_fsdp_weights`
- `axolotl merge_lora`
- `axolotl fetch`

The last command is new functionality which allows the user to download (really, sync) the contents of the `examples/` and `deepspeed_configs/` directories at the top level of the `axolotl` project. For example:

```bash
axolotl fetch examples
axolotl fetch deepspeed_configs
```

An optional argument `--dest [directory]` allows the user to specify an output location; otherwise, this defaults to `examples/` and `deepspeed_configs/` in the local directory.

## Motivation and Context

This change was requested in [this Notion ticket](https://www.notion.so/1431d2ab4f40805ab1ded1be2bf7f025?v=1431d2ab4f408170ab01000c1b203623&p=1451d2ab4f4080c4ba5ef307c883a63d&pm=s&pvs=31). This allows `axolotl` users that have installed the package via `pip` to use the existing CLI commands, and source users to use them with a slightly simpler interface `axolotl [command] ...`.

## How has this been tested?

Ad hoc testing on `axolotlai/axolotl-cloud:main-latest` image on a Runpod A40 instance on this feature branch. I sparsely tested the `preprocess`, `train`, and `fetch` commands.

```bash
# pre install
$ axolotl train examples/openllama-3b/lora.yml 
-bash: axolotl: command not found

$ pip install -e .
...

# works
$ axolotl preprocess examples/openllama-3b/lora.yml 

# works
$ axolotl train examples/openllama-3b/lora.yml 
...

# config file overrides? works with replacing "_" with "-" (click limitation)
$ axolotl train examples/openllama-3b/lora.yml --max-steps 1 --micro-batch-size 1 --val-set-size 0

# etc.
```

TODO:
- [ ] Add `pytest` coverage for the CLI (?)
- [ ] Test via PyPI package installation (?)

## Types of changes

This PR adds a new file `src/axolotl/cli/main.py` which implements the Click CLI commands, and minor changes elsewhere to accommodate them.


src/axolotl/cli/__init__.py
```diff
@@ -380,7 +380,7 @@ def choose_config(path: Path):
 
     if len(yaml_files) == 1:
         print(f"Using default YAML file '{yaml_files[0]}'")
-        return yaml_files[0]
```
djsaunde, line 383: Small bugfix here and on line 394; without converting to a `str`,  this returns a `PosixPath` which the `AxolotlInputConfig` class does not expect.

src/axolotl/utils/data/rl.py
```diff
@@ -6,9 +6,9 @@
 from pathlib import Path
 from typing import Any, List
 
-import yaml
 from datasets import DatasetDict, concatenate_datasets, load_dataset, load_from_disk
 
+import yaml
```
winglian: You might need to install `pyyaml` in your local environment and re-run `pre-commit run --all-files`, since these imports should have been correct already.

src/axolotl/cli/main.py
```diff
@@ -0,0 +1,261 @@
...
+    field_defaults = {field.name: field.default for field in fields(cls)}
+
+    # Update defaults with provided kwargs
+    field_defaults.update(kwargs)
+    return cls(**field_defaults)
+
+
+def create_trainer_args(**kwargs) -> TrainerCliArgs:
+    """Create TrainerCliArgs from cli options"""
+    return create_args_from_dataclass(TrainerCliArgs, **kwargs)
```
winglian: We should also parse out `AxolotlInputConfig` from these since the current behavior allows overriding values from the CLI as well

djsaunde: Updated. Note that `click` doesn't support underscores in argument names so `_` is replaced with `-` in the CLI, which are parsed by `click` back to `_` anyways.

djsaunde:
Question for the team: does it make sense to write tests against this code path? I noticed there weren't tests against the existing `axolotl.cli` module, but I can pad these out if desired. In the interest of moving fast, we can also backlog this.

winglian:
> Question for the team: does it make sense to write tests against this code path? I noticed there weren't tests against the existing `axolotl.cli` module, but I can pad these out if desired. In the interest of moving fast, we can also backlog this.

yeah, let's go ahead and take the time now to add some tests against the cli.

README.md
```diff
@@ -165,6 +165,46 @@ accelerate launch -m axolotl.cli.inference examples/openllama-3b/lora.yml \
...
 
+### Axolotl CLI
+
+If you've installed this package using `pip` from source, we now support a new, more
+streamlined CLI using [click](https://click.palletsprojects.com/en/stable/). Rewriting
+the above commands:
+
+```bash
+# preprocess datasets - optional but recommended
+CUDA_VISIBLE_DEVICES="" axolotl preprocess examples/openllama-3b/lora.yml
```
```suggestion
CUDA_VISIBLE_DEVICES="0" axolotl preprocess examples/openllama-3b/lora.yml
```
this was updated recently https://github.com/axolotl-ai-cloud/axolotl/pull/2114/files

src/axolotl/cli/main.py
```diff
@@ -0,0 +1,322 @@
...
+
+        do_cli(config=config, **kwargs)
+
+
+@cli.command()
+@click.argument("config", type=str)
+@click.option(
+    "--accelerate",
+    is_flag=True,
+    default=True,
```
winglian: let's make the default for this False, I think shard is likely best on single process as it's taking a large model bin/safetensors file and splitting it into sharded files.

winglian:
I think the test errors:

```
    def test_inference_all_options(
        cli_runner, default_config, mock_inference_deps
    ):  # pylint: disable=redefined-outer-name
        """Test inference with all possible options"""
        mock_cfg = MagicMock()
        mock_inference_deps["load_cfg"].return_value = mock_cfg
    
        cli_runner.invoke(
            cli,
            [
                "inference",
                str(default_config),
                "--load-in-8bit",
                "--base-model",
                "base/model/path",
                "--lora-model-dir",
                "lora/path",
                "--prompter",
                "my_prompter",
            ],
        )
    
        # Check all options were passed through
>       cli_args = mock_inference_deps["do_inference"].call_args[1]["cli_args"]
E       TypeError: 'NoneType' object is not subscriptable
```
is because some of the commands run accelerate as a new subprocess, so don't directly call the do_inference function in this scope.

djsaunde:
Assuming tests pass here (they're passing in Runpod at least), this PR is ready for re-review. I think I got the testing mostly right; I'm testing mostly shallowly where I assert that args that are expected are in fact passed to the relevant `train` / `do_inference` / `do_clil` etc. methods in similarly named `axolotl.cli.preprocess` / `axolotl.cli.train` etc. modules. A small subset of tests are more end-to-end and should be useful for preventing unexpected bugs from making it to `main`.

As an aside, once we deprecate the old way of running axolotl commands on the command line (e.g., `accelerate launch axolotl.cli.train ...`), we can probably refactor out a good bit of cruft.

djsaunde:
Adding context from Slack:

> I think something weird is going on with the logging. I couldn't debug it after a few hours of trying so I've moved the tests to tests/e2e/patched for now. I think maybe the test_validations.py could be moved there instead as I think that's what's causing the issue with the caplog logic, but I'm not sure.

I also simplified some of the CLI tests so they're no longer running simple preprocess / train / shard etc. commands end to end for the sake of runtime. We have other test coverage that actually tests this functionality, so I think there's no need to duplicate.

README.md
NanoCode012: I think this change was not applied?

outputs
```diff
@@ -0,0 +1 @@
+/workspace/data/axolotl-artifacts
```
NanoCode012, line 1: May I ask what is this? A symlink?

src/axolotl/cli/utils.py
```diff
@@ -0,0 +1,156 @@
...
+
+        key = key.replace("_", "-")
+
+        if isinstance(value, bool):
+            if value:
+                cmd.append(f"--{key}")
+        else:
+            cmd.extend([f"--{key}", str(value)])
+
+    return cmd
```
NanoCode012, line 90: do we need to sanitize `options` are user inputs?

src/axolotl/cli/utils.py
```diff
@@ -0,0 +1,156 @@
...
+    raw_url = f"{raw_base_url}/{file_path}"
+    dest_file = dest_path / file_path.split(dir_prefix)[-1]
+
+    # Check if file exists and needs updating
+    if dest_file.exists():
+        with open(dest_file, "rb") as file:
+            content = file.read()
+            # Calculate git blob SHA
+            blob = b"blob " + str(len(content)).encode() + b"\0" + content
+            local_sha = hashlib.sha1(blob, usedforsecurity=False).hexdigest()
```
NanoCode012, line 118: Is there a reason we set `usedforsecurity=False`?

src/axolotl/cli/utils.py
```diff
@@ -0,0 +1,156 @@
...
+            file_path = future_to_file[future]
+            try:
+                file_path, status = future.result()
+                files_processed[status].append(file_path)
+            except (requests.RequestException, IOError) as request_error:
+                print(f"Error processing {file_path}: {str(request_error)}")
+                files_processed["error"].append(file_path)
+
+    # Print summary
+    print("\nSync Summary:")
```
NanoCode012: Should all these prints be replaced with the `LOG = logging...` used in other files?

tests/e2e/patched/cli/conftest.py
```diff
@@ -0,0 +1,48 @@
...
+    return path
+
+
+@pytest.fixture(autouse=True)
+def cleanup_model_out():
+    yield
+
+    # Clean up after the test
+    if Path("model-out").exists():
+        shutil.rmtree("model-out")
```
NanoCode012: Should we use a parameter `model_path: str` for the path instead of hardcoding it here?

src/axolotl/cli/main.py
```diff
@@ -0,0 +1,221 @@
...
+        ]
+        if config:
+            base_cmd.append(config)
+        cmd = build_command(base_cmd, kwargs)
+        subprocess.run(cmd, check=True)  # nosec B603
+    else:
+        from axolotl.cli.merge_sharded_fsdp_weights import do_cli
+
+        do_cli(config=config, **kwargs)
+
```
NanoCode012, line 181: I think we're missing `merge_lora` cli?

src/axolotl/cli/utils.py
winglian: i think bandit will flag the use of sha1 if you don't

outputs
winglian, line 1: we can probably remove this symlink from the commit, this gets added by the docker-cloud build and you kind of end up with a broken state and unable to change branches if this is included in the changeset.

src/axolotl/cli/train.py
```diff
@@ -44,7 +44,9 @@ def do_train(cfg, cli_args) -> None:
     else:
         dataset_meta = load_datasets(cfg=cfg, cli_args=cli_args)
 
-    model, tokenizer = train(cfg=cfg, cli_args=cli_args, dataset_meta=dataset_meta)
+    output = train(cfg=cfg, cli_args=cli_args, dataset_meta=dataset_meta)
+    model, tokenizer = output
+    # model, tokenizer = train(cfg=cfg, cli_args=cli_args, dataset_meta=dataset_meta)
```
winglian: is this just for clarity or debugging? can remove the commented out line if we want to keep in this state.

src/axolotl/cli/utils.py
djsaunde: Yup.

src/axolotl/cli/utils.py
djsaunde: Yes, good catch.

tests/e2e/patched/cli/conftest.py
djsaunde: I may not even need this fixture anymore after I simplified my tests. Let me check.

src/axolotl/cli/main.py
djsaunde: Yes, not sure how I missed that...

src/axolotl/cli/train.py
djsaunde: Whoops, was using this to debug, will revert.

src/axolotl/cli/utils.py
djsaunde: Not sure what you mean, can you explain?

src/axolotl/cli/utils.py
NanoCode012: Thanks for clarifying.

djsaunde:
I need to add a way to pass `accelerate launch` args as well.

src/axolotl/cli/utils.py
NanoCode012: Sorry, I mean, since `options` are user inputs, do we need to sanitize it here before passing to `subprocess.run` later on?

djsaunde: If by sanitize you mean remove arguments that are unsupported, this already exists as `click` CLI commands disallow options that aren't explicitly added. E.g.,:

```
axolotl preprocess examples/openllama-3b/lora.yml --foo        
...
Usage: axolotl preprocess [OPTIONS] CONFIG
Try 'axolotl preprocess --help' for help.

Error: No such option: --foo
```

NanoCode012: Oh nice!

winglian:
need to update pytest calls in https://github.com/axolotl-ai-cloud/axolotl/blob/main/.github/workflows/tests.yml#L82 and on line 126 as well (and probably the nightly tests too)

.github/workflows/tests-nightly.yml
```diff
@@ -60,7 +60,8 @@ jobs:
 
       - name: Run tests
         run: |
-          pytest --ignore=tests/e2e/ tests/
+          pytest -n8 --dist loadfile --ignore=tests/e2e/ --ignore=tests/patched/ tests/
+          pytest tests/patched/
```
winglian, line 64:
```suggestion
          pytest -n8 --ignore=tests/e2e/ --ignore=tests/patched/ tests/
          pytest -n1 --dist loadfile tests/patched/
```
I think this is inverted and we want loadfile on the patched tests

.github/workflows/tests.yml
```diff
@@ -79,7 +79,8 @@ jobs:
 
       - name: Run tests
         run: |
-          pytest -n8 --ignore=tests/e2e/ tests/
+          pytest -n8 --dist loadfile --ignore=tests/e2e/ --ignore=tests/patched/ tests/
+          pytest tests/patched/
```
winglian, line 83:
```suggestion
          pytest -n8 --ignore=tests/e2e/ --ignore=tests/patched/ tests/
          pytest -n1 --dist loadfile tests/patched/
```

.github/workflows/tests.yml
```diff
@@ -123,7 +124,8 @@ jobs:
...
```
winglian, line 128:
```suggestion
          pytest -n8 --ignore=tests/e2e/ --ignore=tests/patched/ tests/
          pytest -n1 --dist loadfile tests/patched/
```

.github/workflows/tests-nightly.yml
djsaunde: IIUC `-n1 --dist loadfile` is the same as passing no args (we're using one worker, so we're of course loading each file onto the same worker).

djsaunde: I can go either way on using `--dist loadfile` for the first run, the runtime might be better in either case (would have to test).

winglian: yeah, I think you're right. i think the way we have it now is an evolution of starting with just `--dist loadfile` and eventually just moving it to one worker to prevent the wonky issues that happens when we monkey patch things and leaving the process in a patched state for a test that doesn't expect the patches.

tests/cli/conftest.py
```diff
@@ -0,0 +1,36 @@
...
+datasets:
+  - path: mhenrichsen/alpaca_2k_test
+    type: alpaca
+sequence_len: 2048
+max_steps: 1
+micro_batch_size: 1
+gradient_accumulation_steps: 1
+learning_rate: 1e-3
+special_tokens:
+  pad_token: <|end_of_text|>
```
winglian:
```suggestion
  pad_token: <|endoftext|>
```
see https://huggingface.co/HuggingFaceTB/SmolLM2-135M/blob/main/tokenizer_config.json#L5

winglian (approved):
looks good to go after the last few requested changes.

PR #2118, opened 2024-12-03
Hymba multipack support

bursteratom:
# Description

Support for Hymba multipacking


winglian:
a small e2e test for this with and without packing and validating the train_loss would be good to have.

winglian (changes requested):
Needs e2e test as well as a chat template setup. see https://huggingface.co/nvidia/Hymba-1.5B-Instruct/blob/main/tokenizer_config.json#L40

bursteratom:
rebased in https://github.com/axolotl-ai-cloud/axolotl/pull/2178

PR #2120, opened 2024-12-03
replace tensorboard checks with helper function

winglian:
# Description

There's a lot of duplicate code and boilerplate for checking tensorboard log values. This helper function should make it much easier to check values from e2e trainings without a lot of extra code.


tests/e2e/multigpu/test_eval.py
```diff
@@ -7,13 +7,11 @@
 
 import yaml
 from accelerate.test_utils import execute_subprocess_async
-from tbparse import SummaryReader
 from transformers.testing_utils import get_torch_dist_unique_port
+from utils.tensorboard_ import check_tensorboard
```
djsaunde: Why is `utils` sorted as a 3rd party package?

djsaunde (approved):
Seems like an easy win, just a nit about the `isort` order of the `utils` module.

tests/e2e/multigpu/test_eval.py
winglian: ah, conflict in utils under axolotl and utils under tests maybe? let me see if I can make that work properly

winglian: moving the helper fn and converting to a relative import seems to make it more obvious now.

PR #2123, opened 2024-12-05
fix: chat_template masking due to truncation, consolidate turn build and keys within field

NanoCode012:
# Description

This fixes cases where masking would fail for `chat_template` using the new behavior.

Changes:
- Remove `truncation` while applying `chat_template` as that would cut off conversation and affect masking downstream. We have another section that would drop long conversation already.
~- Removed default `roles` within the prompter to not automatically map roles (in case users intend to keep the original names and wasn't aware we mapped) <- Comment needed for this design choice!~
- Map `roles_to_train` if it exist within `roles`. For example, `roles_to_train: ["gpt"]` but the convo maps it all the `assistant` leading to everything being ignored.
- Consolidates building `turn` across modules and dropping off system prompt if done.
- Refactors `should_train`
- Optimize handling of matching tokens in a turn
- Explicitly truncates chat_template for reward model Bradleyterry. Previously, this was silently truncating.

TODO:
- ~[ ] handle `tool_calling` for assistant response for tools datasets~ we support nous tool calling dataset but not hf one for now.
- ~[] discuss whether to revert removing default `roles`~ temporarily revert for now
- ~[ ] update chat_template doc for explicitly specifying `roles`~
- ~[ ] Fix tests to specify `roles`~

## How has this been tested?

Test with Puffin dataset locally.

### DIFF
--- a/src/axolotl/prompt_strategies/chat_template.py
+++ b/src/axolotl/prompt_strategies/chat_template.py
@@ -32,9 +32,16 @@ class ChatTemplatePrompter(Prompter):
         roles: Optional[Dict[str, List[str]]] = None,
         drop_system_message: bool = False,
     ):
-        self.roles = (
-            {s: t for t, sources in roles.items() for s in sources} if roles else {}
-        )
+        if roles:
+            self.roles = {s: t for t, sources in roles.items() for s in sources}
+        else:
+            self.roles = {
+                "human": "user",
+                "user": "user",
+                "assistant": "assistant",
+                "gpt": "assistant",
+                "system": "system",
+            }
 
         self.message_field_role = message_field_role
         self.message_field_content = message_field_content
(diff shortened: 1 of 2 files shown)


src/axolotl/prompt_strategies/chat_template.py
```diff
@@ -262,40 +254,42 @@ def tokenize_prompt(self, prompt):
...
 
             LOG.debug(f"Should train: {should_train}")
 
             turn_start_idx, turn_end_idx = self.find_turn(
                 conversation_ids=input_ids, turn=index, turn_content=turn
             )
 
+            if turn_start_idx == -1 or turn_end_idx == -1:
+                LOG.warning(f"Failed to find boundaries for turn {index}")
+                continue
```
winglian: I don't think `continue` is correct here. from the original intention, it skips the block below to `# Handle EOS token`

NanoCode012: Thanks for pointing it out. I addressed it in the next commit.

winglian (approved):

PR #2128, opened 2024-12-06
reduce test concurrency to avoid HF rate limiting, test suite parity

winglian:


winglian:
something in this changeset is causing errors/warnings similar to `/root/miniconda3/envs/py3.11/compiler_compat/ld: /usr/local/cuda/lib64/libcufile.so: undefined reference to` in the e2e tests only.

PR #2132, opened 2024-12-06
retry fixtures

winglian:
this is a duplicate of #2128, but adding in chances in smaller chunks to track down the cuda test warnings


winglian:
this PR was mostly for debugging and getting 2128 working

PR #2135, opened 2024-12-06
use separate steps for image build and test run on modal

winglian:
# Description


winglian:
this is more problematic than the value it probably provides 😢

PR #2137, opened 2024-12-06
[docs] Update README Quickstart to use CLI

winglian:
# Description

now that we have a CLI, let's make the Quickstart point to that easier path.

This will need to be the last PR right before we cut a new release. Otherwise, users will see the `pip install axolotl`, but that will point to the previous release and they won't be able to use the CLI.


djsaunde (approved):
LGTM

README.md
```diff
@@ -107,58 +107,50 @@ Get started with Axolotl in just a few steps! This quickstart guide will walk yo
...
+If you're looking for the latest features and updates between releases, you'll need to install
+from source.
 
-# remote yaml files - the yaml config can be hosted on a public URL
-# Note: the yaml config must directly link to the **raw** yaml
-accelerate launch -m axolotl.cli.train https://raw.githubusercontent.com/axolotl-ai-cloud/axolotl/main/examples/openllama-3b/lora.yml
+```bash
+git clone https://github.com/axolotl-ai-cloud/axolotl.git
+cd axolotl
+pip3 install packaging ninja
```
NanoCode012, line 133: Hm, why is that we need `packaging ninja` for git installs but not pip installs?

README.md
```diff
@@ -107,58 +107,50 @@ Get started with Axolotl in just a few steps! This quickstart guide will walk yo
...
+pip3 install -e '.[flash-attn,deepspeed]'
 ```
 
-### Axolotl CLI
-
+### Axolotl CLI Usage
 If you've installed this package using `pip` from source, we now support a new, more
-streamlined CLI using [click](https://click.palletsprojects.com/en/stable/). Rewriting
-the above commands:
+streamlined CLI using [click](https://click.palletsprojects.com/en/stable/).
```
NanoCode012: Is this CLI limited to only via pip installs? I thought it was usable both ways.

examples/llama-3/lora-1b.yml
```diff
@@ -0,0 +1,76 @@
+base_model: NousResearch/Llama-3.2-1B
+
+load_in_8bit: false
+load_in_4bit: false
+strict: false
+
+datasets:
+  - path: teknium/GPT4-LLM-Cleaned
```
NanoCode012, line 8: Maybe we can use a smaller one like 
```suggestion
  - path: mhenrichsen/alpaca_2k_test
```
Edit: on the other hand, this may cause eval sample packing issues, so we can maybe leave it.

examples/llama-3/lora-1b.yml
```diff
@@ -0,0 +1,76 @@
...
+eval_table_size:
+eval_max_new_tokens: 128
+saves_per_epoch: 1
+debug:
+deepspeed:
+weight_decay: 0.0
+fsdp:
+fsdp_config:
+special_tokens:
+  pad_token: "<|end_of_text|>"
```
NanoCode012, line 74: Hm, do we need to set lora modules to train since we override the pad_token?

examples/llama-3/lora-1b.yml
```diff
@@ -0,0 +1,76 @@
...
+xformers_attention:
+flash_attention: true
+
+loss_watchdog_threshold: 5.0
+loss_watchdog_patience: 3
+
+warmup_steps: 10
+evals_per_epoch: 4
+eval_table_size:
+eval_max_new_tokens: 128
```
NanoCode012:
```suggestion
```

Not necessary

README.md
winglian: It was in the previous install instructions for git cloning so I kept it.

README.md
winglian: Yeah. You get the cli from installing either way.

examples/llama-3/lora-1b.yml
winglian: Nah. We don't need it for pad tokens only. We made a fix for that a while back.

README.md
NanoCode012:
```diff
-If you've installed this package using `pip` from source, we now support a new, more
+We now support 
```

Maybe we can update this to show that

NanoCode012:
Also found that the some hyperlinks within readme may be broken/not found. For `Usage` , nightly builds, and the new CLI headers.

PR #2142, opened 2024-12-06
Fix llama type model check

chiragjn:
# Description

I am seeing some unusally high eval losses and garbage predictions (might open another issue for that) and was checking with a debugger, noticed this check was changed when adding multimodal support. `==` seems incorrect here. 

## Motivation and Context

N/A

## How has this been tested?

N/A

N/A

## Types of changes

N/A

## Social Handles (Optional)

N/A


winglian (approved):
thanks!

PR #2153, opened 2024-12-08
add additional fft deepspeed variants

winglian:
# Description

adds deepspeed zero1, zero2, and the offload variants of zero3 to the test matrix for llama. 
also adds train loss checks for sanity
adds fft and qlora checks too across the deepspeed suite
uses max steps 5 instead of the previous 15 steps per test

## Motivation and Context

deepspeed is now v0.16.1, but I don't feel good about the upgrade without some CI in place to test more things


winglian:
see also https://github.com/axolotl-ai-cloud/axolotl/issues/2154

PR #2165, opened 2024-12-09
fix pytorch version check decorator for tests

winglian:


winglian:
version parse already handles this 🤦

PR #2167, opened 2024-12-10
use DataCollatorWithFlattening when not sample packing

winglian:
https://research.ibm.com/blog/hugging-face-training-flash-attention


winglian:
Screenshot 2024-12-12 at 8 11 33 PM
ran a quick experiment against l3-3B with microbatch size of 8. losses stack on top pretty closely, but flattening is good bit faster.

winglian:
I should make this have an auto option that automatically enables this when not using packing with micro batch size > 1

PR #2172, opened 2024-12-10
handle torch_compile set to auto

winglian:
users can get free wins from this, but it can be problematic to provide it depending on the torch version.

### DIFF
--- a/docs/config.qmd
+++ b/docs/config.qmd
@@ -332,7 +332,7 @@ output_dir: ./completed-model
 
 # Whether to use torch.compile and which backend to use
 # setting to `auto` will enable torch compile when torch>=2.5.1
-torch_compile:  Optional[Union[Literal["auto"], bool]]
+torch_compile:  # Optional[Union[Literal["auto"], bool]]
 torch_compile_backend:  # Optional[str]
 
 # Training hyperparameters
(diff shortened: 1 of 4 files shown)


docs/config.qmd
```diff
@@ -331,7 +331,8 @@ comet_experiment_config: # Dictionary for additional configuration settings, see
 output_dir: ./completed-model
 
 # Whether to use torch.compile and which backend to use
-torch_compile:  # bool
+# setting to `auto` will enable torch compile when torch>=2.5.1
+torch_compile:  Optional[Union[Literal["auto"], bool]]
```
```suggestion
torch_compile:  # Optional[Union[Literal["auto"], bool]]
```

Change into comment

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
```diff
@@ -1581,3 +1581,15 @@ def check_adopt_torch_version(cls, data):
...
+    def check_torch_compile_auto(self):
+        if self.torch_compile == "auto" and self.env_capabilities.get("cuda_version"):
+            if version.parse(self.env_capabilities.get("cuda_version")) > version.parse(
+                "2.5.1"
+            ):
+                LOG.info("torch.compile is available, setting torch_compile to True")
+                self.torch_compile = True
+            else:
+                self.torch_compile = False
+        return self
```
NanoCode012: Do we need to handle cases where `self.env_capabilities.get("cuda_version")` is not available? The boolean value won't be set, then `auto` will be passed downstream.

PR #2182, opened 2024-12-13
[feature] add pytorch profiling

winglian:
Screenshot 2024-12-12 at 8 48 59 PM

### DIFF
--- a/src/axolotl/utils/callbacks/profiler.py
+++ b/src/axolotl/utils/callbacks/profiler.py
@@ -1,0 +1,50 @@
+"""
+HF Trainer callback for creating pytorch profiling snapshots
+"""
+from pathlib import Path
+from pickle import dump  # nosec B403
+
+import torch
+from transformers import (
+    TrainerCallback,
+    TrainerControl,
+    TrainerState,
+    TrainingArguments,
+)
+
+
+class PytorchProfilerCallback(TrainerCallback):
+    """
+    PyTorch Profiler callback to create snapshots of GPU memory usage at specified steps.
+    """
+
+    def __init__(self, steps_to_profile: int = 5):
+        self.steps_to_profile = steps_to_profile
+
+    def on_train_begin(
+        self,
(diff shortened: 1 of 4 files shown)


djsaunde (approved):
LGTM, but can we add inline comments and / or some documentation of this config option?

src/axolotl/utils/callbacks/profiler.py
```diff
@@ -0,0 +1,43 @@
...
+
+class PytorchProfilerCallback(TrainerCallback):
+    """
+    PyTorch Profiler callback to create snapshots of GPU memory usage at specified steps.
+    """
+
+    def __init__(self, steps_to_profile: int = 5):
+        self.steps_to_profile = steps_to_profile
+        if self.steps_to_profile:
+            torch.cuda.memory._record_memory_history(  # pylint: disable=protected-access
```
NanoCode012, line 24: Is there any cost to recording all memory here (ie increased ram usage etc)?

src/axolotl/utils/callbacks/profiler.py
```diff
@@ -0,0 +1,43 @@
...
+    def on_step_end(  # pylint: disable=unused-argument
+        self,
+        args: TrainingArguments,  # pylint: disable=unused-argument
+        state: TrainerState,
+        control: TrainerControl,  # pylint: disable=unused-argument
+        **kwargs,  # pylint: disable=unused-argument
+    ):
+        if state.global_step == self.steps_to_profile:
+            snapshot = torch.cuda.memory._snapshot()  # pylint: disable=protected-access
+            with open(Path(args.output_dir) / "snapshot.pickle", "wb") as fout:
```
NanoCode012, line 37: Can we maybe output to a non-pickle file for safety reasons (shared to others , tampered, etc)? How would a person also read this file and display it?

winglian: I don't believe there are any other options that the visualizer supports as input. I added docs but forgot to push them earlier.

PR #2183, opened 2024-12-13
move the setting of PYTORCH_CUDA_ALLOC_CONF to the cli rather than train module

winglian:
noticed in the profiler that it said this wasn't set and moving it around reduced the reserved memory in the configuration I was using from 37GB to 27GB

### DIFF
--- a/src/axolotl/cli/main.py
+++ b/src/axolotl/cli/main.py
@@ -1,4 +1,6 @@
 """CLI definition for various axolotl commands."""
+import os
+
 # pylint: disable=redefined-outer-name
 import subprocess  # nosec B404
 from typing import Optional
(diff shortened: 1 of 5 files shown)


djsaunde (approved):
Maybe some subprocess weirdness? As long as it works.

src/axolotl/cli/main.py
```diff
@@ -48,6 +51,15 @@ def train(config: str, accelerate: bool, **kwargs):
...
     kwargs = {k: v for k, v in kwargs.items() if v is not None}
 
+    # enable expandable segments for cuda allocation to improve VRAM usage
+    torch_version = torch.__version__.split(".")
+    torch_major, torch_minor = int(torch_version[0]), int(torch_version[1])
+    if torch_major == 2 and torch_minor >= 2:
+        if os.getenv("PYTORCH_CUDA_ALLOC_CONF") is None:
+            os.environ[
+                "PYTORCH_CUDA_ALLOC_CONF"
+            ] = "expandable_segments:True,roundup_power2_divisions:16"
```
NanoCode012: I'm just reading about this: `roundup_power2_divisions` config. Is there a reason we're setting 16? I saw that the default was 512.

winglian: The default is 1 from what I've seen. 

The doc you were reading says :  the sizes are rounded up in multiple of blocks size of 512

Which is the block size, not the divisions

Issue #2149, opened 2024-12-07
Error During Model Saving QLORA + FSDP


ghsama:
### Please check that this issue hasn't been reported before.

- [X] I searched previous [Bug Reports](https://github.com/axolotl-ai-cloud/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

It is supposed to save the model without issue after finishing the training.

### Current behaviour

it raises an error that it couldn t find a Paramater in a list, the issue is comming from the funciton  _unflatten_param_groups in  python3.11/site-packages/torch/distributed/fsdp/_optim_utils.py

`
```
[rank0]:     trainer.train(resume_from_checkpoint=resume_from_checkpoint)                                                   [44/1835]
[rank0]:   File "/workspace/axolotl/new_env/lib/python3.11/site-packages/transformers/trainer.py", line 2164, in train               
[rank0]:     return inner_training_loop(                                                                                             
[rank0]:            ^^^^^^^^^^^^^^^^^^^^                                                                                             
[rank0]:   File "/workspace/axolotl/new_env/lib/python3.11/site-packages/transformers/trainer.py", line 2589, in _inner_training_loop
[rank0]:     self._maybe_log_save_evaluate(                                                                                          
[rank0]:   File "/workspace/axolotl/new_env/lib/python3.11/site-packages/transformers/trainer.py", line 3054, in _maybe_log_save_eval
uate                                                                                                                                 
[rank0]:     self._save_checkpoint(model, trial)                                                                                     
[rank0]:   File "/workspace/axolotl/src/axolotl/core/trainer_builder.py", line 990, in _save_checkpoint                              
[rank0]:     saved = super()._save_checkpoint(model, trial, **kwargs)                                                                
[rank0]:             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                
[rank0]:   File "/workspace/axolotl/new_env/lib/python3.11/site-packages/transformers/trainer.py", line 3190, in _save_checkpoint    
[rank0]:     self._save_optimizer_and_scheduler(output_dir)                                                                          
[rank0]:   File "/workspace/axolotl/new_env/lib/python3.11/site-packages/transformers/trainer.py", line 3306, in _save_optimizer_and_
scheduler                                                                                                                            
[rank0]:     save_fsdp_optimizer(                                                                                                    
[...]

PR #2193, opened 2024-12-15
[Fixing #2149] load_from_disk for RL-type training

leeparkuky:
Here’s a draft for the PR message:

---

# Description

This PR enhances the dataset loading functionality by introducing additional checks for loading datasets from disk. If the provided dataset path exists, it now attempts to load the dataset using `load_from_disk`. Additionally, it verifies if a specific split exists when dealing with a `DatasetDict` and raises a clear error message if the split is not found. If the path does not exist or an error occurs, the existing behavior of loading the dataset using `load_dataset` is retained.

## Motivation and Context

This change is required to handle cases where datasets are stored locally on disk, improving flexibility and robustness. It solves potential issues when users provide a valid path to a local dataset that was previously not being handled correctly. The new implementation also adds split validation for `DatasetDict`, preventing silent failures when accessing invalid splits.

## How has this been tested?

The changes have been tested by:
1. Loading a dataset from a valid local path using `load_from_disk`.
2. Verifying split existence when the dataset is of type `DatasetDict`.
3. Raising appropriate errors when an invalid split is provided.
4. Ensuring fallback to the original `load_dataset` when the path does not exist or errors occur.

Test scenarios include both valid and invalid paths, as well as valid and missing dataset splits.

_N/A_

## Types of changes

## Social Handles (Optional)

---

Let me know if you need further refinements!

### DIFF
--- a/src/axolotl/utils/data/rl.py
+++ b/src/axolotl/utils/data/rl.py
@@ -129,11 +129,28 @@ def load_prepare_preference_datasets(cfg):
                     )
                     split_datasets.insert(i, ds)
             else:
-                ds = load_dataset(  # pylint: disable=invalid-name
-                    ds_cfg["path"],
-                    split=ds_cfg["split"],
-                    revision=ds_cfg.get("revision", None),
-                )
+                if Path(ds_cfg["path"]).exists():
+                    try:
+                        ds = load_from_disk(
+                            ds_cfg["path"]
+                        )
+                        if isinstance(ds, DatasetDict):
+                            if ds_cfg["split"] in ds:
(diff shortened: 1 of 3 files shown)


winglian:
@leeparkuky Thanks for the PR! I saw this PR and I think the way to tackle this is via https://github.com/axolotl-ai-cloud/axolotl/pull/2204. I refactored out the dataset loading from remote or disk into its own function that's independent of the SFT transforms. Once this is merged, we can update this PR to simply use the new function provided that should handle the cases that you fixed here.

NanoCode012 (approved):

PR #2197, opened 2024-12-17
convert-diff-transformer CLI command / codepath

djsaunde:
# Description

~This PR implements the differential attention layer from the [Differential Transformer](https://arxiv.org/abs/2410.05258) paper.~

Note that the differential attention layer implementations / modeling code and conversion logic has been moved to this plugin repo: https://github.com/axolotl-ai-cloud/diff-transformer.

## Motivation and Context

We wanted to add this attention implementation to `axolotl` so users can swap out the existing attention layers in their models for this more performant version. We matched the official implementation details as closely as possible, while adopting it to play nicely with the `transformers` attention implementations.

Since we were focused on being able to convert existing LLMs to having these differential attention layers, we wanted a way to not degrade the performance of the (possibly pre-trained) LLM while doing so.

To this end, the conversion process _doubles_ the dimensionality of the query and key projections (since the differential attention requires both a positive and negative component of the attention) and (optionally; pass `--zero-init`) initializes the weights of the negative component to zero, while copying over the weights from the original attention modules to the positive components.

When doing this, the converted network computes the same function as the original (pass `--debug` to confirm this), but may suffer from a vanishing gradient problem. The default behavior is thus to initialize the weights of the negative components of the differential attention layers to 0-centered normally distributed values with a small variance.

Relevant links:

- [Paper](https://arxiv.org/abs/2410.05258)
- [Official implementation](https://github.com/microsoft/unilm/tree/master/Diff-Transformer)

## How has this been tested?

SmolLM2-135m on A40 Runpod instance on this feature branch. Workflow was:

- Convert the model to use either eager or SDPA differential attention
  - With and without `--zero-init` and `--debug` flags for sanity checking exact model conversion (completions, logits, losses)
- Run new `axolotl evaluate` command on the small `mhenrichsen/alpaca_2k_test` dataset with both the original and converted model and check that their evaluation metrics match

For example:

```shell
$ axolotl convert-diff-transformer ../configs/smollm.yaml --output-dir ../converted-model --zero-init --debug
...
[2024-12-17 05:15:26,910] [INFO] [axolotl.cli.convert_attention.convert_diff_transformer:75] [PID:94590] [RANK:0] Converting 
to differential attention...                                                                                                 
[2024-12-17 05:15:26,910] [INFO] [axolotl.integrations.diff_transformer.convert.convert_module:97] [PID:94590] [RANK:0] Conve
rting attention layer 0: LlamaSdpaAttention to LlamaDifferentialSdpaAttention                                                
[2024-12-17 05:15:26,921] [DEBUG] [axolotl.integrations.diff_transformer.convert.copy_attention_weights:64] [PID:94590] [RANK
:0] Copied positive attention weights from LlamaSdpaAttention to LlamaDifferentialSdpaAttention                              
[2024-12-17 05:15:26,921] [INFO] [axolotl.integrations.diff_transformer.convert.convert_module:97] [PID:94590] [RANK:0] Conve
rting attention layer 1: LlamaSdpaAttention to LlamaDifferentialSdpaAttention                                                
[2024-12-17 05:15:26,930] [DEBUG] [axolotl.integrations.diff_transformer.convert.copy_attention_weights:64] [PID:94590] [RANK
:0] Copied positive attention weights from LlamaSdpaAttention to LlamaDifferentialSdpaAttention
...
ANK:0] Converted 30 attention layers to differential attention
[2024-12-17 05:15:27,181] [INFO] [axolotl.cli.convert_attention.convert_diff_transformer:85] [PID:94590] [RANK:0] Testing con
verted model...
[2024-12-17 05:15:27,785] [INFO] [axolotl.cli.convert_attention.test_inference:43] [PID:94590] [RANK:0] Prompt: The quick brown fox                                                                                                                       
[2024-12-17 05:15:28,280] [INFO] [axolotl.cli.convert_attention.convert_diff_transformer:121] [PID:94590] [RANK:0] Generations match!
Model generation:
**************************************************
The quick brown fox jumps over the lazy dog

The quick brown fox jumps over the lazy dog.

The
**************************************************
```

## Types of changes

- `axolotl.integrations.diff_transformer` module, which implements the differential attention layers for the Llama LLM architecture and for various attention implementations (eager, SDPA, Flash Attention 2), and
- `axolotl.cli.integrations.convert_diff_transformer` module (and updates to `axolotl.cli.main`), which implements the `convert-diff-transformer` CLI command, and
- Monkeypatch in `axolotl.cli.integrations.convert_diff_transformer.patches` (to be moved) for updating `LLAMA_ATTENTION_CLASSES` constant in `transformers.models.llama.modeling_llama`.

## TODO

- [x] Test coverage
- [x] Add Flash Attention 2 implementation
- [x] Move monkey patch
- [x] Refactor conversion module as plugin
- [x] Add conversion with same-sized Q, K projections
- [x] Experiments to demonstrate value
  - [ ] Blog post


outputs
winglian, line 1: this symlink shouldn't be in the changeset, let's be sure to include this in .gitignore

model-out/eval_summary.csv
winglian, line 1: Should this be excluded from the changeset too?

djsaunde: Yeah, this was an accident.

src/axolotl/utils/yaml.py
```diff
@@ -0,0 +1,151 @@
...
+        if key not in ordered:
+            ordered[key] = data[key]
+
+    return ordered
+
+
+def dump_yaml_preserved_order(
+    data: Dict, reference_yaml_path: str, output_path: str
+) -> None:
+    """Dump YAML file while preserving nested order and normalized spacing."""
```
winglian, line 122: 🔥

ehartford:
I thought differential transformer requires model architecture change and modeling code change?  Does this somehow automatically implement a modeling.py for the model?

djsaunde:
> I thought differential transformer requires model architecture change and modeling code change? Does this somehow automatically implement a modeling.py for the model?

Good question. I've implemented a monkeypatch in `src/axolotl/monkeypatch/attention/differential.py` that updates the `PreTrainedModel._autoset_attn_implementation` function to be aware of the differential attention implementation. I think it's a bit of a hack, though, so using custom modeling code might be a good change before merge. Happy to hear your thoughts / feedback!

As for the architecture change, we have `src/axolotl/cli/integrations/convert_diff_transformer.py` which does the actually swapping of (llama only, for now) attention layers with differential attention in the model.

src/axolotl/utils/yaml.py
djsaunde: We could similarly have a function to normalize any config yaml file to have some expected ordering / formatting.

ehartford:
monkey patch only works in the context of Axolotl - we will need a modeling.py to make inference work properly in the wild (transformers, TGI, vllm, etc) right?  (If I understand correctly)

ehartford:
I have another question, don't be offended, it's just a question.

Does this really belong as a feature of axolotl, as opposed to perhaps mergekit?  or even a standalone project? 

Just that, axolotl is pretty focused on training - not on modifying model architecture, where mergekit focuses on modifying model architecture.

djsaunde:
I updated the PR to include a `modeling_diff_attn.py` file so that these modeling changes can be used outside of axolotl. Although, as discussed internally, [this PR to huggingface](https://github.com/huggingface/transformers/pull/34083) may have beaten us to the punch. We have a conversion logic, which sets us apart a bit, and can integrate it with this version of differential attention once merged.

djsaunde:
> I have another question, don't be offended, it's just a question.
> 
> Does this really belong as a feature of axolotl, as opposed to perhaps mergekit? or even a standalone project?
> 
> Just that, axolotl is pretty focused on training - not on modifying model architecture, where mergekit focuses on modifying model architecture.

Thanks for the question, definitely does not offend!

I think the value-add of this PR is moreso in the conversion logic than in the attention implementation itself (especially given the aforementioned HF PR), in that we can see users leveraging it to improve the performance of existing models sans differential attention / do some experimentation with models along these lines.

@winglian Care to weigh in here?

djsaunde:
~Should be good to go, we can merge this once the blog post is finalized IMO.~

Never mind, will have to refactor to remove the diff attention modeling code as it's now in upstream transformers. We can keep the conversion code, however.

PR #2200, opened 2024-12-18
refactor trainer to prevent circular dependencies later

winglian:


winglian:
this is included as part of #2202

PR #2201, opened 2024-12-19
remove cicd pytest xdist args

djsaunde:


outputs
winglian, line 1: let's remove this file/symlink

winglian (approved):

PR #2204, opened 2024-12-19
move the dataset loading from remote/disk to a shared function so we can re-use for RL

winglian:


NanoCode012 (approved):
This makes sense. We had inconsistent data loading between training method, so this looks like a good base

PR #2208, opened 2024-12-20
Initial test enablement for LLM Compressor

markurtz:
Still needs to be significantly tested, but starting point to play around with for LLM Compressor. This enables the session mixin for control of the training graph, loading of a teacher, and passing through everything within the axolotl training config.

Example config addition on top of the base config here https://github.com/axolotl-ai-cloud/axolotl/tree/main/examples/llama-3/fft-8b.yaml:
```yaml
....

compressor:
  recipe:
    finetuning_stage:
      run_type: train
      pruning_modifiers:
        ConstantPruningModifier:
          targets: '__ALL__'
          start: 0
  recipe_args: {}
  teacher:
    base_model: NousResearch/Meta-Llama-3.1-8B
    load_in_8bit: false
    load_in_4bit: false
    strict: false
    flash_attention: true
```

after setting up the config file, use the following command:
`axolotl train CONFIG.yaml`

### DIFF
--- a/src/axolotl/core/trainer_builder.py
+++ b/src/axolotl/core/trainer_builder.py
@@ -49,6 +49,7 @@ from trl import (
     RewardTrainer,
 )
 from trl.trainer.utils import RewardDataCollatorWithPadding, pad_to_length
+from llmcompressor.transformers.finetune.session_mixin import SessionManagerMixIn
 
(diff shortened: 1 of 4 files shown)


markurtz:
Accidentally opened against the main rather than our downstream fork. Will push up a proper PR once tested and ready!

PR #2210, opened 2024-12-22
fix untrained tokens if specified explicitly from a list

winglian:
this needs to get paired with a PR in our lgpl contlribs package.


winglian:
needs https://github.com/axolotl-ai-cloud/axolotl-contribs-lgpl/pull/2

PR #2213, opened 2024-12-22
support for custom lr groups for non-embedding modules

winglian:


src/axolotl/core/trainer_builder.py
```diff
@@ -462,11 +466,96 @@ def _wrap_model(self, model, training=True, dataloader=None):
...
+                params["embeddings"][name] = param
+            elif name in decay_parameters:
+                if lr_groups_lookup and any(
+                    group_modules in name for group_modules in lr_groups_lookup
+                ):
+                    lr_group_module = [
+                        group_modules
+                        for group_modules in lr_groups_lookup
+                        if group_modules in name
+                    ][0]
```
NanoCode012: I think we can refactor the check a bit. It seems like we're creating the list twice. The if check is not necessary as the variable would be empty otherwise

Perhaps
```py
lr_group_modules = [group_modules for group_modules in lr_groups_lookup if group_modules in name]
if any(lr_group_modules):
    # use lr_group_modules
```

NanoCode012: Is there a reason we're taking index 0 only? Should we add a check that `assert len(..) == 1`?

docs/lr_groups.qmd
```diff
@@ -0,0 +1,29 @@
...
+    modules:
+      - model.layers.2.self_attn.q_proj.weight
+    lr: 1e-5
+
+learning_rate: 2e-5
+```
+
+In this example, we have a default learning rate of 2e-5 across the entire model, but we have a separate learning rate
+of 1e-6 for all the self attention `o_proj` modules across all layers, and a learning are of 1e-5 to the 3rd layer's
+self attention `q_proj` module.
```
NanoCode012, line 29: Do we need to mention that it uses the same weight decay as global? Should we also open that custom option?

winglian: I think we can add custom weight decay in a follow-up pr

PR #2223, opened 2024-12-26
skip over rows in pretraining dataset

winglian:


NanoCode012:
When would this config be useful?

Could we have this added to `config.qmd` for pretraining only?

PR #2232, opened 2025-01-03
fix(doc): warn merging on cpu would silently cast model to fp16

NanoCode012:
# Description

Reported by Toastino on discord https://discord.com/channels/1104757954588196865/1111279858136383509/1324412403634405386

### DIFF
--- a/README.md
+++ b/README.md
@@ -729,7 +729,7 @@ You may need to use the `gpu_memory_limit` and/or `lora_on_cpu` config options t
 CUDA_VISIBLE_DEVICES="" python3 -m axolotl.cli.merge_lora ...
 ```
 
-although this will be very slow, and using the config options above are recommended instead.
+although this will be very slow, and using the config options above are recommended instead. Furthermore, the model output type will be in fp16 even if bf16 is set.
 
 ## Common Errors 🧰
 


README.md
```diff
@@ -729,7 +729,7 @@ You may need to use the `gpu_memory_limit` and/or `lora_on_cpu` config options t
 CUDA_VISIBLE_DEVICES="" python3 -m axolotl.cli.merge_lora ...
 ```
 
-although this will be very slow, and using the config options above are recommended instead.
+although this will be very slow, and using the config options above are recommended instead. Furthermore, the model output type will be in fp16 even if bf16 is set.
```
salmanmohammadi, line 732: Worth just slightly expanding here on the implications of this?

salmanmohammadi (approved):

README.md
NanoCode012: Hm, I'm not sure what much else to say? Do you have any suggestions?

NanoCode012:
My only thought left for this PR would be to add a warn log within validation as well, but I don't know if it's necessary.

README.md
salmanmohammadi: Something like "This may result in precision errors which could degrade model performance"?

salmanmohammadi:
> My only thought left for this PR would be to add a warn log within validation as well, but I don't know if it's necessary.

Doesn't hurt!

NanoCode012:
Moving to draft as I could not replicate the original issue. This however, may happen if we run with `CUDA_VISIBLE_DEVICES=""` and have `bf16: 'auto'` making it default to fp32/fp16 accordingly.

NanoCode012:
Will close while waiting for reporter to provide more info

PR #2235, opened 2025-01-05
feat: use SequentialSampler if `curriculum_sampling` is enabled with `sample_packing`

v-dicicco:
# Description

The `curriculum_sampling` flag introduced in https://github.com/axolotl-ai-cloud/axolotl/pull/1567 (cc @winglian) preserves dataset ordering by disabling shuffling, which simplifies experimenting with curriculum learning strategies. However, the flag is ignored if we use `sample_packing`. This PR sets the sampler to a `SequentialSampler` if the `curriculum_sampling` flag is enabled.

## How has this been tested?

I've verified that the DataLoader created in this way correctly yields batches with consecutive samples packed together. However, I'm not 100% sure whether other parameters of `MultipackBatchSampler` need adjustment if we potentially don't have a shuffled dataset (e.g., `packing_efficiency_estimate`).

Please note that `MultipackBatchSampler` is already used with a `SequentialSampler` in `_get_eval_sampler()`.

### DIFF
--- a/src/axolotl/core/trainer_builder.py
+++ b/src/axolotl/core/trainer_builder.py
@@ -608,8 +608,14 @@ class AxolotlTrainer(SchedulerMixin, Trainer):
                     self.state.train_batch_size or self.args.per_device_train_batch_size
                 )
                 batch_max_len = train_batch_size * self.args.max_seq_length
+
+            if self.args.curriculum_sampling:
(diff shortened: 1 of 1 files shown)


NanoCode012:
This is a good catch!

I don't think the other params of Multipack should be impacted from this, but pinging @winglian to double check.

winglian (approved):
looks good to me! thanks!

PR #2241, opened 2025-01-07
Process reward models

salmanmohammadi:
## Fixing `num_labels` for reward models
1) Reward models should be initialized with `num_labels=1`. I've added a `num_labels` field which is set in `cfg_kwargs`, rather than `model_kwargs`, as this field is parsed from `config.num_labels` [in transformers](https://github.com/huggingface/transformers/blob/665a4942e47a2d2f7a72c6a4c318b609b8461523/src/transformers/models/qwen2/modeling_qwen2.py#L875). This also allows us to [correctly initialize](https://github.com/huggingface/trl/blob/763738f457f283270772ac9bd5b3e4027fd424d5/examples/scripts/prm.py#L86) process reward models with `num_labels=2`.

## Adding support for process reward model training

I've added support for the `PRMTrainer` from `trl`, and also for the appropriate dataset format. Please see a screenshot of a successful training run below.

image

Resulting in the following trained model: https://huggingface.co/smohammadi/Qwen2.5-3B-MathShepherd


src/axolotl/prompt_strategies/stepwise_supervised.py
```diff
@@ -0,0 +1,118 @@
...
+from typing import Dict, Generator, List, Optional, Union
+
+from transformers import BatchEncoding, PreTrainedTokenizer
+
+from axolotl.prompt_tokenizers import IGNORE_INDEX, PromptTokenizingStrategy
+from axolotl.prompters import Prompter
+from axolotl.utils.dict import DictDefault
+
+
+class StepwiseSupervisedPromptTokenizingStrategy:
```
salmanmohammadi, line 15: @winglian This is how I ended up getting this working - is there any reason why I would specifically need to inherit from `PromptTokenizingStrategy` here? I'm vaguely gathering that using a `Prompter` and `PromptTokenizingStrategy` is important for inference?

src/axolotl/prompt_strategies/stepwise_supervised.py
```diff
@@ -0,0 +1,116 @@
...
+    - completions: a list of `n` completion steps
+    - labels: a list of `n` labels indicating the "correctness" of each step
+    """
+
+    def __init__(
+        self,
+        tokenizer,
+        train_on_inputs: bool = False,
+        sequence_len: int = 2048,
+        step_separator: str = "\n",
```
salmanmohammadi, line 28: @winglian what would be a sensible way to expose this in the config? As `cfg.dataset.step_seperator`, or as a top-level `cfg.step_seperator`?

winglian: I think a dataset level is good for now. I coukd envision a situation where it varies from dataset to dataset

src/axolotl/core/trainer_builder.py
```diff
@@ -1797,11 +1817,13 @@ def build(self, total_num_steps):
...
-            AxolotlTrainingArguments
-            if not self.cfg.reward_model
-            else AxolotlRewardConfig
-        )
+        if self.cfg.reward_model:
+            training_args_cls = AxolotlRewardConfig
+        elif self.cfg.process_reward_model:
+            training_args_cls = AxolotlPRMConfig
+        else:
+            training_args_cls = AxolotlTrainingArguments
```
winglian, line 1870: Now that we have a few of these, should we add a method similar to get_trainer_cls?

src/axolotl/prompt_strategies/stepwise_supervised.py
winglian: If there isn't anything shared between the two, this is good.

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
```diff
@@ -391,7 +391,7 @@ class ModelInputConfig(BaseModel):
 
     base_model: str
     base_model_config: Optional[str] = None
-    cls_model_config: Optional[str] = None
+    cls_model_config: Optional[str] = None  #
```
winglian: Stray #?

tests/e2e/test_process_reward_model_llama.py
```diff
@@ -0,0 +1,66 @@
...
+    """
+    Test case for Llama reward models using LoRA
+    """
+
+    @with_temp_dir
+    def test_rm_fft(self, temp_dir):
+        # pylint: disable=duplicate-code
+        cfg = DictDefault(
+            {
+                "base_model": "JackFram/llama-68m",
```
winglian: I've been preferring using smollm2 135m for e2e tests. It let's us check for train loss values as a gut check.

winglian (commented):
A couple of minor things, but should be good to go after.

tests/e2e/test_process_reward_model_llama.py
salmanmohammadi: Do you have tests I can look at where you check the loss?

winglian: see https://github.com/axolotl-ai-cloud/axolotl/blob/main/tests/e2e/multigpu/test_llama.py#L65-L91

big takeaways are that the yaml needs

```yaml
use_tensorboard: true
```

and tests needs

```python
    check_tensorboard(
        temp_dir + "/runs", "train/train_loss", 2.3, "Train Loss is too high"
    )
```

winglian: I usually set the test to 0.0 and see about where the actual train loss comes out to be

src/axolotl/core/trainer_builder.py
salmanmohammadi: I'll save this for a follow-up as I think the logic around this could do with a bit of a refactor. I'll assign myself an issue.

winglian (approved):
Looks great to me. Thanks!

PR #2243, opened 2025-01-08
assume empty lora dropout means 0.0 and add tests

winglian:
the previous behavior would fail if lora_dropout was None/null. This sets it as 0.0 at runtime if it is null.

### DIFF
--- a/src/axolotl/utils/models.py
+++ b/src/axolotl/utils/models.py
@@ -1322,7 +1322,7 @@ def load_lora(model, cfg, inference=False, config_only=False):
         target_modules=lora_target_modules,
         layers_to_transform=cfg.peft_layers_to_transform,
         layers_pattern=cfg.peft_layers_pattern,
-        lora_dropout=cfg.lora_dropout if cfg.lora_dropout is not None else 0.0,
+        lora_dropout=cfg.lora_dropout,
         fan_in_fan_out=cfg.lora_fan_in_fan_out,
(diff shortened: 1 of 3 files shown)


src/axolotl/utils/models.py
```diff
@@ -1322,7 +1322,7 @@ def load_lora(model, cfg, inference=False, config_only=False):
         target_modules=lora_target_modules,
         layers_to_transform=cfg.peft_layers_to_transform,
         layers_pattern=cfg.peft_layers_pattern,
-        lora_dropout=cfg.lora_dropout,
+        lora_dropout=cfg.lora_dropout if cfg.lora_dropout is not None else 0.0,
```
NanoCode012: Could we perhaps move this higher up in stack? Maybe to the validation stage instead? I think it's better to centralize this.

tests/test_lora.py
```diff
@@ -0,0 +1,51 @@
...
+                "lora_alpha": 16,
+                "lora_dropout": 0.0,
+                "lora_target_linear": True,
+                "micro_batch_size": 1,
+                "gradient_accumulation_steps": 1,
+                "sequence_len": 1024,
+            }
+        )
+        normalize_config(cfg)
+        cli_args = TrainerCliArgs()
```
NanoCode012: Do we need to create this `cli_args`? Seems like you could just pass boolean to `load_model` below.

PR #2245, opened 2025-01-09
add helper to verify the correct model output file exists

winglian:
# Description

Adds a helper function to check for the output model when doing e2e training tests.

## Motivation and Context

It can be cumbersome to remember which file to look for when adding the check to tests, so this handles all that logic.

## How has this been tested?

update tests!

### DIFF
--- a/tests/e2e/integrations/test_cut_cross_entropy.py
+++ b/tests/e2e/integrations/test_cut_cross_entropy.py
@@ -2,9 +2,8 @@
 Simple end-to-end test for Cut Cross Entropy integration
 """
 
-from pathlib import Path
-
 import pytest
+from e2e.utils import check_model_output_exists
 
 from axolotl.cli import load_datasets
 from axolotl.common.cli import TrainerCliArgs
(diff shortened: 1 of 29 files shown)


tests/e2e/utils.py
```diff
@@ -81,3 +83,22 @@ def check_tensorboard(
...
+    if cfg.save_safetensors:
+        if not cfg.adapter:
+            assert (Path(temp_dir) / "model.safetensors").exists()
+        else:
+            assert (Path(temp_dir) / "adapter_model.safetensors").exists()
+    else:
+        if not cfg.adapter:
+            assert (Path(temp_dir) / "pytorch_model.bin").exists()
+        else:
+            assert (Path(temp_dir) / "adapter_model.bin").exists()
```
NanoCode012: Do we need to check for edge case where the model is large and is shared across several files? The naming would not match in this case.

winglian: not really an issue since in our tests, almost everything is a small model that doesn't shard.

NanoCode012:
I saw that you didn't change all files in e2e like `test_dpo.py`. Even though they had `assert (Path(temp_dir) / "checkpoint-20/adapter_model.safetensors").exists()`

Should this be changed to use your new function but pass `Path(temp_dir) / "checkpoint-20"` as input path instead?

tests/e2e/integrations/test_cut_cross_entropy.py
```diff
@@ -2,9 +2,8 @@
 Simple end-to-end test for Cut Cross Entropy integration
 """
 
-from pathlib import Path
-
 import pytest
+from e2e.utils import check_model_output_exists
```
djsaunde: Nit: something weird is going on here with the sort order.

djsaunde (approved):
Just a nit.

PR #2247, opened 2025-01-09
add hf cache caching for GHA

winglian:
this should reduce some of our load to HF during tests and hopefully reduce the number of 503s in CI


winglian:
Screenshot 2025-01-09 at 11 52 19 AM
Screenshot 2025-01-09 at 11 52 42 AM

We can see the cache was saved and reloaded, and the subsequent tests ran about a minute faster since it didn't have to download the data.

djsaunde (approved):
LGTM, this is neat.

PR #2244, opened 2025-01-08
CLI cleanup and documentation

djsaunde:
# Description

This PR implements a cleanup / modularization of the code under `axolotl/cli`, where previously, many of the functions were lumped together in the `axolotl/cli/__init__.py` file. 

## Motivation and Context

These changes allow for better organization / readability and decluttering of the `axolotl.cli` namespace.

## How has this been tested?

CLI tests pass.

## Types of changes

Added doc strings, removed dead code, refactored methods etc. out of `axolotl/cli/__init__.py` into aptly-named modules.

PR #2251, opened 2025-01-10
Ray Train Axolotl Integration 

erictang000:
# Description
This PR aims to introduce a new path for launching training runs in axolotl via an integration with [Ray Train](https://docs.ray.io/en/latest/train/train.html) as an alternative to launching multi-gpu/node runs with `accelerate`. Details on how the Ray Train `TorchTrainer` integrates with Hugging Face `Accelerate` can be found on the [ray docs](https://docs.ray.io/en/latest/train/huggingface-accelerate.html).

We limit the scope of integration to using `ray.train` as a process launcher (instead of using accelerate launch from the cli) and wrapper of the existing `do_train` function. `ray.train` provides an option to do deeper integration to make dataset ingestion more efficient by integration with `ray.data` as well, but that is more invasive to the existing code-base and is not part of this proposal’s scope.

We introduce the following config fields for configuring ray runs

```python
class RayConfig(BaseModel):
    """Ray launcher configuration subset"""

    use_ray: bool
    ray_run_name: Optional[str]
    ray_num_workers: int
    resources_per_worker: Union[dict, str]
```

Note: there seem to be some weird integration points with the `addict` library `Dict` and `ray` object serialization, so there are some edge cases needed for testing what needs to be casted to and from the axolotl `DefaultDict` before and after entering the the ray `TorchTrainer` wrapped code.

## Motivation and Context
This change allows for multi-node runs on axolotl to be run from a single entry point given a ray cluster, rather than having to manually install axolotl and accelerate launch from each node, which is a noted [pain point](https://github.com/huggingface/accelerate/issues/1242) of using accelerate. 

Additionally, the use of `ray.train` allows serious cloud users to gain more control over the physical placement of distributed workers, and provides a unified layer for single / multi-node runs. On top of this it allows leveraging auto-scaling of ray clusters.

## How has this been tested?

This is a separate path for launching axolotl code, so it shouldn't impact core functionality.

To test that the ray code runs (on a 4 gpu ray cluster):
```bash
axolotl train examples/llama-3/lora-1b-ray.yml --use-ray 
```
On a multi-node ray cluster (for example with 2 nodes each with 4 gpus), simply run:
```bash
axolotl train examples/llama-3/lora-1b-ray.yml --use-ray --ray-num-workers 8
```

image

Successful training run on [wandb](https://wandb.ai/erictang000/huggingface/runs/ar1in7nw?nw=nwusererictang000
) from:
```bash
axolotl train examples/llama-3/lora-1b-ray.yml --use-ray 
```

## Social Handles (Optional)

submitted on behalf of x.com/anyscalecompute :)


winglian:
Thanks @erictang000 ! We have a slight CLI refactor happening in #2244 and I'll handle any merge conflicts with this PR after that one gets merged.

erictang000:
@winglian Sounds good thanks!

src/axolotl/utils/config/__init__.py
```diff
@@ -110,6 +111,17 @@ def normalize_config(cfg):
...
+    if cfg.deepspeed:
+        if isinstance(cfg.deepspeed, str) and os.path.exists(cfg.deepspeed):
+            ds_config_path = cfg.deepspeed
+            with open(ds_config_path, encoding="utf-8") as f:
+                cfg.deepspeed = json.load(f)
+    if cfg.fsdp:
+        if isinstance(cfg.fsdp, str) and os.path.exists(cfg.fsdp):
+            fsdp_config_path = cfg.fsdp
+            with open(fsdp_config_path, encoding="utf-8") as f:
+                cfg.fsdp = json.load(f)
```
NanoCode012: What does a fsdp config file look like?

examples/llama-3/lora-1b-ray.yml
```diff
@@ -0,0 +1,79 @@
...
+  - q_proj
+  - v_proj
+  - k_proj
+  - o_proj
+
+wandb_project: axolotl-ray
+wandb_entity: erictang000
+wandb_watch:
+wandb_name: axolotl-ray-llama-1b-single-node
+wandb_log_model:
```
NanoCode012: Could we leave these hyp empty?

src/axolotl/utils/config/__init__.py
erictang000: removed this

examples/llama-3/lora-1b-ray.yml
```diff
@@ -0,0 +1,79 @@
+base_model: NousResearch/Llama-3.2-1B
+# Automatically upload checkpoint and final model to HF
+# hub_model_id: username/custom_model_name
+
+load_in_8bit: false
```
NanoCode012, line 5: We should set this to true for LORA
```suggestion
load_in_8bit: true
```

cicd/Dockerfile.jinja
```diff
@@ -32,9 +32,9 @@ RUN if [ "$NIGHTLY_BUILD" = "true" ] ; then \
     fi
 
 RUN if [ "$AXOLOTL_EXTRAS" != "" ] ; then \
-        pip install --no-build-isolation -e .[deepspeed,flash-attn,optimizers,$AXOLOTL_EXTRAS] $AXOLOTL_ARGS; \
+        pip install --no-build-isolation -e .[deepspeed,flash-attn,optimizers,ray,$AXOLOTL_EXTRAS] $AXOLOTL_ARGS; \
```
NanoCode012, line 35: You would also need to make this change to 
- `docker/Dockerfile`
- `docker/Dockerfile-tests`

src/axolotl/cli/train.py
```diff
@@ -63,7 +65,44 @@ def do_cli(config: Union[Path, str] = Path("examples/"), **kwargs) -> None:
...
+            ray_train_func,
+            train_loop_config=train_loop_config,
+            scaling_config=ScalingConfig(
+                num_workers=parsed_cfg.ray_num_workers,
+                resources_per_worker=parsed_cfg.resources_per_worker.to_dict(),
+                use_gpu=True,
+            ),
+            run_config=RunConfig(
+                name=parsed_cfg.ray_run_name,
+                storage_path=Path("./saves").absolute().as_posix(),
```
NanoCode012:
```suggestion
                storage_path=Path(cfg.output_dir).absolute().as_posix(),
```

Should it specifically use `./saves` to work or can we use the same output dir as in yaml?

src/axolotl/cli/train.py
```diff
@@ -63,7 +65,44 @@ def do_cli(config: Union[Path, str] = Path("examples/"), **kwargs) -> None:
...
+        )
+        return trainer.fit()
+    return do_train(parsed_cfg, parsed_cli_args)
+
+
+def ray_train_func(kwargs: dict):
+    # cast `cfg` back to DictDefault (ray tune deepcopy has issues with DictDefault so needed it to be dict)
+    # also renormalize the config now that TorchTrainer has spawned distributed workers
+    cfg = DictDefault(kwargs["cfg"])
+    normalize_config(cfg)
```
NanoCode012, line 94: Do we need to add unittest to ensure the config is same before&after this cast?

cicd/Dockerfile.jinja
winglian: I'm just making sure it works in tests for now

examples/llama-3/lora-1b-ray.yml
winglian: I think it's fine leaving it as a half precision lora for now, especially when using deepspeed too. Pretty sure deepspeed with 8bit lora's isn't well supported on some hardware (like h100s)

winglian:
@erictang000 I think the last piece we would want for this is a bit of documentation around using Ray Train and possibly ray train w multi-node. Thanks!

SumanthRH:
Let's have a dedicated how-to guide for Ray <> Axolotl integration? A brief comment on setup (A couple of lines deferring to ray docs for details) and then we can mention how to use the CLI with the example yaml. (maybe a screenshot from ray dashboard for training progressing etc would be nice)

How does that sound @winglian

winglian:
@SumanthRH That sounds perfect! thanks!

src/axolotl/utils/config/__init__.py
```diff
@@ -83,7 +83,7 @@ def normalize_config(cfg):
         cfg.device_map = {"": int(os.environ.get("LOCAL_RANK", 0))}
         cfg.batch_size = cfg.batch_size * cfg.world_size
 
-    if cfg.bf16 == "auto":
+    if cfg.bf16 == "auto" and not cfg.use_ray: # if we use ray we want to defer this check to the worker node
```
SumanthRH: This looks ok, but there are a number of other checks right after this that depend on `cfg.bf16` being resolved to true/false. For example:

  https://github.com/axolotl-ai-cloud/axolotl/blob/8945874067ccb872a401a0c6941a6f6b993a9218/src/axolotl/utils/config/__init__.py#L86-L112
  
  
  I would suggest putting the dtype-related if conditions into a helper function 
  
  ```
  def resolve_dtype(cfg):
         ....  
  ``` 
  
  so that you can run the same logic on the worker nodes before training.  
  
  cc @winglian  for thoughts. 
  
  (Context: with ray, the head node - i.e the node that executes the training script is often a CPU-only node with the worker nodes being GPU-enabled. Axolotl's dtype resolution runs on the head node, which may not have a GPU, and thus no GPU-related libraries. We're trying to defer resolving dtypes to worker node).

erictang000: updated!

docs/ray-integration.qmd
```diff
@@ -0,0 +1,94 @@
...
+resources_per_worker:
+    accelerator_type:L40S: 0.001
+```
+
+## Launching training 
+
+You can simply run the following command on the head node: 
+
+```bash 
+axolotl train examples/llama-3/lora-1b-ray.yml --use-ray
```
kouroshHakha, line 84: does `use_ray` have to be present in both the yaml and the cli?

docs/ray-integration.qmd
```diff
@@ -0,0 +1,94 @@
...
+```
+
+This will launch training on the head node and workers will be scheduled automatically by Ray Train to run on the appropriate head/ worker nodes.
```
kouroshHakha:
```suggestion
This will launch training on the head node and workers will be scheduled automatically by Ray Train to run on the appropriate head or worker nodes. 
```

kouroshHakha (commented):
reviewed the docs. 👍

docs/ray-integration.qmd
erictang000: nope in this case it's redundant, the cli argument will overwrite the yaml value

erictang000: either specifying one or the other is fine

kouroshHakha: ok let'e clarify in the docs. It wasn't clear.

SumanthRH: @erictang000

SumanthRH: I added a clarification now. @erictang000 check it out

erictang000: lgtm

kouroshHakha (approved):

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
```diff
@@ -595,6 +595,30 @@ class GradioConfig(BaseModel):
...
+            "help": "The training results will be saved at `saves/ray_run_name`."
+        },
+    )
+    ray_num_workers: int = Field(
+        default=1,
+        metadata={
+            "help": "The number of workers for Ray training. Default is 1 worker."
+        },
+    )
+    resources_per_worker: Union[dict, str] = Field(
```
winglian: @erictang000 if this is set as a `str` type, what are valid values?

erictang000: @winglian good point, should just be dict - fixed!

PR #2253, opened 2025-01-11
[KD] add uld and jsd

kashif:
# Description

Add ULD loss and JSD KD Loss

# Paper

https://arxiv.org/abs/2402.12030

### DIFF
--- a/src/axolotl/integrations/kd/topk_logprob/uld.py
+++ b/src/axolotl/integrations/kd/topk_logprob/uld.py
@@ -1,0 +1,99 @@
+# Copyright 2025 Axolotl AI. All rights reserved.
+#
+# This software may be used and distributed according to
+# the terms of the Axolotl Community License Agreement (the "License");
+# you may not use this file except in compliance with the License.
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


src/axolotl/integrations/kd/topk_logprob/uld.py
```diff
@@ -0,0 +1,99 @@
...
+    student_probs_masked = student_probs[valid_mask]
+
+    # Get masked teacher probabilities
+    teacher_probs_masked = teacher_probs[valid_mask]
+
+    # Sort student probabilities in descending order
+    student_probs_sorted, _ = torch.sort(student_probs_masked, dim=-1, descending=True)
+
+    # For teacher probs, we already have top-K, so just ensure they're sorted
+    teacher_probs_sorted, _ = torch.sort(teacher_probs_masked, dim=-1, descending=True)
```
winglian, line 70: @kashif since the token_ids don't match, would it be better to sort, then. just take top_k of the student distribution?

kashif: so let me check with the definition of wasserstein loss

PR #2254, opened 2025-01-13
fix: use text_column even when not packing for pretraining

NanoCode012:
# Description

When running pretraining without `sample_packing` and using a non-`text` key for dataset, it would error

```
  File "/mnt/storage/axolotl/src/axolotl/utils/data/pretraining.py", line 24, in encode_pretraining
    examples["text"],
KeyError: 'text'
```

as it doesn't read the `text_column` attribute.

reported at https://discord.com/channels/1104757954588196865/1111279858136383509/1327211103104663613

## How has this been tested?

Not tested yet.

### DIFF
--- a/src/axolotl/utils/data/pretraining.py
+++ b/src/axolotl/utils/data/pretraining.py
@@ -203,7 +203,7 @@ def wrap_pretraining_dataset(
             encode_pretraining,
             tokenizer,
             max_tokens,
-            text_column=cfg.pretraining_dataset[0].text_column,
+            text_column=cfg.pretraining_dataset[0].text_column or "text",
         )
 
     if cfg.shuffle_merged_datasets:
--- a/src/axolotl/utils/data/pretraining.py
+++ b/src/axolotl/utils/data/pretraining.py
@@ -18,10 +18,13 @@ LOG = logging.getLogger("axolotl")
 
 
 def encode_pretraining(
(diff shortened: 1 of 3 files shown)


winglian (approved):

src/axolotl/utils/data/pretraining.py
```diff
@@ -196,7 +199,12 @@ def wrap_pretraining_dataset(
         # set this to 1 so downstream data_loader doesn't try to increase the batch again
         cfg.micro_batch_size = 1
     else:
-        encode = functools.partial(encode_pretraining, tokenizer, max_tokens)
+        encode = functools.partial(
+            encode_pretraining,
+            tokenizer,
+            max_tokens,
+            text_column=cfg.pretraining_dataset[0].text_column,
```
winglian:
```suggestion
            text_column=cfg.pretraining_dataset[0].text_column or "text",
```
looks like the e2e needs this?

winglian:
```suggestion
       encode_kwargs = {}
       if cfg.pretraining_dataset[0].text_column is not None:
           encode_kwargs["text_column"] = cfg.pretraining_dataset[0].text_column
        encode = functools.partial(
            encode_pretraining,
            tokenizer,
            max_tokens,
            **encode_kwargs,
```

PR #2260, opened 2025-01-14
feat: add config for optional parameters in a chat message

NJordan72:
# Description

Currently only role/content properties are passed into the chat template (with the exception of some hardcoded optional params that are model specific).  This PR creates a configuration property to add additional optional parameters.

## Motivation and Context

- Chat templates are not restricted to using role/content properties and we should allow users, especially those creating their own custom chat templates to pass in anything via the message.
- OPEN QUESTION: Why restrict the properties passed into the chat template in the first place?  Couldn't we just pass the entire message in and if the template doesn't use a particular property they just become no-ops?

## How has this been tested?

- Added unit test


src/axolotl/prompt_strategies/chat_template.py
```diff
@@ -44,6 +45,13 @@ def __init__(
                 "tool": "tool",
             }
 
+        # Default optional fields, kept for compatibility
+        self._default_optional_fields = ["tool_calls", "name", "tool_call_id"]
```
winglian: Can we include the comments/descriptions on these fields from `get_conversation_thread`?

src/axolotl/prompt_strategies/chat_template.py
```diff
@@ -44,6 +45,13 @@ def __init__(
...
+
+        self.optional_message_fields = self._default_optional_fields + (
+            optional_message_fields or []
+        )
+
```
winglian: Might be helpful to use some sort of set operation or other uniqueness check in case a user duplicates in their configuration a key from the default optional fields

tests/prompt_strategies/test_chat_templates_advanced.py
```diff
@@ -912,5 +912,84 @@ def verify_labels(labels_span, should_train, context_message):
...
+            roles_to_train=["assistant"],
+        )
+
+        res = strategy.tokenize_prompt(test_data)
+        turns = strategy.get_conversation_thread(test_data)
+        labels = res["labels"]
+        input_ids = res["input_ids"]
+
+        LOG.info(f"Labels: {labels}")
+        LOG.info(f"Input IDs: {input_ids}")
```
winglian:
```suggestion
        LOG.debug(f"Labels: {labels}")
        LOG.debug(f"Input IDs: {input_ids}")
```
log probably not needed in the tests, but can convert these to debug if we want to keep them.

tests/prompt_strategies/test_chat_templates_advanced.py
```diff
@@ -912,5 +912,84 @@ def verify_labels(labels_span, should_train, context_message):
...
+
+        # Verify both optional fields are in the tokenized output
+        decoded_output = tokenizer.decode(input_ids)
+
+        LOG.info(f"Decoded output: {decoded_output}")
```
winglian:
```suggestion
        LOG.debug(f"Decoded output: {decoded_output}")
```

winglian (commented):
A few minor nits, but overall looks like a good improvement to chat templates.

NJordan72:
> A few minor nits, but overall looks like a good improvement to chat templates.

fixed.

docs/dataset-formats/conversation.qmd
```diff
@@ -46,6 +46,9 @@ datasets:
     field_messages: conversations
     message_field_role: from
     message_field_content: value
+    optional_message_fields:
+
+
```
NanoCode012: I don't think we need this part here to keep it short and simple
```suggestion
```

tests/prompt_strategies/test_chat_templates_advanced.py
```diff
@@ -912,5 +912,84 @@ def verify_labels(labels_span, should_train, context_message):
...
+        assert all(
+            label == IGNORE_TOKEN_ID for label in turn_labels
+        ), "Expected user message content to be not to be trained on"
+
+        # Verify all content is properly labeled for assistant message
+        start_idx, end_idx = strategy.find_turn(turns=turns, turn_idx=2)
+        turn_labels = labels[start_idx:end_idx]
+        assert not all(
+            label == IGNORE_TOKEN_ID for label in turn_labels
+        ), "Expected assistant message content to be trained on"
```
NanoCode012:
```suggestion
        assert all(
            label != IGNORE_TOKEN_ID for label in turn_labels
        ), "Expected assistant message content to be trained on"
```

I think this is the condition we want?

Edit: This test may fail until you set `train_on_eos: turn`.

docs/config.qmd
```diff
@@ -137,6 +137,8 @@ datasets:
     message_field_role: role
     # Key for content in each message (default:  "content")
     message_field_content: content
+    # Fields that will be passed to the chat template if present in the message. (Optional[List[str]], default: None)
+    optional_message_fields:
```
NanoCode012: Should we say we include `tool_calls`, `name`... by default based on your implementation?

NanoCode012 (commented):
Thanks for the PR. 

> Why restrict the properties passed into the chat template in the first place?

The initial code was written when it was mainly role/content. I expanded it recently for tool_calling datasets, so this is a welcome change.

>  Couldn't we just pass the entire message in and if the template doesn't use a particular property they just become no-ops?

This could be an alternative, but I think the current solution works fine?

docs/config.qmd
NJordan72: Happy to go either way.  Right now we include them as hardcoded, but I'd be happy to refactor it to just make them the default.

NJordan72: I guess the downside of making it the default is that if someone wants to add an additional field, they -- at that point -- have to remember to add back the defaults as well.

It is why my slight preference is just to pass the entire message into the chat template and let the template itself decide what it cares about.

NJordan72: The other option... Would be to parse out all of the arguments from the jinja template and pass in any/all of those.  Jinja provides a pretty easy way to extract the variables from the AST (per documentation, I haven't tried it yet).

We could even warn if the template is expecting something it isn't being passed which might be a nice quality of life addition.

winglian: Also, should this be a `List[str]` (current)

or a `List[Dict[str, str]]` which would allow us to map alternate field names to the field in the jinja template?

The latter more closely matches the meaning of the other args like `message_field_content`, `message_field_role`

NJordan72:
I just updated this to be a little more thoughtful / robust...

- I removed the `optional_message_fields` option and replaced it with a `message_property_mappings` option.
- `message_property_mappings` is basically a more generalized version of the `message_field_role` and `message_field_content` configuration options.  Those continue to work and are just sucked into the `message_property_mappings` if provided.  The idea is you can map any input property to any message property using this configuration.
- I then had to update `chat_template.py` to use the new mappings configuration property to transform the message.
- I added a `JinjaTemplateAnalyzer` that introspects the jinja template to figure out which properties a template is expecting (this is generalized, but it can also look at the properties specifically for the messages, which is what we care about here.
- We then remove any properties from the transformed message that the template does not need/expect.
- I've taken out special handling of `tool_call_id`, etc. as it is no longer needed an is handled by the introspection of the template itself
- I've added test coverage for all of the new stuff and all of the old tests continue to pass

tests/prompt_strategies/test_chat_templates.py
```diff
@@ -74,8 +76,10 @@ def test_llama3(self, llama3_tokenizer, assistant_dataset):
             ChatTemplatePrompter(
                 llama3_tokenizer,
                 chat_template=get_chat_template("llama3"),
-                message_field_role="role",
-                message_field_content="content",
+                message_property_mappings={
+                    "from": "role",
+                    "value": "content",
+                },
```
winglian: I think it might be better to invert the property mappings so that the key is `role` and `content`. Because we always know we need to expect those keys, and the mapping is the field in the dataset that we should expect to find them.

Just for the sake of completeness and being able to verify the previous tests in CI and ensure backwards compatibility, It might be good to keep these tests with `message_field_role`/`message_field_content` and add a new test suite for message_property_mappings, even if only for a single model architecture.

winglian:
Here's the stack trace locally of the broken test in CI currently

```
  File "/Users/wing/Projects/ml/axolotl/src/axolotl/prompt_strategies/chat_template.py", line 275, in tokenize_prompt
    input_ids = self.prompter.build_prompt(turns)
  File "/Users/wing/Projects/ml/axolotl/src/axolotl/prompt_strategies/chat_template.py", line 84, in build_prompt
    return self.tokenizer.apply_chat_template(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/wing/.pyenv/versions/3.11.10/envs/axolotl-3.11/lib/python3.11/site-packages/transformers/tokenization_utils_base.py", line 1683, in apply_chat_template
    rendered_chat = compiled_template.render(
                    ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/wing/.pyenv/versions/3.11.10/envs/axolotl-3.11/lib/python3.11/site-packages/jinja2/environment.py", line 1304, in render
    self.environment.handle_exception()
  File "/Users/wing/.pyenv/versions/3.11.10/envs/axolotl-3.11/lib/python3.11/site-packages/jinja2/environment.py", line 939, in handle_exception
    raise rewrite_traceback_stack(source=source)
  File "<template>", line 1, in top-level template code
jinja2.exceptions.UndefinedError: 'dict object' has no attribute 'role'
```

and the dataset configuration that is working on main but is broken in the tests

```yaml
chat_template: llama3
datasets:
  - path: mlabonne/FineTome-100k
    type: chat_template
    split: train[:20%]
    field_messages: conversations
    message_field_role: from
    message_field_content: value
```

tests/prompt_strategies/test_chat_templates.py
NJordan72: on it!

NJordan72: These specific tests are hard to keep as they were, because the signature of the function is no longer expecting `message_field_role/message_field_content`

Instead we are shuttling `message_field_role/message_field_content` into `message_property_mappings` as part of the config validation.

I have added a test to the model to make sure they are indeed making it into the mappings.

NJordan72:
Still looking into the e2e failure (if this push didn't fix it).

NJordan72:
I tracked down the problem with the failing e2e test.  I'm sure there are some further nits, feedback which I'm happy to incorporate.  Some other thoughts below that explain some of the changes and then maybe a broader topic that we can take off of this thread if there is any interest.

--

I will say that the one challenge I had is there seems to be a lack of consistency as to if the config that is getting passed around to all of the various classes/functions is supposed to be the `Dict` representation of the Pydantic model or the model itself.

Many of the tests are obviously just constructing a `Dict` that represents the config which obviously makes sense as to why they would want to pass that around, but it feels like it would be nicer if the Pydantic model was the lingua franca across the entire project.

It feels like the config is the "special sauce" in some ways for the project.  Instead of having to write 1000 lines of boilerplate code to get something running, just craft a fairly simple (I'm probably being a little generous here) config file and you're on your way.  I know a lot of the complexity of the project lies in how to deal with that config file, but it feels like the config itself is something of a second class citizen insofar that the `Dict` representation of it seems to be used in a bunch of places and there are bits of the config that seem under documented and somewhat unclear.

Maybe I'm not seeing the forest from the trees here (and Python is certainly not my bailiwick) , but I'd be interested in helping tighten things up around this if there is any interest in it.

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
```diff
@@ -311,6 +311,9 @@ class KTODataset(BaseModel):
     revision: Optional[str] = None
 
 
+DatasetConfig = Union[SFTDataset, DPODataset, KTODataset]
```
winglian: Did you intend to incorporate this into the line far below?

```
datasets: Optional[conlist(Union[SFTDataset, DPODataset, KTODataset], min_length=1)] = None  # type: ignore
```

winglian:
@NJordan72 Thanks for your thoughts and insights. We agree that the docs are definitely lacking and we are on a mission this first few months of the year to improve in that area.  The evolution of the project started with a basic DefaultDict with naive validation functions written against it. Only last year did we convert the validation to use pydantic, hence the sort of mismatch of using pydantic up front and then the conversion to a dict after that.  We're definitely open to making the pydantic model the common language through the project, but in places there will still be a mismatch when we're using the lower level dataclasses that we're extending from HF transformers. I'd love your thoughts on how you might approach this.

tests/e2e/patched/test_fa_xentropy.py
```diff
@@ -76,7 +76,9 @@ def test_lora_packing_fa_cross_entropy(self, temp_dir, gradient_accumulation_ste
         else:
             cfg.fp16 = True
 
+        validate_config(cfg)
         normalize_config(cfg)
```
winglian: this is a good call to add the validation in this test as it surfaces a new bug with parsing the chat_dataset. It seems we don't do this consistently across the rest of the tests, and seems like we should add this to more of them to make sure the change for the chat_dataset class doesn't affect other logic.

src/axolotl/prompt_strategies/chat_template.py
```diff
@@ -44,8 +44,10 @@ def __init__(
                 "tool": "tool",
             }
 
-        self.message_field_role = message_field_role
-        self.message_field_content = message_field_content
+        self._chat_template_msg_variables = self.get_chat_template_msg_variables(
+            chat_template
+        )
```
NanoCode012: If I understand correctly, `messages_array_name` refers to the key of the `List[dict]`. Should we change the signature so that `field_messages` is passed to the Prompter as well? To allow passing `messages_array_name=field_messages`?

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
```diff
@@ -184,6 +186,66 @@ class SFTDataset(BaseModel):
...
+            ):
+                raise ValueError(
+                    f"Conflicting message role fields: message_field_role='{data['message_field_role']}' "
+                    f"conflicts with message_property_mappings.role='{data['message_property_mappings']['role']}'"
+                )
+            data["message_property_mappings"]["role"] = (
+                data["message_field_role"] or "role"
+            )
+        elif "role" not in data["message_property_mappings"]:
+            data["message_property_mappings"]["role"] = "role"
```
NanoCode012: After setting the property, should we drop the old fields? (Same for content below)

src/axolotl/prompt_strategies/messages/chat.py
```diff
@@ -51,17 +51,16 @@ def load(tokenizer, cfg, ds_cfg: Optional[Dict[str, Any]] = None):
...
     if field_messages:
         builder_kwargs["conversations_field"] = field_messages
-    if message_field_role:
-        builder_kwargs["message_field_role"] = message_field_role
-    if message_field_content:
-        builder_kwargs["message_field_content"] = message_field_content
+    if message_property_mappings and "role" in message_property_mappings:
+        builder_kwargs["message_field_role"] = message_property_mappings["role"]
+    if message_property_mappings and "content" in message_property_mappings:
+        builder_kwargs["message_field_content"] = message_property_mappings["content"]
```
winglian: These changes should probably reverted and go into chat_template.py. We can address changing the messages type in a later PR.

NJordan72:
fyi, will crank the changes out today and tomorrow.

src/axolotl/prompt_strategies/messages/chat.py
NJordan72: reverted in the most recent revision

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
NJordan72: I think I'll leave the property there for now.  My intention is to come back and implement #2271 in a way that gives us much better type safety and validation logic, at which point it should be easier to to

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
NJordan72: yep. thanks.

src/axolotl/prompt_strategies/chat_template.py
NJordan72: Good catch.  Just updated that.

NJordan72:
@winglian @NanoCode012 looks like the remaining failure is unrelated to my changes here

src/axolotl/utils/data/test.sql.s
winglian, line 1: did you accidentally commit this?

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
```diff
@@ -1644,3 +1669,65 @@ def check_torch_compile_auto(cls, data):
...
+        ):
+            raise ValueError(
+                f"Conflicting message content fields: message_field_content='{data['message_field_content']}' "
+                f"conflicts with message_property_mappings.content='{data['message_property_mappings']['content']}'"
+            )
+        data["message_property_mappings"]["content"] = (
+            data["message_field_content"] or "content"
+        )
+    elif "content" not in data["message_property_mappings"]:
+        data["message_property_mappings"]["content"] = "content"
```
winglian: Since it seems like we're starting a deprecation of message_field_role and message_field_content by referring to them as legacy, we should probably raise a deprecation warning here. Also, we should update https://github.com/axolotl-ai-cloud/axolotl/blob/main/scripts/chat_datasets.py#L33-L54 to reflect the new preferred schema.

NJordan72: added the warning and updated that script

winglian:
@NJordan72 I rebased this from one of your earlier commits against latest main, and cherry-picked a few of your later fixes to clean up the commit history on this. I hope I managed to grab all the rest of your recent changes.

NJordan72:
> @NJordan72 I rebased this from one of your earlier commits against latest main, and cherry-picked a few of your later fixes to clean up the commit history on this. I hope I managed to grab all the rest of your recent changes.

lgtm

src/axolotl/utils/dict.py
```diff
@@ -13,3 +13,26 @@ def __missing__(self, key):
...
+        try:
+            p = object.__getattribute__(self, "__parent")
+            key = object.__getattribute__(self, "__key")
+        except AttributeError:
+            p = None
+            key = None
+        if p is not None:
+            p[key] = self
+            object.__delattr__(self, "__parent")
+            object.__delattr__(self, "__key")
```
NanoCode012, line 38: What does this section do exactly? Why do we need it?

src/axolotl/prompt_strategies/chat_template.py
```diff
@@ -25,10 +26,10 @@ def __init__(
         processor=None,
         chat_template=None,
         max_length=2048,
-        message_field_role: str = "role",
-        message_field_content: str = "content",
+        message_property_mappings: Optional[Dict[str, str]] = None,
         message_field_training: Optional[str] = None,
         message_field_training_detail: Optional[str] = None,
+        messages_array_name: str = "messages",
```
NanoCode012: I'm of the opinion to remove `_messages` and its utility function from ChatTemplateStrategy and to call it via `self.prompter.messages_array_name`. This way, we won't have to pass the same variable to two places.

This change would also require updating all tests to remove `strategy.messages = ...` and within the `load()` within this file.

src/axolotl/prompt_strategies/chat_template.py
```diff
@@ -212,6 +225,10 @@ def __init__(
         self.train_on_eos = train_on_eos
         self.images = "images"
 
+        LOG.info(
```
NanoCode012:
```suggestion
        LOG.debug(
```

Lower the priority log level here.

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
```diff
@@ -893,6 +902,14 @@ def deprecate_sharegpt_datasets(cls, datasets):
...
         return datasets
 
+    @field_serializer("datasets")
+    def datasets_serializer(
+        self, ds_configs: Optional[List[DatasetConfig]]
+    ) -> Optional[List[Dict[str, Any]]]:
+        if ds_configs:
+            return [ds_config.model_dump() for ds_config in ds_configs]
+        return None
+
```
NanoCode012: I did some local tests and noticed a lot of keys which were previously omitted, now exist within the cfg as `key: None`. This messes with some settings like: `"messages_array_name": dataset_config.get("field_messages", "messages")`, returning `messages_array_name: None` (which causes `get_conversation_thread` to fail at  `prompt[self.messages]`).

This doesn't cause error in our tests because we haven't called `strategy.messages` or manually set it ourselves.

Removing this section fixed it for me.

src/axolotl/prompt_strategies/chat_template.py
```diff
@@ -424,61 +441,82 @@ def find_turn(self, turns: list[dict], turn_idx: int):
...
-            "tool_calls",  # tool that 'assistant' calls
-            "name",  # name of tool given by 'tool'
-            "tool_call_id",  # mistral/mixtral requires this
-        ]
         for message in prompt[self.messages]:
+            transformed_message = self.transform_message(message)
+
             turn = {
-                "role": self.prompter.roles[message[self.prompter.message_field_role]],
+                **transformed_message,
```
NanoCode012, line 492: Would we need to worry about any nested copying here? Should we set `turn = transformed_message` then set the other two fields after?

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
```diff
@@ -214,26 +225,40 @@ def check_chat_template_config(cls, data):
...
     data_files: Optional[List[str]] = None
+    trust_remote_code: Optional[bool] = False
     revision: Optional[str] = None
+    field_system: Optional[str] = None
+    field_prompt: Optional[str] = None
+    field_completion: Optional[str] = None
+    field_messages: Optional[str] = None
+    field_label: Optional[bool] = None
+    prompt_format: Optional[str] = None
+    completion_format: Optional[str] = None
```
NanoCode012: I almost missed this as we didn't have test to cover it.

This would change the nested layer for UserDefined for DPO/KTO. What is the motivation behind this change?

src/axolotl/utils/dict.py
winglian: I added this, something is happening in the distributed handling where the config get's pickled, but because there is now a DictDefault as a nested dataset configuration, it has to unpickle that and it fails.

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
NJordan72: Reverted.

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
NJordan72: Added a exlude_none to the model_dump

src/axolotl/prompt_strategies/chat_template.py
NJordan72: Removed.

NJordan72:
@winglian @NanoCode012 sorry for the spam here, but I rebased this with main and fixed what i think are the remaining problems.

winglian:
thanks, I rebased it again, @NanoCode012 let's get this merged this week

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
```diff
@@ -683,17 +702,15 @@ class Config:
     ] = None  # whether to use weighting in DPO trainer. If none, default is false in the trainer.
     dpo_use_logits_to_keep: Optional[bool] = None
 
-    datasets: Optional[conlist(Union[SFTDataset, DPODataset, KTODataset, StepwiseSupervisedDataset], min_length=1)] = None  # type: ignore
-    test_datasets: Optional[conlist(Union[SFTDataset, DPODataset, KTODataset, StepwiseSupervisedDataset], min_length=1)] = None  # type: ignore
+    datasets: Optional[conlist(DatasetConfig, min_length=1)] = None  # type: ignore
+    test_datasets: Optional[conlist(DatasetConfig, min_length=1)] = None  # type: ignore
```
NanoCode012, line 706: This will have merge conflict with the #2309

NanoCode012 (approved):
Thank you the PR. Sorry it took some time for the review as it had many files.

I rebased it again and removed ref to the configs in docs/yml.

We might need to address type conflict with PR #2309 mentioned.

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
NJordan72: If you want to merge this one first, I can address the conflict in #2309.

NanoCode012: Sounds good!

tests/test_validation_dataset.py
```diff
@@ -302,3 +302,22 @@ def test_dataset_sharegpt_deprecation(self, minimal_cfg):
...
+            minimal_cfg
+            | {
+                "datasets": [
+                    {
+                        "path": "mhenrichsen/alpaca_2k_test",
+                        "type": "alpaca",
+                        "message_property_mappings": {
+                            "role": "role",
+                            "content": "content",
+                        },
```
winglian, line 317: Is this just to validate that the config validates? Not actually verifying against the dataset, right?

NJordan72: correct

PR #2263, opened 2025-01-16
option to not concatenate during pretraining

winglian:
currently, the default for pertaining without sample packing is to soft pack/concatenate samples together still, but not handle cross attention contamination. This adds an option to simply disable concatenation too during pretraining.

### DIFF
--- a/src/axolotl/utils/data/pretraining.py
+++ b/src/axolotl/utils/data/pretraining.py
@@ -22,6 +22,7 @@ def encode_pretraining(
     max_tokens: int,
     examples: Dict[str, List],
     text_column: str = "text",
+    concatenate: bool = True,
 ) -> Dict[str, List]:
     res = tokenizer(
         examples[text_column],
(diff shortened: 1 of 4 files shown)


src/axolotl/utils/data/pretraining.py
```diff
@@ -198,6 +206,14 @@ def wrap_pretraining_dataset(
...
         # set this to 1 so downstream data_loader doesn't try to increase the batch again
         cfg.micro_batch_size = 1
+    elif cfg.pretraining_sample_concatenation is False:
+        encode = functools.partial(
+            encode_pretraining,
+            tokenizer,
+            max_tokens,
+            text_column=cfg.pretraining_dataset[0].text_column or "text",
+            concatenate=False,
+        )
```
djsaunde:
```suggestion
    else cfg.pretraining_sample_concatenation is False:
        encode = functools.partial(
            encode_pretraining,
            tokenizer,
            max_tokens,
            text_column=cfg.pretraining_dataset[0].text_column or "text",
            concatenate=cfg.pretraining_sample_concatenation is True,
        )
```
And remove the code below.

djsaunde (approved):
Just a non-blocking comment.

Should we also add a line to https://axolotl-ai-cloud.github.io/axolotl/docs/config.html?

PR #2269, opened 2025-01-20
add support for include_tokens_per_second in training args

winglian:
# Description

### DIFF
--- a/docs/config.qmd
+++ b/docs/config.qmd
@@ -371,7 +371,7 @@ save_total_limit: # Checkpoints saved at a time
 # e.g., when 1 epoch is 1000 steps => `num_epochs: 2` and `max_steps: 100` will train for 100 steps
 max_steps:
 
-# bool of whether to include tokens trainer per second in the training metrics
+# bool of whether to include tokens trainer per second in the training metrics. This iterates over the entire dataset once, so it takes some time.
 include_tokens_per_second:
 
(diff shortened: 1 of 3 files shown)


docs/config.qmd
```diff
@@ -369,6 +369,9 @@ save_total_limit: # Checkpoints saved at a time
 # e.g., when 1 epoch is 1000 steps => `num_epochs: 2` and `max_steps: 100` will train for 100 steps
 max_steps:
 
+# bool of whether to include tokens trainer per second in the training metrics
```
```suggestion
# bool of whether to include tokens trainer per second in the training metrics. This iterates over the entire dataset once, so it takes some time.
```

Add more details to doc following hf doc: https://huggingface.co/docs/transformers/en/main_classes/trainer#transformers.TrainingArguments.include_tokens_per_second

src/axolotl/core/trainer_builder.py
```diff
@@ -1428,6 +1428,11 @@ def build(self, total_num_steps):
         )
 
         training_arguments_kwargs = {}
+
+        training_arguments_kwargs[
+            "include_tokens_per_second"
+        ] = self.cfg.include_tokens_per_second
```
NanoCode012:
```suggestion
        if self.cfg.include_tokens_per_second is not None:
            training_arguments_kwargs[
                "include_tokens_per_second"
            ] = self.cfg.include_tokens_per_second
```

The trainer expects boolean with False being default. Since we default to None, maybe this is better? (spacing in edit might be off a bit)

PR #2273, opened 2025-01-21
Hf trainer refactor

winglian:


winglian:
multigpu tests: https://github.com/axolotl-ai-cloud/axolotl/actions/runs/12891388223

PR #2278, opened 2025-01-22
Pretrain multipack

winglian:
multipack for pretraining had some wonky edge cases that were broken and untested

### DIFF
--- a/src/axolotl/core/trainer_builder.py
+++ b/src/axolotl/core/trainer_builder.py
@@ -1880,6 +1880,8 @@ class HFCausalTrainerBuilder(TrainerBuilderBase):
         if training_args.pretraining:
             if self.cfg.pretraining_sample_concatenation is False:
                 return DataCollatorForSeq2Seq(self.tokenizer, **kwargs)
+            if self.cfg.micro_batch_size > 1:
+                return DataCollatorForSeq2Seq(self.tokenizer, **kwargs)
             return None
 
         if self.cfg.model_config_type == "mamba":
--- a/src/axolotl/utils/data/pretraining.py
+++ b/src/axolotl/utils/data/pretraining.py
@@ -245,7 +245,7 @@ def encode_packed_pretraining(
     examples: Dict[str, List],
     max_seq_length: int = 2048,
     batch_size: int = 4,
-    multipack_attn: Optional[bool] = False,
+    multipack_attn: Optional[bool] = True,
 ) -> Dict[str, List]:
     # pylint: disable=duplicate-code
     # tokenize all the examples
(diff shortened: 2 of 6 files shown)


src/axolotl/core/trainer_builder.py
```diff
@@ -1879,6 +1879,8 @@ def build_collator(
         if training_args.pretraining:
             if self.cfg.pretraining_sample_concatenation is False:
                 return DataCollatorForSeq2Seq(self.tokenizer, **kwargs)
+            if self.cfg.micro_batch_size > 1:
+                return DataCollatorForSeq2Seq(self.tokenizer, **kwargs)
```
NanoCode012, line 1884: Are we expecting this to be the original `micro_batch_size` or would it have been overwritten in `wrap_pretraining_dataset` to be set to 1?

src/axolotl/utils/data/pretraining.py
```diff
@@ -248,8 +246,6 @@ def encode_packed_pretraining(
     max_seq_length: int = 2048,
     batch_size: int = 4,
     multipack_attn: Optional[bool] = False,
```
NanoCode012: Should `multipack_attn` default to True following Pydantic model?

tests/e2e/test_llama_pretrain.py
```diff
@@ -68,3 +75,12 @@ def test_pretrain(self, temp_dir, sample_packing):
 
         train(cfg=cfg, dataset_meta=dataset_meta)
         check_model_output_exists(temp_dir, cfg)
+        loss_threshold = 3.5
+        if sample_packing and not pretrain_multipack_attn:
+            loss_threshold = 6.5
```
NanoCode012, line 80: Why would the threshold be this high without multipack_attn?

tests/test_packed_pretraining.py
```diff
@@ -87,9 +88,9 @@ def test_packing_stream_dataset(self):
             assert data["labels"].shape == torch.Size(
                 [1, original_bsz * cfg.sequence_len]
             )
-            assert data["attention_mask"].shape == torch.Size(
-                [1, original_bsz * cfg.sequence_len]
-            )
+            # assert data["attention_mask"].shape == torch.Size(
```
NanoCode012, line 93: Perhaps , we can add `assert "attention_mask" not in data`

tests/e2e/test_llama_pretrain.py
winglian: it's the same as concatenation so their is cross-sample contamination resulting in a higher loss

PR #2282, opened 2025-01-23
Num epochs float

mashdragon:
# Description

Lets `num_epochs` be a float in the config instead of restricting it to ints. Huggingface takes the parameter as a float anyway.

## Motivation and Context

Often you want to train on 1.2 epochs or 0.4 epochs instead of 1 or 2. This lets you do that.

## How has this been tested?

I ran preprocess + train once on a config that had `sample_packing` set to `True`. The training began for a few steps and then I ended it.

(The extremely light testing is due to me having not enough time and minimal compute resources.)

### DIFF
--- a/src/axolotl/utils/config/models/input/v0_4_1/__init__.py
+++ b/src/axolotl/utils/config/models/input/v0_4_1/__init__.py
@@ -489,7 +489,7 @@ class HyperparametersConfig(BaseModel):
     adam_beta1: Optional[float] = None
     adam_beta2: Optional[float] = None
     max_grad_norm: Optional[float] = None
-    num_epochs: int = Field(default=1)
+    num_epochs: float = Field(default=1.0)
 
     @field_validator("batch_size")
     @classmethod
--- a/src/axolotl/utils/trainer.py
+++ b/src/axolotl/utils/trainer.py
@@ -374,7 +374,7 @@ def calculate_total_num_steps(cfg, train_dataset, update=True):
         if cfg.sample_packing_eff_est:
             total_num_steps = (
                 # match count to len est in dataloader
-                (
+                int(
                     math.floor(
                         0.99
                         * cfg.total_num_tokens


winglian (approved):

NanoCode012 (changes requested):
Given that num_epochs can be < 1 now, we may want to check here as well https://github.com/axolotl-ai-cloud/axolotl/blob/8fb72cbc0b94129141bae5fa4d84edd23b648af6/src/axolotl/utils/config/__init__.py#L113-L120

mashdragon:
I'd add the changes myself, but truthfully I'm not sure what to do. Why doesn't the math work out? In any case we could maybe just do `max(1, cfg.num_epochs)` but I don't think that's what we want, right?

NanoCode012:
My thought on this section was: if num_epochs is a fraction (say `0.3` with `saves_per_epoch: 2`), this would cause the `save_steps=1.6` which would then silently not be passed to the trainer.

I'm unclear myself on the best way to handle this. Perhaps we require a validation check: `num_epochs > 1` to use `saves_per_epoch`?

mashdragon:
I was thinking about it from a different angle: If I want it to save with a frequency of 4 times every epoch (`saves_per_epoch` = 4) when I train with `num_epochs` = 4, then when I reduce it to `num_epochs` = 0.5 I expect it to save only twice.

If you have `num_epochs` = 0.3, then you would need to set `saves_per_epoch` to `6.667` for it to save twice within the 0.3 epochs. Personally, it feels to me that we should keep this behaviour. Otherwise aren't we redefining what an epoch is?

mashdragon:
We could also just log a warning if `saves_per_epoch` pushes `save_steps` out of range.

NanoCode012:
> If I want it to save with a frequency of 4 times every epoch (`saves_per_epoch` = 4) when I train with `num_epochs` = 4, then when I reduce it to `num_epochs` = 0.5 I expect it to save only twice.

Yes, I agree with you here. It's just, the calculation for `save_steps` is wrong if num_epochs < 1. In earlier example, `0.3 with saves_per_epoch: 2` would have `save_steps: 1.6` which would attempt to save every 1.6 steps, instead of at the right point (`epoch=0.5`). This example is not the best (given `epoch=0.3 < 0.5` when it should've saved), but I hope you get what I mean.

mashdragon:
Yes, sorry - I think I am just very confused because I am not quite sure what `save_steps` entails, and if `save_steps` is 1.6, the control path `save_steps < 1.0` would not be followed, and so there would not be a problem since the option wouldn't even be used, right? Or is there a big difference between, say, `num_epochs = 1, saves_per_epoch = 1` and omitting `saves_per_epoch` from the config altogether?

In a more usual case, let's say we have `num_epochs = 2` and `saves_per_epoch = 2`, then `save_steps = 0.25`. "Steps" is a strange term because steps are usually atomic, but in this code even a regular case has fractional steps ( < 1), so I don't really know what steps means here.

I have "Allow edits by maintainers" enabled if you would like to apply changes yourself, but I will be happy to fix it once I fully understand.

NanoCode012 (approved):
Hey, I discussed with wing a bit about this. The issue is more with our original code than your PR, so we plan to fix it in a separate PR to keep things more organized and get this through first.

Apologies for any confusion.

PR #2293, opened 2025-01-28
Defaulting to `fused=True` AdamW

salmanmohammadi:
# Description

Changing optimizers to `adamw_torch_fused` as the default, as it is generally faster (in both single-device and distributed settings)


winglian (approved):
once the single and multi-gpu tests pass, should be good to go

winglian:
rebased against main to include your latest changes to drop support for 2.3.1

PR #2296, opened 2025-01-29
better handling of multipack dataset length

winglian:
by sampling multiple times per rank, we can more confidently ascertain the minimum dataset packing length.


src/axolotl/utils/samplers/multipack.py
```diff
@@ -195,42 +201,15 @@ def calc_sample_packing_eff_est(estimates: List[float]):
...
-                world_size
-                * math.floor(
-                    0.99
-                    * lengths_sum_per_device
-                    / efficiency
-                    // (self.batch_max_len * self.batch_size)
-                )
-                - 1
-            ),
-        )
```
winglian, line 236: fyi, this is dead code

PR #2307, opened 2025-02-03
GRPO

winglian:
wip towards adding support for GRPO


src/axolotl/core/trainers/grpo/__init__.py
```diff
@@ -0,0 +1,64 @@
...
+        return grpo_args_kwargs
+
+    @classmethod
+    def set_trainer_kwargs(cls, cfg):
+        trainer_kwargs = {}
+        if cfg.grpo_reward_funcs:
+            reward_funcs = []
+            for reward_func_module in cfg.grpo_reward_funcs:
+                # use importlib to dynamically load the reward function from the module
+                reward_func_module_name = reward_func_module.split(".")[-1]
```
salmanmohammadi: worth inspecting/asserting the signature of the reward fns?

winglian: how exactly do you have this in mind?

winglian: also `RewardFunc` also accepts strings which could be a pretrained model id, so we should probably wrap with a try / except ModuleNotFoundError.

salmanmohammadi: done!

src/axolotl/utils/config/models/input/v0_4_1/trl.py
```diff
@@ -0,0 +1,32 @@
+"""
+GRPO specific configuration args
+"""
+from typing import List, Optional
+
+from pydantic import BaseModel, Field
+
+
+class TrlConfig(BaseModel):
```
salmanmohammadi: This looks great : )

src/axolotl/core/trainers/grpo/trainer.py
```diff
@@ -0,0 +1,306 @@
...
+                        apply_chat_template(x, reward_processing_class)["text"]
+                        for x in messages
+                    ]
+                else:
+                    texts = [p + c for p, c in zip(prompts, completions)]
+                reward_inputs = reward_processing_class(
+                    texts,
+                    return_tensors="pt",
+                    padding=True,
+                    padding_side="right",
```
abodacs: why there padding_side mismatched between `reward_inputs` and `prompt_inputs`?

src/axolotl/utils/trainer.py
```diff
@@ -576,7 +576,7 @@ def prepare_opinionated_env(cfg):
 def setup_trainer(
     cfg, train_dataset, eval_dataset, model, tokenizer, processor, total_num_steps
 ):
-    if cfg.rl in ("dpo", "ipo", "orpo", "kto", "simpo"):
+    if cfg.rl in ("dpo", "grpo", "ipo", "orpo", "kto", "simpo"):
```
salmanmohammadi: I think we can just check for `if cfg.rl` here right?

src/axolotl/core/trainers/grpo/trainer.py
winglian: this was a copy/paste from upstream trl trainer as I needed this PR to land to simplify our end. https://github.com/huggingface/trl/pull/2817/files I've removed this method from this class now that the referenced PR is merged.

winglian:
I also need https://github.com/axolotl-ai-cloud/axolotl/pull/2193 merged to handle the named config from the dataset we're using for the grpo test.

PR #2311, opened 2025-02-05
feat: add torch2.6 to ci

NanoCode012:
# Description

Added support for torch 2.6.

Questions:
- [ ] I see that we build python 3.10 torch 2.4 in our `base.yml` but not used anywhere else. Should we remove it?
- [ ] I saw python 3.10 still included in some test but simultaneously excluded in our `tests-nightly.yml`. Should we remove it?

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,6 +26,11 @@ jobs:
             pytorch: 2.5.1
             axolotl_extras:
             is_latest: true
+          - cuda: 124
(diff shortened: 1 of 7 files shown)


winglian:
Once this CI https://github.com/axolotl-ai-cloud/axolotl/actions/runs/13159506894/job/36724412671 finishes, we can re-run the docker based tests too

winglian (approved):

PR #2319, opened 2025-02-07
Update faq.qmd

bursteratom:
Added Q&A for being stuck on saving preprocessed datasets

# Description

### DIFF
--- a/docs/faq.qmd
+++ b/docs/faq.qmd
@@ -22,4 +22,4 @@ description: Frequently asked questions
 
 **Q: The codes is stuck on saving preprocessed datasets.**
 
-> A: This is usually an issue with the GPU. This can be resolved through setting the os environment variable `CUDA_VISIBLE_DEVICES=0`. If you are on runpod, this is usually pod issue and starting a new pod should solve the issue. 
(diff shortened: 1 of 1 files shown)


docs/faq.qmd
```diff
@@ -19,3 +19,7 @@ description: Frequently asked questions
 **Q: AttributeError: 'DummyOptim' object has no attribute 'step'**
 
 > A: You may be using deepspeed with single gpu. Please don't set `deepspeed:` in yaml or cli.
+
+**Q: The codes is stuck on saving preprocessed datasets.**
+
+> A: This is usually an issue with the machine. On runpod, starting a new pod should solve the issue.
```
winglian: also, doing preprocessing on CPU-only (e.g. `CUDA_VISIBLE_DEVICES=""`,  or using `NCCL_P2P_DISABLE=1` will solve these issues

bursteratom: Thank you for the additional info! I've updated the q&a.

NanoCode012:
I usually see this issue for multi-gpu, which is solved by setting `CUDA_VISIBLE_DEVICES=0` would help.

I used to recall, setting it to run fully on CPU may have conflict with our `bf16: True` hardware check.

Therefore, my usual recommendation is to set `CUDA_VISIBLE_DEVICES=0` over `CUDA_VISIBLE_DEVICES=""`

bursteratom:
@NanoCode012 @winglian yup, just incorporated your suggested changes. Lmk if this is in good shape

NanoCode012 (approved):

PR #2322, opened 2025-02-09
Add `bos_token` and `add_generation_prompt` to the alpaca chat template

minpeter:
# Description

Fixes and resolves issues with the alpaca template used when using the axolotl inference feature.

AS-IS
```
### Instruction: Describe the structure of an atom.

<Completion starts here>
```

TO-BE
```
<bos_token>### Instruction:
Describe the structure of an atom.

### Response:
<Completion starts here>
```

## Social Handles (Optional)

discord: minpeter


NanoCode012 (approved):
Thanks for catching that!

xzuyn:
That template would still have issues. It can't handle system turns, and also leaves `\n\n` after the last turn.
```
<bos_token>### Instruction:
<user turn>

### Response:
<assistant turn><eos_token>

### Instruction:
<user turn>

### Response:
<assistant turn><eos_token>


```

---

This template would fix those. Due to the way the original alpaca format handles system prompts, I've limited it to only having a system prompt on the first turn.

```jinja
{{ bos_token }}{% for message in messages %}{% if message['role'] == 'system' and loop.first %}{{ message['content'] }}{% elif message['role'] == 'user' %}{{ '### Instruction:\n' + message['content'] }}{% elif message['role'] == 'assistant' %}{{ '### Response:\n' + message['content'] + eos_token }}{% endif %}{% if not loop.last %}{{ '\n\n' }}{% endif %}{% endfor %}{% if add_generation_prompt %}{{ '\n\n### Response:\n' }}{% endif %}
```
```
<bos_token><optional system turn>

### Instruction:
<user turn>

### Response:
<assistant turn><eos_token>

### Instruction:
<user turn>

### Response:
<assistant turn><eos_token>
```

minpeter:
> That template would still have issues. It can't handle system turns, and also leaves `\n\n` after the last turn.
> 
> ```
> <bos_token>### Instruction:
> <user turn>
> 
> ### Response:
> <assistant turn><eos_token>
> 
> ### Instruction:
> <user turn>
> 
> ### Response:
> <assistant turn><eos_token>
> ```
> 
> This template would fix those. Due to the way the original alpaca format handles system prompts, I've limited it to only having a system prompt on the first turn.
> 
> ```
> {{ bos_token }}{% for message in messages %}{% if message['role'] == 'system' and loop.first %}{{ message['content'] }}{% elif message['role'] == 'user' %}{{ '### Instruction:\n' + message['content'] }}{% elif message['role'] == 'assistant' %}{{ '### Response:\n' + message['content'] + eos_token }}{% endif %}{% if not loop.last %}{{ '\n\n' }}{% endif %}{% endfor %}{% if add_generation_prompt %}{{ '\n\n### Response:\n' }}{% endif %}
> ```
> 
> ```
> <bos_token><optional system turn>
> 
> ### Instruction:
> <user turn>
> 
> ### Response:
> <assistant turn><eos_token>
> 
> ### Instruction:
> <user turn>
> 
> ### Response:
> <assistant turn><eos_token>
> ```

Adding \n\n in the last turn is not a big deal. In fact, the generation is stopped by eos before that part is executed. However, if you think the above change is essential, you can apply it. However, from my point of view, it seems like an unnecessary increase in complexity.

minpeter:
Please review the changes and proceed with merging..! .cc @NanoCode012

NanoCode012:
Since we're adding an eos_token as well to separate turns (along others), let me think this through as we're modifying a common template.

NanoCode012 (approved):
Okay, I think it's fine. Thanks for the PR!

PR #2332, opened 2025-02-13
Fix sample packing producing longer sequences than specified by `sequence_len`

tobmi1:
# Description

Return tokenized sequence lengths in `get_dataset_lengths` directly from the `Dataset` object instead of accessing the underlying PyArrow table. Filtering the dataset (e.g., by applying `drop_long_seq_in_dataset`) would update the `Dataset` object but not the underlying table. Thus, the sequence lengths returned by `get_dataset_lengths` would no longer correspond to the correct sequences and the sample packing algorithm may produce samples longer than `cfg.sequence_len`.

## Motivation and Context

Training with sample packing and `pad_to_sequence_len=True` should result in stable VRAM usage. However, due to this bug, sequence length may differ between batches, which caused increased VRAM usage and even OOMs in the middle of training in my case. Fine-tuning Phi-3.5 with `sequence_len=4096`, `micro_batch_size=1`, `gradient_accumulation_steps=4`, and `gradient_checkpointing=False` went from random OOMs after ~500 steps to using stable ~42GB VRAM on a single A100 80GB with the fix.

Note that pretraining does not seem to be affected since the underlying in-memory table is updated along with the `IterableDataset` when applying any filters.

Pre-processing the dataset in a separate step and then loading the prepared dataset for training should theoretically also circumvent this issue, but I was not able confirm this yet.

## How has this been tested?

I extended `test_packed_batch_sampler.py` to use `drop_long_seq_in_dataset` before applying the `MultipackBatchSampler` and parametrized the test to additionally use a `sequence_len` of 512 so that some samples are actually dropped from the dataset. Running the extended test without the fix fails, while it runs successfully on my setup with the fix.

My env:
OS: Windows 10 (WSL Ubuntu 24.04)
Python: 3.12.9
Torch: 2.5.1 (CPU-only)

### DIFF
--- a/src/axolotl/utils/samplers/utils.py
+++ b/src/axolotl/utils/samplers/utils.py
@@ -5,12 +5,12 @@ import numpy as np
 
 
 def get_dataset_lengths(dataset):
-    if "length" in dataset.data.column_names:
-        lengths = np.array(dataset.data.column("length"))
-    elif "position_ids" in dataset.data.column_names:
-        position_ids = dataset.data.column("position_ids")
+    if "length" in dataset.column_names:
+        lengths = np.array(dataset["length"])
+    elif "position_ids" in dataset.column_names:
+        position_ids = dataset["position_ids"]
         lengths = np.array([x[-1] + 1 for x in position_ids])
     else:
-        input_ids = dataset.data.column("input_ids")
+        input_ids = dataset["input_ids"]
         lengths = np.vectorize(len)(np.array(input_ids, dtype=object))
     return lengths
--- a/src/axolotl/utils/samplers/utils.py
+++ b/src/axolotl/utils/samplers/utils.py
@@ -12,5 +12,5 @@ def get_dataset_lengths(dataset):
         lengths = np.array([x[-1] + 1 for x in position_ids])
     else:
         input_ids = dataset["input_ids"]
-        lengths = np.vectorize(len)(np.array(input_ids, dtype=object))
+        lengths = np.array([len(seq) for seq in input_ids])
     return lengths
--- a/tests/test_packed_batch_sampler.py
+++ b/tests/test_packed_batch_sampler.py
@@ -7,6 +7,7 @@ from transformers import AutoTokenizer
 from axolotl.datasets import TokenizedPromptDataset
 from axolotl.prompt_strategies.completion import load
 from axolotl.utils.collators import V2BatchSamplerDataCollatorForSeq2Seq
+from axolotl.utils.data.utils import drop_long_seq_in_dataset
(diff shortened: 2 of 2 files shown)


NanoCode012 (dismissed):
Thank you for catching this. I wonder whether this is the cause for those silent vram increases.

NanoCode012:
Tests are failing with 

```
>           dataset, _ = load_tokenized_prepared_datasets(
                self.tokenizer, cfg, prepared_path
            )

tests/test_datasets.py:231: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
src/axolotl/utils/data/sft.py:325: in load_tokenized_prepared_datasets
    dataset = drop_long_seq_in_dataset(dataset, cfg)
src/axolotl/utils/data/utils.py:175: in drop_long_seq_in_dataset
    min_input_len = np.min(get_dataset_lengths(dataset))
src/axolotl/utils/samplers/utils.py:15: in get_dataset_lengths
    lengths = np.vectorize(len)(np.array(input_ids, dtype=object))

...

            inputs = [arg.flat[0] for arg in args]
>           outputs = func(*inputs)
E           TypeError: object of type 'int' has no len()
```

src/axolotl/utils/samplers/utils.py
```diff
@@ -5,12 +5,12 @@
...
-        position_ids = dataset.data.column("position_ids")
+    if "length" in dataset.column_names:
+        lengths = np.array(dataset["length"])
+    elif "position_ids" in dataset.column_names:
+        position_ids = dataset["position_ids"]
         lengths = np.array([x[-1] + 1 for x in position_ids])
     else:
-        input_ids = dataset.data.column("input_ids")
+        input_ids = dataset["input_ids"]
         lengths = np.vectorize(len)(np.array(input_ids, dtype=object))
```
NanoCode012: Maybe try this change? Untested!

```suggestion
        lengths = np.array([len(seq) for seq in input_ids])
```

Since np.vectorize is not for perf but convenience (https://numpy.org/doc/stable/reference/generated/numpy.vectorize.html) , it shouldn't be much different.

tobmi1: Thank you for the suggestion! The tests in `test_datasets.py` are now running fine on my machine. However, a bunch of others (e.g. in `tests/patched/test_validation.py`) are still failing for seemingly unrelated reasons. I hope this is just due to my env with WSL

NanoCode012 (approved):
Checked the test, confirmed error appears with prev code.

PR #2344, opened 2025-02-18
Bumping 0.15.1 TRL version for GRPO+PEFT fix

salmanmohammadi:


winglian (approved):
ran some quick tests and this looks good to go

PR #2347, opened 2025-02-19
Correctly reference mount paths

### LINKED ISSUE #2346: Cloud running with Modal is broken due to mount issue

### Expected Behavior

The following cloud config should work:

```yaml
provider: modal
gpu: l40s
gpu_count: 1
timeout: 86400
volumes:
- mount: /workspace/cache
  name: axolotl-cache
env:
- WANDB_API_KEY
- HF_TOKEN
```

### Current behaviour

Running `axolotl train --config config.yaml --cloud ./cloud_config.yaml` crashes once booted inside Modal, saying that the `config.yaml` doesn't exist in the Docker container. This is because — starting from line 262 in `src/axolotl/cli/cloud/modal_.py` — the config paths are all hardcoded to `/workspace/artifacts/axolotl`, but on line 48 of `src/axolotl/cli/__init__.py`, all local directory data is mounted to `/workspace/mounts` instead.

Mounting to `/workspace/mounts` makes sense to me personally: I assume hardcoding `/workspace/artifacts/axolotl` is an accidental artifact left in from testing this behavior on axolotl itself. Nonetheless, it's broken and AFAICT there is no valid configuration that could make it work, since the mounting code copies the files to one place (which is hardcoded), and the training code expects the files to be in a different place (also hardcoded).

### Steps to reproduce

Take any working axolotl config, and the `cloud_config.yaml` above — or any `cloud_config.yaml` — and try to run it. It won't work because the hardcoded paths are out of sync.

### Config yaml

```yaml
literally any config.yaml
```

### Possible solution

Instead of hardcoding the lookup paths to `/workspace/artifacts/axolotl`, hardcode them to `/workspace/mounts`. I cloned the Axolotl repo, did this, and it works.

- [x] Linux
- [ ] macOS
- [ ] Windows

### Python Version

3.11

### axolotl branch-commit

8dfadc2b3c7f85e4208b74eb93c32f440c7e25b4

### Acknowledgements

reissbaker:
# Description

Fixes the mount paths referenced in the Axolotl code that runs inside the Modal containers to correctly match the mount path set in `src/axolotl/cli/cloud/__init__.py`.

## Motivation and Context

Fixes https://github.com/axolotl-ai-cloud/axolotl/issues/2346

## How has this been tested?

I ran `axolotl train ./config.yaml --cloud ./cloud_config.yaml` and it no longer crashed immediately as soon as the container booted inside Modal due to the hardcoded mismatched mount paths.

FWIW, it appears that `preprocess` and `lm_eval` actually never mount any local files or directories at all in `__init__.py`, and so I suspect they will continue to be broken since none of the configs will be copied over (they already are broken in master), but this fixes `train`, at least.

I'm working on a fix for `preprocess`, which I can add as a followup PR as it's a more involved change.

## Types of changes

Bugfix

## Social Handles (Optional)

[@reissbaker](https://x.com/reissbaker) on Twitter

### DIFF
--- a/src/axolotl/cli/cloud/modal_.py
+++ b/src/axolotl/cli/cloud/modal_.py
@@ -261,9 +261,7 @@ class ModalCloud(Cloud):
 
 def _preprocess(config_yaml: str, volumes=None):
     Path("/workspace/mounts").mkdir(parents=True, exist_ok=True)
-    with open(
-        "/workspace/mounts/config.yaml", "w", encoding="utf-8"
-    ) as f_out:
+    with open("/workspace/mounts/config.yaml", "w", encoding="utf-8") as f_out:
         f_out.write(config_yaml)
     run_folder = "/workspace/mounts"
     run_cmd(
@@ -274,9 +272,7 @@ def _preprocess(config_yaml: str, volumes=None):
 
 
 def _train(config_yaml: str, accelerate: bool = True, volumes=None, **kwargs):
-    with open(
-        "/workspace/mounts/config.yaml", "w", encoding="utf-8"
-    ) as f_out:
+    with open("/workspace/mounts/config.yaml", "w", encoding="utf-8") as f_out:
         f_out.write(config_yaml)
     run_folder = "/workspace/mounts"
     if accelerate:
@@ -294,9 +290,7 @@ def _train(config_yaml: str, accelerate: bool = True, volumes=None, **kwargs):
 
 
 def _lm_eval(config_yaml: str, volumes=None):
-    with open(
-        "/workspace/mounts/config.yaml", "w", encoding="utf-8"
-    ) as f_out:
+    with open("/workspace/mounts/config.yaml", "w", encoding="utf-8") as f_out:
         f_out.write(config_yaml)
     run_folder = "/workspace/mounts"
     run_cmd(
--- a/src/axolotl/cli/cloud/modal_.py
+++ b/src/axolotl/cli/cloud/modal_.py
@@ -295,12 +295,12 @@ def _train(config_yaml: str, accelerate: bool = True, volumes=None, **kwargs):
 
 def _lm_eval(config_yaml: str, volumes=None):
     with open(
-        "/workspace/artifacts/axolotl/config.yaml", "w", encoding="utf-8"
+        "/workspace/mounts/config.yaml", "w", encoding="utf-8"
     ) as f_out:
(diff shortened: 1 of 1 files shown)


winglian:
Thanks, aren't mount paths specific to the configured cloud configuration?  see https://github.com/axolotl-ai-cloud/axolotl-cookbook/blob/main/grpo/cloud.yaml#L2-L6 and https://github.com/axolotl-ai-cloud/axolotl/blob/main/examples/cloud/modal.yaml#L3-L6


Happy to accept this PR, but is this something we can solve with docs? is /workspace/mounts the default for modal already?

reissbaker:
You are referencing volume mounts, which are configured via the `cloud_config.yaml`. However, *where Axolotl itself uploads its configuration* appears to be hardcoded in Axolotl, in two different places:

1. Where it actually uploads the configuration ([line 48 of `__init__.py`](https://github.com/axolotl-ai-cloud/axolotl/blob/8dfadc2b3c7f85e4208b74eb93c32f440c7e25b4/src/axolotl/cli/cloud/__init__.py#L48): `/workspace/mounts/config.yaml`)
2. Where it later tries to read the configuration from ([line 278, among others, of `modal_.py`](https://github.com/axolotl-ai-cloud/axolotl/blob/8dfadc2b3c7f85e4208b74eb93c32f440c7e25b4/src/axolotl/cli/cloud/modal_.py#L278): `/workspace/artifacts/axolotl/config.yaml`)

This means that the Modal integration is completely broken: it's impossible to have a working config, since Axolotl itself hardcodes two different paths (as described in the linked issue, #2346 ). Every attempt to use this feature will instantly crash on Modal.

Totally happy for you to make it configurable where Axolotl uploads its config files. However, currently there's no way in the `cloud_config.yaml` to specify where you want it to upload those files: it's just hardcoded in Axolotl. The only configurable thing in the `cloud_config.yaml` is volume mounts; none of them specify that they should be used for anything specific e.g. config file uploads (and you can have multiple... At which point, how do you decide which to use?). This was meant as a small bugfix PR to get the existing feature to work.

winglian:
Oh, I see what you mean now. Thanks for walking me through it 🙏 . good to go!

winglian (approved):

reissbaker:
Thanks @winglian! For some reason, here's what I see on my end:

image

Do I need specific permissions to be able to merge? Or, happy if you hit the merge button if I'm not able to.

PR #2359, opened 2025-02-24
Telemetry implementation

djsaunde:
# Description

Telemetry for Axolotl. Implementation of system described in #2323.

Features the following:
- `axolotl.telemetry.manager`: `TelemetryManager` class that interacts with the PostHog Python SDK and sends telemetry. Handles opt-out (via `DO_NOT_TRACK` or `AXOLOTL_DO_NOT_TRACK` env vars), organization whitelisting (all `base_model`s except for those that are whitelisted are redacted), and path sanitization (`_path` or `_dir` config (key, value) pairs are redacted).
- `axolotl.telemetry.errors`: Function decorator implementation for tracking errors in various important methods in the codebase (`preprocess`, `train`, `evaluate`, `inference`, etc.). Includes path sanitization to remove local paths, but keeps axolotl paths / other Python library paths for debugging purposes.
- `axolotl.telemetry.runtime_metrics`: Dataclass for metrics tracked during model training, including timing metrics, memory metrics, and elapsed steps / epochs.
- `axolotl.telemetry.callbacks`: Includes callback for trainer to send telemetry related to the aforementioned runtime metrics class.

## Motivation and Context

See #2323.

## How has this been tested?

Pytest tests for each added module + some manual testing and viewing of telemetry in PostHog console.


src/axolotl/telemetry/manager.py
```diff
@@ -0,0 +1,252 @@
...
+
+        # Create new dict with redacted values
+        redacted = {k: redact_value(v, k) for k, v in properties.items()}
+
+        return redacted
+
+    def _get_system_info(self) -> dict[str, Any]:
+        """Collect system information"""
+        gpu_info = []
+        if torch.cuda.is_available():
```
NanoCode012, line 280: Do we need to consider for NPU/MPS?

src/axolotl/telemetry/manager.py
```diff
@@ -0,0 +1,252 @@
...
+        if not properties:
+            return {}
+
+        # NOTE: Keep this up to date with any config schema changes
+        path_indicators = {"path", "dir"}
+
+        def redact_value(value: Any, key: str = "") -> Any:
+            """Recursively sanitize values, redacting those with path-like keys"""
+            # Special case: base_model should be redacted if org is not whitelisted
+            if key == "base_model":
```
NanoCode012: Could we include `base_model_config`, `tokenizer_config`, `lora_model_dir`, `resume_from_checkpoint`, `hub_model_id`, `pretraining_dataset`(can be str path to hub) ?

Secondly, I'm not sure which part this fits in but we would also need to redact wandb / mlflow / comet (especially if comet_api_key is passed via yaml)

NanoCode012 (commented):
Just quick look, do we need to disable telemetry on CI runs or move them to a separate group tracking to not skew the data?

src/axolotl/telemetry/manager.py
djsaunde: `lora_model_dir` should be covered by the `path_indicators` logic, but thanks for pointing out these other keys.

And yes, let's definitely redact all API keys.

djsaunde: Actually, the `pretrained_dataset` `path` will be redacted due to the `path_indicators` logic as well.

NanoCode012: > Actually, the `pretrained_dataset` `path` will be redacted due to the `path_indicators` logic as well.

Oh sorry I wasn't clear. These are the cases I mean`pretraining_dataset: hub_org/hub_name`. Due to old legacy code, it also supports being a `str`. Is this covered? (We should probably deprecate this path though).

djsaunde: Ah, no, I didn't realize this. I will update to cover.

docs/telemetry.qmd
```diff
@@ -0,0 +1,59 @@
...
+
+The telemetry system will block training startup for 15 seconds to ensure users are
+aware of data collection, unless telemetry is explicitly enabled or disabled.
+
+## Opt-Out Mechanism
+
+Telemetry is **enabled by default** on an opt-out basis. To disable it, set either:
+
+- `AXOLOTL_DO_NOT_TRACK=1` (Axolotl-specific)
+- `DO_NOT_TRACK=1` (Global standard)
```
winglian: It would be good to link to https://consoledonottrack.com/

src/axolotl/telemetry/whitelist.yaml
```diff
@@ -0,0 +1,12 @@
...
+  - "huggingface"
+  - "nvidia"
+  - "facebook"
+  - "mistralai"
+  - "briaai"
+  - "unsloth"
+  - "NousResearch"
+  - "allenai"
+  - "amd"
+  - "tiiuae"
```
winglian, line 18: [axolotl-ai-co](https://huggingface.co/axolotl-ai-co/) :D

winglian: deepseek-ai
microsoft (phi, etc)
google
huggingfaceTB
Qwen
allenai

src/axolotl/telemetry/manager.py
```diff
@@ -0,0 +1,269 @@
...
+            cls._instance._initialized = False
+
+        return cls._instance
+
+    def __init__(self):
+        """Telemetry manager initializer"""
+        if self._initialized:
+            return
+
+        self.enabled, self.explicit_enable = self._check_telemetry_enabled()
```
winglian, line 137: would these be better as a @property? Is there a world where the env var for do not track gets dynamically set sometime after and we need to account for that then?

src/axolotl/telemetry/errors.py
```diff
@@ -0,0 +1,160 @@
...
+                    f"{module.__name__}.{func.__name__}" if module else func.__name__
+                )
+
+                # Get stack trace
+                stack_trace = "".join(
+                    traceback.format_exception(
+                        type(exception), exception, exception.__traceback__
+                    )
+                )
+                stack_trace = sanitize_stack_trace(stack_trace)
```
winglian, line 147: iirc, we often have a stack trace on each rank/process. It might be helpful to include rank in the event somewhere to help deduplicate this.

src/axolotl/telemetry/manager.py
```diff
@@ -0,0 +1,269 @@
...
+            "true",
+        ) and do_not_track.lower() not in ("1", "true")
+
+        # Show warning (and sleep on all ranks) unless explicitly enabled
+        if enabled and not explicit_enabled:
+            if is_main_process():
+                LOG.warning(ENABLED_WARNING)
+            time.sleep(ENABLED_WARNING_SLEEP_SECONDS)
+
+        # Only rank 0 will send telemetry
```
winglian, line 203: Aren't there cases where we can get an OOM on another rank that fails the parent? would that not get logged?

src/axolotl/telemetry/errors.py
djsaunde: Current implementation only sends telemetry on the main process.

src/axolotl/telemetry/manager.py
djsaunde: The parent exception would get logged in this case. Will that be enough to debug?

src/axolotl/telemetry/manager.py
djsaunde: > Is there a world where the env var for do not track gets dynamically set sometime after and we need to account for that then?

Hm, not that I'm aware of.

djsaunde: What's the motivation for a property?

src/axolotl/train.py
```diff
@@ -39,13 +41,16 @@
...
 
+
+@send_errors
 def train(
     *, cfg: DictDefault, dataset_meta: TrainDatasetMeta
 ) -> Tuple[Union[PeftModel, PreTrainedModel], PreTrainedTokenizer]:
     # Load tokenizer
     LOG.debug(
-        f"loading tokenizer... {cfg.tokenizer_config or cfg.base_model_config}",
+        f"loading tokenizer... {cfg.tokenizer_cocnfig or cfg.base_model_config}",
```
salmanmohammadi: cocnfig

djsaunde: Not sure I'm following?

src/axolotl/telemetry/manager.py
salmanmohammadi: > What's the motivation for a property?

My guess here is to ensure immutability on these fields

djsaunde: Hm, good point, although if a user wanted to modify these for any reason, why wouldn't they just set the relevant env variable? Can't think of another reason why we'd want immutability

src/axolotl/telemetry/whitelist.yaml
djsaunde: Thanks! Was meaning to pad out this list more...

src/axolotl/telemetry/manager.py
djsaunde: Good point!

I wonder if we should setup testing for different accelerators in our CI/CD. Otherwise, I will have to test all the possible hardware manually, which I'm not looking forward to 😅 @winglian any thoughts here?

src/axolotl/telemetry/callbacks.py
```diff
@@ -0,0 +1,161 @@
...
+                self.tracker.update_memory_metrics()
+
+                # Prepare metrics to report
+                metrics = {
+                    "step": step,
+                    "epoch": self.current_epoch,
+                    "progress": state.epoch,  # Fractional epoch progress
+                    "loss": state.log_history[-1].get("loss", 0)
+                    if state.log_history
+                    else 0,
```
salmanmohammadi: nit: move this out of the dict

src/axolotl/telemetry/manager.py
winglian: This is likely fine and I'm overthinking it.

src/axolotl/train.py
winglian: typo `cocnfig` -> `config`

src/axolotl/telemetry/manager.py
djsaunde: I suppose we could send all telemetry _but stack trace telemetry_ in the main process.

src/axolotl/telemetry/manager.py
```diff
@@ -0,0 +1,252 @@
...
+            "os": platform.system(),
+            "python_version": platform.python_version(),
+            "pytorch_version": torch.__version__,
+            "transformers_version": transformers.__version__,
+            "axolotl_version": axolotl.__version__,
+            "cpu_count": psutil.cpu_count(),
+            "memory_total": psutil.virtual_memory().total,
+            "gpu_count": len(gpu_info),
+            "gpu_info": gpu_info,
+        }
```
salmanmohammadi, line 362: Versioning for our other dependencies? TRL would be good - the others may be platform-dependent (e.g. liger)

src/axolotl/telemetry/manager.py
salmanmohammadi: Yeah I'd +1 excluding stack trace telemetry

djsaunde: @SalmanMohammadi What do you mean? I don't think Wing is advocating for excluding stack trace telemetry here. In fact, this was part of the RFC

src/axolotl/train.py
djsaunde: Oh weird, read it wrong and thought you were asking me to add in this typo

src/axolotl/telemetry/manager.py
salmanmohammadi: Ah totally misread - sorry!

djsaunde:
Example events in PostHog:

image

E.g., expanding the properties under `system-info`:

image

winglian:
It might be possible that some of the environment manipulation is breaking deepspeed as hf needs information set into the environment about some of the deepspeed options.

djsaunde:
Closing for now in favor of #2366.

PR #2361, opened 2025-02-25
Top-k logprobs using triton kernel

winglian:


winglian:
I still need to test correctness of the gradients and add some tests.

PR #2362, opened 2025-02-25
Fix(doc): address missing doc changes

NanoCode012:
# Description

- Fix some minor heading mistakes
- Mentions multiple times on `chat_template` eos_token setting

### DIFF
--- a/docs/dataset-formats/conversation.qmd
+++ b/docs/dataset-formats/conversation.qmd
@@ -156,5 +156,5 @@ datasets:
 ```
 
 ::: {.callout-tip}
-It is not necessary to use both `message_field_training` and `message_field_training_detail` at a time.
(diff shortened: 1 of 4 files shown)


docs/dataset-formats/conversation.qmd
```diff
@@ -151,4 +155,6 @@ datasets:
     message_field_training_detail: train_detail
 ```
 
-Tip: It is not necessary to use both `message_field_training` and `message_field_training_detail` at a time.
+::: {.callout-tip}
+It is not necessary to use both `message_field_training` and `message_field_training_detail` at a time.
```
salmanmohammadi:
```suggestion
It is not necessary to set both `message_field_training` and `message_field_training_detail` at once.
```

salmanmohammadi (approved):
bueno

PR #2364, opened 2025-02-26
feat: add eos_tokens and train_on_eot for chat_template EOT parsing

### LINKED ISSUE #2308: Mistral-Small-3 support

### 🔖 Feature description

https://youtu.be/Ikon2p8Strs

### ✔️ Solution

https://youtu.be/Ikon2p8Strs

### ❓ Alternatives

https://youtu.be/Ikon2p8Strs

### 📝 Additional Context

https://youtu.be/Ikon2p8Strs

### Acknowledgements

NanoCode012:
# Description

Introduce new `eot_tokens` and `train_on_eot` fields to extend the capabilities of `chat_template` prompting to support arbitrary number of delimiters.

Adds token check within jinja template to ensure that the `eos_token` and `eot_token` exists.

## Motivation and Context

Newer templates use separate tokens for EOT and EOS which causes confusion with the naming and usage of prior fields.

For example, the `MistralV7-Tekken` formatted prompt below uses `[/SYSTEM_PROMPT]` , `[/INST]`, `</s>` to delimit different sections of the prompt.

```
<s>[SYSTEM_PROMPT]<system prompt>[/SYSTEM_PROMPT][INST]<user message>[/INST]<assistant response></s>[INST]<user message>[/INST]
```

The current chat_template parser expects only 1 delimiter.

Furthermore, this will allow users to keep the original EOS token when finetuning instead of overwriting it with the EOT tokens

## Implementation

The new code would follow the following order:

1. Check `eot_tokens` exist in template. If not provided, checks if EOS in template either hardcoded or as variable `eos_token`.

2. `train_on_eot` -> checks for all `eot_tokens` if `train_on_eos` is provided, else it'll check for EOS token.
3. `train_on_eos` -> only checks for EOS token to mask or unmask

It is assumed that there is only 1 `eot_token` per turn.

## Backwards Compatibility

If `train_on_eot` and `eot_tokens` aren't provided, the values will be loaded from `train_on_eos` and `tokenizer.eos_token` respectively.

This ensures the same legacy behavior.

If either `train_on_eot` or `eot_tokens` are provided, it will not load from the respective variables.

## New Errors Thrown

`eot_tokens` contain `eos_token` and has conflict between `train_on_eot` and `train_on_eos` <- Value Error

EOS tokens not found in chat_template <- Log Warning.
EOT tokens not found in chat_template <- Log Warning
EOT tokens not added to tokenizer <- ValueError

## Alternatives

There was an initial discussion to use `tokenizer.eos_token: List[str]` which applies for generation, however, it's only a `str` when in other modes.

## Discussion / ToDos

Questions:
- [x] Should we support multiple eot_tokens per turn? A: Single as that's the common case.
- [x] Do we need to force single-token eot_tokens only? A: Yes and keep as-is.
- [x] Do we warn if EOT token not appear in chat_template or just debug log? A: Keep as warn for now, no need escalate to error.
- [x] How to handle chat_templates (like mistral) with default system prompt + train on system role? We won't be able to detect that role if when turn building, `system` turn doesn't exist. A: Since this is uncommon case, we can ignore.
- [x] Automatic EOT token detection? or pre-populate with common delimiters? A: No, let users specify.
- [x] The current implementation feels too heavy / too many extra fields to handle this masking. Could we make it simpler? A: The defaults does not require specifying `train_on___` (at most `eot_tokens: `), so normal users won't need to mess with this.

Docs:
- [x] Update to clarify new fields with examples

Tests to add:
- [x] Check conflict eos_token in eot_token and mismatch train_on_eos / train_on_eot
- [x] Test backward compatibility that eot_token inherits eos_token
- [x] Check token not exist in template
- [x] ~Test trainable with system role ~ Unneeded as is edge case we don't need to worry about.

### DIFF
--- a/docs/config.qmd
+++ b/docs/config.qmd
@@ -183,7 +183,7 @@ datasets:
     # IMPORTANT: The following fields determine which parts of the conversation to train on.
     # Priority order: message_field_training > message_field_training_detail > train_on_inputs or role in roles_to_train
     # See examples at `docs/dataset-formats/conversation.qmd`
-    # Note: If the below 4 fields are set to empty, defaults to training only on the last message.
+    # Note: If the below 5 fields are empty, defaults to training only on the last message.
 
     # Optional[List[str]]. Roles to train on. The tokens from these roles will be considered for the loss.
     roles_to_train: ["assistant"]  # default
@@ -192,7 +192,13 @@ datasets:
     # - turn (default): train on the EOS token at the end of each trainable turn
     # - last: train on the last EOS token in the conversation
     # TIP: Please make sure that your `tokenizer.eos_token` is same as EOS/EOT token in template. Otherwise, set `eos_token` under `special_tokens`.
-    train_on_eos: last
+    train_on_eos: turn
+    # Optional[str]. Which EOT (End-of-Turn) tokens to train on in the conversation. Possible values are:
+    # - all: train on all EOT tokens
+    # - turn: train on the EOT token at the end of each trainable turn
+    # - last: train on the last EOT token in the conversation
+    # If not specified, defaults to the value of train_on_eos for backward compatibility.
+    train_on_eot:
     # The key in the message turn that indicates via boolean whether tokens of a turn should be considered for training. Useful to selectively train on certain turns besides the `roles_to_train`.
     message_field_training: training
     # The key in the message turn that contains the training details. Useful to selectively train on certain tokens in a turn.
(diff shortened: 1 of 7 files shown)


kubernetes-bad:
Was there any recent progress on this? Seems like it's pretty close to the finish line.

NanoCode012:
@kubernetes-bad , thanks for ping on this. I didn't get to finish the test suite on it, but the PR should just work.

I'll look into this tmr/day after.

winglian (approved):
looks good from a code perspective. I want to run a few tests manually if you haven't already

NanoCode012:
> looks good from a code perspective. I want to run a few tests manually if you haven't already

Thanks! I think I did manual tests from way back but after the recent rebases. I'll do some today

NanoCode012:
Manual test seem to be working fine

PR #2012, opened 2024-11-04
refactor(optimizer): use `optimizer_cls_and_kwargs` for custom optim

NanoCode012:
# Breaking change

- Deprecates `alternate_optimizer`. Move to use `optimizer` instead to consolidate.

# Description

Refactors the optimizer handling to offload from `create_optimizer` into the Trainer instead. Only `lora_plus` requires overriding that function now.

- `ao_adamw_4bit` is available on `transformers` as `adamw_torch_4bit`.
- `lion_pytorch` is available on `transformers` as `lion_32bit`.
- Removed default handling of `loraplus_lr_embedding` to let it inherit from Model Input class instead
- Updated docs to include new upstream optimizers

TODO:
- [x] Verify `adamw_anyprecision` is not affected
- [x] Verify Galore is not affected
- [ ] Discuss whether to support other `optimi` optimizers https://optimi.benjaminwarner.dev/which_optimizer/
- [x] Update transformers to commit that includes the linked PR
- [ ] Test custom optimizers
- [ ] Test upstream optimizers

## Motivation and Context

This is due to upstream PR opening a new config https://github.com/huggingface/transformers/pull/34358/

## Next steps

Once some time passes:
- Deprecate `lion_pytorch` and remove the dependency

PR #2367, opened 2025-02-28
Optimizer refactor and add Muon support

winglian:
supersedes #2012 

see: https://arxiv.org/pdf/2502.16982v1

### DIFF
--- a/src/axolotl/core/trainers/base.py
+++ b/src/axolotl/core/trainers/base.py
@@ -14,6 +14,7 @@ from typing import Dict, Literal, Optional
 import torch
 from datasets import Dataset
 from peft.optimizers import create_loraplus_optimizer
+from torch import nn
 from torch.optim.lr_scheduler import OneCycleLR
 from torch.utils.data import BatchSampler, DataLoader, RandomSampler, SequentialSampler
 from transformers import Trainer
@@ -22,6 +23,7 @@ from transformers.utils import is_sagemaker_mp_enabled
 from trl import CPOTrainer, KTOTrainer, ORPOTrainer, PRMTrainer, RewardTrainer
 from trl.trainer.utils import pad_to_length
 
+from axolotl.integrations.base import BaseOptimizerFactory
 from axolotl.monkeypatch.relora import ReLoRAScheduler
 from axolotl.utils.samplers import MultipackBatchSampler, get_dataset_lengths
 from axolotl.utils.schedulers import (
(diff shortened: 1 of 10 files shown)


src/axolotl/train.py
```diff
@@ -208,7 +208,7 @@ def terminate_handler(_, __, model_weakref):
             model = model.merge_and_unload()
         else:
             # final model weights have already been saved by `ReLoRACallback.on_train_end`
-            return model, tokenizer
+            return model, tokenizer, trainer
```
NanoCode012: I noticed that `trainer` is now returned (which seems to be a core api change), but I'm not seeing where it's actually used except to be `del trainer` in cli call.

Edit: I saw it being used in tests. Would it be better to have a separate test that calls `setup_trainer` to check the optimizer and keep current training test as-is?

src/axolotl/core/trainers/base.py
```diff
@@ -314,50 +294,47 @@ def create_optimizer(self):
...
+                            {
+                                p.data_ptr(): p.numel() for p in module.parameters()
+                            }.values()
+                        )
+                        LOG.info(f"skipped {module}: {skipped/2**20}M params")
+                        manager.register_module_override(
+                            module, "weight", {"optim_bits": 32}
+                        )
+                        LOG.debug(f"bitsandbytes: will optimize {module} in fp32")
+                LOG.info(f"skipped: {skipped/2**20}M params")
```
NanoCode012, line 349: Do you mean `AdamW8bit`? Is there a new change upstream to require this?

winglian: This is copy paste from upstream

src/axolotl/train.py
winglian: So the trainer doesn't actually setup the optimizer until training is kicked off. I just tried it with the trainer refactor and grabbing the trainer after the setup step, and the optimizer attribute isn't set yet.

NanoCode012: Oh I see!

src/axolotl/core/trainers/base.py
```diff
@@ -326,50 +306,47 @@ def create_optimizer(self):
...
+                        "optimizer_dict"
                     )
-                )
-            elif self.args.alternate_optimizer == "ao_adamw_4bit":
-                from torchao.prototype.low_bit_optim import AdamW4bit
 
-                self.optimizer = (  # pylint: disable=attribute-defined-outside-init
-                    AdamW4bit(optimizer_grouped_parameters, **optimizer_kwargs)
+                self.optimizer = optimizer_cls(
+                    optimizer_grouped_parameters, **optimizer_kwargs
```
NanoCode012, line 328: Here `optimizer_grouped_parameters` is a `list[dict]`. Is it correct to overwrite it via what is being popped? Does it only need to override a specific key/element?

NanoCode012: Nevermind, it seems that it's how upstream handles it too.

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
```diff
@@ -64,6 +64,18 @@ class ChatTemplate(str, Enum):
...
 
+class CustomSupportedOptimizers(str, Enum):
+    """Custom supported optimizers"""
+
+    optimi_adamw = "optimi_adamw"  # pylint: disable=invalid-name
+    ao_adamw_4bit = "ao_adamw_4bit"  # pylint: disable=invalid-name
+    ao_adamw_8bit = "ao_adamw_8bit"  # pylint: disable=invalid-name
+    ao_adamw_fp8 = "ao_adamw_fp8"  # pylint: disable=invalid-name
+    adopt_adamw = "adopt_adamw"  # pylint: disable=invalid-name
+    lion_pytorch = "lion_pytorch"  # pylint: disable=invalid-name
```
NanoCode012, line 75:
```suggestion
```

I think we can remove lion_pytorch here as Lion is supported in PyTorch and we don't even include the dependency for this anymore.

NanoCode012 (approved):
I checked that the changes from prior PR transferred and were updated following latest transformers `create_optimizer`.

PR #2370, opened 2025-02-28
fix(test): replace jackfram llama with smollm

NanoCode012:
# Description

Replace the old jackfram llama 68m with the smollm 135m due to breaking tests.

CI broken: https://github.com/axolotl-ai-cloud/axolotl/actions/runs/13533778077

### DIFF
--- a/tests/e2e/test_lora_llama.py
+++ b/tests/e2e/test_lora_llama.py
@@ -28,8 +28,7 @@ class TestLoraLlama(unittest.TestCase):
         # pylint: disable=duplicate-code
         cfg = DictDefault(
             {
-                "base_model": "JackFram/llama-68m",
-                "tokenizer_type": "LlamaTokenizer",
+                "base_model": "HuggingFaceTB/SmolLM2-135M-Instruct",
                 "sequence_len": 1024,
                 "load_in_8bit": True,
                 "adapter": "lora",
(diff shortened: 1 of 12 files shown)


winglian:
need to add
```
                "special_tokens": {
                    "pad_token": "<|endoftext|>",
                },
```
and remove `"model_type": "AutoModelForCausalLM",` and `"tokenizer_type": "LlamaTokenizer",` from the configs

tests/e2e/test_lora_llama.py
```diff
@@ -28,15 +28,15 @@ def test_lora(self, temp_dir):
...
-                "tokenizer_type": "LlamaTokenizer",
+                "base_model": "HuggingFaceTB/SmolLM2-135M-Instruct",
                 "sequence_len": 1024,
                 "load_in_8bit": True,
                 "adapter": "lora",
                 "lora_r": 8,
                 "lora_alpha": 16,
                 "lora_dropout": 0.05,
                 "lora_target_linear": True,
+                "lora_modules_to_save": ["lm_head", "embed_tokens"],
```
winglian, line 39: this isn't needed if you correct the pad_token. I'm not a fan of changing the test case for a model swap, but rather we should add a new test case to include lora_modules_to_save. otherwise we risk having regressions down the road

winglian: should also remove the unk_token, bos_tokenm, and eos_token from the configs where you swap this. Would also not swap in an instruct model since the original model you're replacing isn't an instruct model

PR #2371, opened 2025-02-28
`train.py` refactor

djsaunde:
# Description

Quick refactor for `axolotl.train` to make it more modular / readable / documented.

## Motivation and Context

Part of a larger codebase refactor effort over the next week or so.

## How has this been tested?

Should be able to merge if existing tests pass. I'll also run a few test runs as a sanity check.

## Types of changes

Moving code out of `train` into functions, adding docstrings, typing, etc.


src/axolotl/train.py
```diff
@@ -291,22 +314,158 @@ def terminate_handler(_, __, model_weakref):
...
+def save_initial_configs(
+    cfg: DictDefault,
+    tokenizer: PreTrainedTokenizer,
+    model: PreTrainedModel,
+    peft_config: PeftConfig | None,
+):
+    """
+    Save initial configurations before training.
 
-def pretrain_hooks(_cfg, _trainer):
```
NanoCode012, line 297: Do we need to keep this? I saw this hook for a very long time, but I don't recall ever us using it @winglian .

src/axolotl/train.py
```diff
@@ -255,34 +272,40 @@ def terminate_handler(_, __, model_weakref):
...
+                    param_name = (
+                        "dataset_name"
+                        if (
+                            cfg.rl is not None
+                            or cfg.reward_model
+                            or cfg.process_reward_model
+                        )
+                        else "dataset_tags"
+                    )
+                    model_card_kwarg[param_name] = dataset_tags
```
NanoCode012: I think we should leave a comment to remind us why we added this, I think it was due to trl using `dataset_tags` and HF using `dataset_name`? https://github.com/axolotl-ai-cloud/axolotl/pull/2101/files#r1856723756

src/axolotl/train.py
```diff
@@ -187,32 +188,49 @@ def terminate_handler(_, __, model_weakref):
...
             # final model weights have already been saved by `ReLoRACallback.on_train_end`
-            return model, tokenizer
+            return
 
-    # TODO do we need this fix? https://huggingface.co/docs/accelerate/usage_guides/fsdp#saving-and-loading
-    # only save on rank 0, otherwise it corrupts output on multi-GPU when multiple processes attempt to write the same file
     if cfg.fsdp:
+        # TODO: do we need this fix? https://huggingface.co/docs/accelerate/usage_guides/fsdp#saving-and-loading
+        # only save on rank 0, otherwise it corrupts output on multi-GPU when multiple
+        # processes attempt to write the same file
```
NanoCode012, line 236: Re: this TODO, do we need to call their `merge_fsdp_weights` instead?

NanoCode012 (commented):
The refactor looks great!

src/axolotl/train.py
djsaunde: Oh yeah, I actually removed this in my telemetry PR. Unless @winglian has any reason to keep

salmanmohammadi:
beautiful

src/axolotl/train.py
```diff
@@ -291,22 +314,158 @@ def terminate_handler(_, __, model_weakref):
...
+    trainer = setup_trainer(
+        cfg,
+        train_dataset,
+        eval_dataset,
+        (model, model_ref, peft_config),
+        tokenizer,
+        processor,
+        total_num_steps,
+    )
+
```
winglian, line 460: Is there a good way to split off the function here up to setting up the trainer? Could be useful in the optimizer refactor where it's helpful to have just the trainer to assert the optimizer that was constructed. There's a good bit of setup up to this point that happens that this function handles

djsaunde: Yeah, I can do this.

src/axolotl/train.py
djsaunde: It's not used anywhere AFAICT so I think safe to remove.

djsaunde: https://github.com/search?q=org%3Aaxolotl-ai-cloud%20pretrain_hooks&type=code

src/axolotl/train.py
djsaunde: Thanks for the pointer!

djsaunde: Looks like there's actually no equivalent arg for dataset tags in the TRL trainer (and subclasses).

See: https://github.com/huggingface/trl/blob/98de0e7c62a7e7f683fd55aa8b4957245ff90fea/trl/trainer/orpo_trainer.py#L1045.

Compare with: https://github.com/huggingface/transformers/blob/3e83ee75ec7a5291c95a7172d90665eed450bfba/src/transformers/trainer.py#L4592

I think we should just remove for RL trainers?

src/axolotl/train.py
djsaunde: No idea, just copying this TODO from the existing code. We could make a ticket to look into this.

djsaunde:
Can anyone rubber stamp this?

NanoCode012 (approved):

PR #2384, opened 2025-03-05
various fixes 20250305

winglian:
validation fix for spectrum and fsdp
use offload instead of unsloth value for activation checkpointing

### DIFF
--- a/src/axolotl/integrations/spectrum/args.py
+++ b/src/axolotl/integrations/spectrum/args.py
@@ -17,7 +17,7 @@ Module for handling Spectrum input arguments.
 """
 from typing import Optional
 
-from pydantic import BaseModel
+from pydantic import BaseModel, model_validator
 
 
 class SpectrumArgs(BaseModel):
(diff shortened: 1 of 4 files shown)


src/axolotl/integrations/spectrum/args.py
```diff
@@ -27,3 +27,20 @@ class SpectrumArgs(BaseModel):
...
     spectrum_top_fraction: Optional[float] = 0.5
     spectrum_model_name: Optional[str] = None
+
+    @model_validator(mode="before")
+    @classmethod
+    def check_fsdp_use_orig_params(cls, data):
+        if (
+            data.get("fsdp")
+            and data.get("fsdp_config")
+            and data["fsdp_config"].get("use_orig_params") is False
```
NanoCode012: Do we need to check for None value as well as None may default to False? https://pytorch.org/docs/stable/fsdp.html#torch.distributed.fsdp.FullyShardedDataParallel

NanoCode012 (approved):

PR #2390, opened 2025-03-07
fix: create mount folder on modal if not exist

NanoCode012:
# Description

As a side effect of https://github.com/axolotl-ai-cloud/axolotl/pull/2347 , if a user has the local dev env and runs on modal, the `/workspace/mounts` dir won't be available as it won't be locally mounted to prevent axolotl codebase conflict.

`_train` and `_lm_eval` missed creating the mount folder. `_preprocess` is ok.

### DIFF
--- a/src/axolotl/cli/cloud/modal_.py
+++ b/src/axolotl/cli/cloud/modal_.py
@@ -270,6 +270,7 @@ def _preprocess(config_yaml: str, volumes=None):
 
 
 def _train(config_yaml: str, accelerate: bool = True, volumes=None, **kwargs):
+    Path("/workspace/mounts").mkdir(parents=True, exist_ok=True)
     with open("/workspace/mounts/config.yaml", "w", encoding="utf-8") as f_out:
         f_out.write(config_yaml)
     run_folder = "/workspace/mounts"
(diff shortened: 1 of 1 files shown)


salmanmohammadi (approved):
Did you have a chance to test this out?

NanoCode012:
Yep, it errored for me and this fixed it using the example `cloud_config.yml`

For more detail:

This line would mount `cwd` to `/workspace/mounts` if it doesn't detect `src/axolotl` (not in dev env).

https://github.com/axolotl-ai-cloud/axolotl/blob/ae663741566ab27bb25af25ef3d2f16b6334e492/src/axolotl/cli/cloud/__init__.py#L46-L48

Afterward, this line tries to create a config at this location. (Only `/workspace` is available at this time)
https://github.com/axolotl-ai-cloud/axolotl/blob/ae663741566ab27bb25af25ef3d2f16b6334e492/src/axolotl/cli/cloud/modal_.py#L272-L274

---

A bit unrelated but I found us not passing `cwd=` when not running accelerate on L163. https://github.com/axolotl-ai-cloud/axolotl/blob/8dfadc2b3c7f85e4208b74eb93c32f440c7e25b4/src/axolotl/cli/main.py#L126-L173

salmanmohammadi:
Nice sleuthing : ) 
> A bit unrelated but I found us not passing cwd= when not running accelerate on L163. 

Is this expected behaviour?

NanoCode012:
> > A bit unrelated but I found us not passing cwd= when not running accelerate on L163.
> 
> Is this expected behaviour?

Honestly, I'm not sure why we even need to pass `cwd` as a param. Could we not just call `os.getcwd()` when we need it?

NanoCode012:
Merging this in first. Regarding the other issue mentioned, it can be addressed in separate PR.

PR #2394, opened 2025-03-07
bump axolotl-contribs-lgpl

winglian:
just a test PR to see if the updated dependency is causing test failures


winglian:
here's the changeset from when it was previously working ...
https://github.com/axolotl-ai-cloud/axolotl-contribs-lgpl/compare/cedbf68a357a693366b275723fc8fd6ab5232714...main

winglian:
fixed, and integrated into other PR

PR #2397, opened 2025-03-08
grab sys prompt too from dataset

winglian:
add sys prompt if separate column

### DIFF
--- a/docs/config.qmd
+++ b/docs/config.qmd
@@ -154,6 +154,10 @@ datasets:
     # Key containing the messages (default: "messages")
     field_messages: messages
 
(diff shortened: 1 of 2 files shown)


NanoCode012:
Which datasets use a separate `system` field?

winglian:
https://huggingface.co/datasets/bespokelabs/Bespoke-Stratos-17k?row=0

NanoCode012 (approved):
Tested to work after a rebase locally.

~We should also add `field_system` to the chat_template section within `config.qmd`~

Edit: added the commit.

PR #2398, opened 2025-03-09
chore(doc): add faq when having no default chat_template

NanoCode012:
# Description

A user asked about this on discord, so adding it to FAQ when using `type: chat_template` without template in tokenizer.

### DIFF
--- a/docs/dataset-formats/conversation.qmd
+++ b/docs/dataset-formats/conversation.qmd
@@ -75,7 +75,7 @@ datasets:
 ```
 
 ::: {.callout-tip}
-If you receive an error like `chat_template choice is tokenizer_default but tokenizer's chat_template is null.`, it means the tokenizer does not have a default chat_template. Follow the examples below instead to set a custom chat_template.
+If you receive an error like "`chat_template` choice is `tokenizer_default` but tokenizer's `chat_template` is null.", it means the tokenizer does not have a default `chat_template`. Follow the examples below instead to set a custom `chat_template`.
 :::
 
 2. Using the `gemma` chat template to override the tokenizer_config.json's chat template on OpenAI messages format, training on all assistant messages.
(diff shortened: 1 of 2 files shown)


docs/dataset-formats/conversation.qmd
```diff
@@ -74,6 +74,10 @@ datasets:
     train_on_eos:
 ```
 
+::: {.callout-tip}
+If you receive an error like `chat_template choice is tokenizer_default but tokenizer's chat_template is null.`, it means the tokenizer does not have a default chat_template. Follow the examples below instead to set a custom chat_template.
+:::
+
 2. Using the `gemma` chat template to override the tokenizer_config.json's chat template on OpenAI messages format, training on all assistant messages.
```
salmanmohammadi: nit : )
```suggestion
::: {.callout-tip}
If you receive an error like "`chat_template` choice is `tokenizer_default` but tokenizer's `chat_template` is null.", it means the tokenizer does not have a default `chat_template`. Follow the examples below instead to set a custom `chat_template`.
:::

2. Using the `gemma` chat template to override the tokenizer_config.json's chat template on OpenAI messages format, training on all assistant messages.
```

docs/faq.qmd
```diff
@@ -52,3 +52,7 @@ description: Frequently asked questions
 **Q: The EOS/EOT token is incorrectly being masked or not being masked.**
 
 > A: This is because of the mismatch between `tokenizer.eos_token` and EOS/EOT token in template. Please make sure to set `eos_token` under `special_tokens` to the same EOS/EOT token as in template.
+
+**Q: `chat_template choice is tokenizer_default but tokenizer's chat_template is null. Please add a chat_template in tokenizer config`**
```
salmanmohammadi:
```suggestion

**Q: "`chat_template` choice is `tokenizer_default` but tokenizer's `chat_template` is null. Please add a `chat_template` in tokenizer config"**
```

salmanmohammadi (approved):

PR #2401, opened 2025-03-11
Feat: minor docs improvements for RLHF and faq on embeddings

NanoCode012:
# Description

Added following new questions on discord.

### DIFF
--- a/docs/config.qmd
+++ b/docs/config.qmd
@@ -30,6 +30,8 @@ tokenizer_legacy:
 # Resize the model embeddings when new tokens are added to multiples of 32
 # This is reported to improve training speed on some models
 resize_token_embeddings_to_32x:
+# Optional[bool] Whether to shrink the embeddings to len(tokenizer). By default, we won't shrink.
+shrink_embeddings:
 
 # (Internal use only)
 # Used to identify which the model is based on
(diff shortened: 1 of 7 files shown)


docs/config.qmd
```diff
@@ -205,10 +207,46 @@ test_datasets:
     data_files:
       - /workspace/data/eval.jsonl
 
-# use RL training: 'dpo', 'ipo', 'kto'
+# use RL training: 'dpo', 'ipo', 'kto', 'simpo', 'orpo', 'grpo'
 rl:
-# whether to perform weighting if doing DPO training. Boolean.
-dpo_use_weighting:
+rl_beta:  # Optional[float]. The beta parameter for the RL training.
```
winglian, line 212: I think the recommended way is to use `trl.beta` now?

NanoCode012: Is that the recommended approach now? I was under the impression we're leaving the `trl: ` namespace for GRPO specifically.

NanoCode012: From internal discussion, we will be keeping both for now and ensuring they are the same.

docs/getting-started.qmd
```diff
@@ -36,27 +36,33 @@ The YAML configuration file controls everything about your training. Here's what
...
+
 See our [Config options](config.qmd) for more details.
 
 ### Training {#sec-training}
 
 When you run `axolotl train`, Axolotl:
 
 1. Downloads the base model
-2. (If specified) applies LoRA adapter layers
+2. (If specified) applies q/LoRA adapter layers
```
salmanmohammadi: nit:
```suggestion
2. (If specified) applies QLoRA/LoRA adapter layers
```

docs/rlhf.qmd
```diff
@@ -298,7 +298,7 @@ The input format is a simple JSON input with customizable fields based on the ab
 
 ### IPO
 
-As IPO is just DPO with a different loss function, all supported options for DPO works here.
+As IPO is just DPO with a different loss function, all supported dataset formats for [DPO](#dpo) works.
```
salmanmohammadi: optional nit
```suggestion
As IPO is just DPO with a different loss function, all supported dataset formats for [DPO](#dpo) are also supported for IPO.
```

salmanmohammadi (approved):
lgtm!

Issue #2395, opened 2025-03-07
LoRA example from quickstart guide not working with Docker container


daniel-dona:
### Please check that this issue hasn't been reported before.

- [x] I searched previous [Bug Reports](https://github.com/axolotl-ai-cloud/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

Train the example LoRA

### Current behaviour

Runtime exception

```
# export CUDA_VISIBLE_DEVICES=1
# axolotl train examples/llama-3/lora-1b.yml
The following values were not passed to `accelerate launch` and had defaults used instead:
        `--num_processes` was set to a value of `1`
        `--num_machines` was set to a value of `1`
        `--mixed_precision` was set to a value of `'no'`
        `--dynamo_backend` was set to a value of `'no'`
To avoid this warning pass in values for each of the problematic parameters or run `accelerate config`.
[2025-03-07 18:49:23,348] [INFO] [datasets.<module>:54] [PID:1523] PyTorch version 2.5.1+cu124 available.
[2025-03-07 18:49:24,651] [INFO] [real_accelerator.py:219:get_accelerator] Setting ds_accelerator to cuda (auto detect)
[2025-03-07 18:49:24,741] [INFO] [root.spawn:60] [PID:1523] gcc -pthread -B /root/miniconda3/envs/py3.11/compiler_compat -DNDEBUG -fwrapv -O2 -Wall -fPIC -O2 -isystem /root/miniconda3/envs/py3.11/include -fPIC -O2 -isystem /root/miniconda3/envs/py3.11/include -fPIC -c /tmp/tmp5m6kkyqp/test.c -o /tmp/tmp5m6kkyqp/test.o
[2025-03-07 18:49:24,784] [INFO] [root.spawn:60] [PID:1523] gcc -pthread -B /root/miniconda3/envs/py3.11/compiler_compat /tmp/tmp5m6kkyqp/test.o -laio -o /tmp/tmp5m6kkyqp/a.out
[...]

PR #2403, opened 2025-03-11
use max of 32 dataset processes if not explicit

winglian:
should help with #2395 @NanoCode012


NanoCode012:
The reason I mentioned 32 was because we had (in several places), a max of 64. 

https://github.com/axolotl-ai-cloud/axolotl/blob/4a736986fa6f3387f21391f186f472926754dac1/src/axolotl/datasets.py#L49
https://github.com/axolotl-ai-cloud/axolotl/blob/4a736986fa6f3387f21391f186f472926754dac1/src/axolotl/core/datasets/chat.py#L46

We should probably consolidate this.

PR #2404, opened 2025-03-11
Sequential sample packing

DreamGenX:
# Description

This change adds support for next-fit bin packing in MultipackBatchSampler. What this means is that we can now use sample packing while preserving the order of the examples.

## Motivation and Context

Current sample packing affects order of examples in ways that can affect training results. This PR adds a new flag:
`--sample_pack_sequentially` which uses a simple greedy / sequential next-fit bin packing.

If you use `--sample_pack_sequentially` alone, the order of the examples will be determined by the underlying `RandomSampler`. If you use `--sample_pack_sequentially` with `--curriculum_sampling`, the order of the examples will be the same as in the training data. Both options make sense and differ from the current settings.

For example, previously it was not possible to do proper curriculum learning since the bin packing algorithm would reorder the examples anyway.

## How has this been tested?

I tested it by running the new config: `examples/llama-3/lora-1b-sample-packing-sequentially.yml`
On this dataset, packing efficiency is high (>97%).

---

I would also like to bring attention to a potential bug in the multipack code:
- [Here](https://github.com/axolotl-ai-cloud/axolotl/blob/4a736986fa6f3387f21391f186f472926754dac1/src/axolotl/utils/trainer.py#L463) we set `group_size` and `bin_size`
- But [it's never used](https://github.com/axolotl-ai-cloud/axolotl/blob/4a736986fa6f3387f21391f186f472926754dac1/src/axolotl/utils/samplers/multipack.py#L111).

### DIFF
--- a/examples/llama-3/lora-1b-sample-packing-sequentially.yml
+++ b/examples/llama-3/lora-1b-sample-packing-sequentially.yml
@@ -1,0 +1,80 @@
+base_model: meta-llama/Llama-3.2-1B
+# optionally might have model_type or tokenizer_type
+model_type: LlamaForCausalLM
+tokenizer_type: AutoTokenizer
+# Automatically upload checkpoint and final model to HF
+# hub_model_id: username/custom_model_name
+
+load_in_8bit: true
+load_in_4bit: false
+strict: false
+
+datasets:
+  - path: mhenrichsen/alpaca_2k_test
+    type: alpaca
+  - path: mhenrichsen/alpaca_2k_test
+    type: alpaca
+dataset_prepared_path:
+val_set_size: 0.0
+output_dir: ./outputs/lora-out
+
+test_value: true
+
+sequence_len: 4096
+sample_packing: true
+sample_packing_sequentially: true
+curriculum_sampling: true
+eval_sample_packing: false
+pad_to_sequence_len: true
+
+adapter: lora
+lora_model_dir:
+lora_r: 32
+lora_alpha: 16
+lora_dropout: 0.05
+lora_target_linear: true
+lora_fan_in_fan_out:
+lora_modules_to_save:
+  - embed_tokens
+  - lm_head
+
+wandb_project:
+wandb_entity:
+wandb_watch:
+wandb_name:
+wandb_log_model:
+
+gradient_accumulation_steps: 4
+micro_batch_size: 2
+num_epochs: 4
+optimizer: adamw_bnb_8bit
+lr_scheduler: cosine
+learning_rate: 0.0002
+
+train_on_inputs: false
+group_by_length: false
+bf16: auto
+fp16:
+tf32: false
+
+gradient_checkpointing: true
+early_stopping_patience:
+resume_from_checkpoint:
+local_rank:
+logging_steps: 1
+xformers_attention:
+flash_attention: true
+s2_attention:
+
+warmup_steps: 10
+evals_per_epoch: 4
+eval_table_size:
+eval_max_new_tokens: 128
+saves_per_epoch: 1
+debug:
+deepspeed:
+weight_decay: 0.0
+fsdp:
+fsdp_config:
+special_tokens:
+  pad_token: <|end_of_text|>
--- a/src/axolotl/utils/samplers/multipack.py
+++ b/src/axolotl/utils/samplers/multipack.py
@@ -8,7 +8,7 @@ from typing import Any, Iterable, List, Union
 
 import numba
 import numpy as np
-from torch.utils.data import BatchSampler, Sampler
+from torch.utils.data import BatchSampler, Sampler, SequentialSampler
 
 from axolotl.utils.distributed import reduce_and_broadcast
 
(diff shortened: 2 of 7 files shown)


winglian:
@DreamGenX can you make this PR editable my maintainers? I'm happy to help get the linting and tests passing for this

DreamGenX:
@winglian done, thank you

src/axolotl/utils/trainer.py
```diff
@@ -455,13 +455,18 @@ def calculate_total_num_steps(cfg, train_dataset, update=True):
             else:
                 sampler_batch_size = cfg.micro_batch_size
                 batch_max_len = cfg.sequence_len
+            if cfg.curriculum_sampling:
```
winglian, line 459: would it make sense to have pydantic validation to warn if curriculum sampling is not enabled when sample_packing_sequentially is enabled?

DreamGenX: There's a LOG.warn in the dataset class:

```
        if self.sequential and not isinstance(sampler, SequentialSampler):
            LOG.warn(
                "using sequential sample packing with non-sequential sampler, did you want to also enable curriculum_sampling?"
            )
```

Though it aslo makes sense to use the sequential packing without curriculum sampling. That way you get the same order as if you did not do packing. Can be useful way to get the benefits of packing with minimal diff.

winglian:
I rebased again against the latest main since there were still some conflicts and linting issues. Hopefully I didn't screw up the implementation.

examples/llama-3/lora-1b-sample-packing-sequentially.yml
```diff
@@ -0,0 +1,80 @@
...
+# hub_model_id: username/custom_model_name
+
+load_in_8bit: true
+load_in_4bit: false
+strict: false
+
+datasets:
+  - path: mhenrichsen/alpaca_2k_test
+    type: alpaca
+  - path: mhenrichsen/alpaca_2k_test
```
NanoCode012, line 15: Is this double dataset intentional?

examples/llama-3/lora-1b-sample-packing-sequentially.yml
```diff
@@ -0,0 +1,80 @@
...
+    type: alpaca
+dataset_prepared_path:
+val_set_size: 0.0
+output_dir: ./outputs/lora-out
+
+test_value: true
```
NanoCode012, line 21: Maybe forgot to remove this?

examples/llama-3/lora-1b-sample-packing-sequentially.yml
```diff
@@ -0,0 +1,80 @@
...
+
+sequence_len: 4096
+sample_packing: true
+sample_packing_sequentially: true
+curriculum_sampling: true
```
NanoCode012, line 26: Could we add both of these to documentation `docs/config.qmd` to explain them?

src/axolotl/utils/samplers/multipack.py
```diff
@@ -103,6 +103,55 @@ def allocate(
...
+    - result: List of batches for the current rank
+    - total_used: Number of actual example tokens
+    - total_slots: Maximum theoretical number of example tokens (number of bins * bin capacity)
+    """
+    result = []
+    total_used = 0
+
+    # First, do sequential packing into bins
+    all_bins = []
+    current_bin = [0 for i in range(0)]  # numba hint
```
NanoCode012, line 127: What is this for loop for?

src/axolotl/utils/samplers/multipack.py
```diff
@@ -136,6 +187,11 @@ def __init__(
         # the minimum packed dataset length across all ranks determined by a gather/broadcast
         self.len_across_ranks = None
 
+        if self.sequential and not isinstance(sampler, SequentialSampler):
+            LOG.warn(
+                "using sequential sample packing with non-sequential sampler, did you want to also enable curriculum_sampling?"
+            )
+
```
NanoCode012, line 194: Similar to the earlier discussion, could we move this validation to the schema `src/axolotl/utils/schemas/config.py` and not here?

examples/llama-3/lora-1b-sample-packing-sequentially.yml
DreamGenX: that's how it was in the original config, it just doubles the data

src/axolotl/utils/samplers/multipack.py
DreamGenX: it's for numba to understand that it's array of int (hence the comment), note that using `current_bin: list[int] = []` is not enough

examples/llama-3/lora-1b-sample-packing-sequentially.yml
DreamGenX: yep, seems unnecessary -- it's a leftover from the config i forked

DreamGenX:
@winglian @NanoCode012 BTW, when i run `train`, I see both "using non-sequential sample packing" and "using sequential sample packing" in the logs (I think due to eval, which is strange), but during pre-process I see only `using sequential sample packing` as expected. I don't know why, because all occurences of the MultiPackBatchSampler get the `sequential` flag (even the instance used for eval), except for the one used for pre-train... but maybe I missed something.

examples/llama-3/lora-1b-sample-packing-sequentially.yml
NanoCode012: Oh, I think you were basing off a config meant for deduplication `examples/llama-3/lora-1b-deduplicate-sft.yml`. We won't need this for your case.

src/axolotl/utils/samplers/multipack.py
NanoCode012: You can change the check to be: if sample_pack_sequential and not curriculum_sampling instead

winglian (approved):

PR #2170, opened 2024-12-10
Multimodal integration - pixtral/llava/qwen2-vl

bursteratom:
# Description

rudimentary support for pixtral/llava/qwen2-vl

PR #2406, opened 2025-03-12
Feat: Add support for gemma3_text and add e2e for gemma2

NanoCode012:
# Description

Requires transformers tag https://github.com/huggingface/transformers/tree/v4.49.0-Gemma-3

```
pip install git+https://github.com/huggingface/transformers@v4.49.0-Gemma-3
```

Upstream PR has just been merged: https://github.com/huggingface/transformers/pull/36658

## How has this been tested?

No, waiting on E2E to verify


requirements.txt
```diff
@@ -13,7 +13,7 @@ liger-kernel==0.5.3
 packaging==23.2
 
 peft==0.14.0
-transformers==4.49.0
+transformers @ git+https://github.com/huggingface/transformers@v4.49.0-Gemma-3
```
salmanmohammadi: should we point to main if support for Gemma3 has landed?

NanoCode012: They recommended this branch. We can also point to main `50d3530aa04e7a7d003e6b255a98f79fd0447357`. I just set this initially to see how E2E fairs first.

winglian:
We may want to add validation that this doesn't work with liger atm

rob-x-ai:
Successfully fine-tuned the 1B model with this PR—thanks!

Looking to fine-tune the other Gemma-3 models (4B, 12B, and 27B). Since they are omni-modal, I'm trying to remove the vision layers to get a text-only version to work with.

Any updates or guidance on handling this?

winglian:
I've rebased this against the latest main and updated it to use `4.45.0` release of `transformers`

NanoCode012:
Thanks for the rebase. This PR is only the 1B text only. I'm working on a fork of #2170 that would allow multi-modal.

> Since they are omni-modal, I'm trying to remove the vision layers to get a text-only version to work with.

@kroonen  You could load as-if they're text-only, but it's possible that when save, you would lose the vision tower.

winglian (approved):

src/axolotl/integrations/liger/__init__.py
```diff
@@ -114,3 +114,5 @@ def pre_model_load(self, cfg):
                 modeling_mod.CrossEntropyLoss = LigerCrossEntropyLoss
             if cfg.liger_fused_linear_cross_entropy:
                 modeling_mod.DeepseekV2ForCausalLM.forward = deepseekv2_lce_forward
+        # elif cfg.model_config_type in ["gemma3_text", "deepseek_v3"]:
+        #     raise ValueError(f"Unsupported model config type: {cfg.model_config_type}")
```
NanoCode012: I think this PR only needs this fixed, it should error if trying to use liger with theses two model configs types

winglian: Wdym?

NanoCode012: It should error if liger is set and model type is one of these two.

NanoCode012: I feel silly, the code was right but I thought something was wrong.

PR #2412, opened 2025-03-13
Sequence parallelism

djsaunde:
# Description

This PR implements sequence parallelism via [`ring-flash-attn`](https://github.com/zhuzilin/ring-flash-attention). Specifically, their [hf_adapter.py](https://github.com/zhuzilin/ring-flash-attention/blob/main/ring_flash_attn/adapters/hf_adapter.py) module is used to patch transformers flash attention with `llama3_flash_attn_varlen_func`, the SP implementation from the [llama3 tech report](https://arxiv.org/abs/2407.21783). This _technically_ isn't ring attention, but is the most performant SP variant in most cases.

I think since the batch API (non-sample packing case) is a special case of the varlen API (sample packing case), these changes should be sufficient to cover both cases, but this should be validated with tests.

## Motivation and Context

SP is necessary for long context post-training where the VRAM on a single card results in OOM for a single sequence. If a user has >1 GPUs, they can run longer context post-training by enabling this option.

The attention is distributed across the GPUs according to the set `sequence_parallel_degree` (i.e., if `sequence_parallel_degree = 4`, then sequences are split into 4 equal-length chunks). Attention is computed on each of the sub-sequences, and then comm is done inter-GPU in order to complete the attention computation.

## How has this been tested?

`pytest` coverage (not super comprehensive) and functional tests.

## Types of changes

- `ring-flash-attn` `hf_adapter.py` integration
- Data collation changes (sequence splitting, position ID adjustment)
- `AxolotlTrainer` sampler, dataloader changes
  - Refactor multipack sampler logic to helper method
  - `DistributedSampler` for SP case
    - Setting rank = SP group ID allows us to sample data according to SP group
  - Data loader (in the SP case) is not prepared for distributed training by the accelerator object
    - Distribution already handled by the `DistributedSampler`
- Bonus: added `random_init` flag to load model without pretrained weights
- Bonus: a bit of cleanup


src/axolotl/utils/models.py
```diff
@@ -548,6 +553,14 @@ def apply_patches(self) -> None:
 
             patch_self_attn_lora(self.cfg)
 
+        if self.cfg.sequence_parallel_degree > 1:
```
```suggestion
        if self.cfg.sequence_parallel_degree and self.cfg.sequence_parallel_degree > 1:
```
should fix the NoneType comparison exception in the e2e tests

djsaunde:
Quick test for OOM alleviation:

### Running on 1 x H100 SXM (SP disabled / not possible)

Config (note that `sequence_len: 16384`):

```yaml
base_model: HuggingFaceTB/SmolLM2-1.7B
# Automatically upload checkpoint and final model to HF
# hub_model_id: username/custom_model_name

load_in_8bit: false
load_in_4bit: false
strict: false

datasets:
    type: alpaca
val_set_size: 0.1
output_dir: ./outputs/lora-out

sequence_len: 16384
sample_packing: true
eval_sample_packing: true
pad_to_sequence_len: true

micro_batch_size: 1
num_epochs: 1
optimizer: adamw_8bit
lr_scheduler: cosine
learning_rate: 0.0002

train_on_inputs: false
group_by_length: false
bf16: auto
fp16:
tf32: false

gradient_checkpointing: false
early_stopping_patience:
resume_from_checkpoint:
local_rank:
logging_steps: 1
xformers_attention:
flash_attention: true

warmup_steps: 10
evals_per_epoch: 4
saves_per_epoch: 1
debug:
deepspeed:
weight_decay: 0.0
fsdp:
fsdp_config:
special_tokens:
  pad_token: "<|end_of_text|>"

sequence_parallel_degree: 1

seed: 0

random_init: true
```

This configuration OOMs during the backward pass on the second step of training (~why not on the first step? probably worth looking into~ this can be explained by the fact that optimizer states are lazily loaded (i.e., don't exist in memory until the end of step 1 of training, likely causing the OOM during the forward pass of step 2)).

### Running on 2 x H100 SXM (SP enabled with SP degree = 2)

The above config remains unchanged except for the following change:

```yaml
sequence_parallel_degree: 2
```

This configuration does not OOM and training runs with ~1.6s / iteration.

```bash
-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 560.35.05              Driver Version: 560.35.05      CUDA Version: 12.6     |
|-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA H100 80GB HBM3          On  |   00000000:18:00.0 Off |                    0 |
| N/A   45C    P0            490W /  700W |   40826MiB /  81559MiB |     99%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
|   1  NVIDIA H100 80GB HBM3          On  |   00000000:AB:00.0 Off |                    0 |
| N/A   47C    P0            493W /  700W |   40950MiB /  81559MiB |     99%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+

+-----------------------------------------------------------------------------------------+
| Processes:                                                                              |
|  GPU   GI   CI        PID   Type   Process name                              GPU Memory |
|        ID   ID                                                               Usage      |
|=========================================================================================|
+-----------------------------------------------------------------------------------------+
```

As an aside, this FFT seems to be taking up more memory than it should, I'm probably not configuring something properly (feel free to comment!) (edit: memory is reduced by ~2x by enable gradient checkpointing 🤦).

djsaunde:
Possible nice-to-have: build docs as part of PR. We currently can't see the diff in docs in PRs where it matters (i.e., this PR).

djsaunde:
Similar to the above, but with 4 x H100s, 32K context length, and SP degree = 4:

```
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 565.57.01              Driver Version: 565.57.01      CUDA Version: 12.7     |
|-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA H100 80GB HBM3          On  |   00000000:12:00.0 Off |                    0 |
| N/A   44C    P0            159W /  700W |   71108MiB /  81559MiB |      1%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
|   1  NVIDIA H100 80GB HBM3          On  |   00000000:43:00.0 Off |                    0 |
| N/A   45C    P0            169W /  700W |   71204MiB /  81559MiB |    100%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
|   2  NVIDIA H100 80GB HBM3          On  |   00000000:46:00.0 Off |                    0 |
| N/A   41C    P0            156W /  700W |   71204MiB /  81559MiB |    100%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
|   3  NVIDIA H100 80GB HBM3          On  |   00000000:86:00.0 Off |                    0 |
| N/A   40C    P0            163W /  700W |   70724MiB /  81559MiB |    100%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
```

src/axolotl/monkeypatch/attention/ring_attn.py
```diff
@@ -0,0 +1,68 @@
+"""Ring attention group registration and flash attention patching."""
```
NanoCode012: Should we should cite the following repos?

https://github.com/haoliuhl/ringattention?tab=readme-ov-file#reference

https://github.com/zhuzilin/ring-flash-attention https://github.com/deepspeedai/DeepSpeed/blob/master/blogs/deepspeed-ulysses/README.md

src/axolotl/core/trainers/base.py
```diff
@@ -376,12 +387,19 @@ def __init__(
...
         self.train_data_collator = self.data_collator
         self._stored_metrics = defaultdict(lambda: defaultdict(list))
         if self.args.orpo_alpha:
             self.loss_fct = torch.nn.CrossEntropyLoss(reduction="none")
 
+        if self.args.sequence_parallel_degree > 1:
+            from axolotl.monkeypatch.attention.ring_attn import get_ring_attn_group
+
+            self.ring_attn_group = get_ring_attn_group()
+
```
winglian: Is it feasible to have some sort of SequenceParallelMixin for trainers so we can extend this to other trainers, or is that something we need to address on a trainer case by case basis?

setup.py
```diff
@@ -124,9 +119,7 @@ def get_package_version():
         ],
     },
     extras_require={
-        "flash-attn": [
-            "flash-attn==2.7.4.post1",
-        ],
+        "flash-attn": ["flash-attn==2.7.4.post1", "ring-flash-attn>=0.1.4"],
```
salmanmohammadi: Should `ring-flash-attn` be an additional install option?

docs/config.qmd
```diff
@@ -611,6 +614,11 @@ ddp_timeout:
 ddp_bucket_cap_mb:
 ddp_broadcast_buffers:
 
+# Sequence parallelism
+# Set to a divisor of the number of GPUs available to split sequences into chunks of equal size.
+# Use in long context training to prevent OOM when sequences cannot fit into a single GPU's VRAM.
+sequence_parallel_degree:
```
salmanmohammadi, line 629: Could you provide an example value of what this could be set to?

src/axolotl/core/trainers/base.py
djsaunde: Yeah, I've been thinking about this. I think it's possible but it may be ugly, specifically the logic around sampler / dataloader. I'll take a stab at it, though.

setup.py
djsaunde: Yeah, probably!

src/axolotl/core/trainers/base.py
```diff
@@ -36,7 +38,18 @@
 if is_sagemaker_mp_enabled():
     import smdistributed.modelparallel.torch as smp
 
-LOG = logging.getLogger("axolotl.core.trainer_builder")
+try:
+    from ring_flash_attn import update_ring_flash_attn_params
+except ImportError:
+    # pylint: disable=unused-argument
+    def update_ring_flash_attn_params(*args, **kwargs):
```
salmanmohammadi: This is clever : )

src/axolotl/core/trainers/base.py
```diff
@@ -394,142 +412,262 @@ def _wrap_model(self, model, training=True, dataloader=None):
...
+        # Otherwise prepare with accelerator
+        return self.accelerator.prepare_data_loader(dataloader)
+
+    def get_train_dataloader(self) -> DataLoader:
+        """Get dataloader for training"""
+        train_dataset = self.train_dataset
+        data_collator = self.data_collator  # type: ignore
+
+        # Handle dataset preprocessing
+        if is_datasets_available() and isinstance(train_dataset, datasets.Dataset):
```
salmanmohammadi: Since datasets is a core requirement, do we need to make this check?

src/axolotl/core/trainers/base.py
```diff
@@ -394,142 +412,262 @@ def _wrap_model(self, model, training=True, dataloader=None):
...
+            if self.args.sample_packing and not self.args.pretraining:
+                train_dataset = train_dataset.remove_columns(["length"])
+            if not (self.args.sample_packing and not self.args.pretraining):
```
salmanmohammadi: de morgan's law? : )

src/axolotl/core/trainers/base.py
```diff
@@ -394,142 +412,262 @@ def _wrap_model(self, model, training=True, dataloader=None):
...
-
-            if isinstance(eval_sampler, BatchSampler):
-                dataloader_params["batch_sampler"] = eval_sampler
-                del dataloader_params["batch_size"]
-            else:
-                dataloader_params["sampler"] = eval_sampler
-                dataloader_params["drop_last"] = self.args.dataloader_drop_last
+            # Handle dataset preprocessing for SP
+            if self.args.sequence_parallel_degree > 1:
+                if is_datasets_available() and isinstance(
```
salmanmohammadi: similar comment above about needing to check for datasets

src/axolotl/core/trainers/base.py
```diff
@@ -394,142 +412,262 @@ def _wrap_model(self, model, training=True, dataloader=None):
...
+            )
+
+            return dataloader
 
         return super().get_eval_dataloader(eval_dataset)
 
     def _get_bench_sampler(
         self, bench_dataset: Dataset
-    ) -> Optional[torch.utils.data.Sampler]:
+    ) -> torch.utils.data.Sampler | None:
```
salmanmohammadi, line 325: I totally forgot that we can do this now we're on Python 3.10 😍

src/axolotl/core/trainers/base.py
djsaunde: Hmm I copied it from the superclass method without really thinking about it. I think we can remove but I'll double check

src/axolotl/core/trainers/base.py
djsaunde: wait until you see my pydantic models refactor! 🚀

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
```diff
@@ -1700,6 +1703,20 @@ def check_kto_config(cls, data):
 
         return data
 
+    @field_validator("sequence_parallel_degree", mode="before")
+    @classmethod
+    def check_sequence_parallel_config(cls, value, info):
+        if not value:
+            value = 1
+
+        if value > 1:
```
salmanmohammadi: Worth checking here if ring-flash-attn is installed so we fail early?

src/axolotl/monkeypatch/attention/ring_attn.py
```diff
@@ -0,0 +1,68 @@
...
+
+from typing import Any
+
+import torch.distributed as dist
+from accelerate.logging import get_logger
+from ring_flash_attn import substitute_hf_flash_attn
```
salmanmohammadi: Do we need to check if ring_flash_attn is available here (to avoid import errors when it isnt installed)?

src/axolotl/train.py
```diff
@@ -165,6 +166,18 @@ def terminate_handler(_, __, model_weakref):
         )
 
 
+def train_context_manager(enable=False) -> contextlib.AbstractContextManager:
```
salmanmohammadi: I'm curious what motivated this?

src/axolotl/utils/collators/batching.py
```diff
@@ -1,14 +1,64 @@
...
+) -> torch.Tensor:
+    """
+    Adjust position IDs for a sliced sequence to maintain proper relative positions.
+    This handles the case where position IDs might not be contiguous due to sample packing.
+    """
+    # Convert to tensor if not already
+    if not isinstance(position_ids, torch.Tensor):
+        position_ids = torch.tensor(
+            position_ids,
+            device=position_ids.device if hasattr(position_ids, "device") else None,
```
salmanmohammadi: if `position_ids` isn't a tensor by this point, will it ever have the device attribute?

src/axolotl/monkeypatch/attention/ring_attn.py
djsaunde: No, because this module doesn't get imported until after this is checked earlier in the sequence parallelism logic.

src/axolotl/train.py
djsaunde: It _slightly_ simplifies the `train` function logic so we're not writing `trainer.train` twice. Maybe an overambitious attempt to DRY?

src/axolotl/utils/collators/batching.py
djsaunde: Oops, totally. I'm not even sure this will ever be a list tbh

docs/config.qmd
winglian: maybe something along the lines of "with 8 gpus, valid values include 2 for 4" (since 1 does nothing)

src/axolotl/utils/config/models/input/v0_4_1/__init__.py
djsaunde: Yeah, I like this thought. I think in general we can do a lot of these import checks (that we're currently doing elsewhere in the codebase) at the pydantic model level.

setup.py
```diff
@@ -17,11 +17,7 @@ def parse_requirements():
         lines = [r.strip() for r in requirements_file.readlines()]
         for line in lines:
             is_extras = (
-                "flash-attn" in line
-                or "flash-attention" in line
-                or "deepspeed" in line
-                or "mamba-ssm" in line
-                or "lion-pytorch" in line
+                "deepspeed" in line or "mamba-ssm" in line or "lion-pytorch" in line
```
NanoCode012, line 20: I think we dropped support for lion-pytorch in favor of lion in base torch

requirements.txt
```diff
@@ -64,3 +64,6 @@ schedulefree==1.3.0
 
 axolotl-contribs-lgpl==0.0.6
 axolotl-contribs-mit==0.0.3
+
+# for sequence parallelism
+yunchang==0.6.0
```
NanoCode012: Should this be included with FA as it depends on FA? The repo mentioned it's built on ring-flash-attention repo, do we still need both?

docs/config.qmd
```diff
@@ -32,6 +32,9 @@ tokenizer_legacy:
 resize_token_embeddings_to_32x:
 # Optional[bool] Whether to shrink the embeddings to len(tokenizer). By default, we won't shrink.
 shrink_embeddings:
+# Whether to load the model with randomly initialized weights. Useful for
+# pre-training a model from scratch or debugging purposes.
+random_init:
```
```suggestion
random_init_weights:
```

Would this be a better name?

tests/test_exact_deduplication.py
```diff
@@ -279,6 +279,7 @@ def setUp(self) -> None:
                 "batch_size": 10,
                 "micro_batch_size": 10,
                 "num_epochs": 1,
+                "sequence_parallel_degree": 1,
```
NanoCode012: Is this needed for this test?

src/axolotl/train.py
salmanmohammadi: maybe slightly haha, but it's your call :)

src/axolotl/monkeypatch/attention/ring_attn.py
djsaunde: Can't hurt to cite `ring-flash-attn`. I wouldn't necessarily cite ringattention or deepspeed ulysses here, though.

setup.py
djsaunde: probably should cover that in a separate PR, then.

requirements.txt
djsaunde: yup! I actually made this change locally already

docs/config.qmd
djsaunde: Yes, I like it!

tests/test_exact_deduplication.py
djsaunde: Yes, but only because this config is unnormalized. Normalizing it (after adding a `base_model` attribute) causes the tests to pass even when `sequence_parallel_degree` is removed.

src/axolotl/core/trainers/base.py
djsaunde: Added `SequenceParallelMixin`. Not 100% satisfied with how we're doing these mixins but I think it's an improvement over what I had earlier.

winglian (approved):

Issue #2415, opened 2025-03-14
DPO with the chat_template strategy fails due to extra columns being passed


lhl:
### Please check that this issue hasn't been reported before.

- [x] I searched previous [Bug Reports](https://github.com/axolotl-ai-cloud/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

When running DPO with the ChatTemplate strategy [using the exact fields from the docs](https://axolotl-ai-cloud.github.io/axolotl/docs/rlhf.html#chat_template.default):
```
rl: dpo
datasets:
  - path: ...
    split: train
    type: chat_template.default
    field_messages: "messages"
    field_chosen: "chosen"
    field_rejected: "rejected"
    message_property_mappings:
      role: role
      content: content
    roles:
      user: ["user"]
      assistant: ["assistant"]
      system: ["system"]
```

We should be able to DPO.

### Current behaviour

Instead, I get this error:
```
[rank2]: Traceback (most recent call last):
[rank2]:   File "/fsx/ubuntu/miniforge3/envs/axolotl/lib/python3.12/site-packages/multiprocess/pool.py", line 125, in worker
[rank2]:     result = (True, func(*args, **kwds))
[rank2]:                     ^^^^^^^^^^^^^^^^^^^
[rank2]:   File "/fsx/ubuntu/miniforge3/envs/axolotl/lib/python3.12/site-packages/datasets/utils/py_utils.py", line 678, in _write_generator_to_queue
[rank2]:     for i, result in enumerate(func(**kwargs)):
[rank2]:                      ^^^^^^^^^^^^^^^^^^^^^^^^^
[rank2]:   File "/fsx/ubuntu/miniforge3/envs/axolotl/lib/python3.12/site-packages/datasets/arrow_dataset.py", line 3446, in _map_single
[rank2]:     example = apply_function_on_filtered_inputs(example, i, offset=offset)
[rank2]:               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank2]:   File "/fsx/ubuntu/miniforge3/envs/axolotl/lib/python3.12/site-packages/datasets/arrow_dataset.py", line 3338, in apply_function_on_filtered_inputs
[rank2]:     processed_inputs = function(*fn_args, *additional_args, **fn_kwargs)
[...]

PR #2416, opened 2025-03-14
DPO with the chat_template fix for trl errors (removes extra columns)

lhl:
This fixes my reported issue #2415 where trl is not happy about the original rows being submitted.

When using the [documented fields](https://axolotl-ai-cloud.github.io/axolotl/docs/rlhf.html#chat_template.default), trl will give an error because the original fields are submitted and trl expects *only* prompt (optional), chosen, and rejected for the DPOTrainer.

The only way you could use the existing code without an error is if you made your original fields named prompt, chosen, and rejected I guess? In any case, it seems the best way to fix this (and to allow the field mapping to continue to work) is to remove the original fields from being sbumitted.

To do this, I simply add the original fields to `remove_columns` in the kwargs.

### DIFF
--- a/src/axolotl/prompt_strategies/dpo/chat_template.py
+++ b/src/axolotl/prompt_strategies/dpo/chat_template.py
@@ -91,4 +91,9 @@ def default(
 
         return result
 
-    return transform_fn
+    # Then when you call map, specify the original columns to remove
+    map_kwargs = {
+        "remove_columns": [field_messages, field_chosen, field_rejected] 
+    }
+
+    return transform_fn, map_kwargs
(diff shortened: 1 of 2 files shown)


kubernetes-bad:
I cherry-picked this into latest main and it indeed fixes the problem with extra columns in TRL! 🚀

NanoCode012:
@kubernetes-bad , thanks for your results on this. 

Sorry I haven't gotten to this @lhl. I'll give this a repro run to see.

lhl:
Hey @NanoCode012 let me resubmit after syncing and doing a feature branch...

PR #2420, opened 2025-03-17
Fixing KTO+QLoRA+multi-GPU

salmanmohammadi:


salmanmohammadi:
@NanoCode012 I've traced the issue upstream to TRL. Gradient checkpointing with `use_reentrant=True` is broken for both LoRA and QLoRA, e.g.
```bash
trl kto     \
--dataset_name trl-lib/kto-mix-14k    \
 --model_name_or_path Qwen/Qwen2.5-0.5B     \
--per_device_train_batch_size 2     \
--num_train_epochs 1    \
 --learning_rate 5e-7    \
 --lr_scheduler_type=cosine    \
 --gradient_accumulation_steps 1    \
 --logging_steps 10     \
--eval_steps 500    \
 --output_dir=kto-aligned-model-lora    \
 --warmup_ratio 0.1   \
  --report_to wandb    \
 --bf16    \
 --logging_first_step     \
--use_peft    \
 --load_in_4bit    \
 --lora_target_modules=all-linear   \
  --lora_r=16    \
 --lora_alpha=16 \
--gradient_checkpointing_use_reentrant True \
--gradient_checkpointing True
```
will produce the same error.

NanoCode012:
Would it be better to change the validation to check if multi gpu + rl + use_reentrant to warn/error that it may cause hanging?

NanoCode012 (approved):
I think this is good patch for now till we can isolate root cause upstream

PR #2422, opened 2025-03-18
chore(doc): add instructions on adding custom integrations

NanoCode012:
# Description

Added more info about adding custom integrations

Request from user on discord: https://discord.com/channels/1104757954588196865/1111279858136383509/1351513981784817716

### DIFF
--- a/docs/custom_integrations.qmd
+++ b/docs/custom_integrations.qmd
@@ -77,7 +77,7 @@ To add a new integration, please follow these steps:
 
 ::: {.callout-tip}
 
-See [src/axolotl/integrations/cut_cross_entropy](https://github.com/axolotl-ai-cloud/axolotl/tree/main/src/axolotl/integrations/cut_cross_entropy) for an example of a new integration.
+See [src/axolotl/integrations/cut_cross_entropy](https://github.com/axolotl-ai-cloud/axolotl/tree/main/src/axolotl/integrations/cut_cross_entropy) for an example of a minimal integration.
 
 :::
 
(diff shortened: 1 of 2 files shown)


docs/custom_integrations.qmd
```diff
@@ -55,3 +55,28 @@ sections = [
...
+::: {.callout-note}
+
+See `BasePlugin` at [src/axolotl/integrations/base.py](https://github.com/axolotl-ai-cloud/axolotl/blob/main/src/axolotl/integrations/base.py) for the possible hooks.
+
+:::
+
+To add a new integration, please follow these steps:
+
+1. Create a new folder in the `src/axolotl/integrations` directory.
+2. Add any relevant files (`LICENSE`, `README.md`, `ACKNOWLEDGEMENTS.md`, etc.) to the new folder.
```
ChaosEternal, line 66: I am not a copyright expert, but my suggestion here is for those "in-tree" plugins, they should use the same license as the whole axolotl (for simplicity)

NanoCode012: Currently, we extract all the readme info to the https://axolotl-ai-cloud.github.io/axolotl/docs/custom_integrations.html to be clear about this. But yes, we would only add those permissible as well in this repo.

docs/custom_integrations.qmd
```diff
@@ -55,3 +55,53 @@ sections = [
 for section_name, folder_name in sections:
     print(print_section(section_name, folder_name))
 ```
+
+## Adding a new integration
+
+Adding new plugins allows hooking to multiple parts of the training pipeline for custom features.
```
salmanmohammadi:
```suggestion
Plugins can be used to customize the behavior of the training pipeline through [hooks](https://en.wikipedia.org/wiki/Hooking). See [`axolotl.integrations.BasePlugin`](https://github.com/axolotl-ai-cloud/axolotl/blob/main/src/axolotl/integrations/base.py) for the possible hooks.

```
I would personally put this inline rather than as a callout

salmanmohammadi (approved):

PR #2426, opened 2025-03-20
add 12.8.1 cuda to the base matrix

winglian:
builds the nightly PyTorch against cuda 12.8.1 which is included by default on bare metal lambda


NanoCode012:
Is this a bit early? I don't see pytorch supporting this version yet.

winglian:
> Is this a bit early? I don't see pytorch supporting this version yet.

click on nightly tab. this is more of a smoke test/check, and shouldn't block PRs

PR #2419, opened 2025-03-17
Autodoc generation with quartodoc

djsaunde:
# Description

Title.

## Motivation and Context

Writing [(Google-style) docstrings](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) can now get parsed into the "API Reference" section in our docs.

Modules and classes / functions underneath them are documented only if they are specified in the `quartodoc:` config in the `_quarto.yml` file.

## How has this been tested?

Locally.

## Screenshots (if appropriate)

image

image

image

image

## Types of changes

- quartodoc config
- list / organization of documented modules
- custom CSS
- small codebase changes for autodoc purposes

## Follow ups

Running `quarto build` flags some docstrings as not fitting the Google-style docstring format. Other modules / classes / functions have _no_ docstrings, which should be added eventually.

Custom CSS can and should be improved!
- IMO, headers are too large
- spacing under function signatures is too large

PR #2432, opened 2025-03-21
add preview-docs workflow

djsaunde:
# Description

Adding workflow for docs preview on PR open / change / reopen.

## Motivation and Context

Helpful to see diffs in generated docs, especially for big changes such as in #2419.

## How has this been tested?

TODO.

### DIFF
--- a/.github/workflows/preview-docs.yml
+++ b/.github/workflows/preview-docs.yml
@@ -1,0 +1,49 @@
+name: Preview
+on:
+  workflow_dispatch:
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+permissions:
+  checks: write
+  contents: write
+  deployments: write
+  issues: write
+  discussions: write
+  pages: write
+  pull-requests: write
+  statuses: write
+
+jobs:
+  preview:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v4
+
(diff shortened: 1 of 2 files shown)


.github/workflows/preview-docs.yml
```diff
@@ -0,0 +1,49 @@
+name: Preview
+on:
+  workflow_dispatch:
+  pull_request:
```
winglian, line 4: Probably manual dispatch is all we want? PRs would simply clobber each other

djsaunde: [actions-netlify](https://github.com/nwtgck/actions-netlify) should generate a unique URL per PR!

djsaunde:
@winglian I requested to install the Netlify GitHub app to the Axolotl repo, so if / when we do that we can procede with this PR. Probably will want to weigh the cost / benefit of adding this first, though.

winglian:
Were you able to get the Netlify added?

djsaunde:
@winglian I requested to install the Netlify app to this org, did you see a request come through? I can resend.

image

djsaunde:
nvm, I have org-level perms now, looks like I can do it.

djsaunde:
@winglian Should be good to go once tests pass (actually, tests don't need to pass IMO, no changes made here that impact them).

winglian (approved):

PR #2435, opened 2025-03-21
Feat: Rework multimodal support (mllama, llava, pixtral, qwen2, qwen25, gemma3, mistral3)

NanoCode012:
# Description

This PR continues the work in #2170 and adds multimodal support for:
- mllama
- llava15
- pixtral
- qwen2
- qwen25
- gemma3
- mistral3

We reworked the data collator to use `apply_chat_template` to tokenize via the new transformers 4.49.0 update.

Requires `transformers==4.50.0` released today.

## How has this been tested?

The example configs are ran on cloud pod (but with lower settings to save vram) and verified both train/eval loss decrease appropriately.

We will run against external benchmarks as well.


docs/config.qmd
```diff
@@ -578,6 +578,14 @@ resume_from_checkpoint:
 # Be careful with this being turned on between different models.
 auto_resume_from_checkpoints: false
 
+## Multimodal section
+# int | tuple[int, int] | None . Size to resize images to, width x height.
```
djsaunde, line 590: I like the typing indication here in the config docs. I wonder if there's a better way to do this, e.g.,

```
image_size: int | tuple[int, int] | None = None
```

This would just be duplicating the typing indication in the Axolotl config pydantic model(s).

Maybe even better would be to just generate these docs from the Pydantic models themselves. We could use `Field`s with `json_schema_extra` populated with the comments that we have here.

Just an idea. It might help reduce duplication of efforts.

docs/multimodal.qmd
djsaunde, line 1: Add to `_quarto.yml`?

docs/multimodal.qmd
```diff
@@ -1,28 +1,171 @@
...
+
+Here are the hyperparams you'll need to use to finetune a multimodal model.
 
 ```yaml
-base_model: alpindale/Llama-3.2-11B-Vision-Instruct
 processor_type: AutoProcessor
+
 skip_prepare_dataset: true
+remove_unused_columns: false
+sample_packing: false
```
djsaunde: I would write a quick comment as to why these settings are needed.

ehartford:
thanks for this; I'm currently stuck using llama factory for this feature.

src/axolotl/processing_strategies/__init__.py
djsaunde, line 1: Nit: don't put meaningful code in `__init__.py` files, they're meant for module initialization.

If you make this change, and you want e.g. `ProcessingStrategy` to be import-able from `axolotl.processing_strategies`, import it in `__init__.py`.

Better yet, just define `axolotl/utils/processing_strategies.py` and remove the folder since this is just a single file.

src/axolotl/processing_strategies/__init__.py
```diff
@@ -0,0 +1,272 @@
...
+        Converts different conversation formats to OpenAI format with 'messages'.
+        Supports two formats:
+        1. OpenAI format with 'messages'
+        2. Legacy format with 'conversations'
+
+        Args:
+            examples: list of conversation dictionaries
+
+        Returns:
+            dict in OpenAI format with 'messages' key
```
djsaunde: nit: list of dicts

src/axolotl/processing_strategies/__init__.py
```diff
@@ -0,0 +1,272 @@
...
+            result = deepcopy(example)
+            result.pop("conversations")
+            result["messages"] = messages
+            return result
+
+        def convert_messages_to_multimedia_messages(messages: list[dict]) -> list[dict]:
+            """Convert regular messages format to Messages format with content type"""
+
+            new_messages = []
+            for convo in messages:
```
djsaunde: nit:

```suggestion
            for message in messages:
```

src/axolotl/utils/collators/mm_chat.py
```diff
@@ -2,15 +2,16 @@
...
 from typing import Any, Optional, Union
 
-from PIL import Image
-from transformers import PreTrainedTokenizerBase, ProcessorMixin
+import torch
+from transformers import PreTrainedTokenizerBase
 from transformers.data.data_collator import DataCollatorMixin
 from transformers.utils import PaddingStrategy
 
+from ...processing_strategies import ProcessingStrategy
```
djsaunde: nit: absolute imports are preferred

src/axolotl/utils/collators/mm_chat.py
```diff
@@ -19,174 +20,76 @@ class MultiModalChatDataCollator(DataCollatorMixin):
...
-        self, examples: list[Union[list[int], Any, dict[str, Any]]]
-    ) -> dict[str, Any]:
-        # Handle dict or lists with proper padding and conversion to tensor.
+    def torch_call(self, examples: list[dict]) -> dict[str, Any]:
+        return self.process_rows(examples)
 
-        return self.__class__.process_rows(
-            examples, self.processor, self.chat_template, self.max_images
-        )
+    def process_rows(
```
djsaunde, line 37: nit: missing return type

src/axolotl/utils/collators/mm_chat.py
```diff
@@ -19,174 +20,76 @@ class MultiModalChatDataCollator(DataCollatorMixin):
...
+        self,
+        examples: list[dict],
+    ):
+        # Preprocess the examples
+        examples = self.processing_strategy.preprocess(examples)
```
djsaunde: should we consider making the ProcessingStrategy class callable (implement `__call__` in place of`preprocess`) ?

src/axolotl/utils/collators/mm_chat.py
```diff
@@ -19,174 +20,76 @@ class MultiModalChatDataCollator(DataCollatorMixin):
...
-
-            return images
+            # TODO: Check if need handling for len(input_ids) > sequence_len
 
-        # Preprocess the examples
-        examples = _preprocess(examples)
+            # Add the processed tensors to our batch
+            for key in result.keys():
+                if key not in batch:
+                    batch[key] = []
```
djsaunde, line 66: If we're making this check, I think we can init the batch dict as `{}`.

src/axolotl/utils/config/__init__.py
```diff
@@ -158,7 +158,8 @@ def normalize_config(cfg):
 
     cfg.is_multimodal = (
         hasattr(model_config, "model_type")
-        and model_config.model_type in ["llava", "mllama"]
+        and model_config.model_type
+        in ["llava", "mllama", "qwen2_vl", "qwen2_5_vl", "gemma3", "mistral3"]
```
djsaunde: Nit: constantize this.

docs/config.qmd
NanoCode012: Yep, I agree, that's why I'm really excited with the quartodocs :)

docs/multimodal.qmd
NanoCode012: I was just updating existing file

docs/config.qmd
djsaunde: I'm not sure quartodoc can do it, but it should be relatively easy to do with the script if not!

src/axolotl/utils/models.py
```diff
@@ -70,8 +74,12 @@
 LOG = logging.getLogger(__name__)
 
 MULTIMODEL_AUTO_MODEL_MAPPING = {
```
djsaunde: Just realized there's a typo here.

```suggestion
MULTIMODAL_AUTO_MODEL_MAPPING = {
```

winglian:
should probably merge/rebase from main to get the fixes from #2436 to fix the ring attention test

winglian (approved):

PR #2437, opened 2025-03-23
Updates for trl 0.16.0 - mostly for GRPO

winglian:


salmanmohammadi:
Very nice - does it run OK?

src/axolotl/core/trainers/grpo/__init__.py
```diff
@@ -52,6 +52,18 @@ def set_training_args_kwargs(cls, cfg):
             if trl.vllm_max_model_len:
                 grpo_args_kwargs["vllm_max_model_len"] = trl.vllm_max_model_len
 
+            if trl.vllm_server_host:
```
salmanmohammadi: Are these arguments compatible with `use_vllm`? Should we ensure it's one or the other - the user has either spawned their own vLLM process or wants to handle it in the trainer?

winglian: Trl no longer spawns the vllm server in the trainer and relies on the end user to manage that via trl vllm-serve.

salmanmohammadi: Ugh that's annoying.

Would it be worth checking if the configured vllm server is online if `use_vllm` and erroring out otherwise before we kick of the trainer?

winglian: the grpo trainer will also do this at runtime, so we probably don't need to do that separately.

src/axolotl/core/trainers/grpo/trainer.py
```diff
@@ -2,108 +2,68 @@
...
-        self, model: PreTrainedModel, args: GRPOConfig
-    ) -> PreTrainedModel:
-        """Enables gradient checkpointing for the model."""
-        # pylint: disable=unused-argument,redefined-builtin
-        gradient_checkpointing_kwargs = args.gradient_checkpointing_kwargs or {}
-        use_reentrant = (
-            "use_reentrant" not in gradient_checkpointing_kwargs
-            or gradient_checkpointing_kwargs["use_reentrant"]
+    @profiling_decorator
+    def _move_model_to_vllm(self):
```
winglian, line 25: we still have to override this method b/c the unwrapped model still seems to have a double prefix and I can't convince upstream that this is an issue.

src/axolotl/cli/vllm_serve.py
```diff
@@ -0,0 +1,122 @@
...
+        :param cli_args: additional command-line arguments
+
+    Returns:
+        process_id: the process id of the started VLLM server
+    """
+    cfg = load_cfg(config)
+    model = cfg.base_model
+    tensor_parallel_size = cli_args.tensor_parallel_size
+    host = cli_args.host
+    port = cli_args.port
```
NanoCode012: Should this be passable via config yaml?

src/axolotl/cli/args.py
```diff
@@ -35,6 +35,61 @@ class TrainerCliArgs:
     num_processes: Optional[int] = field(default=None)
 
 
+@dataclass
+class VllmServeCliArgs:
+    """Dataclass with CLI arguments for `axolotl vllm-serve` command."""
```
NanoCode012, line 40: Could we have documentation on how GRPO should be run now? Are users expected to run vllm_serve first?

src/axolotl/core/trainers/grpo/trainer.py
NanoCode012: Why does it have double prefixes ? Is it a side effect from grpo or zero3?

Do we need to add a link to the upstream Issue for future ref?

tests/e2e/multigpu/solo/test_grpo.py
```diff
@@ -98,18 +98,38 @@ def test_llama_dora(self, temp_dir, num_gpus):
...
+            "NCCL_P2P_LEVEL": "LOC",
+            "CUDA_VISIBLE_DEVICES": "1",
+            **current_env,
+        }
+        vllm_process_id = start_vllm(
+            cfg.base_model,
+            env=env,
+            quiet=True,
+            wait=60,
+            gpu_memory_utilization=0.15,
```
NanoCode012, line 186: Do you need to limit this GPU to use this little memory?

src/axolotl/cli/args.py
```diff
@@ -35,6 +35,61 @@ class TrainerCliArgs:
...
+        },
+    )
+    enable_prefix_caching: Optional[bool] = field(
+        default=None,
+        metadata={
+            "help": "Whether to enable prefix caching in vLLM. If set to `True`, ensure that the model and the "
+            "hardware support this feature."
+        },
+    )
+    wait: Optional[int] = field(
```
salmanmohammadi: Should we set a reasonable default here?

src/axolotl/cli/main.py
```diff
@@ -14,7 +14,12 @@
 from dotenv import load_dotenv
 
 import axolotl
-from axolotl.cli.args import EvaluateCliArgs, PreprocessCliArgs, TrainerCliArgs
+from axolotl.cli.args import (
+    EvaluateCliArgs,
+    PreprocessCliArgs,
+    TrainerCliArgs,
+    VllmServeCliArgs,
```
salmanmohammadi, line 21: very optional nit: `vLLMServeCliArgs`

src/axolotl/cli/vllm_serve.py
```diff
@@ -0,0 +1,122 @@
...
+
+    gpu_memory_utilization = (
+        cfg.trl.vllm_gpu_memory_utilization or cli_args.gpu_memory_utilization
+    )
+    dtype = cfg.trl.vllm_dtype or cli_args.dtype
+    max_model_len = cfg.trl.vllm_max_model_len or cli_args.max_model_len
+    enable_prefix_caching = (
+        cfg.trl.enable_prefix_caching or cli_args.enable_prefix_caching
+    )
+    wait = cli_args.wait
```
salmanmohammadi: Yeah I wonder if we should just grab these from `cfg.trl.vllm`?

src/axolotl/cli/vllm_serve.py
```diff
@@ -0,0 +1,122 @@
...
+            except requests.exceptions.RequestException:
+                pass
+
+            time.sleep(1)
+
+    if wait and not started:
+        print(
+            f"VLLM server process did not start within {wait} seconds. Please check your server logs."
+        )
+        process.kill()
```
salmanmohammadi: Do we also need to cleanup the vLLM process?

src/axolotl/cli/vllm_serve.py
```diff
@@ -0,0 +1,122 @@
...
+    if max_model_len := kwargs.get("max_model_len"):
+        cmd.extend(["--max-model-len", str(max_model_len)])
+    if kwargs.get("enable_prefix_caching"):
+        cmd.extend(["--enable-prefix-caching"])
+
+    # print out the command to be executed
+    print(" ".join(cmd))
+
+    # start `trl vllm-serve` command in the background and capture the process id
+    process = subprocess.Popen(  # pylint: disable=consider-using-with
```
salmanmohammadi: I'm slightly confused here - If I run `trl vllm_serve` it opens a subprocess which does `trl vllm_serve`? Where does the `vllm serve` command happen?

src/axolotl/core/trainers/grpo/__init__.py
salmanmohammadi: How are these args configured through `trl vllm_serve`?

src/axolotl/core/trainers/grpo/trainer.py
```diff
@@ -2,108 +2,68 @@
...
-    def _enable_gradient_checkpointing(
-        self, model: PreTrainedModel, args: GRPOConfig
-    ) -> PreTrainedModel:
-        """Enables gradient checkpointing for the model."""
-        # pylint: disable=unused-argument,redefined-builtin
-        gradient_checkpointing_kwargs = args.gradient_checkpointing_kwargs or {}
-        use_reentrant = (
-            "use_reentrant" not in gradient_checkpointing_kwargs
-            or gradient_checkpointing_kwargs["use_reentrant"]
+    @profiling_decorator
```
salmanmohammadi, line 24: TLDR why we're overriding?

src/axolotl/core/trainers/grpo/__init__.py
salmanmohammadi: Ah I see they're configured through the top-level config.
So if the user hasn't done `trl vllm_serve` does the GRPO TRL trainer handle it?

src/axolotl/core/trainers/grpo/trainer.py
winglian: I tried to get trl to include proper handling of some weird double wrapping that's happening see https://github.com/huggingface/trl/pull/2818, but they changed it in my PR 🤷

src/axolotl/cli/args.py
winglian: yeah, I'll make a follow up PR on GRPO docs

tests/e2e/multigpu/solo/test_grpo.py
winglian: it's a small model and if you don't vllm will by default allocate 90% of the GPU memory, which doesn't leave much left for actual training

src/axolotl/cli/args.py
winglian: This is getting dropped from the CLI portion since the CLI doesn't work with subprocessing it, so it will inherently be blocking/waiting.

src/axolotl/cli/vllm_serve.py
winglian: is process.kill not sufficient?

src/axolotl/cli/vllm_serve.py
winglian: originally it was `axolotl vllm-serve config.yaml` that called `trl vllm-serve ...` , the reason being is that we can automatically handle the model name and other things from our own yaml and set those on trl. Now we simply call the module directly and don't subprocess it from the cli. We still subprocess it for tests though.

src/axolotl/cli/vllm_serve.py
```diff
@@ -31,16 +31,18 @@ def do_vllm_serve(
...
-    tensor_parallel_size = cli_args.get("tensor_parallel_size")
-    host = cli_args.get("host")
-    port = cli_args.get("port")
 
-    gpu_memory_utilization = cfg.trl.vllm_gpu_memory_utilization or cli_args.get(
+    tensor_parallel_size = cfg.vllm.tensor_parallel_size or cli_args.get(
+        "tensor_parallel_size"
+    )
+    host = cfg.vllm.host or cli_args.get("host")
+    port = cfg.vllm.port or cli_args.get("port")
```
winglian: this should be reversed, prefer cli overrides

src/axolotl/cli/vllm_serve.py
```diff
@@ -31,16 +31,18 @@ def do_vllm_serve(
...
+    gpu_memory_utilization = cfg.vllm.gpu_memory_utilization or cli_args.get(
         "gpu_memory_utilization"
     )
-    dtype = cfg.trl.vllm_dtype or cli_args.get("dtype")
-    max_model_len = cfg.trl.vllm_max_model_len or cli_args.get("max_model_len")
-    enable_prefix_caching = cfg.trl.enable_prefix_caching or cli_args.get(
+    dtype = cfg.vllm.dtype or cli_args.get("dtype")
+    max_model_len = cfg.vllm.max_model_len or cli_args.get("max_model_len")
+    enable_prefix_caching = cfg.vllm.enable_prefix_caching or cli_args.get(
```
winglian: @SalmanMohammadi Do we need to add a pydantic model for vllm?

src/axolotl/core/trainers/grpo/__init__.py
winglian: the trl trainer simply errors iirc, but the trainer no longer has support for starting the vllm server

src/axolotl/cli/vllm_serve.py
winglian: I went ahead and added this, so should be good to go

src/axolotl/cli/vllm_serve.py
```diff
@@ -0,0 +1,121 @@
...
+        process_id: the process id of the started VLLM server
+    """
+    cfg = load_cfg(config)
+    model = cfg.base_model
+
+    tensor_parallel_size = cfg.vllm.tensor_parallel_size or cli_args.get(
+        "tensor_parallel_size"
+    )
+    host = cfg.vllm.host or cli_args.get("host")
+    port = cfg.vllm.port or cli_args.get("port")
```
winglian:
```suggestion
    host = cli_args.get("host") or cfg.vllm.host
    port = cli_args.get("port") or cfg.vllm.port
```

salmanmohammadi (approved):

src/axolotl/cli/args.py
salmanmohammadi: I've updated the docs.

PR #2446, opened 2025-03-25
expandable segments by default

casper-hansen:
## Motivation and Context

This helped me train with 65k context in axolotl. A common problem at this context length is high memory pressure and fragmentation, which the following default will help alleviate.

## How has this been tested?

I tested with/without and the job crashes without this in PyTorch. At this point, this optimization is common (e.g. it's on by default in Unsloth).

There is a [long blog post](https://www.gilesthomas.com/2024/07/fine-tuning-5) written about this and it concludes that this variable should generally be used.

### DIFF
--- a/src/axolotl/cli/train.py
+++ b/src/axolotl/cli/train.py
@@ -1,7 +1,10 @@
 """CLI to run training on a model."""
 
+import os  # noqa
+# Reduce memory fragmentation in PyTorch for less VRAM usage and more stable training.
+os.environ["PYTORCH_CUDA_ALLOC_CONF"] = "expandable_segments:True"
+
 import logging
-import os
 from pathlib import Path
 from typing import Union
 


salmanmohammadi:
Hi @casper-hansen. Thanks so much for opening this. We already set `expandable_segments` to True [here](https://github.com/axolotl-ai-cloud/axolotl/blob/a9b0733f2c6be3a41ce4b8af8481599f0cf78064/src/axolotl/utils/__init__.py#L42), but we don't override the user's choice if set. 

I'm curious why your job crashes - do you already have `expandable_segments` set in your environment?

casper-hansen:
> Hi @casper-hansen. Thanks so much for opening this. We already set `expandable_segments` to True [here](https://github.com/axolotl-ai-cloud/axolotl/blob/a9b0733f2c6be3a41ce4b8af8481599f0cf78064/src/axolotl/utils/__init__.py#L42), but we don't override the user's choice if set.

Ahh, I didn't see this. 

> I'm curious why your job crashes - do you already have `expandable_segments` set in your environment?

I think it may have been a user error. Closing this for now!

winglian:
So I think we need to add set_pytorch_cuda_alloc_conf to do_train as we broke the feature when we wrapped the train in the axolotl cli

casper-hansen:
> So I think we need to add set_pytorch_cuda_alloc_conf to do_train as we broke the feature when we wrapped the train in the axolotl cli

Not sure where `axolotl train` ends up in the code, but that's what I'm using

PR #2447, opened 2025-03-25
Set the pytorch_cuda_alloc_conf env in the train module

winglian:
Thanks to @casper-hansen for #2446 as a nudge to check where we call `set_pytorch_cuda_alloc_conf()`. It used to get set correctly when we called train, but after the cli refactor, it got moved there, but the cli subsequently calls a subprocess to launch `-m axolotl.cli.train` so we need to make sure it gets included in that path.


djsaunde (approved):
Yup. Meant to do this earlier 😓

salmanmohammadi (approved):

PR #2449, opened 2025-03-27
gemma3 packing fixes

winglian:
transformers needs attention mask dropped in order for sliding window attention to work properly with flash attention


ehartford:
Thank you

ehartford:
Thank you

PR #2451, opened 2025-03-27
add multimodal chat conversion for  multiple choice format

bursteratom:
# Description


NanoCode012:
I think this is an uncommon dataset format to be adding support for?

PR #2455, opened 2025-03-31
Fix(doc): Clarify doc on attention configs and missing pad_token

NanoCode012:
# Description

### DIFF
--- a/src/axolotl/utils/data/shared.py
+++ b/src/axolotl/utils/data/shared.py
@@ -238,7 +238,8 @@ def load_dataset_w_config(
             trust_remote_code=config_dataset.trust_remote_code,
             **load_ds_kwargs,
         )
-    else:
+    elif config_dataset.data_files:
+        fp: str | list[str] | None = None
         if isinstance(config_dataset.data_files, str):
             fp = hf_hub_download(
                 repo_id=config_dataset.path,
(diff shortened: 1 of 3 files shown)


salmanmohammadi (approved):

src/axolotl/utils/data/shared.py
```diff
@@ -238,7 +238,8 @@ def load_dataset_w_config(
             trust_remote_code=config_dataset.trust_remote_code,
             **load_ds_kwargs,
         )
-    else:
+    elif config_dataset.data_files:
+        fp: str | list[str] | None = None
```
winglian, line 242: if we're removing the final else, it leaves the possibility of an unhandled case.

winglian (changes requested):

src/axolotl/utils/data/shared.py
winglian: maybe let's remove this change from this PR and leave this a doc-only PR and address this in a different PR

NanoCode012: I don’t think so. There’s an “ if not ds:” below. Also, this data_files error is raised even if user doesn’t use it as it’s the last else condition that all falls to.

winglian (approved):

PR #2456, opened 2025-03-31
feat: update xformers for new attention interface

NanoCode012:
# Description

Updated xformers llama attention patch for transformers v4.50.0


winglian:
let's make sure we add a quick e2e test for xformers alongside this

NanoCode012:
I initially tested this with a simple lora 1b low-seqlen and there was not much speed gain (<5%), in fact, there was more vram usage.

I think I'll need to test with higher seq length and model to get a better idea

PR #2460, opened 2025-03-31
Release update 20250331

winglian:
# Description

### DIFF
--- a/.github/workflows/tests-nightly.yml
+++ b/.github/workflows/tests-nightly.yml
@@ -46,7 +46,7 @@ jobs:
 
       - name: Install PyTorch
         run: |
-          pip3 install torch==${{ matrix.pytorch_version }} --index-url https://download.pytorch.org/whl/cpu
+          pip3 install torch==${{ matrix.pytorch_version }}
 
       - name: Update requirements.txt
         run: |
@@ -58,8 +58,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          pip3 install --upgrade pip
-          pip3 install --upgrade packaging==23.2
+          pip3 show torch
           pip3 install --no-build-isolation -U -e .
           python scripts/unsloth_install.py | sh
           python scripts/cutcrossentropy_install.py | sh
@@ -75,8 +74,9 @@ jobs:
 
       - name: Run tests
         run: |
-          pytest -n8 --dist loadfile --ignore=tests/e2e/ --ignore=tests/patched/ tests/
-          pytest tests/patched/
+          pytest -v -n8 --dist loadfile --ignore=tests/e2e/ --ignore=tests/patched/ --ignore=tests/cli/ tests/
+          pytest -v tests/patched/
+          pytest -v tests/cli/
 
       - name: cleanup pip cache
         run: |
--- a/.github/workflows/tests-nightly.yml
+++ b/.github/workflows/tests-nightly.yml
@@ -33,6 +33,15 @@ jobs:
       - name: Check out repository code
         uses: actions/checkout@v4
 
+      - name: Restore HF cache
+        id: hf-cache-restore
+        uses: actions/cache/restore@v4
+        with:
+          path: |
+            /home/runner/.cache/huggingface/hub/datasets--*
+            /home/runner/.cache/huggingface/hub/models--*
+          key: ${{ runner.os }}-hf-hub-cache-v2
+
       - name: Setup Python
         uses: actions/setup-python@v5
         with:
@@ -71,6 +80,10 @@ jobs:
       - name: Ensure axolotl CLI was installed
         run: |
           axolotl --help
+
+      - name: Pre-Download dataset fixture
+        run: |
+          huggingface-cli download --repo-type=dataset axolotl-ai-internal/axolotl-oss-dataset-fixtures
 
       - name: Run tests
         run: |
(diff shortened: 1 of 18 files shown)


.github/workflows/tests-nightly.yml
```diff
@@ -46,7 +55,7 @@ jobs:
 
       - name: Install PyTorch
         run: |
-          pip3 install torch==${{ matrix.pytorch_version }} --index-url https://download.pytorch.org/whl/cpu
```
NanoCode012, line 49: Do you recall why we installed cpu? Was it due to some multiple torch conflict?

.github/workflows/tests-nightly.yml
```diff
@@ -58,8 +67,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          pip3 install --upgrade pip
```
NanoCode012, line 61: Do we want to keep this line to update pip?

src/axolotl/core/trainers/mixins/rng_state_loader.py
```diff
@@ -0,0 +1,63 @@
...
+        else:
+            rng_file = os.path.join(checkpoint, "rng_state.pth")
+            if not os.path.isfile(rng_file):
+                LOG.info(
+                    "Didn't find an RNG file, if you are resuming a training that was launched in a distributed "
+                    "fashion, reproducibility is not guaranteed."
+                )
+                return
+
+        with safe_globals():
```
NanoCode012, line 53:
```suggestion
        # Use safe_globals to ensure numpy RNG states can be deserialized safely under PyTorch 2.6+,
        # which requires allowlisted classes when loading with weights_only=True.
        with safe_globals():
```

src/axolotl/core/trainers/mixins/rng_state_loader.py
```diff
@@ -0,0 +1,63 @@
...
+
+        is_distributed = self.args.parallel_mode == ParallelMode.DISTRIBUTED
+        if torch.cuda.is_available():
+            set_rng_state_for_device(
+                "CUDA", torch.cuda, checkpoint_rng_state, is_distributed
+            )
+        if is_torch_npu_available():
+            set_rng_state_for_device(
+                "NPU", torch.npu, checkpoint_rng_state, is_distributed
+            )
```
NanoCode012, line 67: Do we need for MPS `is_torch_mps_available`?

src/axolotl/core/trainers/mixins/rng_state_loader.py
```diff
@@ -0,0 +1,63 @@
+"""
+Temporary fix/override for bug in resume from checkpoint
+
+See https://github.com/huggingface/transformers/pull/37162
```
NanoCode012, line 4:
```suggestion
"""
Temporary fix/override for bug in resume from checkpoint

See https://github.com/huggingface/transformers/pull/37162

TODO: Remove when upstream added PR to release
```

src/axolotl/utils/schemas/config.py
```diff
@@ -1270,3 +1270,13 @@ def check_beta_and_trl_beta_match(cls, data):
...
                 raise ValueError("beta and trl.beta must match or one must be removed")
         return data
+
+    @model_validator(mode="after")
+    def check_min_torch_version(self):
+        if self.env_capabilities and self.env_capabilities.torch_version:
+            torch_version = self.env_capabilities.torch_version
+            if version.parse(torch_version) < version.parse("2.5.1"):
+                LOG.warning(
+                    "torch=={torch_version} may not be supported in future versions. Please consider "
```
```suggestion
                    f"torch=={torch_version} may not be supported in future versions. Please consider "
```

Missing f-string

tests/conftest.py
```diff
@@ -108,43 +118,43 @@ def download_argilla_distilabel_intel_orca_dpo_dataset():
...
+# @disable_hf_offline
+# def dataset_fozzie_alpaca_dpo_dataset_rev_ea82cff(
+#     download_fozzie_alpaca_dpo_dataset,
+# ):  # pylint: disable=unused-argument,redefined-outer-name
+#     return load_dataset(
+#         "fozziethebeat/alpaca_messages_2k_dpo_test", split="train", revision="ea82cff"
+#     )
 
 
 @pytest.fixture(scope="session", autouse=True)
```
NanoCode012, line 160: As these other fixtures are also auto-downloaded, should we include them in the consolidated dataset?

Should we also create a single repo for model downloading?

NanoCode012 (commented):
Really like the change to use a single consolidated dataset!

.github/workflows/tests-nightly.yml
winglian: it gets upgraded above in line 53

.github/workflows/tests-nightly.yml
winglian: I don't. Seems to work now.

tests/conftest.py
winglian: I'll tackle models/tokenizers in a follow-up PR. right now we're only seeing issues with datasets.

winglian: Some of the other downloads are used in e2e tests which tend to not fail. I don't think they will benefit from consolidating them into a single dataset as that is going to make the e2e tests more complicated, but I'll take a look after this release.

src/axolotl/core/trainers/mixins/rng_state_loader.py
winglian: wasn't in the upstream method

PR #2469, opened 2025-04-01
FSDP2 support

winglian:
# Description

### DIFF
--- a/src/axolotl/integrations/liger/__init__.py
+++ b/src/axolotl/integrations/liger/__init__.py
@@ -27,6 +27,7 @@ from axolotl.integrations.base import BasePlugin
 
 from ...utils.distributed import zero_only
 from .args import LigerArgs  # pylint: disable=unused-import. # noqa: F401
+from .utils import patch_with_compile_disable
 
 LOG = logging.getLogger("axolotl.integrations.liger")
 
@@ -40,6 +41,17 @@ class LigerPlugin(BasePlugin):
         return "axolotl.integrations.liger.LigerArgs"
 
     def pre_model_load(self, cfg):
+        if cfg.torch_compile:
+            import liger_kernel.ops.fused_linear_cross_entropy
+
+            patch_with_compile_disable(
+                liger_kernel.ops.fused_linear_cross_entropy,
+                "fused_linear_cross_entropy_forward",
+            )
+            patch_with_compile_disable(
+                liger_kernel.ops.fused_linear_cross_entropy,
+                "fused_linear_cross_entropy_backward",
+            )
         from liger_kernel.transformers.cross_entropy import LigerCrossEntropyLoss
         from liger_kernel.transformers.functional import liger_cross_entropy
         from liger_kernel.transformers.geglu import LigerGEGLUMLP
--- a/src/axolotl/integrations/liger/__init__.py
+++ b/src/axolotl/integrations/liger/__init__.py
@@ -42,6 +42,7 @@ class LigerPlugin(BasePlugin):
 
     def pre_model_load(self, cfg):
         if cfg.torch_compile:
+            # torch compile will unnecessarily attempt to optimize the triton kernel unless explicitly disabled
             import liger_kernel.ops.fused_linear_cross_entropy
 
             patch_with_compile_disable(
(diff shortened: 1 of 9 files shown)


src/axolotl/integrations/liger/__init__.py
```diff
@@ -40,6 +41,17 @@ def get_input_args(self):
...
+            import liger_kernel.ops.fused_linear_cross_entropy
+
+            patch_with_compile_disable(
+                liger_kernel.ops.fused_linear_cross_entropy,
+                "fused_linear_cross_entropy_forward",
+            )
+            patch_with_compile_disable(
+                liger_kernel.ops.fused_linear_cross_entropy,
+                "fused_linear_cross_entropy_backward",
+            )
```
djsaunde: Can you add a quick note as to why this is needed?

src/axolotl/utils/schemas/config.py
```diff
@@ -933,6 +933,7 @@ def check_fsdp_offload_w_8bit_optimizer(cls, data):
             and "8bit" in data.get("optimizer", "")
             and data.get("fsdp_config")
             and data["fsdp_config"].get("fsdp_offload_params")
+            and str(data["fsdp_config"].get("fsdp_version")) != "2"
```
djsaunde, line 953: Is this only support for FSDP 2? I wonder why we had this check before?

djsaunde (commented):
Are additional tests needed?

src/axolotl/utils/schemas/config.py
winglian: So 8bit optimizers didn't work with FSDP1, so this original check was looking for it. with FSDP2, it no longer uses FlatParams  but DTensors which allows us to support it. This check is modified to only continue to warn if it's not FSDP2.

src/axolotl/utils/trainer.py
```diff
@@ -556,6 +558,10 @@ def setup_fsdp_envs(cfg):
         os.environ["FSDP_TRANSFORMER_CLS_TO_WRAP"] = (
             cfg.fsdp_config.fsdp_transformer_layer_cls_to_wrap
         )
+    if cfg.fsdp_config.fsdp_reshard_after_forward is not None:
+        os.environ["FSDP_RESHARD_AFTER_FORWARD"] = (
+            "true" if cfg.fsdp_config.fsdp_reshard_after_forward else "false"
+        )
```
salmanmohammadi, line 564:
```suggestion
        os.environ["FSDP_RESHARD_AFTER_FORWARD"] = cfg.fsdp_config.fsdp_reshard_after_forward
```

salmanmohammadi:
I think we need to make sure loading/saving checkpoints work and catch any optimizers which won't work. Optionally also allow for meta-device initialization (does accelerate support this?)

Here's a good reference for things we should include https://github.com/pytorch/torchtitan/blob/main/docs/fsdp.md

src/axolotl/utils/trainer.py
winglian: That change could leave that env value as `None`, which isn't the intention

salmanmohammadi: Ah mb

winglian:
Looks like we're waiting on https://github.com/huggingface/transformers/pull/37147

EDIT: merged, now need it in a release

S1ro1:
@SalmanMohammadi 
> I think we need to make sure loading/saving checkpoints work and catch any optimizers which won't work. Optionally also allow for meta-device initialization (does accelerate support this?)
> 
> Here's a good reference for things we should include https://github.com/pytorch/torchtitan/blob/main/docs/fsdp.md

I'll be adding support for extra stuff when I have capacity to match torchtitan fully, open to collab so we don't waste capacity working on the same stuff.

Re meta initialization: Closest we are to this is `--fsdp_cpu_ram_efficient_loading` flag, which loads the model on rank 0 and has `torch.emtpy` on other ranks, note that this isn't effectively using meta device, all ranks have stuff on CPU, though non-main ranks are empty tensors. Also to note, this works only for transformers models.

winglian:
thanks @S1ro1, looks like transformers checkpoint saves of FSDP2 is broken atm, so we're blocked on that still

S1ro1:
> thanks @S1ro1, looks like transformers checkpoint saves of FSDP2 is broken atm, so we're blocked on that still

I'll take a look, I think it just uses gathering to full_state_dict which I wanted to do in a smarter way to not spike the mem usage, though the straight-forward solution should on-par with FSDP1 and simple enough

S1ro1:
@winglian what exactly do you need supported? When using `transformers.Trainer` everything worked for me. What method of saving are we talking about exactly.

PR #2470, opened 2025-04-02
fix: disable SP during merge

NanoCode012:
# Description

Issue reported at: https://discord.com/channels/1104757954588196865/1111279858136383509/1356787809620459550

Solution reported to work at: https://discord.com/channels/1104757954588196865/1111279858136383509/1356817423117844490

### DIFF
--- a/src/axolotl/cli/merge_lora.py
+++ b/src/axolotl/cli/merge_lora.py
@@ -74,8 +74,10 @@ def do_cli(config: Union[Path, str] = Path("examples/"), **kwargs) -> None:
         load_in_8bit=False,
         load_in_4bit=False,
         flash_attention=False,
+        sequence_parallel_degree=None,
         deepspeed=None,
         fsdp=None,
+        fsdp_config=None,
         **kwargs,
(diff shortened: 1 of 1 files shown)


NanoCode012:
One change to also consider is whether to disable it during inference as our inference is supposedly single gpu (from last I checked)

salmanmohammadi:
I wonder if it's possible to sanitise configs in a way where we define the acceptable fields for a particular script and automatically discard other fields so we don't run into errors like these (and the RL one you posted a fix for)?

winglian (approved):

salmanmohammadi (approved):

djsaunde:
> I wonder if it's possible to sanitise configs in a way where we define the acceptable fields for a particular script and automatically discard other fields so we don't run into errors like these (and the RL one you posted a fix for)?

Yeah, this would be good to have. We talked previously about having the config being aware of which codepath we're running (train, inference, etc.) and I think that could neatly solve this issue by loading only the relevant config for the codepath.

PR #2473, opened 2025-04-02
fix: gemma3 loss in forward pass

NanoCode012:
# Description

Gemma3 in transformers has really high loss due to using old loss calculation.

Tuning on text-only:

image

Tuning on text-vision:

image

`patched` refers to this branch

transformers: `v4.50.3`

Nvidia 4090 `torch 2.6.0+cu124`

### DIFF
--- a/src/axolotl/monkeypatch/gemma3.py
+++ b/src/axolotl/monkeypatch/gemma3.py
@@ -1,5 +1,7 @@
 """Monkeypatch for gemma3 conditional generation forward to fix loss exploding"""
 
+# pylint: disable=duplicate-code
+
 from typing import Optional, Tuple, Union
 
 import torch
--- a/src/axolotl/monkeypatch/gemma3.py
+++ b/src/axolotl/monkeypatch/gemma3.py
@@ -99,7 +99,7 @@ def new_forward(
 
     is_training = token_type_ids is not None and labels is not None
 
-    # Replace image id woth PAD if the image token if OOV, to avoid index-errors
+    # Replace image id with PAD if the image token is OOV, to avoid index-errors
     if input_ids is not None and self.config.image_token_index >= self.vocab_size:
         special_image_mask = input_ids == self.config.image_token_index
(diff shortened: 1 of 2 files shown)


NanoCode012:
Need some eyes on the handling of `attention_mask` , shifting done properly, and whether we need to update `num_items_in_batch`?

winglian:
Screenshot 2025-04-02 at 7 38 38 AM

The fixed train loss and grad norm seems to batch the previous baseline using CCE. So I think this is probably good to go

winglian (approved):

src/axolotl/monkeypatch/gemma3.py
```diff
@@ -0,0 +1,238 @@
...
+        if output_hidden_states is not None
+        else self.config.output_hidden_states
+    )
+    return_dict = (
+        return_dict if return_dict is not None else self.config.use_return_dict
+    )
+
+    is_training = token_type_ids is not None and labels is not None
+
+    # Replace image id woth PAD if the image token if OOV, to avoid index-errors
```
salmanmohammadi:
```suggestion
    # Replace image id with PAD if the image token is OOV, to avoid index-errors
```

src/axolotl/monkeypatch/gemma3.py
```diff
@@ -0,0 +1,238 @@
...
+    if input_ids is not None and self.config.image_token_index >= self.vocab_size:
+        special_image_mask = input_ids == self.config.image_token_index
+        llm_input_ids = input_ids.clone()
+        llm_input_ids[special_image_mask] = 0
```
salmanmohammadi, line 106: should we grab the pad token here rather than hardcode 0?

salmanmohammadi (approved):

src/axolotl/monkeypatch/gemma3.py
winglian: looks like this part of the patch is simply duplicated from upstream in `modeling_gemma3.py`

PR #284, opened 2023-07-17
Fix bug in dataset loading

emmatyping:
This fixes a bug when loading datasets. `d.data_files` is a list, so it cannot be directly passed to `hf_hub_download`, which only takes a single filename.

PR #2482, opened 2025-04-04
Feat: Add doc on loading datasets and support for Azure/OCI

NanoCode012:
# Description

- Added doc on dataset loading
- Added azure/oci filesystem support
- Delete unknown `train_on_split` param

## How has this been tested?

Azure/OCI uses fsspec, so it should just work as we depend on upstream implementation.

### DIFF
--- a/src/axolotl/utils/data/shared.py
+++ b/src/axolotl/utils/data/shared.py
@@ -6,7 +6,7 @@ from pathlib import Path
 from typing import Optional, Union
 
 from datasets import Dataset, DatasetDict, load_dataset, load_from_disk
-from huggingface_hub import snapshot_download
+from huggingface_hub import hf_hub_download, snapshot_download
 from huggingface_hub.errors import (
     HFValidationError,
     RepositoryNotFoundError,
@@ -251,7 +251,35 @@ def load_dataset_w_config(
             trust_remote_code=config_dataset.trust_remote_code,
             **load_ds_kwargs,
         )
-
+    elif config_dataset.data_files:
+        fp: str | list[str] | None = None
+        if isinstance(config_dataset.data_files, str):
+            fp = hf_hub_download(
+                repo_id=config_dataset.path,
+                repo_type="dataset",
+                filename=config_dataset.data_files,
+                revision=config_dataset.revision,
+            )
+        elif isinstance(config_dataset.data_files, list):
+            fp = []
+            for file in config_dataset.data_files:
+                fp.append(
+                    hf_hub_download(
+                        repo_id=config_dataset.path,
+                        repo_type="dataset",
+                        filename=file,
+                        revision=config_dataset.revision,
+                    )
+                )
+        else:
+            raise ValueError("data_files must be either a string or list of strings")
+        ds = load_dataset(
+            "json",
+            name=config_dataset.name,
+            data_files=fp,
+            streaming=streaming,
+            **load_ds_kwargs,
+        )
     if not ds:
         raise ValueError("unhandled dataset load")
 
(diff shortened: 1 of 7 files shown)


src/axolotl/utils/data/shared.py
```diff
@@ -125,28 +122,44 @@ def load_dataset_w_config(
...
+        # storage_options = {
+        #     "tenant_id": AZURE_STORAGE_TENANT_ID,
+        #     "client_id": AZURE_STORAGE_CLIENT_ID,
+        #     "client_secret": AZURE_STORAGE_CLIENT_SECRET,
+        # }
+        # # Gen 2
+        # storage_options = {
+        #     "account_name": AZURE_STORAGE_ACCOUNT_NAME,
+        #     "account_key": AZURE_STORAGE_ACCOUNT_KEY,
+        # }
```
winglian, line 148: we could also warn in the pydantic validation if we detect `adl://` or `abfs://` but the requisite env vars aren't set

src/axolotl/utils/data/shared.py
```diff
@@ -238,35 +251,7 @@ def load_dataset_w_config(
...
-                )
-        else:
-            raise ValueError("data_files must be either a string or list of strings")
-        ds = load_dataset(
-            "json",
-            name=config_dataset.name,
-            data_files=fp,
-            streaming=streaming,
-            **load_ds_kwargs,
-        )
```
winglian: I don't think this is the PR to remove this in. Let's keep the PR's small and focused. If anything, this should get moved into the `elif ds_from_hub:` conditional rather than being removed. (First introduced in #284)

NanoCode012: Reverted! Will move this to separate PR

src/axolotl/utils/data/shared.py
```diff
@@ -96,20 +96,17 @@ def load_dataset_w_config(
...
-                "s3:// paths require aiobotocore and s3fs to be installed"
-            ) from exc
+            raise ImportError("s3:// paths require s3fs to be installed") from exc
 
-        # Takes credentials from ~/.aws/credentials for default profile
-        s3_session = aiobotocore.session.AioSession(profile="default")
-        storage_options = {"session": s3_session}
+        # Reads env, credentials from ~/.aws/credentials, or IAM metadata provider
+        # https://s3fs.readthedocs.io/en/latest/index.html?highlight=storage_options#credentials
+        storage_options = {"anon": False}
```
winglian, line 109: isn't anon: True a valid option too for publicly exposed buckets/objects?

winglian (approved):

src/axolotl/utils/data/shared.py
NanoCode012: Hm, it's a bit tricky as they can also fallback to several cred combinations. I think the users would get an error if the files can't be reached `ds not found` etc..

src/axolotl/utils/data/shared.py
NanoCode012: I'm not sure why, but S3, we need `anon: False` for it to auto-pick from env https://s3fs.readthedocs.io/en/latest/api.html#s3fs.core.S3FileSystem

Same for Azure.

If public buckets are off interest, I'm proposing we enable `storage_options: dict` config to be passable

winglian:
The failing multi-gpu tests are due to hitting the timeout, but aren't blockers for merging this. I'm addressing that in another PR

src/axolotl/utils/data/shared.py
winglian: thanks, yeah, let's just release this as is and see who asks for that support

PR #2483, opened 2025-04-04
fix: cohere cce scaling wrong tensor

NanoCode012:
# Description

Evaluated on `CohereForAI/c4ai-command-r7b-12-2024` using `examples/cohere/command-r-7b-qlora.yml` with slight modification to reduce vram requirements.

Loss fixed

image

Memory is the same before/after change
image


winglian (approved):
lgtm

PR #2484, opened 2025-04-04
Fix(doc): use zero3_bf16.json if using bf16

NanoCode012:
# Description


docs/multi-gpu.qmd
```diff
@@ -51,6 +51,12 @@ We provide default configurations for:
 - ZeRO Stage 2 (`zero2.json`)
 - ZeRO Stage 3 (`zero3.json`)
 
+::: {.callout-tip}
```
winglian, line 54: We should probably include the other configurations we have in `deepspeed_configs/` with explanations of the differences.

winglian:
looks like you pushed this branch on top of the Azure branch?

PR #2485, opened 2025-04-04
check if fixture exists in the cache already

winglian:
Due to pytest worker concurrency (n=8), it hits the HF API for each worker. This is almost always already cached, so we can avoid some of the API checks with this PR

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -14,10 +14,15 @@ import datasets
 import pytest
 import requests
 from huggingface_hub import snapshot_download
+from huggingface_hub.errors import LocalEntryNotFoundError
 from tokenizers import AddedToken
 from transformers import AutoTokenizer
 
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -47,6 +52,11 @@ def wrapper(*args, **kwargs):  # pylint: disable=inconsistent-return-statements
 @retry_on_request_exceptions(max_retries=3, delay=5)
 @disable_hf_offline
 def snapshot_download_w_retry(*args, **kwargs):
+    with hf_offline_context(True):
+        try:
+            return snapshot_download(*args, **kwargs)
+        except LocalEntryNotFoundError:
+            pass
     return snapshot_download(*args, **kwargs)
```
djsaunde: Can you add a doc string explaining what this does? It's a little confusing

djsaunde (approved):
Just a nit, otherwise I think this makes sense.

PR #2486, opened 2025-04-04
simplify the example configs to be more minimal and less daunting

winglian:
There are a lot of extraneous options that we present in the configs that may overwhelm new users. Let's drop them when possible


djsaunde (approved):
Awesome!

Follow up ideas:
- normalized ordering / grouping for config YAML
- smoke tests for each YAML in the config (I don't think we should have an example that errors OOTB)

PR #2488, opened 2025-04-05
fix tokenizer overrides w gemma3

winglian:
gemma3 token overrides were getting appended to the id list rather than replaced

### DIFF
--- a/src/axolotl/utils/models.py
+++ b/src/axolotl/utils/models.py
@@ -283,6 +283,13 @@ def modify_tokenizer_files(
                         raise ValueError(
                             f"Token ID {token_id} not found in added_tokens"
                         )
+            if "model" in tokenizer_data and "vocab" in tokenizer_data["model"]:
+                for token_id, new_value in token_id_mappings.items():
+                    for entry_val, entry_id in tokenizer_data["model"]["vocab"].items():
+                        if entry_id == token_id:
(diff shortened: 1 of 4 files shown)


src/axolotl/utils/models.py
```diff
@@ -283,6 +283,13 @@ def modify_tokenizer_files(
                         raise ValueError(
                             f"Token ID {token_id} not found in added_tokens"
                         )
+            if "model" in tokenizer_data and "vocab" in tokenizer_data["model"]:
+                for token_id, new_value in token_id_mappings.items():
+                    for entry_val, entry_id in tokenizer_data["model"]["vocab"].items():
+                        if entry_id == token_id:
+                            del tokenizer_data["model"]["vocab"][entry_val]
+                            tokenizer_data["model"]["vocab"][new_value] = token_id
+                            break
```
NanoCode012, line 292: Should we add a 

```
else:
                        # Reaching this section means the token_id was not found in tokenizer.json->model->vocab
                         raise ValueError(
                             f"Token ID {token_id} not found in model vocab"
                         )
```

winglian: I don't think so, this seemed to be specific to gemma3, and didn't fail with llama

winglian: to be specific, gemma3 needed both files modified.

PR #2489, opened 2025-04-05
replace attention in the yaml config with an enum

winglian:
simplication to use a single field an enum value for attention backends. We can start to deprecate the old fields in a future release


src/axolotl/utils/schemas/config.py
```diff
@@ -220,6 +220,10 @@ class AxolotlInputConfig(
         },
     )
 
+    attention: AttentionBackend = Field(
+        default=AttentionBackend.flash,
```
NanoCode012: Not sure whether to default to enabling Flash? Some models don't work with flash (maybe Pixtral)

src/axolotl/utils/schemas/config.py
```diff
@@ -412,6 +416,76 @@ def check_gptq_w_revision(cls, data):
...
+        ):
+            return data
+        if data.get("attention") == AttentionBackend.s2 and data.get("s2_attention"):
+            return data
+        if data.get("attention") == AttentionBackend.sdpa and data.get("sdp_attention"):
+            return data
+        if data.get("attention") == AttentionBackend.xformers and data.get(
+            "xformers_attention"
+        ):
+            return data
```
NanoCode012: Is this check needed?

I think this function could be re-arranged.

First, check for mismatch (third block -> first block). Then set attention if not done (second block). This first block can be removed?

src/axolotl/utils/schemas/config.py
```diff
@@ -412,6 +416,76 @@ def check_gptq_w_revision(cls, data):
             )
         return data
 
+    @model_validator(mode="before")
+    @classmethod
+    def normalize_attn(cls, data):  # pylint: disable=too-many-return-statements
+        # cases where both are set and already match
+        if data.get("attention") == AttentionBackend.eager and data.get(
```
salmanmohammadi: nit: grab the `attention` field at the top rather than lots of `data.get("attention")`

src/axolotl/utils/schemas/config.py
```diff
@@ -412,6 +416,76 @@ def check_gptq_w_revision(cls, data):
             )
         return data
 
+    @model_validator(mode="before")
```
salmanmohammadi, line 443: Should we add a deprecation warning that `attention` will be the reccomended way to configure the attention backend in future releases?

src/axolotl/utils/schemas/config.py
salmanmohammadi: Should all these be `or`?

src/axolotl/utils/schemas/config.py
```diff
@@ -434,6 +438,83 @@ def check_gptq_w_revision(cls, data):
...
+                raise ValueError("attention mismatch with sdp_attention already set")
+        if data.get("xformers_attention"):
+            if not attention:
+                data["attention"] = AttentionBackend.xformers
+                return data
+            if attention != AttentionBackend.xformers:
+                raise ValueError(
+                    "attention mismatch with xformers_attention already set"
+                )
+
```
NanoCode012, line 489: Do we need a default fallback here?

src/axolotl/utils/schemas/config.py
```diff
@@ -434,6 +438,83 @@ def check_gptq_w_revision(cls, data):
             )
         return data
 
+    @model_validator(mode="before")
+    @classmethod
+    def normalize_attn(cls, data):  # pylint: disable=too-many-return-statements
```
NanoCode012, line 445: I think we can refactor this to use mapping. Just a sample refactor

```python
@classmethod
def normalize_attn(cls, data):
    attention = data.get("attention")
    
    # Define mapping between enum values and flag names
    backend_mapping = {
        AttentionBackend.eager: "eager_attention",
        AttentionBackend.flash: "flash_attention",
        AttentionBackend.flex: "flex_attention",
        AttentionBackend.s2: "s2_attention",
        AttentionBackend.sdpa: "sdp_attention",
        AttentionBackend.xformers: "xformers_attention",
    }
    
    # Check if any attention flag is set
    any_flag_set = any(data.get(flag_name) for flag_name in backend_mapping.values())
    
    # CASE 1: attention is set but no flags are set - set the corresponding flag
    if attention and not any_flag_set:
        flag_name = backend_mapping.get(attention)
        if flag_name:
            data[flag_name] = True
        return data
    
    # CASE 2: no attention set but flags are set - set attention from flags
    if not attention:
        LOG.warning(
            "*_attention will be deprecated soon. One of `attention: eager | flash | flex | s2 | sdp | xformers` is recommended"
        )
        
        # Find the first True flag and set attention accordingly
        for backend, flag_name in backend_mapping.items():
            if data.get(flag_name):
                data["attention"] = backend
                return data
    
    # CASE 3: both attention and flags are set - check for consistency
    if attention:
        expected_flag = backend_mapping.get(attention)
        for backend, flag_name in backend_mapping.items():
            # If a flag is set that doesn't match the attention value
            if data.get(flag_name) and flag_name != expected_flag:
                raise ValueError(f"attention mismatch with {flag_name} already set")
    
    return data
```

(would also need to handle fallback)

examples/replit-3b/config-lora.yml
```diff
@@ -38,7 +38,7 @@ tf32: true
 gradient_checkpointing:
 resume_from_checkpoint:
 logging_steps: 1
-flash_attention:
+attention: flash
```
NanoCode012:
```suggestion
attention: eager
```

This model is mpt-based and I don't think they work with flash. They hardcoded the attention forward.

https://github.com/huggingface/transformers/blob/main/src/transformers/models/mpt/modeling_mpt.py

NanoCode012 (commented):
Did a quick look through the example yaml changes. They seem fine except one.

NanoCode012:
Refactored and fixed conflicts.

NanoCode012 (approved):

PR #2495, opened 2025-04-06
SP cu_seqlens fix, refactor

djsaunde:
# Description

I realized that the cu_seqlens calculation was just a placeholder and incorrect for multipacking. This PR fixes that. I was also able to move more logic into the SP data collation logic and simplify it.

## Motivation and Context

This issue was raised on discord by @lhl. Thanks for the catch!

## How has this been tested?

Running `axolotl train` for `SmolLM2-1.7B` on 2 x H100s:

SP = 2 vs. SP = 1 shows nearly identical training losses:

W B Chart 4_5_2025, 8_40_56 PM

Diffs can _likely_ be chalked up to numerical differences / implementation details between the `ring-flash-attn` and `transformers` flash attention implementations.

Also tested with SP = 8 and training losses are similarly close:

W B Chart 4_5_2025, 9_52_22 PM

## Types of changes

- fix for `cu_seqlens` calculation
- refactor
- smoke test for SP

### DIFF
--- a/src/axolotl/monkeypatch/attention/ring_attn.py
+++ b/src/axolotl/monkeypatch/attention/ring_attn.py
@@ -6,7 +6,9 @@ package, specifically the `hf_adapter.substitute_hf_flash_attn` function to patc
 their sequence parallel version of Flash Attention 2.
 """
 
+import torch
 import torch.distributed as dist
+import torch.nn.functional as F
 from accelerate.logging import get_logger
 
 from axolotl.logging_config import configure_logging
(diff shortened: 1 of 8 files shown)


src/axolotl/monkeypatch/attention/ring_attn.py
```diff
@@ -98,3 +100,72 @@ def register_ring_attn(sequence_parallel_degree: int, heads_k_stride: int | None
     substitute_hf_flash_attn(
         process_group=get_ring_attn_group(), heads_k_stride=heads_k_stride
     )
+
+
+def calculate_packed_seq_lens(position_ids: torch.Tensor) -> torch.Tensor:
```
winglian: Were you aware of `get_cu_seqlens_from_pos_ids` in `axolotl.monkeypatch.utils`? We have unit tests already for that one too, so you may be able to reuse that or replace it if this is more performant and matches the functionality from the unit tests.

djsaunde: Nope! Updated to use.

winglian (approved):

djsaunde:
FYI, I detected another bug with grad accumulation, will fix on this PR before merging.

djsaunde:
> FYI, I detected another bug with grad accumulation, will fix on this PR before merging.

I've tracked this down to how the `transformers.train.Trainer` computes the loss for logging across ranks: it just averages them. This works fine for the data parallel setting, but in SP, where the last sub-sequence generally has fewer tokens and higher (or, at least, more variable) losses, its contribution to the average with the other ranks' losses is too high; i.e., it should be weighted by the number of non-padding tokens out of the total in the full sequence.

I checked this by comparing losses on a single-GPU run vs. SP with ranks 2 and 4, and computing the re-weighted loss as described above. The logging is not straightforward to patch, so perhaps we should just live with this?

This also seems to imply that the contribution to the gradient update for the last sub-sequence will be outsized. Maybe we can handle this by implementing a pre-optimizer callback that rescales it contribution somehow, but it's not immediately clear if a callback can do this.

Also worth noting is that the transformers trainer just grabs the gradient norm from the 0-th rank, so it can't be expected to match the value in the single GPU or DDP settings.

djsaunde:
Oh, and this seems to explain the differences between the losses in both the `gradient_accumulation_steps = 1` and `gradient_accumulation_steps > 1` settings. The differences in losses gets accumulated as the gradient accumulation steps increases.

winglian:
Maybe document this as a TODO/errata and just warn and link to errata when gradient accumulation steps > 1 and sequence_parallel_size > 1? I'm happy enough to ship this as is. It should be less of an issue as long as packing is used and the packing efficiency is high

PR #2503, opened 2025-04-08
fix: preprocess yielding whole dataset to each worker

### LINKED ISSUE #2496: Prepared datasets take up 32x more disk space than normal

### Expected Behavior

The dataset should only take up 1x its size. Additionally, you would not need to deduplicate if this bug was fixed. Note that the reason for filing this bug report is that I have identified why it's happening. 

### Current behaviour

Due to a bug in axolotl's preprocessing functions, the dataset ends up taking 32x the size when saved to disk. The number 32x is simply from the default `cfg.dataset_processes`. _This happens because multiple workers are iterating over the full dataset as opposed to only the worker's share of the dataset._

https://github.com/axolotl-ai-cloud/axolotl/blob/38df5a36ea7c40787c95886d01fafe117dfadc62/src/axolotl/utils/data/sft.py#L336-L346

### Steps to reproduce

Use any of the example configs in axolotl. Check size of dataset of main branch versus my suggested fix below.

### Possible solution

Note that I don't have time to create a PR, figured I would just drop a patch here to fix it.

Fix this by appropriately yielding:

```python
    def gen_from_iter_ds(_ds, worker_id: List[int], num_workers: List[int]):
        for i, item in enumerate(_ds):
            if i % num_workers[0] == worker_id[0]:
                yield item

    dataset = Dataset.from_generator(
        functools.partial(gen_from_iter_ds, dataset),
        features=dataset.features,
        num_proc=num_workers,
        split="train",
        gen_kwargs={
            "worker_id": list(range(num_workers)),
            "num_workers": [num_workers] * num_workers
        }
    )
    dataset.save_to_disk(str(prepared_hash_path))
```

- [x] Linux
- [ ] macOS
- [ ] Windows

### Python Version

3.10

### axolotl branch-commit

main

### Acknowledgements

NanoCode012:
# Description

Closes #2496 

## Social Handles (Optional)

Thanks to casperhansen for fix.

### DIFF
--- a/src/axolotl/utils/data/sft.py
+++ b/src/axolotl/utils/data/sft.py
@@ -332,16 +332,23 @@ def load_tokenized_prepared_datasets(
         if cfg.local_rank == 0 and not cfg.skip_prepare_dataset:
             LOG.info(f"Saving merged prepared dataset to disk... {prepared_ds_path}")
             if isinstance(dataset, IterableDataset):
+                num_workers = cfg.dataset_processes
 
-                def gen_from_iter_ds(_ds, _=None):
-                    yield from _ds
+                def gen_from_iter_ds(_ds, worker_id: List[int], num_workers: List[int]):
+                    """Generator function to correctly splice the dataset for each worker"""
+                    for i, item in enumerate(_ds):
+                        if i % num_workers[0] == worker_id[0]:
+                            yield item
 
                 ds_from_iter = Dataset.from_generator(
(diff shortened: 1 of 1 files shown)


winglian:
When does the dataset we're saving become an iterable dataset?

NanoCode012:
That's the question I had in mind as well. Isn't it for pretraining only but this is sft.py

However, the fix does make sense hence I kept this up.

djsaunde (approved):

PR #2511, opened 2025-04-09
fix: allow merge lora on pre-quantized model

NanoCode012:
# Description

When merging adapter to pre-quantized model (like `axolotl-quants/Llama-4-Scout-17B-16E-Linearized-bnb-nf4-bf16`), there would be two errors:
- `ValueError: model_config.quantization_config is set but `gptq` flag is not.` due to pre-quantization
- `ValueError: You cannot cast a bitsandbytes model in a new `dtype`.` as it's still a bnb model at that point.

This PR fixes that and allows merging adapter to quantized NF4 model.

## How has this been tested?

Tested to run with adapter trained on llama scout nf4 and merged.

### DIFF
--- a/src/axolotl/cli/merge_lora.py
+++ b/src/axolotl/cli/merge_lora.py
@@ -32,7 +32,13 @@ def do_merge_lora(*, cfg: DictDefault) -> None:
 
     LOG.info("Running merge of LoRA with base model...")
     model = model.merge_and_unload(progressbar=True)
-    model.to(dtype=cfg.torch_dtype)
+    try:
+        model.to(dtype=cfg.torch_dtype)
+    except ValueError as e:
+        LOG.warning("Failed to convert model to dtype %s", cfg.torch_dtype)
+        LOG.warning("Ignore this if the base_model is pre-quantized.")
+        LOG.warning("Error raised: %s", e)
+
     model.generation_config.do_sample = True
 
     if cfg.local_rank == 0:
--- a/src/axolotl/utils/models.py
+++ b/src/axolotl/utils/models.py
@@ -151,11 +151,12 @@ def check_model_config(cfg: DictDefault, model_config: PretrainedConfig):
             "Please make sure to point to a GPTQ model."
         )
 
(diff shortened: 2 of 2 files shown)


src/axolotl/cli/merge_lora.py
```diff
@@ -32,7 +32,13 @@ def do_merge_lora(*, cfg: DictDefault) -> None:
 
     LOG.info("Running merge of LoRA with base model...")
     model = model.merge_and_unload(progressbar=True)
-    model.to(dtype=cfg.torch_dtype)
+    try:
+        model.to(dtype=cfg.torch_dtype)
+    except ValueError as e:
+        LOG.warning("Failed to convert model to dtype %s", cfg.torch_dtype)
+        LOG.warning("Ignore this if the base_model is pre-quantized.")
+        LOG.warning("Error raised: %s", e)
```
salmanmohammadi, line 40: can just do `raise e`

src/axolotl/utils/models.py
```diff
@@ -151,11 +151,12 @@ def check_model_config(cfg: DictDefault, model_config: PretrainedConfig):
...
-        raise ValueError(
-            "model_config.quantization_config is set but `gptq` flag is not. "
-            "Please use the `gptq` flag to train quantized model or point to a non-quantized model."
-        )
+    # if quant_config_exists, transformers will use the quant_config to load the model
+    # if not cfg.gptq and quant_config_exists and not cfg.load_in_4bit:
+    #     raise ValueError(
+    #         "model_config.quantization_config is set but `gptq` flag is not. "
+    #         "Please use the `gptq` flag to train quantized model or point to a non-quantized model."
+    #     )
```
salmanmohammadi: just delete?

salmanmohammadi (approved):
does it work? : )

src/axolotl/cli/merge_lora.py
NanoCode012: In this case, I want it to allow and just pass the error along. It'll be saved as NF4 back

PR #2513, opened 2025-04-10
feat: add CNAME

NanoCode012:
# Description

We need to add this file to root alongside the quarto.yml. Then, change the DNS as CNAME, not A record. 

https://quarto.org/docs/publishing/github-pages.html#custom-domain

I think even gh pages doc site is broken

### DIFF
--- a/CNAME
+++ b/CNAME
@@ -1,0 +1,1 @@
+docs.axolotl.ai


NanoCode012:
Force merge as the doc site is kind of broken at the moment.

PR #2520, opened 2025-04-11
batch api HF adapter for ring-flash-attn; cleanup and improvements

djsaunde:
# Description

Added an adapter similar to [`ring-flash-attn`'s](https://github.com/zhuzilin/ring-flash-attention/blob/c587a0def9eaac7709e0d439757033178f30069f/ring_flash_attn/adapters/hf_adapter.py#L93) for the ring attention batch API.

We should probably upstream this adapter at some point.

## Motivation and Context

This allows us to support sequence parallelism for the non-sample packed case, and allows for microbatch size > 1 (in that case). This will be useful for e.g. GRPO training, where we're currently not sample packing and use >1 microbatch size.

## How has this been tested?

Varied microbatch_size in [1, 2, 4] and 

Added smoke tests for these added features.

Example with varying SP in 1, 2, 4 and fixing microbatch size at 2:

W B Chart 4_10_2025, 11_59_51 PM

FYI: losses ~match due to a fix in the logged loss calculation in the huggingface trainer (we may upstream).

## TODO / follow-ups

- [ ] Gradient explosion with `pad_to_sequence_len: false`
- [x] Additional HF adapters for the remaining `ring-flash-attn` batch functions (`zigzag` and `stripe`)
- [ ] Additional HF adapters for the remaining varlen functions (`ring`, `zigzag`)
- [ ] Move to `https://github.com/feifeibear/long-context-attention` instead?
  - This might just entail adding HF adapters for these that match what we're currently using + testing for perf gains / different scenarios

### DIFF
--- a/docs/sequence_parallelism.qmd
+++ b/docs/sequence_parallelism.qmd
@@ -28,7 +28,7 @@ sequence_parallel_degree: 4  # Split sequences across 4 GPUs
 # Optional; strides across the key dimension. Larger values use more memory but should make training faster.
 heads_k_stride: 1
 # Optional; one of "varlen_llama3", "batch_ring", "batch_zigzag", "batch_stripe". Defaults to
-"varlen_llama3" when `sample_packing: true`, and "batch_ring" otherwise.
+# "varlen_llama3" when `sample_packing: true`, and "batch_ring" otherwise.
 ring_attn_func:
 ```
 
--- a/docs/sequence_parallelism.qmd
+++ b/docs/sequence_parallelism.qmd
@@ -27,6 +27,9 @@ To enable sequence parallelism, add the following to your configuration file:
 sequence_parallel_degree: 4  # Split sequences across 4 GPUs
 # Optional; strides across the key dimension. Larger values use more memory but should make training faster.
 heads_k_stride: 1
+# Optional; one of "varlen_llama3", "batch_ring", "batch_zigzag", "batch_stripe". Defaults to
+"varlen_llama3" when `sample_packing: true`, and "batch_ring" otherwise.
+ring_attn_func:
 ```
 
 The `sequence_parallel_degree` should be a divisor of the total number of GPUs. For example:
--- a/src/axolotl/monkeypatch/attention/ring_attn/patch.py
+++ b/src/axolotl/monkeypatch/attention/ring_attn/patch.py
@@ -6,6 +6,8 @@ package, specifically the `hf_adapter.substitute_hf_flash_attn` function to patc
 their sequence parallel version of Flash Attention 2.
 """
 
+from enum import Enum
+
 import torch
 import torch.distributed as dist
 from accelerate.logging import get_logger
@@ -15,6 +17,7 @@ from axolotl.monkeypatch.utils import get_cu_seqlens_from_pos_ids
 
 configure_logging()
 LOG = get_logger(__name__)
+
 
 RING_ATTN_GROUP = None
 
@@ -40,7 +43,23 @@ def set_ring_attn_group(ring_attn_group: dist.ProcessGroup | None):
     RING_ATTN_GROUP = ring_attn_group
 
 
-def register_ring_attn(sequence_parallel_degree: int, heads_k_stride: int | None):
+class RingAttnFunc(Enum):
+    """Enum class for supported `ring-flash-attn` implementations"""
+
+    # VARLEN_RING = "varlen_ring"
+    # VARLEN_ZIGZAG = "varlen_zigzag"
+    VARLEN_LLAMA3 = "varlen_llama3"
+    BATCH_RING = "batch_ring"
+    # BATCH_ZIGZAG = "batch_zigzag"
+    # BATCH_STRIPE = "batch_stripe"
+
+
+def register_ring_attn(
+    sequence_parallel_degree: int,
+    heads_k_stride: int | None,
+    sample_packing: bool,
+    ring_attn_func: RingAttnFunc | None,
+):
     """
     Create ring attention group and substitute flash attn with ring flash attn.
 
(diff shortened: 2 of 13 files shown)


djsaunde:
SP = 4, microbatch size = 2, batch ring vs. batch stripe vs. batch zigzag ~match. There are some notable diffs for zigzag on a few steps, however 🤔

W B Chart 4_11_2025, 12_47_58 AM (1)

[wandb](https://wandb.ai/axolotl-ai/seq-parallel-batch?nw=nwuserdanaxolotl&panelDisplayName=train%2Floss&panelSectionName=train)

docs/sequence_parallelism.qmd
```diff
@@ -27,6 +27,9 @@ To enable sequence parallelism, add the following to your configuration file:
 sequence_parallel_degree: 4  # Split sequences across 4 GPUs
 # Optional; strides across the key dimension. Larger values use more memory but should make training faster.
 heads_k_stride: 1
+# Optional; one of "varlen_llama3", "batch_ring", "batch_zigzag", "batch_stripe". Defaults to
+"varlen_llama3" when `sample_packing: true`, and "batch_ring" otherwise.
```
salmanmohammadi:
```suggestion
# "varlen_llama3" when `sample_packing: true`, and "batch_ring" otherwise.
```

src/axolotl/monkeypatch/attention/ring_attn/patch.py
```diff
@@ -40,22 +43,58 @@ def set_ring_attn_group(ring_attn_group: dist.ProcessGroup | None):
     RING_ATTN_GROUP = ring_attn_group
 
 
-def register_ring_attn(sequence_parallel_degree: int, heads_k_stride: int | None):
+class RingAttnFunc(str, Enum):
+    """Enum class for supported `ring-flash-attn` implementations"""
+
+    # VARLEN_RING = "varlen_ring"
+    # VARLEN_ZIGZAG = "varlen_zigzag"
```
SalmanMohammadi, line 50:
```suggestion
```
nit

src/axolotl/monkeypatch/attention/ring_attn/patch.py
```diff
@@ -40,22 +43,58 @@ def set_ring_attn_group(ring_attn_group: dist.ProcessGroup | None):
     RING_ATTN_GROUP = ring_attn_group
 
 
-def register_ring_attn(sequence_parallel_degree: int, heads_k_stride: int | None):
+class RingAttnFunc(str, Enum):
+    """Enum class for supported `ring-flash-attn` implementations"""
+
+    # VARLEN_RING = "varlen_ring"
```
salmanmohammadi, line 49: how come these are commented out?

src/axolotl/monkeypatch/attention/ring_attn/patch.py
```diff
@@ -40,22 +43,58 @@ def set_ring_attn_group(ring_attn_group: dist.ProcessGroup | None):
...
+    BATCH_RING = "batch_ring"
+    BATCH_ZIGZAG = "batch_zigzag"
+    BATCH_STRIPE = "batch_stripe"
+
+
+def register_ring_attn(
+    sequence_parallel_degree: int,
+    heads_k_stride: int | None,
+    sample_packing: bool,
+    ring_attn_func: str | None,
```
salmanmohammadi: nit: could type with `RingAttnFunc`? Or just indicate in the docstring it should be one of those values.

src/axolotl/utils/collators/batching.py
```diff
@@ -157,18 +157,20 @@ def apply_sequence_parallelism(
...
+            update_ring_attn_params(
+                input_ids=batch["input_ids"], position_ids=batch.get("position_ids")
+            )
 
         # Slice batch for sequence parallel processing
-        keys_to_slice = ["input_ids", "attention_mask", "labels", "position_ids"]
-        for key in keys_to_slice:
-            if key in batch:
+        for key in batch:
+            if batch[key].size(1) == total_seq_len:
```
salmanmohammadi: Will there ever be keys in the batch that fulfill ` batch[key].size(1) == total_seq_len` but which we don't want to slice?

tests/e2e/multigpu/test_sp.py
```diff
@@ -17,8 +17,14 @@
 class TestSequenceParallelism:
     """Test case for training with sequence parallelism enabled"""
 
-    def test_sequence_parallel_training(self, temp_dir):
-        # pylint: disable=duplicate-code
+    def _run_sequence_parallel_test(
```
salmanmohammadi, line 21: Worth throwing a `parametrize` here to exercise the different options for this test?

salmanmohammadi (approved):

src/axolotl/monkeypatch/attention/ring_attn/patch.py
djsaunde: not yet supported

src/axolotl/utils/collators/batching.py
djsaunde: I don't think so? The thought behind this was to extend it to arbitrary sequence data.

src/axolotl/monkeypatch/attention/ring_attn/patch.py
djsaunde: would need another adapter module, but would be pretty straightforward to add to this PR

tests/e2e/multigpu/test_sp.py
djsaunde: Maybe, though I don't want to add too many tests + extend the runtime too much.

src/axolotl/monkeypatch/attention/ring_attn/patch.py
winglian: Yeah. Enum or Literal type

PR #2534, opened 2025-04-16
Feat: HQQ integration

bursteratom:
# Description

### DIFF
--- a/src/axolotl/utils/schemas/config.py
+++ b/src/axolotl/utils/schemas/config.py
@@ -383,7 +383,7 @@ class AxolotlInputConfig(
     @classmethod
     def check_hqq_config_redundancy(cls, data):
         if (data.get("load_in_4bit") or data.get("load_in_8bit")) and data.get(
-            "hqq_nbits"
+            "use_hqq"
         ):
 
             raise ValueError(
--- a/src/axolotl/utils/schemas/config.py
+++ b/src/axolotl/utils/schemas/config.py
@@ -44,7 +44,6 @@ from axolotl.utils.schemas.model import (
 )
 from axolotl.utils.schemas.multimodal import MultiModalConfig
 from axolotl.utils.schemas.peft import LoraConfig, ReLoRAConfig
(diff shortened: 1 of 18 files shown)


src/axolotl/utils/schemas/quant.py
```diff
@@ -0,0 +1,51 @@
...
+            "nbits": cfg.hqq_nbits,
+            "group_size": cfg.hqq_group_size,
+        }
+
+    hqq_target_module = cfg.hqq_target_module
+    if not isinstance(cfg.hqq_target_module, list):
+        hqq_target_module = [hqq_target_module]
+
+    hqq_quant_config_kwargs = {"dynamic_config": {}}
+    for module in hqq_target_module:
```
winglian: I think this isn't what we want. Just let the user specify the entire dynamic config. Dynamic config is for when you want different quantization for different models. As this currently is implemented, it is not very different from just specifying it without this.

src/axolotl/utils/schemas/config.py
```diff
@@ -59,6 +60,7 @@ class AxolotlInputConfig(
     ModelOutputConfig,
     LoraConfig,
     ReLoRAConfig,
+    HQQConfig,
```
winglian: I don't see this used anywhere. And the hqq_ prefix doesn't match the proposed schema we discussed

bursteratom: oh yeah that was what I originally roughly implemented for debugging, will update as we discussed

src/axolotl/utils/schemas/quant.py
bursteratom: Yeah this is not final but an illustration of my thinking that for each set of unique config we specify which model layers will be using that, so users don't have to copy and paste the same quant config for two model layers, reducing user input

bursteratom:
@winglian the multigpu tests error seem to be an issue with the CI. When doing the pytest locally on 2xA100 multigpu tests all pass

kallewoof:
Is this being worked on?

PR #2535, opened 2025-04-16
Fixed Rex Scheduler Warm Up

Catgat:
# Description

Fixes an issue with the Rex scheduler where the warmup was incorrect due to the last step calculation going negative on training step 1.

## Motivation and Context

This change fixes the warm up for Rex which was incorrectly spiking the LR on the first turn. This occurs if your final LR is greater than zero ( with cosine_min_lr_ratio > 0 )

## How has this been tested?

I ran a training job both with and without the changes. With the changes I observed step 1 LR was correctly calculated.

Green line is after changes. 
Red line is before changes.
image

### DIFF
--- a/src/axolotl/utils/schedulers.py
+++ b/src/axolotl/utils/schedulers.py
@@ -40,7 +40,7 @@ class RexLR(LRScheduler):
         self.max_lr = max_lr
         self.total_steps = total_steps
         self.num_warmup_steps = num_warmup_steps
-        self.last_step = last_step - 1
+        self.last_step = max(last_step - 1,0)
 
         # Ensure each parameter group has an "initial_lr" key to avoid issues when resuming.
         for group in optimizer.param_groups:
--- a/src/axolotl/utils/schedulers.py
+++ b/src/axolotl/utils/schedulers.py
@@ -40,7 +40,7 @@ class RexLR(LRScheduler):
         self.max_lr = max_lr
         self.total_steps = total_steps
         self.num_warmup_steps = num_warmup_steps
-        self.last_step = max(last_step - 1,0)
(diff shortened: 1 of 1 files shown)


NanoCode012:
hey @xzuyn , do you recall why you had `last_step` here to be `-1` in first place?

I notice we have a weird param handling around `last_step` and `last_epoch`

https://github.com/axolotl-ai-cloud/axolotl/blob/1d6347043b5912deb0802f1622346700ceff9721/src/axolotl/utils/schedulers.py#L52-L58

and passing 
https://github.com/axolotl-ai-cloud/axolotl/blob/1d6347043b5912deb0802f1622346700ceff9721/src/axolotl/utils/schedulers.py#L49-L50

but using `last_step` as a step counter instead of epoch counter in 
https://github.com/axolotl-ai-cloud/axolotl/blob/1d6347043b5912deb0802f1622346700ceff9721/src/axolotl/utils/schedulers.py#L60-L81

xzuyn:
> warmup was incorrect due to the last step calculation going negative on training step 1

I think it happens with specific total step counts or some configurations, I didn't have this problem initially when I tested. Sad that I missed it.

> do you recall why you had `last_step` here to be `-1` in first place?

https://github.com/IvanVassi/REX_LR/blob/main/lr_scheduler.py#L4C69-L4C82
I was just adapting the original code with as little changes as possible to get stuff working right. Other schedulers I've seen have `-1` as well.

> but using last_step as a step counter instead of epoch counter in

I can't remember why I even added that in the first place, I might've done it when I was wrapping my head around why it was spiking at the end of training. Other schedulers use `last_epoch` normally, even tho its counting the steps not epochs. Could probably just go back to `last_epoch` to simplify things.

PR #2539, opened 2025-04-18
Add Post_model_load, post_lora_load, post_train, post_train_unload function calls

divyanshuaggarwal:
Add Post_model_load, post_lora_load, post_train, post_train_unload function calls in the training pipeline to help integrations module.

# Description

Added  function calls to use integrations more effectively

## Motivation and Context
While using integrations for my use case, I could not use the above functions as they were never called in the training pipeline.

## How has this been tested?
tested it in my environment. 

## Types of changes
code change

## Social Handles (Optional)
twitter: @divyanshuaggarw

### DIFF
--- a/src/axolotl/cli/train.py
+++ b/src/axolotl/cli/train.py
@@ -1,5 +1,6 @@
 """CLI to run training on a model."""
 
+import gc
 import logging
 import os
 from pathlib import Path
@@ -49,11 +50,11 @@ def do_train(cfg: DictDefault, cli_args: TrainerCliArgs):
 
     model, tokenizer, trainer = train(cfg=cfg, dataset_meta=dataset_meta)
 
-    plugin_manager = PluginManager.get_instance()
-    plugin_manager.post_train(cfg, model)
-
     del model, tokenizer, trainer
 
+    gc.collect()
+
+    plugin_manager = PluginManager.get_instance()
     plugin_manager.post_train_unload(cfg)
 
 
--- a/src/axolotl/cli/train.py
+++ b/src/axolotl/cli/train.py
@@ -48,9 +48,12 @@ def do_train(cfg: DictDefault, cli_args: TrainerCliArgs):
         dataset_meta = load_datasets(cfg=cfg, cli_args=cli_args)
 
     model, tokenizer, trainer = train(cfg=cfg, dataset_meta=dataset_meta)
-    del model, tokenizer, trainer
 
     plugin_manager = PluginManager.get_instance()
+    plugin_manager.post_train(cfg, model)
+    
+    del model, tokenizer, trainer
+    
     plugin_manager.post_train_unload(cfg)
 
 
--- a/src/axolotl/cli/train.py
+++ b/src/axolotl/cli/train.py
@@ -51,9 +51,9 @@ def do_train(cfg: DictDefault, cli_args: TrainerCliArgs):
 
     plugin_manager = PluginManager.get_instance()
     plugin_manager.post_train(cfg, model)
-    
+
     del model, tokenizer, trainer
-    
+
     plugin_manager.post_train_unload(cfg)
 
 
--- a/src/axolotl/integrations/base.py
+++ b/src/axolotl/integrations/base.py
@@ -82,11 +82,8 @@ class BasePlugin:
         """
         Performs actions after the model is built/loaded, but before any adapters are applied.
 
-        Parameters:
-        cfg (dict): The configuration for the plugin.
-
-        Returns:
-        None
+        Args:
+            cfg (dict): The configuration for the plugin.
         """
 
     def post_model_load(self, cfg, model):  # pylint: disable=unused-argument
--- a/src/axolotl/integrations/base.py
+++ b/src/axolotl/integrations/base.py
@@ -457,7 +457,7 @@ class PluginManager:
             if plugin_callbacks:
                 callbacks.extend(plugin_callbacks)
         return callbacks
-        
+
     def post_train(self, cfg, model):
         """
         Calls the post_train method of all registered plugins.
@@ -471,7 +471,7 @@ class PluginManager:
         """
         for plugin in self.plugins.values():
             plugin.post_train(cfg, model)
-            
+
     def post_train_unload(self, cfg):
         """
         Calls the post_train_unload method of all registered plugins.
--- a/src/axolotl/integrations/base.py
+++ b/src/axolotl/integrations/base.py
@@ -457,7 +457,21 @@ class PluginManager:
             if plugin_callbacks:
                 callbacks.extend(plugin_callbacks)
         return callbacks
+        
+    def post_train(self, cfg, model):
+        """
+        Calls the post_train method of all registered plugins.
 
+        Parameters:
+        cfg (dict): The configuration for the plugins.
+        model (object): The loaded model.
+
+        Returns:
+        None
+        """
+        for plugin in self.plugins.values():
+            plugin.post_train(cfg, model)
+            
     def post_train_unload(self, cfg):
         """
         Calls the post_train_unload method of all registered plugins.
--- a/src/axolotl/integrations/base.py
+++ b/src/axolotl/integrations/base.py
@@ -36,9 +36,10 @@ class BasePlugin:
     Methods:
     register(cfg): Registers the plugin with the given configuration.
     pre_model_load(cfg): Performs actions before the model is loaded.
-    post_model_load(cfg, model): Performs actions after the model is loaded.
+    post_model_build(cfg, model): Performs actions after the model is loaded, but before LoRA adapters are applied.
     pre_lora_load(cfg, model): Performs actions before LoRA weights are loaded.
     post_lora_load(cfg, model): Performs actions after LoRA weights are loaded.
+    post_model_load(cfg, model): Performs actions after the model is loaded, inclusive of any adapters.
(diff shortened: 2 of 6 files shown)


winglian (approved):
thanks for filling in the gaps on this!

src/axolotl/train.py
```diff
@@ -495,6 +495,10 @@ def train(
         processor,
     ) = setup_model_and_trainer(cfg, dataset_meta)
 
+    plugin_manager = PluginManager.get_instance()
```
winglian: This needs to be imported above.

src/axolotl/cli/train.py
```diff
@@ -48,9 +48,12 @@ def do_train(cfg: DictDefault, cli_args: TrainerCliArgs):
         dataset_meta = load_datasets(cfg=cfg, cli_args=cli_args)
 
     model, tokenizer, trainer = train(cfg=cfg, dataset_meta=dataset_meta)
-    del model, tokenizer, trainer
 
     plugin_manager = PluginManager.get_instance()
+    plugin_manager.post_train(cfg, model)
```
winglian: `src/axolotl/cli/train.py:53: error: "PluginManager" has no attribute "post_train"  [attr-defined]`

perhaps there were some additions you didn't commit?

winglian (changes requested):

src/axolotl/cli/train.py
divyanshuaggarwal: Thanks for pointing, updated with the fix.

src/axolotl/train.py
divyanshuaggarwal: added.

src/axolotl/train.py
```diff
@@ -495,6 +496,10 @@ def train(
...
+    plugin_manager.post_model_load(cfg, model)
+    plugin_manager.post_lora_load(cfg, model)
```
NanoCode012: Seems like these two have no differences? Should `.post_model_load` be run prior to loading any lora adapters?

winglian:
@divyanshuaggarwal I rebased this PR on the latest main and slightly tweaked the hooks and naming to add better granularity and meaning. Let me know what you think. thanks!

divyanshuaggarwal:
> @divyanshuaggarwal I rebased this PR on the latest main and slightly tweaked the hooks and naming to add better granularity and meaning. Let me know what you think. thanks!

@winglian  It looks good to me.

src/axolotl/train.py
djsaunde: Agreed with @NanoCode012!

src/axolotl/utils/models.py
```diff
@@ -1331,6 +1330,10 @@ def load_model(self) -> Tuple[PreTrainedModel, Optional[PeftConfig]]:
                 before_kbit_train_or_finetune=False,
             )
 
+        plugin_manager = PluginManager.get_instance()
+        plugin_manager.post_model_build(self.cfg, self.model)
```
salmanmohammadi: When would we need both of these calls?

src/axolotl/integrations/base.py
```diff
@@ -77,6 +78,17 @@ def pre_model_load(self, cfg):  # pylint: disable=unused-argument
...
 
+    def post_model_build(self, cfg, model):  # pylint: disable=unused-argument
+        """
+        Performs actions after the model is built/loaded, but before any adapters are applied.
+
+        Parameters:
+        cfg (dict): The configuration for the plugin.
+
+        Returns:
+        None
```
djsaunde: nit

```suggestion
        Args:
            cfg (dict): The configuration for the plugin.
```

src/axolotl/integrations/base.py
```diff
@@ -329,9 +341,25 @@ def pre_model_load(self, cfg):
...
+        """
+        Calls the post_model_build method of all registered plugins after the model has been built/loaded,
+        but before any adapters have been applied.
+
+        Parameters:
+        cfg (dict): The configuration for the plugins.
+        model (object): The loaded model.
+
+        Returns:
+        None
```
djsaunde: nit

```suggestion
        Args:
            cfg (dict): The configuration for the plugins.
            model (object): The loaded model.
```

src/axolotl/utils/models.py
winglian: While practically they could be made the same, I think it might be semantically more obvious that one is specific to LoRA use cases. Happy to merge them and document if that is better.

salmanmohammadi: Could just remove the call to `plugin_manager.pre_lora_load(self.cfg, self.model)` here and keep both

src/axolotl/integrations/base.py
djsaunde: see https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings

src/axolotl/utils/models.py
```diff
@@ -1427,9 +1430,15 @@ def load_adapter(model, cfg, adapter, inference=False):
     if hasattr(model, "enable_input_require_grads"):
         model.enable_input_require_grads()
     if adapter in ["lora", "qlora"]:
-        return load_lora(model, cfg, inference=inference)
+        res = load_lora(model, cfg, inference=inference)
+        plugin_manager = PluginManager.get_instance()
+        plugin_manager.post_lora_load(cfg, res[0])
+        return res
```
djsaunde:
```suggestion
        model, lora_config = load_lora(model, cfg, inference=inference)
        plugin_manager = PluginManager.get_instance()
        plugin_manager.post_lora_load(cfg, model)
        return model, lora_config
```

salmanmohammadi (approved):
One comment about potentially removing an extra call to the plugin, but otherwise LGTM

src/axolotl/utils/models.py
```diff
@@ -1427,9 +1430,15 @@ def load_adapter(model, cfg, adapter, inference=False):
...
     if adapter == "llama-adapter":
-        return load_llama_adapter(model, cfg)
+        res = load_llama_adapter(model, cfg)
+        plugin_manager = PluginManager.get_instance()
+        plugin_manager.post_lora_load(cfg, res[0])
+        return res
```
djsaunde:
```suggestion
        model, lora_config = load_llama_adapter(model, cfg)
        plugin_manager = PluginManager.get_instance()
        plugin_manager.post_lora_load(cfg, model)
        return model, lora_config
```

src/axolotl/utils/models.py
djsaunde: IMO the `post_model_build` call should come immediately after `skip_move_to_device = self.build_model(qlora_fsdp)`.

src/axolotl/train.py
```diff
@@ -495,6 +496,9 @@ def train(
         processor,
     ) = setup_model_and_trainer(cfg, dataset_meta)
 
+    plugin_manager = PluginManager.get_instance()
+    plugin_manager.post_model_load(cfg, model)
```
djsaunde: This is is called twice. I prefer the other call at the end of the `load_model` function.

src/axolotl/utils/models.py
djsaunde: Consider also setting `PLUGIN_MANAGER = PluginManager.get_instance()` as a module-level constant.

src/axolotl/utils/models.py
```diff
@@ -74,6 +75,7 @@
 from axolotl.utils.model_shard_quant import load_sharded_model, load_sharded_model_quant
 
 LOG = logging.getLogger(__name__)
+PLUGIN_MANAGER = PluginManager.get_instance()
```
winglian, line 78: @djsaunde I'm not 100% that this should get instantiated here when the module is loaded/imported, but it doesn't seem to have any major side effects rn, but something to keep an eye on.

winglian (approved):
thanks! Looks good to me. Will merge tomorrow to give a chance for any additional feedback.

src/axolotl/utils/models.py
NanoCode012: Should we keep it consistent with how we fetch this instance?

In some places, we take this global var, in others, we fetch it locally.

src/axolotl/utils/models.py
```diff
@@ -1392,7 +1395,7 @@ def load_model(self) -> Tuple[PreTrainedModel, Optional[PeftConfig]]:
             gc.collect()
             torch.cuda.empty_cache()
 
-        # TODO resume_from_checkpoint handling
```
NanoCode012, line 1395: Do we need to keep this comment?

tests/e2e/integrations/test_hooks.py
```diff
@@ -0,0 +1,184 @@
...
+    """
+
+    def test_plugin_hooks(self, temp_dir):
+        # pylint: disable=duplicate-code
+        cfg = DictDefault(
+            {
+                "base_model": "HuggingFaceTB/SmolLM2-135M",
+                "plugins": [
+                    "tests.e2e.integrations.test_hooks.LogHooksPlugin",
+                ],
```
NanoCode012, line 123: Is this plugins path accessible?

NanoCode012: I don't see the `integrations` folder being run in CI https://github.com/axolotl-ai-cloud/axolotl/actions/runs/14699492579/job/41248447774

winglian: Runs in the e2e tests since it requires an actual train run

PR #2545, opened 2025-04-22
Fix(doc): add delinearize instruction

NanoCode012:
# Description

### DIFF
--- a/docs/cli.qmd
+++ b/docs/cli.qmd
@@ -201,13 +201,13 @@ See [LM Eval Harness](https://github.com/EleutherAI/lm-evaluation-harness) for m
 
 ### delinearize-llama4
 
-Delinearizes a Llama 4 linearized model into a regular HuggingFace Llama 4 model.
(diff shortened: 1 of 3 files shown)


docs/cli.qmd
```diff
@@ -199,6 +199,17 @@ output_dir: # Directory to save evaluation results
 
 See [LM Eval Harness](https://github.com/EleutherAI/lm-evaluation-harness) for more details.
 
+### delinearize-llama4
+
+Delinearizes a Llama 4 linearized model into a regular HuggingFace Llama 4 model.
+
+```bash
+axolotl delinearize-llama4 --model path/to/model_dir --output path/to/output_dir
```
salmanmohammadi, line 207: Do we need to point out how to do this for quantized delinearized models?

NanoCode012: From my reading, for training adapter with quantized linearized models, the user should merge the adapter to the non-quantized linearized version, then delinearize it. Is this flow correct, @winglian ?

winglian: correct, merge with unquantized linearized, then delinearize

salmanmohammadi (approved):
lgtm!

PR #2551, opened 2025-04-23
add base docker image with pytorch 2.7.0 and variant for cuda 12.8

winglian:
we include FA2 in the base build b/c it doesn't have a wheel and would take forever to build otherwise.


salmanmohammadi (approved):

salmanmohammadi:
Does this mean we can drop 2.4.1?

winglian:
the plan is to officially drop 2.4.1 when we release 0.9.0

PR #2553, opened 2025-04-23
Sequence parallel training context manager

djsaunde:
# Description

We push the logic for splitting the data along the sequence dimension into a context manager which, on entrance, registers a forward hook on the model to be trained which splits the data prior to the forward pass. The hook is removed on context manager exit, once training is complete.

Inspired by [this blog post](https://discuss.pytorch.org/t/distributed-w-torchtitan-breaking-barriers-training-long-context-llms-with-1m-sequence-length-in-pytorch-using-context-parallel/215082).

## Motivation and Context

To avoid having SP logic in the data collator(s) (to separate concerns), and to make SP more easily extensible to the RL trainers.

## How has this been tested?

Some manual testing with SP = 2 gives identical loss curves to `main` with the same settings, and matches the SP = 1 setting; need to do more testing with SP > 2.

May need to add / modify pytest tests.


src/axolotl/core/trainers/mixins/sequence_parallel.py
```diff
@@ -87,3 +96,106 @@ def _sp_get_eval_sampler(self, eval_dataset) -> Sampler | None:
...
+        # Will store hook handles for removal
+        self.hook_handle: RemovableHandle | None = None
+
+    def __enter__(self):
+        # Define a forward pre-hook to apply sequence parallelism with kwargs support
+        def sequence_parallel_pre_hook(
+            module, args, kwargs
+        ):  # pylint: disable=unused-argument
+            # Apply sequence parallelism to kwargs
+            kwargs = self.apply_sequence_parallelism(kwargs)
```
winglian: This method accepts `batch`, is it the same as `kwargs` here?

src/axolotl/core/trainers/mixins/sequence_parallel.py
```diff
@@ -87,3 +96,106 @@ def _sp_get_eval_sampler(self, eval_dataset) -> Sampler | None:
...
+        )
+
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        # Remove the forward pre-hook
+        self.hook_handle.remove()
+        self.hook_handle = None
+
+    def apply_sequence_parallelism(
```
winglian: Would moving this out of the context manager make it easier to test? We could use `functools.partial()` possibly to ensure we define some of the things like world_size, attn_func, and local_rank as args

winglian:
@djsaunde the test fixes in https://github.com/axolotl-ai-cloud/axolotl/pull/2554 should resolve a lot of the test failures you're seeing in this PR.

salmanmohammadi (approved):
Honestly no real concerns here, awesome PR

src/axolotl/core/trainers/mixins/sequence_parallel.py
djsaunde: Yup. The items in a batch are passed in the HF trainer as kwargs:

```
outputs = model(**inputs)
```

src/axolotl/core/trainers/mixins/sequence_parallel.py
djsaunde: done!

djsaunde:
Added a post-hook for gathering model outputs across ranks. This should make the GRPO SP implementation way simpler :)

src/axolotl/utils/trainer.py
```diff
@@ -348,7 +348,7 @@ def drop_no_trainable_tokens(sample):
                     load_from_cache_file=not cfg.is_preprocess,
                     desc="Add position_id column (PoSE)",
                 )
-    elif cfg.sample_packing or cfg.sequence_parallel_degree > 1:
+    elif cfg.sample_packing:
```
winglian, line 351: Just to clarify my understanding, the new implementation doesn't need position_ids anymore?

djsaunde: Position IDs are only needed in the sample packed case!

winglian:
Seems like this PR is breaking the `tests/e2e/multigpu/test_llama.py::TestMultiGPULlama::test_ds_zero3_packed` case for qlora=False cases. The multi-gpu ci didn't run against this PR unfortunately. Running this test with the commit before this was merged passes.

```bash
pytest -n1 -v tests/e2e/multigpu/test_llama.py::TestMultiGPULlama::test_ds_zero3_packed[False-deepspeed_configs/zero3_bf16.json-1]
```

djsaunde:
> Seems like this PR is breaking the `tests/e2e/multigpu/test_llama.py::TestMultiGPULlama::test_ds_zero3_packed` case for qlora=False cases. The multi-gpu ci didn't run against this PR unfortunately. Running this test with the commit before this was merged passes.
> 
> ```shell
> pytest -n1 -v tests/e2e/multigpu/test_llama.py::TestMultiGPULlama::test_ds_zero3_packed[False-deepspeed_configs/zero3_bf16.json-1]
> ```

Testing now

PR #2558, opened 2025-04-25
fix: crash when pretraining_dataset with dispatch_batches is false

chiwanpark:
# Description

After #2536, when we use `pretraining_dataset` without setting `accelerator_config.dispatch_batches` to true explicitly, we got `UnboundedLocalError` for `iter_ds` variable in [line 145 of src/axolotl/utils/data/sft.py](https://github.com/axolotl-ai-cloud/axolotl/blob/main/src/axolotl/utils/data/sft.py#L145).

To fix this error, this PR changes the default value of `accelerator_config.dispatch_batches` to true if `pretraining_dataset` is set.

### DIFF
--- a/src/axolotl/utils/data/sft.py
+++ b/src/axolotl/utils/data/sft.py
@@ -134,10 +134,9 @@ def prepare_dataset(cfg, tokenizer, processor=None, preprocess_iterable=None):
                     "csv", data_files=f.name, split="train", streaming=True
                 )
         else:
-            if is_local_main_process():
-                iter_ds = load_dataset(
-                    path, streaming=True, split=split, name=name, data_files=data_files
-                )
+            iter_ds = load_dataset(
+                path, streaming=True, split=split, name=name, data_files=data_files
+            )
 
         if skip:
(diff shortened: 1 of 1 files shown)


NanoCode012:
I'm not really clear on the usage of the dispatch batches.

So, if I'm understanding it right. If it's true, non-main ranks would get a dummy dataset initially and main ranks would get actual dataset.

However, if it's false, the non-main ranks would not get anything at all?

If true is the proper value for this, when would false be appropriate?

winglian:
I think the root cause is simply to remove line 137 here https://github.com/axolotl-ai-cloud/axolotl/pull/2536/files#diff-dc284316c84f597e4e7126d804cf388e9df82a9ad29bb6f56a48e255d1cf5d54R137 so we don't change the default behavior.

chiwanpark:
@NanoCode012 @winglian Thanks for the review. I've reverted the default value of `dispatch_batches` and updated this PR as suggested in the comment.

winglian (approved):

PR #2559, opened 2025-04-25
chore(doc): minor update docker tags on doc

NanoCode012:
# Description

Update the docs to include new 2.7 and cuda128 support.

### DIFF
--- a/docs/docker.qmd
+++ b/docs/docker.qmd
@@ -28,6 +28,8 @@ main-base-py{python_version}-cu{cuda_version}-{pytorch_version}
 
 Tags examples:
 
+- `main-base-py3.11-cu128-2.7.0`
+- `main-base-py3.11-cu126-2.7.0`
 - `main-base-py3.11-cu124-2.6.0`
 - `main-base-py3.11-cu124-2.5.1`
 - `main-base-py3.11-cu124-2.4.1`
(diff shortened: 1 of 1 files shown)


docs/docker.qmd
```diff
@@ -68,6 +70,7 @@ There may be some extra tags appended to the image, like `-vllm` which installs
 
 Tags examples:
 
+- `main-py3.11-cu126-2.7.0`
```
salmanmohammadi, line 73: can we also add cu128 here too?
```suggestion
- `main-py3.11-cu128-2.7.0`
- `main-py3.11-cu126-2.7.0`
```

also I'm slightly confused how these examples are different from the tags above

NanoCode012: We don't build 128 for our main/cloud image yet

salmanmohammadi (approved):

docs/docker.qmd
winglian: > also I'm slightly confused how these examples are different from the tags above

So we build the base images that's just pytorch + cuda with 12.8 support, but we don't build axolotl with 12.8 yet because we don't have CI in place yet to test those, so my preference is to not ship axolotl images with 12.8 until we have some CI for it.

winglian (approved):

PR #2565, opened 2025-04-25
add e2e smoke test for using activation/gradient checkpointing with offload

winglian:
follow-on for #2563 to have a e2e check for offload option. The CI should probably fail for this at this stage until that PR is merged.


winglian:
Also tested on runpod with manual fixes from #2563 and the new test passes

salmanmohammadi (approved):

PR #2570, opened 2025-04-27
replace references to random 68m model w 135m smollm2

winglian:


winglian:
This can supersede #2370

tests/conftest.py
```diff
@@ -79,9 +79,9 @@ def download_smollm2_135m_model():
 
 
 @pytest.fixture(scope="session", autouse=True)
-def download_llama_68m_random_model():
+def download_smollm2_135m_gptq_model():
     # download the model
-    snapshot_download_w_retry("JackFram/llama-68m", repo_type="model")
+    snapshot_download_w_retry("lilmeaty/SmolLM2-135M-Instruct-GPTQ", repo_type="model")
```
NanoCode012, line 84: I only see this hub model being used once. Should this be replaced with the non-gptq version as that's used more often given this fixture will always run too?

winglian: not really, the test is specifically for loading a model that's been converted to GPTQ

NanoCode012:
Should we switch to https://huggingface.co/HuggingFaceTB/SmolLM2-135M-Instruct as it already has `pad_token` set?

winglian:
We're already mostly using the base model across a lot of things already so let's stick with that.

PR #2572, opened 2025-04-28
support val_set_size for splitting test split from train with DPO

winglian:
report from user that val_set_size didn't work with DPO and bugs with evals. see https://github.com/huggingface/trl/pull/3377

### DIFF
--- a/src/axolotl/core/trainers/dpo/trainer.py
+++ b/src/axolotl/core/trainers/dpo/trainer.py
@@ -3,15 +3,29 @@ DPO trainer for axolotl
 """
 
 import gc
+import random
 from functools import wraps
-from typing import Any, Dict, Union
+from typing import Any, Dict, Optional, Union
 
+import pandas as pd
 import torch
+import wandb
+from accelerate import PartialState
+from datasets import Dataset, IterableDataset
 from peft.optimizers import create_loraplus_optimizer
 from torch import nn
-from transformers import Trainer
+from torch.utils.data import DataLoader
+from transformers import (
+    BaseImageProcessor,
+    FeatureExtractionMixin,
+    PreTrainedTokenizerBase,
+    ProcessorMixin,
+    Trainer,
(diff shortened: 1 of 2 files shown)


src/axolotl/utils/data/rl.py
```diff
@@ -204,7 +204,12 @@ def load_split(dataset_cfgs, _cfg):
             else:
                 eval_dataset = load_split(cfg.test_datasets, cfg)
         if not eval_dataset:
-            eval_dataset = None
+            if cfg.val_set_size:
+                ds_w_test_split = train_dataset.train_test_split(
+                    test_size=cfg.val_set_size, seed=cfg.seed
+                )
```
NanoCode012: Do we need to calculate hash like it's done in SFT?

https://github.com/axolotl-ai-cloud/axolotl/blob/8b33ae1c4f5078e0b3d2cb31bb6856692aa3d993/src/axolotl/utils/data/sft.py#L419-L439

src/axolotl/core/trainers/dpo/trainer.py
```diff
@@ -124,3 +196,67 @@ def training_step(
...
+                ],
+            )
+            if "wandb" in self.args.report_to and self.accelerator.is_main_process:
+                wandb.log({"game_log": wandb.Table(data=table)})
+
+            if "comet_ml" in self.args.report_to:
+                log_table_to_comet_experiment(
+                    name="game_log.csv",
+                    table=table,
+                )
```
NanoCode012, line 251: Given that, `wandb` requires `is_main_process`, how about comet ml? Or is this out of scope?

winglian: Out of scope for now.

winglian:
wandb: https://wandb.ai/axolotl-ai/dpo-fixes/runs/30kushnc

PR #2573, opened 2025-04-28
fix(doc): clarify vllm usage with grpo

NanoCode012:
# Description

Just minor update as the doc was a bit confusing.

We should though update the cookbook as that used the old method. Should we point to Salman's GRPO with WASM instead?

### DIFF
--- a/docs/rlhf.qmd
+++ b/docs/rlhf.qmd
@@ -538,7 +538,7 @@ CUDA_VISIBLE_DEVICES=0,1 axolotl train grpo.yaml --num-processes 2
 ```
 
 ::: {.callout-note}
-Due to trl's implementation with vllm, the vllm instance must use the last N GPUs instead of the first N GPUs. This is why in the example above, we use `CUDA_VISIBLE_DEVICES=2,3` for the vllm instance.
+Due to trl's implementation with vLLM, the vLLM instance must use the last N GPUs instead of the first N GPUs. This is why in the example above, we use `CUDA_VISIBLE_DEVICES=2,3` for the vLLM instance.
 :::
 
 #### Reward functions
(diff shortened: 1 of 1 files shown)


docs/rlhf.qmd
```diff
@@ -539,6 +537,10 @@ Your `vLLM` instance will now attempt to spin up, and it's time to kick off trai
 CUDA_VISIBLE_DEVICES=0,1 axolotl train grpo.yaml --num-processes 2
 ```
 
+::: {.callout-note}
+Due to trl's implementation with vllm, the vllm instance must use the last N GPUs instead of the first N GPUs. This is why in the example above, we use `CUDA_VISIBLE_DEVICES=2,3` for the vllm instance.
+:::
```
salmanmohammadi:
```suggestion
Due to trl's implementation with vLLM, the vLLM instance must use the last N GPUs instead of the first N GPUs. This is why in the example above, we use `CUDA_VISIBLE_DEVICES=2,3` for the vLLM instance.
:::
```
nit

salmanmohammadi (approved):

docs/rlhf.qmd
```diff
@@ -539,6 +537,10 @@ Your `vLLM` instance will now attempt to spin up, and it's time to kick off trai
 CUDA_VISIBLE_DEVICES=0,1 axolotl train grpo.yaml --num-processes 2
 ```
 
+::: {.callout-note}
+Due to trl's implementation with vLLM, the vLLM instance must use the last N GPUs instead of the first N GPUs. This is why in the example above, we use `CUDA_VISIBLE_DEVICES=2,3` for the vLLM instance.
```
salmanmohammadi:
```suggestion
Due to TRL's implementation with vLLM, the vLLM instance must use the last N GPUs instead of the first N GPUs. This is why in the example above, we use `CUDA_VISIBLE_DEVICES=2,3` for the vLLM instance.
```

Issue #2585, opened 2025-04-29
Not able to run evaluate


locovamos:
### Please check that this issue hasn't been reported before.

- [x] I searched previous [Bug Reports](https://github.com/axolotl-ai-cloud/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

I need to evaluate my finetuned model by using evaluate command.

### Current behaviour

I'm getting always the following when executing the evaluate:

Traceback (most recent call last):
  File "/home/axolotl-test/axolotl/bin/axolotl", line 8, in <module>
    sys.exit(main())
  File "/home/axolotl-test/axolotl/lib/python3.10/site-packages/axolotl/cli/main.py", line 337, in main
    cli()
  File "/home/axolotl-test/axolotl/lib/python3.10/site-packages/click/core.py", line 1161, in __call__
    return self.main(*args, **kwargs)
  File "/home/axolotl-test/axolotl/lib/python3.10/site-packages/click/core.py", line 1082, in main
    rv = self.invoke(ctx)
  File "/home/axolotl-test/axolotl/lib/python3.10/site-packages/click/core.py", line 1697, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/axolotl-test/axolotl/lib/python3.10/site-packages/click/core.py", line 1443, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/axolotl-test/axolotl/lib/python3.10/site-packages/click/core.py", line 788, in invoke
    return __callback(*args, **kwargs)
  File "/home/axolotl-test/axolotl/lib/python3.10/site-packages/axolotl/cli/utils.py", line 66, in wrapper
    return func(*args, **filtered_kwargs)
  File "/home/axolotl-test/axolotl/lib/python3.10/site-packages/axolotl/cli/main.py", line 212, in evaluate
    do_cli(config=config, **kwargs)
  File "/home/axolotl-test/axolotl/lib/python3.10/site-packages/axolotl/cli/evaluate.py", line 59, in do_cli
    do_evaluate(parsed_cfg, parsed_cli_args)
  File "/home/axolotl-test/axolotl/lib/python3.10/site-packages/axolotl/cli/evaluate.py", line 42, in do_evaluate
    evaluate(cfg=cfg, dataset_meta=dataset_meta)
  File "/home/axolotl-test/axolotl/lib/python3.10/site-packages/axolotl/evaluate.py", line 104, in evaluate
[...]

PR #2586, opened 2025-04-29
Fix eval + add smoke test

djsaunde:
# Description

Title.

## Motivation and Context

#2585.

## How has this been tested?

Manually and with smoke test.


src/axolotl/cli/evaluate.py
```diff
@@ -29,10 +31,14 @@ def do_evaluate(cfg: DictDefault, cli_args: TrainerCliArgs) -> None:
         cfg: Dictionary mapping `axolotl` config keys to values.
         cli_args: CLI arguments.
     """
+    # Enable expandable segments for cuda allocation to improve VRAM usage
+    set_pytorch_cuda_alloc_conf()
```
salmanmohammadi, line 35: Is there a way to have this called whenever the CLI is invoked, so we don't have to set it for every command?

winglian (approved):

salmanmohammadi (approved):

src/axolotl/cli/evaluate.py
djsaunde: Yes, but not everything is run through the CLI. Some folks still use the `python -m axolotl.cli.train` approach, I think. Maybe we should deprecate that, though that would give users less flexibility for what launcher they use.

If we went this route, the code change could be as simple as:

```python
@click.group()
@click.version_option(version=axolotl.__version__, prog_name="axolotl")
def cli():
    """Axolotl CLI - Train and fine-tune large language models"""
    # Enable expandable segments for cuda allocation to improve VRAM usage
    set_pytorch_cuda_alloc_conf()
```

PR #2588, opened 2025-04-29
support for qwen3 with lora kernels

winglian:


djsaunde (approved):
Awesome stuff!

I wonder which other common models we should support.

djsaunde:
I have a script for testing all the model types in `transformers`, we could apply it to find model types to support: https://gist.github.com/djsaunde/7f53602746f6c9a810ddd6b6393d27cb.

PR #2589, opened 2025-04-29
auto-enable lora kernels where possible

djsaunde:
# Description

Title.

## Motivation and Context

Let's be opinionated here and get speed / VRAM improvements for users that are unaware of these options.

## How has this been tested?

TODO.

### DIFF
--- a/src/axolotl/utils/schemas/config.py
+++ b/src/axolotl/utils/schemas/config.py
@@ -1318,7 +1318,7 @@ class AxolotlConfigWCapabilities(AxolotlInputConfig):
 
     @model_validator(mode="before")
     @classmethod
-    def auto_enable_lora_kernels(cls, data):
+    def check_auto_enable_lora_kernels(cls, data):
         # Only proceed if using LoRA or QLoRA adapter
         if data.get("adapter") in ["lora", "qlora"]:
             # Skip if already set, using unsloth optimizations, or using 8-bit
--- a/src/axolotl/utils/schemas/config.py
+++ b/src/axolotl/utils/schemas/config.py
@@ -1049,6 +1049,7 @@ class AxolotlInputConfig(
             or data.get("lora_qkv_kernel")
             or data.get("lora_o_kernel")
         ):
+            print(data.get("lora_mlp_kernel"))
             if data.get("adapter") == "lora" and data.get("load_in_8bit"):
                 raise ValueError(
                     "lora_mlp_kernel, lora_mlp_kernel, and lora_mlp_kernel are not compatible with 8-bit LoRA"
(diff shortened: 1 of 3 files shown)


src/axolotl/utils/schemas/config.py
```diff
@@ -1315,6 +1315,54 @@ def check_multigpu_lora_kernels(cls, data):
...
+
+                if data.get("lora_qkv_kernel") is None:
+                    data["lora_qkv_kernel"] = True
+
+                if data.get("lora_o_kernel") is None:
+                    data["lora_o_kernel"] = True
+
+                LOG.warning(
+                    "Auto-enabling LoRA kernel optimizations for faster training. "
+                    + "Please explicitly set `lora_*_kernel` config values to `False` to disable. "
```
winglian:
```suggestion
                    + "Please explicitly set `lora_*_kernel` config values to `false` to disable. "
```

yaml false vs python False

src/axolotl/utils/schemas/config.py
```diff
@@ -1049,6 +1049,7 @@ def check_lora_8bit(cls, data):
             or data.get("lora_qkv_kernel")
             or data.get("lora_o_kernel")
         ):
+            print(data.get("lora_mlp_kernel"))
```
winglian: stray debugging

src/axolotl/utils/schemas/config.py
```diff
@@ -1315,6 +1316,54 @@ def check_multigpu_lora_kernels(cls, data):
...
+    @model_validator(mode="before")
+    @classmethod
+    def check_auto_enable_lora_kernels(cls, data):
+        # Only proceed if using LoRA or QLoRA adapter
+        if data.get("adapter") in ["lora", "qlora"]:
+            # Skip if already set, using unsloth optimizations, or using 8-bit
+            unsloth_fields = ["unsloth_lora_mlp", "unsloth_lora_qkv", "unsloth_lora_o"]
+            kernel_fields = ["lora_mlp_kernel", "lora_qkv_kernel", "lora_o_kernel"]
+            if (
+                all(data.get(k) is not None for k in kernel_fields)
```
winglian: Not a blocker, but should this be `any(...)`? If they made the choice to enable even one, they've probably considered they didn't want the others perhaps?

winglian (approved):
good to go once the stray `print` is statement is removed.

src/axolotl/utils/schemas/config.py
djsaunde: Good point. I think I could go either way, but your idea seems slightly more user friendly. I'll make this change.

PR #2591, opened 2025-04-29
Handle other reasoning trace dataset formats

winglian:
Some datasets have different thinking delimiters as well as delimiters around the non-thinking content. Let's split that and handle it for passing to the chat template for reasoning_content


djsaunde (approved):
Code changes seem reasonable, though I admit this is out of my wheelhouse a bit.

NanoCode012 (approved):

PR #2592, opened 2025-04-29
replace zero_only with simpler if statement

winglian:
my understanding of context managers is incorrect, and it would still run the code on non-zero ranks. Replacing with simpler if statement, and using simpler environ check to prevent un-necessarily doing a init distributed.

### DIFF
--- a/src/axolotl/integrations/cut_cross_entropy/__init__.py
+++ b/src/axolotl/integrations/cut_cross_entropy/__init__.py
@@ -25,7 +25,7 @@ import torch
 
 from axolotl.integrations.base import BasePlugin
 from axolotl.utils import get_pytorch_version
-from axolotl.utils.distributed import zero_only
+from axolotl.utils.distributed import is_main_process
 
 from .args import CutCrossEntropyArgs  # pylint: disable=unused-import. # noqa: F401
 
@@ -76,7 +76,7 @@ class CutCrossEntropyPlugin(BasePlugin):
                 cce_patch,
             )
 
-            with zero_only():
+            if is_main_process(use_distributed=False):
                 LOG.info(
                     f"Applying Cut Cross Entropy to model type: {cfg.model_config_type}"
                 )
--- a/src/axolotl/integrations/cut_cross_entropy/__init__.py
+++ b/src/axolotl/integrations/cut_cross_entropy/__init__.py
@@ -76,7 +76,7 @@ class CutCrossEntropyPlugin(BasePlugin):
                 cce_patch,
             )
 
-            if is_main_process(use_distributed=False):
+            if is_main_process(use_environ=True):
                 LOG.info(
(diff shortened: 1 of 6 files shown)


src/axolotl/utils/distributed.py
```diff
@@ -99,17 +103,6 @@ def cleanup_distributed():
...
-@contextmanager
-def zero_only():
-    """
-    Context manager that only runs the enclosed block on the main rank.
-    """
-    if is_main_process():
-        yield
-    else:
-        yield None
-
```
NanoCode012, line 111: That's a good catch. I mistakenly thought it would only run on main too, if yield is not None.

src/axolotl/utils/distributed.py
```diff
@@ -69,17 +69,21 @@ def barrier():
         dist.barrier()
 
 
-def is_main_process():
+def is_main_process(use_distributed=True):
     """
     Check if the current process is the main process. If not in distributed mode,
     always return `True`.
```
NanoCode012, line 75: This comment should be adjusted that for new params and avoid conflict name between `use_distributed` and `is_distributed`

NanoCode012: I'm a bit confused what's difference myself here.

winglian: renamed.

src/axolotl/utils/distributed.py
```diff
@@ -69,17 +69,27 @@ def barrier():
...
+
+    Args:
+    - use_environ (bool, optional): Use environment variable to determine main process.
+
+    Returns:
+    - bool: `True` if the current process is the main process, `False` otherwise.
     """
+    if use_environ:
+        return os.environ.get("LOCAL_RANK", "0") == "0"
```
djsaunde, line 84: Nice, this allows us to use this function before init-ing deepspeed, right?

djsaunde (approved):

salmanmohammadi (approved):

src/axolotl/integrations/cut_cross_entropy/__init__.py
```diff
@@ -76,7 +76,7 @@ def pre_model_load(self, cfg):
                 cce_patch,
             )
 
-            with zero_only():
+            if is_main_process(use_environ=True):
                 LOG.info(
```
salmanmohammadi, line 80: Not for this PR but we might want to consider adding a `log_rank_zero` function

src/axolotl/utils/distributed.py
winglian: exactly

src/axolotl/integrations/cut_cross_entropy/__init__.py
winglian: agreed. maybe something similar to how accelerate does it https://github.com/huggingface/accelerate/blob/v1.3.0/src/accelerate/logging.py#L22-L37

PR #2593, opened 2025-04-29
upload the deepspeed json to wandb

winglian:
artifact: https://wandb.ai/axolotl-ai/qwen3-distributed/runs/0stfrsk4/files/tmp/deepspeed_config_1vfn1i35.json

### DIFF
--- a/src/axolotl/utils/callbacks/__init__.py
+++ b/src/axolotl/utils/callbacks/__init__.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import gc
+import json
 import logging
 import os
 import traceback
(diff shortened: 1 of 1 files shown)


src/axolotl/utils/callbacks/__init__.py
```diff
@@ -808,11 +809,44 @@ def on_train_begin(
...
+                        suffix=".json",
+                        prefix="deepspeed_config_",
+                    ) as temp_file:
+                        skip_upload = False
+                        if isinstance(args.deepspeed, dict):
+                            json.dump(args.deepspeed, temp_file, indent=4)
+                        elif isinstance(args.deepspeed, str) and os.path.exists(
+                            args.deepspeed
+                        ):
+                            copyfile(args.deepspeed, temp_file.name)
```
NanoCode012, line 833: I noticed this being referenced as `temp_file` for name in json dump but `temp_file.name` elsewhere. Does this have to be adjusted to `temp_file.name`?

winglian: in json.dump, that uses a a file reference, where copyfile simply needs the file names

winglian: see docs for[ json.dump](https://docs.python.org/3/library/json.html#json.dumps) where the arg is a file-pointer

NanoCode012: Thanks for clarifying!

NanoCode012 (approved):

PR #2595, opened 2025-04-30
patch to convert LR from tensor to float when using DS

winglian:
# Description

Saving checkpoints when using DS is broken atm. See https://github.com/huggingface/transformers/pull/37881

### DIFF
--- a/src/axolotl/monkeypatch/trainer/lr.py
+++ b/src/axolotl/monkeypatch/trainer/lr.py
@@ -1,0 +1,42 @@
+"""
+monkeypatch for Trainer _get_learning_rate method
+"""
+
+import logging
+
+import torch
+
+LOG = logging.getLogger(__name__)
+
+
+# TODO remove this patch once https://github.com/huggingface/transformers/pull/37881 is included in a release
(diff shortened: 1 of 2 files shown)


NanoCode012 (approved):

src/axolotl/monkeypatch/trainer/lr.py
```diff
@@ -0,0 +1,42 @@
...
+                last_lr = 0
+            else:
+                raise
+    else:
+        if isinstance(self.lr_scheduler, torch.optim.lr_scheduler.ReduceLROnPlateau):
+            last_lr = self.optimizer.param_groups[0]["lr"]
+        else:
+            last_lr = self.lr_scheduler.get_last_lr()[0]
+
+    if torch.is_tensor(last_lr):
```
NanoCode012, line 34: Do we need to check if var exists?

```suggestion
    if last_lr and torch.is_tensor(last_lr):
```

NanoCode012: Actually, shouldn't be required as the other cases would resolve to a value

PR #2600, opened 2025-04-30
don't automatically enable lora kernels for RL training

winglian:
#2589 didn't trigger the multigpu CI and is causing failures with the RL trainers + PEFT. Let's disable for now on RL since it's not as battle tested.  see recent failures here https://github.com/axolotl-ai-cloud/axolotl/actions/runs/14755394008/job/41422234427?pr=2592

### DIFF
--- a/src/axolotl/utils/schemas/config.py
+++ b/src/axolotl/utils/schemas/config.py
@@ -1319,6 +1319,9 @@ class AxolotlConfigWCapabilities(AxolotlInputConfig):
     @classmethod
     def check_auto_enable_lora_kernels(cls, data):
         # Only proceed if using LoRA or QLoRA adapter
+        if data.get("rl"):
+            # RL trainers not tested so don't enable kernels by default
+            return data
         if data.get("adapter") in ["lora", "qlora"]:
             # Skip if already set, using unsloth optimizations, or using 8-bit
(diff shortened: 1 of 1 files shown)


winglian:
manually triggered multigpu ci for this PR here: https://github.com/axolotl-ai-cloud/axolotl/actions/runs/14756663042

src/axolotl/utils/schemas/config.py
```diff
@@ -1319,6 +1319,9 @@ def check_multigpu_lora_kernels(cls, data):
     @classmethod
     def check_auto_enable_lora_kernels(cls, data):
         # Only proceed if using LoRA or QLoRA adapter
+        if data.get("rl"):
+            # RL trainers not tested so don't enable kernels by default
```
salmanmohammadi, line 1323: Can we log a warning to indicate that the kernel is being disabled and a link to this PR, rather than silently overriding?

salmanmohammadi (approved):

src/axolotl/utils/schemas/config.py
winglian: It's not really disabling the kernel, it's just not auto-enabling it if the user didn't specify it

salmanmohammadi: Oh I missed that. Carry on : )

Issue #2601, opened 2025-04-30
vLLM incompatibility


dhruvmullick:
### Please check that this issue hasn't been reported before.

- [x] I searched previous [Bug Reports](https://github.com/axolotl-ai-cloud/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

I should be able to launch a vllm server, but since the 0.9.0 upgrade, that's erroring out. 

The associated vLLM version (0.7.3) doesn't support a ```worker_extension_cls``` keyword argument during initialization. 

The latest vLLM code does have this parameter https://github.com/vllm-project/vllm/blob/main/vllm/engine/arg_utils.py#L353

### Current behaviour

Getting the error: 

```
[2025-04-30 16:34:57,684] [INFO] [real_accelerator.py:219:get_accelerator] Setting ds_accelerator to cuda (auto detect)
[2025-04-30 16:35:00,099] [INFO] [datasets.<module>:54] [PID:2567] PyTorch version 2.5.1+cu124 available.
/opt/conda/lib/python3.11/site-packages/axolotl/monkeypatch/relora.py:17: DeprecationWarning: `TorchScript` support for functional optimizers is deprecated and will be removed in a future PyTorch release. Consider using the `torch.compile` optimizer instead.
  from torch.distributed.optim import ZeroRedundancyOptimizer
[2025-04-30 16:35:01,264] [INFO] [root.register:307] [PID:2567] Attempting to load plugin: axolotl.integrations.liger.LigerPlugin
[2025-04-30 16:35:01,265] [INFO] [root.register:310] [PID:2567] Plugin loaded successfully: axolotl.integrations.liger.LigerPlugin
[2025-04-30 16:35:01,329] [WARNING] [axolotl.utils.schemas.config.hint_lora_8bit:847] [PID:2567] [RANK:0] We recommend setting `load_in_8bit: true` for LORA finetuning
[2025-04-30 16:35:02,277] [DEBUG] [axolotl.resolve_dtype:65] [PID:2567] [RANK:0] bf16 support detected, enabling for this configuration.
[2025-04-30 16:35:02,388] [INFO] [axolotl.normalize_config:234] [PID:2567] [RANK:0] cuda memory usage baseline: 0.000GB (+0.551GB misc)
Process Process-1:
Traceback (most recent call last):
  File "/opt/conda/lib/python3.11/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
[...]

PR #2605, opened 2025-05-01
Fix: pin same vllm version for all torch

NanoCode012:
# Description

VLLM should support torch2.5-2.7 equally. Prev, torch2.5 would install 0.7.4 which would break GRPO as seen in #2601

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ Features:
 
 - NVIDIA GPU (Ampere or newer for `bf16` and Flash Attention) or AMD GPU
 - Python 3.11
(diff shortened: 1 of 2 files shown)


winglian:
If we want to support only 0.8.5, then we need to remove vllm support from every python version except 2.6.0 since technically it's not supported on those versions.

salmanmohammadi:
> If we want to support only 0.8.5, then we need to remove vllm support from every python version except 2.6.0 since technically it's not supported on those versions.

Unless there's other conflicts my vote is to only support vllm for 2.6 torch

PR #2606, opened 2025-05-01
fix: run preview-docs only when md/qmd changes

NanoCode012:
# Description

The preview docs CI should only run if the markdown or qmd file changes.

### DIFF
--- a/.github/workflows/preview-docs.yml
+++ b/.github/workflows/preview-docs.yml
@@ -4,6 +4,11 @@ on:
   pull_request:
     types: [opened, synchronize, reopened]
 
+    # Run the workflow only when one of these files changes
+    paths:
+      - '**/*.md'      # any Markdown file
+      - '**/*.qmd'     # any Quarto file
+
 permissions:
   checks: write
   contents: write
--- a/.github/workflows/preview-docs.yml
+++ b/.github/workflows/preview-docs.yml
@@ -8,6 +8,7 @@ on:
     paths:
       - '**/*.md'      # any Markdown file
       - '**/*.qmd'     # any Quarto file
+      - '_quarto.yaml'
 
 permissions:
   checks: write


.github/workflows/preview-docs.yml
```diff
@@ -4,6 +4,11 @@ on:
   pull_request:
     types: [opened, synchronize, reopened]
 
+    # Run the workflow only when one of these files changes
+    paths:
+      - '**/*.md'      # any Markdown file
```
winglian, line 9: Does this cover the base README.md too?

.github/workflows/preview-docs.yml
```diff
@@ -4,6 +4,11 @@ on:
   pull_request:
     types: [opened, synchronize, reopened]
 
+    # Run the workflow only when one of these files changes
+    paths:
```
djsaunde, line 8: Can we add `_quarto.yaml` here as well? If we make quarto config changes or add a section to the `quartodoc` autodoc bit we'll want to see the changes in preview.

.github/workflows/preview-docs.yml
djsaunde: Yup, from the root dir:

```shell
✗ ls **/*.md
FAQS.md
README.md
TODO.md
...
```

djsaunde (approved):
Just one requested change; otherwise should be good to go.

winglian:
Actually, don't we have Autodoc from code annotations? In that case we can't limit to markdown/quarto. @djsaunde

djsaunde:
@winglian Yeah, that's true, although I think we could just not preview for docstring changes, seems a bit minimal? I could go either way on this.

.github/workflows/preview-docs.yml
NanoCode012: Added thanks!

NanoCode012:
> Actually, don't we have Autodoc from code annotations? In that case we can't limit to markdown/quarto.

I think the main use case for the preview was to mainly see doc changes. So, I think we can ignore these for now too?

salmanmohammadi:
> > Actually, don't we have Autodoc from code annotations? In that case we can't limit to markdown/quarto.
> 
> I think the main use case for the preview was to mainly see doc changes. So, I think we can ignore these for now too?

Yeah once we have more % coverage of docstrings I would say we could start linting/previewing for it

winglian (approved):

winglian:
> > > Actually, don't we have Autodoc from code annotations? In that case we can't limit to markdown/quarto.
> > 
> > 
> > I think the main use case for the preview was to mainly see doc changes. So, I think we can ignore these for now too?
> 
> Yeah once we have more % coverage of docstrings I would say we could start linting/previewing for it

sgtm. thanks for the feedback!

PR #2607, opened 2025-05-01
automatically set pad_to_sequence_len when use packing

winglian:
Let's be opinionated and set `pad_to_sequence_len` to True if not explicitly set when sample packing is enabled. This reduces the number of options the end user needs to set, but can be disabled still if they don't want it.

### DIFF
--- a/src/axolotl/utils/schemas/config.py
+++ b/src/axolotl/utils/schemas/config.py
@@ -512,10 +512,17 @@ class AxolotlInputConfig(
     @model_validator(mode="before")
     @classmethod
     def hint_sample_packing_padding(cls, data):
-        if data.get("sample_packing") and not data.get("pad_to_sequence_len"):
-            LOG.warning(
-                "`pad_to_sequence_len: true` is recommended when using sample_packing"
-            )
+        if data.get("sample_packing"):
+            pad_to_sequence_len = data.get("pad_to_sequence_len")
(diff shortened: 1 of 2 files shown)


djsaunde:
stupid question: what does this actually do? why is it good to do? might be good to update docs with this.

winglian:
So what happens when you don't pad is you get tensors of different lengths from batch to batch, and during training there seems to be a memory leak in pytorch due to having to allocate different sizes on each step

djsaunde:
> So what happens when you don't pad is you get tensors of different lengths from batch to batch, and during training there seems to be a memory leak in pytorch due to having to allocate different sizes on each step

Hm, I would think enabling sample packing would pack to the passed `sequence_len`, I need to review the logic.

salmanmohammadi:
> > So what happens when you don't pad is you get tensors of different lengths from batch to batch, and during training there seems to be a memory leak in pytorch due to having to allocate different sizes on each step
> 
> Hm, I would think enabling sample packing would pack to the passed `sequence_len`, I need to review the logic.

Yeah I'm thinking along the same lines. I have two questions here:
- Do we split samples across packs if a sample exceeds the maximum sequence length?
- What is the interaction between `pad_to_seq_len` and sample packing?

NanoCode012 (approved):

PR #2611, opened 2025-05-01
Logging config for colab

winglian:
configure_logging doesn't play nicely with Google Colab, and seems to mostly choke when called more than once. Moving this to `axolotl.cli` seems like the most sane code patch to call it once. 

This pr also allows for passing a DictDefault to load_config so we can easily set the params in colab, and then reuse the load_cfg to revalidate it

### DIFF
--- a/src/axolotl/cli/__init__.py
+++ b/src/axolotl/cli/__init__.py
@@ -2,4 +2,7 @@
 
 import os
 
+from axolotl.logging_config import configure_logging
+
 os.environ["HF_HUB_ENABLE_HF_TRANSFER"] = "1"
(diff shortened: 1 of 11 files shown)


djsaunde (approved):
Cool, a nice follow-up would to add logging config options and parse them in the CLI itself (e.g., log level, which is currently hardcoded to debug).

salmanmohammadi (approved):

salmanmohammadi:
> Cool, a nice follow-up would to add logging config options and parse them in the CLI itself (e.g., log level, which is currently hardcoded to debug).

++1

PR #2612, opened 2025-05-01
qwen3 and qwen3_moe support for liger kernels

winglian:


NanoCode012:
Qwen3MLP does not use `intermediate_size`, so I defaulted it back to normal LigerMLP.

I also fixed Qwen3MoE returning causal output instead of moe causal output.

winglian:
thanks @NanoCode012 , wandb parity tests here https://wandb.ai/axolotl-ai/qwen3-testing-liger/overview
Screenshot 2025-05-02 at 9 24 39 AM

PR #2614, opened 2025-05-02
Adds example for training a TTS model on top of a LLM.

mhenrichsen:
# Description

Adds an example of training a LLM to output audio.

## Motivation and Context
Just a pretty cool concept that is now trainable with axolotl.

## How has this been tested?
Trained a model. Check wandb: 
https://wandb.ai/mhenrichsen/orph/runs/qh5ijvoq?nw=nwusermhenrichsen


examples/orpheus/finetune.yml
```diff
@@ -0,0 +1,52 @@
...
+    type: 
+dataset_prepared_path: last_run_prepared
+val_set_size: 0.01
+output_dir: ./outputs/out
+
+sequence_len: 8192
+sample_packing: true
+pad_to_sequence_len: true
+
+wandb_project: orph
```
NanoCode012:
```suggestion
wandb_project:
```

examples/orpheus/finetune.yml
```diff
@@ -0,0 +1,52 @@
...
+plugins:
+  - axolotl.integrations.liger.LigerPlugin
+liger_rope: true
+liger_rms_norm: true
+liger_glu_activation: true
+liger_fused_linear_cross_entropy: true
+
+datasets:
+  - path: <your-hf-dataset-id>
+    type:
```
NanoCode012: Should we leave a comment here, so user's don't accidentally think they need to put sth for `type: `?

```suggestion
    type:  # leave empty to load pre-tokenized
```

examples/orpheus/finetune.yml
```diff
@@ -0,0 +1,52 @@
...
+logging_steps: 1
+flash_attention: true
+
+warmup_steps: 20
+evals_per_epoch: 5
+saves_per_epoch: 5
+weight_decay: 0.05
+
+special_tokens:
+  pad_token: <custom_token_7>
```
NanoCode012: Do we need to set pad_token? If yes, is this the one that's a good default?

examples/orpheus/README.md
```diff
@@ -0,0 +1,340 @@
+# Finetuning LLMs to output audio
+
+In this example, we finetune Orpcanopylabs/orpheus-tts-0.1-pretrained (a LLaMA 3.2 3b model) to output audio.
```
NanoCode012, line 3: Do you have the hardware requirements to run this config?

NanoCode012 (commented):
I wasn't aware this was even possible! Thanks for making this PR.

examples/orpheus/finetune.yml
mhenrichsen: Yes. The original orpheus model is trained with the same pad token, so we have to match.

examples/orpheus/README.md
mhenrichsen: Updated.

NanoCode012: Thanks!

examples/orpheus/finetune.yml
```diff
@@ -0,0 +1,52 @@
+base_model: canopylabs/orpheus-tts-0.1-pretrained
```
NanoCode012: I may be doing sth wrong, but I could not find this model?

mhenrichsen: https://huggingface.co/canopylabs/orpheus-3b-0.1-pretrained

Seems like they have changed the name since I trained.

Will fix.

NanoCode012 (approved):

examples/orpheus/finetune.yml
NanoCode012: Thanks!

PR #2615, opened 2025-05-02
fix(trl): remove access to invalid property

NanoCode012:
# Description

Seen this a while back. TRL updated their `tokenize_row` to return `input_ids` only, no more labels/attention_mask. https://github.com/huggingface/trl/blob/fc2b041b58f6fbe766dceaec819bc5a8f9d209da/trl/trainer/dpo_trainer.py#L631-L635

Reported at: https://discord.com/channels/1104757954588196865/1117071926926512248/1367778621145612370

## How has this been tested?

Untested!


winglian:
Duplicate of 2616

PR #2617, opened 2025-05-02
allow plugins to return their own dataset

winglian:
# Description

### DIFF
--- a/src/axolotl/cli/preprocess.py
+++ b/src/axolotl/cli/preprocess.py
@@ -18,6 +18,7 @@ from axolotl.cli.checks import check_accelerate_default_config, check_user_token
 from axolotl.cli.config import load_cfg
 from axolotl.common.const import DEFAULT_DATASET_PREPARED_PATH
 from axolotl.common.datasets import load_datasets, load_preference_datasets
+from axolotl.integrations.base import PluginManager
 from axolotl.utils.dict import DictDefault
 from axolotl.utils.trainer import disable_datasets_caching
 
@@ -47,7 +48,10 @@ def do_preprocess(cfg: DictDefault, cli_args: PreprocessCliArgs) -> None:
         cfg.dataset_prepared_path = DEFAULT_DATASET_PREPARED_PATH
 
     with disable_datasets_caching():
-        if cfg.rl:
+        plugin_manager = PluginManager.get_instance()
+        if plugin_manager.load_datasets(cfg, preprocess=True):
+            pass
+        elif cfg.rl:
             load_preference_datasets(cfg=cfg, cli_args=cli_args)
         else:
             load_datasets(cfg=cfg, cli_args=cli_args)
--- a/src/axolotl/cli/train.py
+++ b/src/axolotl/cli/train.py
@@ -44,12 +44,12 @@ def do_train(cfg: DictDefault, cli_args: TrainerCliArgs):
         check_user_token()
 
     plugin_manager = PluginManager.get_instance()
-    if dataset_meta := plugin_manager.load_datasets(cfg, preprocess=False):
-        pass
-    elif cfg.rl:
-        dataset_meta = load_preference_datasets(cfg=cfg, cli_args=cli_args)
-    else:
-        dataset_meta = load_datasets(cfg=cfg, cli_args=cli_args)
+    dataset_meta = plugin_manager.load_datasets(cfg, preprocess=False)
+    if not dataset_meta:
+        if cfg.rl:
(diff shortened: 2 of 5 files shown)


src/axolotl/cli/preprocess.py
```diff
@@ -47,7 +48,10 @@ def do_preprocess(cfg: DictDefault, cli_args: PreprocessCliArgs) -> None:
         cfg.dataset_prepared_path = DEFAULT_DATASET_PREPARED_PATH
 
     with disable_datasets_caching():
-        if cfg.rl:
+        plugin_manager = PluginManager.get_instance()
+        if plugin_manager.load_datasets(cfg, preprocess=True):
```
djsaunde, line 52: This seems to assume that the plugin's `load_datasets` function saves the preprocessed dataset. Should we warn something to this effect?

src/axolotl/cli/train.py
```diff
@@ -43,7 +43,10 @@ def do_train(cfg: DictDefault, cli_args: TrainerCliArgs):
...
         check_user_token()
 
-    if cfg.rl:
+    plugin_manager = PluginManager.get_instance()
+    if dataset_meta := plugin_manager.load_datasets(cfg, preprocess=False):
+        pass
+    elif cfg.rl:
         dataset_meta = load_preference_datasets(cfg=cfg, cli_args=cli_args)
     else:
         dataset_meta = load_datasets(cfg=cfg, cli_args=cli_args)
```
djsaunde: Maybe slightly cleaner?

```suggestion
    plugin_manager = PluginManager.get_instance()
    dataset_meta = plugin_manager.load_datasets(cfg, preprocess=False)
    if not dataset_meta:
      if cfg.rl:
          dataset_meta = load_preference_datasets(cfg=cfg, cli_args=cli_args)
      else:
          dataset_meta = load_datasets(cfg=cfg, cli_args=cli_args)
```

src/axolotl/integrations/base.py
```diff
@@ -68,6 +70,18 @@ def get_input_args(self):
         Returns a pydantic model for the plugin's input arguments.
         """
 
+    def load_datasets(self, cfg, preprocess: bool = False):
```
djsaunde: nit

```suggestion
    def load_datasets(self, cfg, preprocess: bool = False) -> TrainDatasetMeta | None:
```

src/axolotl/integrations/base.py
```diff
@@ -68,6 +70,18 @@ def get_input_args(self):
...
+        """
+        Loads and preprocesses the dataset for training.
+
+        Parameters:
+        cfg (dict): The configuration for the plugin.
+        preprocess (bool): Whether this is the preprocess step of the datasets.
+
+        Returns:
+        dataset_meta (TrainDatasetMeta | None): The metadata for the training dataset.
+        """
```
djsaunde: nit

```suggestion
        """
        Loads and preprocesses the dataset for training.

        Args:
            cfg: The configuration for the plugin.
            preprocess: Whether this is the preprocess step of the datasets.

        Returns:
            dataset_meta: The metadata for the training dataset.
        """
```

djsaunde: similarly for other methods in this file.

src/axolotl/integrations/base.py
```diff
@@ -338,6 +364,23 @@ def get_input_args(self):
...
+        cfg (dict): The configuration for the plugins.
+        preprocess (bool): Whether this is preprocess step of the datasets.
+
+        Returns:
+        dataset_meta (TrainDatasetMeta | None): The dataset metadata loaded from all registered plugins.
+        """
+        for plugin in self.plugins.values():
+            dataset_meta = plugin.load_datasets(cfg, preprocess)
+            if dataset_meta is not None:
+                return dataset_meta
```
djsaunde: What if multiple plugins load datasets? Probably should assert here about uniqueness.

djsaunde (approved):
Just some nits / minor comments; otherwise good to go.

src/axolotl/cli/preprocess.py
winglian: Probably should just document that the plugin is responsible for the entire lifecycle? For the downstream case in mind of an iterable dataset from an API, there would be no saved cached dataset

src/axolotl/integrations/base.py
winglian: So can't actually import `TrainDatasetMeta` in this as it causes a circular import

PR #2618, opened 2025-05-02
[WIP] feat: add heartbeat endpoint

AlpinDale:
This PR adds a heartbeat monitoring endpoint to axolotl.

# Description

When the plugin is enabled, axolotl will setup a heartbeat server on a localhost port. Users can then make use of that to check for the status of their run -- whether it's running, or crashed/finished. Makes setting up alerts easier too.

## Motivation and Context

Sometimes runs crash or finish overnight, and we're left with idle GPUs.

## How has this been tested?
Still WIP.


winglian:
This is great Alpin! One other fun thing might be an endpoint that triggers a checkpoint save on the end of the next step as well as terminating the run (although may want rudimentary security for that)

PR #2619, opened 2025-05-03
xformers attention with packing

winglian:
# Description
This PR adds packing support when using xformers attention for training. xformers supports fp16 attention, allowing us to train on T4 devices. We have to patch over FA2, as there are some attention checks that handle the attention masking that aren't easily overridable as we would have to do it for every model arch.

### DIFF
--- a/src/axolotl/monkeypatch/attention/__init__.py
+++ b/src/axolotl/monkeypatch/attention/__init__.py
@@ -1,0 +1,19 @@
+"""
+attention module for attention monkeypatches
+"""
+
+from transformers.integrations.flash_attention import flash_attention_forward
+
+
+def patch_xformers_attn_over_fa2():
+    from transformers.modeling_utils import ALL_ATTENTION_FUNCTIONS
+
+    from .xformers import xformers_attention_forward
+
+    ALL_ATTENTION_FUNCTIONS["flash_attention_2"] = xformers_attention_forward
+
+
+def unpatch_xformers_attn_over_fa2():
+    from transformers.modeling_utils import ALL_ATTENTION_FUNCTIONS
+
+    ALL_ATTENTION_FUNCTIONS["flash_attention_2"] = flash_attention_forward()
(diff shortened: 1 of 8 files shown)


src/axolotl/utils/callbacks/__init__.py
```diff
@@ -868,3 +868,29 @@ def on_epoch_end(
...
+            self, args, state, control, **kwargs
+        ):  # pylint: disable=unused-argument
+            """
+            handle T4 gpu, we need to convert attention to eager for inference
+            """
+            if "Tesla T4" in self.gpu_name and self.cfg.xformers_attention:
+                trainer.model.eval()
+            trainer.model.config._attn_implementation = (  # pylint: disable=protected-access
+                "eager"
+            )
```
NanoCode012: Do you mean to indent this block?

src/axolotl/utils/config/__init__.py
```diff
@@ -70,6 +70,9 @@ def resolve_dtype(cfg):
             if cfg.fp16 is None and not cfg.float16:
                 cfg.fp16 = True
 
+    if cfg.fp16 and cfg.bf16 == "auto":
+        cfg.bf16 = False
+
```
NanoCode012, line 75: Would be good to note this behavior of `auto` in documentation, `config.qmd`.

There may be unexpected consequences for users if they have both enabled, but can use bf16.

src/axolotl/monkeypatch/attention/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+
+    from .xformers import xformers_attention_forward
+
+    ALL_ATTENTION_FUNCTIONS["flash_attention_2"] = xformers_attention_forward
+
+
+def unpatch_xformers_attn_over_fa2():
+    from transformers.modeling_utils import ALL_ATTENTION_FUNCTIONS
+
+    ALL_ATTENTION_FUNCTIONS["flash_attention_2"] = flash_attention_forward()
```
NanoCode012, line 19: I'm curious why we don't add a new attention implementation in the monkeypatch itself?

winglian: we need to set it to use flash attention because there are other places the modeling code makes decisions based off of this.
Screenshot 2025-05-06 at 8 54 37 AM

src/axolotl/utils/config/__init__.py
winglian: docs updated

PR #2621, opened 2025-05-03
Configurable embeddings upcast

winglian:
# Description
PEFT will upcast the embeddings to float32, but this can OOM on low resource GPUs and isn't always strictly necessary. We can make it configurable to skip the upcast.

### DIFF
--- a/src/axolotl/monkeypatch/peft/utils.py
+++ b/src/axolotl/monkeypatch/peft/utils.py
@@ -1,0 +1,77 @@
+"""
+Patch prepare_model_for_kbit_training to not upcast everything
+"""
+
+import inspect
+import logging
+
+import peft
+
+from axolotl.monkeypatch.utils import detab_code
+
+LOG = logging.getLogger(__name__)
+
+ORIGINAL_PREPARE_CODE = """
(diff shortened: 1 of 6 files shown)


src/axolotl/monkeypatch/peft/utils.py
```diff
@@ -24,7 +24,7 @@
         for name, param in model.named_parameters():
             if (
                 (param.dtype == torch.float16) or (param.dtype == torch.bfloat16)
-            ) and param.__class__.__name__ != "Params4bit" and "norm" in name:
+            ) and param.__class__.__name__ != "Params4bit" and all(embed_name not in name for embed_name in ["embed_tokens", "lm_head"]):
```
salmanmohammadi, line 27: did this work OK with models which have tied embeddings?

salmanmohammadi (approved):

src/axolotl/monkeypatch/peft/utils.py
winglian: just did a quick notebook check with talk like a pirate, and qwen3-4b (tied) works as expected.

PR #2625, opened 2025-05-03
chunked cross entropy loss

winglian:
in testing with qwen3, reduced single gpu vram from 22-26GB -> 15-18GB

## Summary by CodeRabbit

- **New Features**
  - Introduced an option to enable memory-efficient chunked cross-entropy loss during training, with configurable chunk sizes.
  - Added new configuration fields to control chunked cross-entropy behavior.
- **Bug Fixes**
  - Improved import handling for plugin integration.
- **Tests**
  - Added tests to verify the correctness of the chunked cross-entropy loss implementation.

### DIFF
--- a/src/axolotl/integrations/cut_cross_entropy/__init__.py
+++ b/src/axolotl/integrations/cut_cross_entropy/__init__.py
@@ -71,7 +71,7 @@ class CutCrossEntropyPlugin(BasePlugin):
         if cfg.cut_cross_entropy:
             self._check_requirements()
 
-            from axolotl.integrations.cut_cross_entropy.monkeypatch.patch import (
+            from .monkeypatch.patch import (
                 cce_patch,
             )
 
--- a/src/axolotl/monkeypatch/loss/chunked.py
+++ b/src/axolotl/monkeypatch/loss/chunked.py
@@ -1,0 +1,129 @@
+"""
+chunked ce loss
+"""
+
+from typing import List, Optional
+
+import torch
+import torch.nn.functional as F
+
+
+# copied and modified from torchtune.modules.loss.CEWithChunkedOutputLoss
+class CEWithChunkedOutputLoss(torch.nn.Module):
+    """
+    Cross-entropy with chunked outputs that saves memory by only upcasting one chunk at a time.
+
+    For more details, please refer to: https://github.com/pytorch/torchtune/pull/1390
+    """
+
+    def __init__(self, num_output_chunks: int = 8, ignore_index: int = -100):
+        super().__init__()
+        self.num_output_chunks = num_output_chunks
+        self.ignore_index = ignore_index
+
+    def compute_cross_entropy(
+        self,
+        logits: torch.Tensor,
+        labels: torch.Tensor,
+        normalize: bool = True,  # pylint: disable=unused-argument
+    ) -> torch.Tensor:
+        """
+        Upcast logits to fp32 and compute cross entropy loss.
+        """
+        return F.cross_entropy(
+            logits.float(), labels, ignore_index=self.ignore_index, reduction="sum"
+        )
+
+    def forward(
+        self, logits: List[torch.Tensor], labels: torch.Tensor, reduction="sum"
+    ) -> torch.Tensor:
+        """
+        Args:
+            logits (List[torch.Tensor]): List of chunked logits of length
+                ``self.num_output_chunks``, where each chunk has shape
+                ``(batch_size, num_tokens / num_output_chunks, vocab_size)``.
+            labels (torch.Tensor): Ground truth labels of shape ``(batch_size, num_tokens)``.
+            reduction (str): The reduction to apply to the output.
+
+        Returns:
+            torch.Tensor: Cross entropy loss of shape (1,).
+        """
+
+        total_elements = (labels != self.ignore_index).sum()
+
+        # chunk and reshape labels (bsz, num_tokens, vocab) -> [(bsz*num_tokens/num_chunks, vocab)]
+        labels = [
+            target_chunk.reshape(-1)
+            for target_chunk in labels.chunk(self.num_output_chunks, dim=1)
+        ]
+        # reshape logits [(bsz, num_tokens/num_chunks, vocab)] -> [(bsz*num_tokens/num_chunks, vocab)]
+        logits = [
+            logit_chunk.reshape(-1, logit_chunk.size(-1)) for logit_chunk in logits
+        ]
+
+        # compute one chunk at a time
+        total_loss = 0.0
+        for logits_chunk, labels_chunk in zip(logits, labels):
+            total_loss += self.compute_cross_entropy(logits_chunk, labels_chunk)
+
+        if reduction == "sum":
+            return total_loss
+        return total_loss / total_elements
+
+
+def build_chunked_ce_loss_fn(num_output_chunks: int = 8, ignore_index: int = -100):
+    loss_fn_ce = CEWithChunkedOutputLoss(num_output_chunks, ignore_index)
+    loss_fn_ce.compute_cross_entropy = torch.compile(
+        loss_fn_ce.compute_cross_entropy, backend="inductor"
+    )
+    return loss_fn_ce
+
+
+def patch_chunked_ce_loss_fn(num_output_chunks: int = 8, ignore_index: int = -100):
+    import transformers.loss.loss_utils
+
+    loss_fn_ce = build_chunked_ce_loss_fn(num_output_chunks, ignore_index)
+
+    def chunked_fix_cross_entropy(
+        source,
+        target,
+        num_items_in_batch: int = None,
+        ignore_index: int = -100,
+        **kwargs,
+    ):  # pylint: disable=unused-argument
+        reduction = "sum" if num_items_in_batch is not None else "mean"
+        logit_chunks = [  # pylint: disable=unnecessary-comprehension
+            chunk for chunk in source.chunk(loss_fn_ce.num_output_chunks, dim=1)
+        ]
+        loss = loss_fn_ce(logit_chunks, target, reduction=reduction)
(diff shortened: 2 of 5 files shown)


src/axolotl/integrations/cut_cross_entropy/__init__.py
```diff
@@ -72,7 +72,7 @@ def pre_model_load(self, cfg):
         if cfg.cut_cross_entropy:
             self._check_requirements()
 
-            from axolotl.integrations.cut_cross_entropy.monkeypatch.patch import (
+            from .monkeypatch.patch import (
```
NanoCode012, line 74: Why are we changing to relative import here?

src/axolotl/monkeypatch/loss/chunked.py
```diff
@@ -0,0 +1,134 @@
...
+chunked ce loss
+"""
+
+from typing import List, Optional
+
+import torch
+import torch.nn.functional as F
+
+
+# copied and modified from torchtune.modules.loss.CEWithChunkedOutputLoss
```
NanoCode012, line 11: Do we need to add license block?

tests/test_chunked_xentropy.py
```diff
@@ -0,0 +1,40 @@
...
+    batch_size = 1
+
+    lm_head = nn.Linear(model_dim, vocab_size)
+    hidden_state = torch.randn(batch_size, seq_len, model_dim)
+    labels = torch.randint(low=0, high=vocab_size, size=(batch_size, seq_len))
+    return lm_head, hidden_state, labels, vocab_size
+
+
+def test_chunked_forward(chunked_fixtures):  # pylint: disable=redefined-outer-name
+    lm_head, hidden_state, labels, vocab_size = chunked_fixtures
```
NanoCode012, line 26: Should we set seed for RNG? (referring to torchtune's 

```py
from torchtune.utils.seed import set_seed


 @pytest.fixture(autouse=True)
 def random():
     set_seed(42)
```

src/axolotl/utils/schemas/config.py
```diff
@@ -242,6 +242,9 @@ class AxolotlInputConfig(
     unsloth_rms_norm: bool | None = None
     unsloth_rope: bool | None = None
 
+    chunked_cross_entropy: bool | None = None
+    chunked_cross_entropy_num_chunks: int | None = None
+
```
NanoCode012: Can this be added to docs `config.qmd`?

src/axolotl/monkeypatch/loss/chunked.py
NanoCode012: The source is BSD3

src/axolotl/monkeypatch/loss/chunked.py
```diff
@@ -0,0 +1,134 @@
...
+        ]
+
+        # compute one chunk at a time
+        total_loss = 0.0
+        for logits_chunk, labels_chunk in zip(logits, labels):
+            total_loss += self.compute_cross_entropy(logits_chunk, labels_chunk)
+
+        if reduction == "sum":
+            return total_loss
+        return total_loss / total_elements
```
NanoCode012, line 71: Should check if `mean` for second return and raise value/notimplemented error for other reduction

src/axolotl/monkeypatch/loss/chunked.py
```diff
@@ -0,0 +1,134 @@
...
+    ) -> torch.Tensor:
+        """
+        Upcast logits to fp32 and compute cross entropy loss.
+        """
+        return F.cross_entropy(
+            logits.float(), labels, ignore_index=self.ignore_index, reduction="sum"
+        )
+
+    def forward(
+        self, logits: List[torch.Tensor], labels: torch.Tensor, reduction="sum"
```
NanoCode012, line 38:
```suggestion
        self, logits: List[torch.Tensor], labels: torch.Tensor, reduction="mean"
```

This should be mean to follow regular defaults? https://pytorch.org/docs/stable/generated/torch.nn.CrossEntropyLoss.html#torch.nn.CrossEntropyLoss

src/axolotl/monkeypatch/loss/chunked.py
```diff
@@ -0,0 +1,134 @@
...
+        if reduction == "sum":
+            return total_loss
+        return total_loss / total_elements
+
+
+def _build_chunked_ce_loss_fn(num_output_chunks: int = 8, ignore_index: int = -100):
+    loss_fn_ce = CEWithChunkedOutputLoss(num_output_chunks, ignore_index)
+    loss_fn_ce.compute_cross_entropy = torch.compile(
+        loss_fn_ce.compute_cross_entropy, backend="inductor"
+    )
```
NanoCode012, line 78: Should we also read env var like in torchtune?

```python
backend = os.environ.get("TORCH_COMPILE_BACKEND", "inductor")
```

src/axolotl/monkeypatch/loss/chunked.py
```diff
@@ -0,0 +1,134 @@
...
+    loss_fn_ce = _build_chunked_ce_loss_fn(num_output_chunks, ignore_index)
+
+    def chunked_fix_cross_entropy(
+        source,
+        target,
+        num_items_in_batch: int = None,
+        ignore_index: int = -100,
+        **kwargs,
+    ):  # pylint: disable=unused-argument
+        reduction = "sum" if num_items_in_batch is not None else "mean"
```
NanoCode012, line 92: Is this correct? Could you explain the relationship between `num_items_in_batch` and `reduction`?

src/axolotl/monkeypatch/loss/chunked.py
```diff
@@ -0,0 +1,134 @@
...
+        logit_chunks = [  # pylint: disable=unnecessary-comprehension
+            chunk for chunk in source.chunk(loss_fn_ce.num_output_chunks, dim=1)
```
NanoCode012, line 94: Is this doing nothing, as current transformers already calculate the logits before hand?

NanoCode012: I saw this comment on the linked PR

> It is important to notice that chunking must happen BEFORE the linear output, or the memory is not released:

Does it refer to the chunk CE inside the class or the logits chunk calculation?

src/axolotl/monkeypatch/loss/chunked.py
NanoCode012: Does that mean, we need to add a test with `num_items_in_batch` for `reduction=sum`?

NanoCode012 (commented):
Nice implementation. Left some comments on parts I'm unsure of

src/axolotl/monkeypatch/loss/chunked.py
NanoCode012: Same for in the `get_*` function below.

src/axolotl/monkeypatch/loss/chunked.py
salmanmohammadi, line 11: This implementation has since been deprecated for https://github.com/pytorch/torchtune/blob/main/torchtune/modules/loss/cross_entropy_loss.py which we should follow instead

winglian: Let's tackle that as a follow-on. The newer version in TT fuses with the linear by passing the weight of the embedding, but this isn't as straightforward and requires individual patches for each model architecture. This is a simpler patch that will work for all models

src/axolotl/monkeypatch/loss/chunked.py
salmanmohammadi: Do we not want to mean across all the chunks, rather than per-chunk? The actual normalisation happens below

src/axolotl/integrations/cut_cross_entropy/__init__.py
winglian: absolutely import would break in colab for some reason. Also, it's the pattern we're typically using in plugins.

tests/test_chunked_xentropy.py
winglian: not really necessary?

src/axolotl/monkeypatch/loss/chunked.py
winglian: this is the stupid gradient accumulation think that Daniel Han brought up that forced this pattern.

src/axolotl/monkeypatch/loss/chunked.py
winglian: This is the chunking, the optimization is that by chunking before we upcast to float32, we reduce the VRAM overhead.

src/axolotl/monkeypatch/loss/chunked.py
```diff
@@ -0,0 +1,134 @@
...
+            total_loss += self.compute_cross_entropy(logits_chunk, labels_chunk)
+
+        if reduction == "sum":
+            return total_loss
+        return total_loss / total_elements
+
+
+def _build_chunked_ce_loss_fn(num_output_chunks: int = 8, ignore_index: int = -100):
+    loss_fn_ce = CEWithChunkedOutputLoss(num_output_chunks, ignore_index)
+    loss_fn_ce.compute_cross_entropy = torch.compile(
```
NanoCode012, line 76: Another question, would this work for all hardware? Do we need to check for compatibility?

salmanmohammadi: Yeah unlike flex attention there's no cudagraphs/autotuning here so inductor should be pretty hardware friendly

salmanmohammadi (approved):
I think this looks good - the loss/grad norms lined up OK with the run you tested on?

Let's also make sure we add the license/code attribution link

NanoCode012:
Rebased. Added license, arg to docs, and use TORCH_COMPILE_BACKEND for backend if available

PR #2627, opened 2025-05-04
Print axolotl art if train is called outside of cli

winglian:
## Motivation and Context

putting together some google colabs that don't use the cli directly, but still want some pretty art :D

## How has this been tested?

TBD

### DIFF
--- a/src/axolotl/common/datasets.py
+++ b/src/axolotl/common/datasets.py
@@ -48,6 +48,7 @@ def load_datasets(
     *,
     cfg: DictDefault,
     cli_args: PreprocessCliArgs | TrainerCliArgs | None = None,
+    debug: bool = False,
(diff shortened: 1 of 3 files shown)


src/axolotl/common/datasets.py
```diff
@@ -48,6 +48,7 @@ def load_datasets(
     *,
     cfg: DictDefault,
     cli_args: PreprocessCliArgs | TrainerCliArgs | None = None,
+    debug: bool = False,
```
NanoCode012, line 51: Why do we need this when it can already exist in `cfg.debug` and `cli_args.debug`?

winglian: makes it easier to manually debug this when calling from colab

PR #2631, opened 2025-05-05
Multipack parallel bin packing

winglian:
Resurrecting @dsesclei 's previous PR #1516 to get better packing efficiency without the loss issues we were seeing post merge where we had to revert in #1672. 

I also had Claude add comments and rename variables so this module is easier to read.

### Benchmarks

I threw this at FineTome-100k dataset and on `main` the packing efficiency was 0.8978517270429655, needing 721 train steps (after 21 steps: `21/721 [02:05<1:09:16,  5.94s/it]`). With this PR, it's now an efficiency of 0.9983726460415066 needing 650 train steps ( after 21 steps:`21/650 [02:11<1:03:33,  6.06s/it]`), resulting in a 10% speedup which is mostly free lunch since we have to pad to the full sequence length already.

### DIFF
--- a/src/axolotl/utils/samplers/multipack.py
+++ b/src/axolotl/utils/samplers/multipack.py
@@ -61,7 +61,8 @@ def pack_group(
     group_offset: int,
     bin_capacity: int,
     max_bins: int,
-    safe_mode: bool = False,
+    bin_size: int,
+    safe_mode: bool = True,
 ):
     """
     Pack a group of sequences into bins using First-Fit Decreasing algorithm
@@ -71,6 +72,7 @@ def pack_group(
         group_offset: Offset to apply to indices when returning results
         bin_capacity: Maximum capacity of each bin
         max_bins: Maximum number of bins to use
+        bin_size: Maximum number of sequences per bin
         safe_mode: If True, use a more conservative packing approach
 
     Returns:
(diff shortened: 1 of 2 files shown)


NanoCode012:
How does the loss chart look before/after?

src/axolotl/utils/samplers/multipack.py
```diff
@@ -13,267 +16,408 @@
...
+    Pack a group of sequences into bins using First-Fit Decreasing algorithm
 
-    indices = np.argsort(a)[::-1]
-    a = a[indices]
+    Args:
+        sequence_lengths: Array of sequence lengths
+        group_offset: Offset to apply to indices when returning results
+        bin_capacity: Maximum capacity of each bin
+        max_bins: Maximum number of bins to use
+        safe_mode: If True, use a more conservative packing approach
```
NanoCode012, line 76: When would this be False?

Should we keep the defaults for this the same? `MultipackBatchSampler` sets this to true.

src/axolotl/utils/samplers/multipack.py
```diff
@@ -13,267 +16,408 @@
...
-        sequential: bool = False,
-        **kwargs,
+        batch_size: int,  # Number of bins per batch
+        batch_max_len: int,  # Maximum sequence length (bin capacity)
+        lengths: np.ndarray,  # Sequence lengths
+        packing_efficiency_estimate: float = 1.0,  # Initial efficiency estimate
+        drop_last: bool = False,  # Whether to drop final batches (might be incomplete)
+        num_count_samples: int = 16,  # Number of times to estimate batch count
+        sequential: bool = False,  # Whether to use sequential packing
+        group_size: int = 100_000,  # Size of groups for parallel packing
```
NanoCode012, line 240: Where is this number from? Does it need to be configurable?

NanoCode012 (commented):
Looking through this a few times. The main change is the swap to parallel packing. The rest were refactors.

What changed specifically from the original PR to fix the loss issue?

src/axolotl/utils/samplers/multipack.py
winglian: Updated to be configurable. These were the original defaults from the previous PR

winglian:
numba.njit is killing me on this (for the sequential cases)

PR #2637, opened 2025-05-06
Fix: improve error message on failed dataset load

NanoCode012:
# Description

Improve error message when failing to load dataset. A user reported to not being able to understand the original error

### DIFF
--- a/src/axolotl/utils/data/shared.py
+++ b/src/axolotl/utils/data/shared.py
@@ -281,6 +281,9 @@ def load_dataset_w_config(
             **load_ds_kwargs,
         )
     if not ds:
-        raise ValueError("unhandled dataset load")
+        raise ValueError(
+            "Cannot load dataset from the given config. Please check (if applicable) path / name / data_files. "
+            "This is not from the dataset type."
+        )
 
     return ds
--- a/src/axolotl/utils/data/shared.py
+++ b/src/axolotl/utils/data/shared.py
@@ -282,9 +282,9 @@ def load_dataset_w_config(
         )
     if not ds:
         raise ValueError(
-            f"Cannot load dataset from the given config (path: {config_dataset.path}). "
-            "Please check (if applicable) path / name / data_files. "
-            "This is not from the dataset type."
+            "The dataset could not be loaded. This could be due to a misconfigured dataset path "
+            f"({config_dataset.path}). Try double-check your path / name / data_files. "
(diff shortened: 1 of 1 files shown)


src/axolotl/utils/data/shared.py
```diff
@@ -281,6 +281,10 @@ def load_dataset_w_config(
             **load_ds_kwargs,
         )
     if not ds:
-        raise ValueError("unhandled dataset load")
+        raise ValueError(
```
salmanmohammadi, line 284: So when we've hit this error, `load_datasets` has returned `None` rather than raised an error? Do you have any examples of when this would happen?

NanoCode012: Yes, any typos for path/datafiles etc

src/axolotl/utils/data/shared.py
```diff
@@ -281,6 +281,10 @@ def load_dataset_w_config(
...
+            f"Cannot load dataset from the given config (path: {config_dataset.path}). "
+            "Please check (if applicable) path / name / data_files. "
+            "This is not from the dataset type."
+        )
```
salmanmohammadi:
```suggestion
            f"The dataset type has been verified, but the dataset could not  "
            "be loaded. This could be due to a misconfigured dataset path "
            "({config_dataset.path}). Try double-check your path / name/ data_files."
        )
```

salmanmohammadi (approved):
I would've thought `load_datasets` would throw an error if files couldn't be found but this is a good stopgap

src/axolotl/utils/data/shared.py
NanoCode012: Hm, at this stage in data process, the dataset type is not being used yet and hence would not be the cause of the error. It's only just trying to load the data from disk.

The stage after this would be loading the dataset type wrapper, so I don't think we can say, it's verified.

NanoCode012:
> I would've thought `load_datasets` would throw an error if files couldn't be found but this is a good stopgap

This is because we check if the path exists before attempting to load any local path

https://github.com/axolotl-ai-cloud/axolotl/blob/0d71b0aa5fd4a35a011a574a9233eecb4f2793bb/src/axolotl/utils/data/shared.py#L176-L178

src/axolotl/utils/data/shared.py
NanoCode012: Updated the message!

PR #2638, opened 2025-05-06
make sure gc_steps is used for all trainers

winglian:
It was only configured to run for SFT, but should be available for all.


salmanmohammadi (approved):

tcapelle:
merge merge

PR #2641, opened 2025-05-06
swap tinymodels that have safetensors for some ci tests

winglian:
There were some changes upstream with forcing checks on models when using torch.load that is breaking our CI against upstream.  

Failures mostly show up in the lora kernel tests https://github.com/axolotl-ai-cloud/axolotl/actions/runs/14805596453/job/41573587577

Kicked off an upstream CI job here: ~https://github.com/axolotl-ai-cloud/axolotl/actions/runs/14868230112~

latest: https://github.com/axolotl-ai-cloud/axolotl/actions/runs/14872074870

### DIFF
--- a/src/axolotl/utils/gradient_checkpointing/__init__.py
+++ b/src/axolotl/utils/gradient_checkpointing/__init__.py
@@ -1,15 +1,36 @@
 """custom checkpointing utils"""
 
+import importlib
 from functools import partial
+
+from packaging import version
 
 from axolotl.utils.gradient_checkpointing.unsloth import (
     Unsloth_Offloaded_Gradient_Checkpointer,
 )
 
+transformers_version = version.parse(importlib.metadata.version("transformers"))
+if transformers_version > version.parse("4.51.3"):
+    from transformers.modeling_layers import GradientCheckpointingLayer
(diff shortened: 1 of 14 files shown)


djsaunde (approved):

src/axolotl/utils/gradient_checkpointing/__init__.py
```diff
@@ -1,15 +1,36 @@
...
+if transformers_version > version.parse("4.51.3"):
+    from transformers.modeling_layers import GradientCheckpointingLayer
+
+    def uses_gc_layers(decoder_layer):
+        return isinstance(decoder_layer.func.__self__, GradientCheckpointingLayer)
+
+else:
+
+    def uses_gc_layers(_):
+        return False
```
NanoCode012, line 22: Would it be better to just have the transformers check within the function itself instead of doing if else and declaring it twice?

winglian: If we put the transformers check inside, then transformers check happens on every train step and layer of the model. This way it only happens once.

salmanmohammadi (approved):

PR #2642, opened 2025-05-07
Add missing __init__.py file to `cut_cross_entropy` integration

BitPhinix:
It's currently not present resulting in `ModuleNotFoundError: No module named 'axolotl.integrations.cut_cross_entropy.monkeypatch'` when installing axolotl via `uv` and using the `cut_cross_entropy` integration


winglian (approved):

winglian:
Thanks for catching that @BitPhinix !

PR #2608, opened 2025-05-01
Rank 0-only logging

salmanmohammadi:
Just a start on removing unnecessary duplicate logging across multiple ranks. Will update as I find more cases - @maintainers feel free to add to this branch.

```bash
root@eebe878229e8:/workspace/axolotl# axolotl train examples/llama-3/lora-1b.yml 
[2025-05-19 17:03:05,194] [INFO] [real_accelerator.py:219:get_accelerator] Setting ds_accelerator to cuda (auto detect)
[2025-05-19 17:03:05,283] [INFO] [root.spawn:77] [PID:98907] gcc -pthread -B /root/miniconda3/envs/py3.11/compiler_compat -DNDEBUG -fwrapv -O2 -Wall -fPIC -O2 -isystem /root/miniconda3/envs/py3.11/include -fPIC -O2 -isystem /root/miniconda3/envs/py3.11/include -fPIC -c /tmp/tmpnwaw2628/test.c -o /tmp/tmpnwaw2628/test.o
[2025-05-19 17:03:05,310] [INFO] [root.spawn:77] [PID:98907] gcc -pthread -B /root/miniconda3/envs/py3.11/compiler_compat /tmp/tmpnwaw2628/test.o -laio -o /tmp/tmpnwaw2628/a.out
[2025-05-19 17:03:05,350] [INFO] [root.spawn:77] [PID:98907] gcc -pthread -B /root/miniconda3/envs/py3.11/compiler_compat -DNDEBUG -fwrapv -O2 -Wall -fPIC -O2 -isystem /root/miniconda3/envs/py3.11/include -fPIC -O2 -isystem /root/miniconda3/envs/py3.11/include -fPIC -c /tmp/tmpc82bup43/test.c -o /tmp/tmpc82bup43/test.o
[2025-05-19 17:03:05,376] [INFO] [root.spawn:77] [PID:98907] gcc -pthread -B /root/miniconda3/envs/py3.11/compiler_compat /tmp/tmpc82bup43/test.o -L/usr/local/cuda -L/usr/local/cuda/lib64 -lcufile -o /tmp/tmpc82bup43/a.out
The following values were not passed to `accelerate launch` and had defaults used instead:
        `--num_processes` was set to a value of `2`
                More than one GPU was found, enabling multi-GPU training.
                If this was unintended please pass in `--num_processes=1`.
        `--num_machines` was set to a value of `1`
        `--mixed_precision` was set to a value of `'no'`
        `--dynamo_backend` was set to a value of `'no'`
[...]

PR #2643, opened 2025-05-07
SP GRPO support + batch SP fixes

djsaunde:
# Description

This PR implements support for sequence parallelism (SP) for the GRPO trainer. This includes a custom sampler, similar to the one implemented in TRL, but with logic for replicating samples across processes in the same SP group.

To make this happen, we need to override a lot more code from the GRPO trainer (e.g., the `__init__` function). This could be greatly improved by refactoring the trainer upstream to be a lot more modular and therefore more easily extensible; most of the code we're taking from the superclass is unchanged.

We were also able to fix the issue in the `pad_to_sequence_len: false` case where, for specific input lengths, the `ring_flash_attn` batch ring attention function experienced extremely large or `inf` gradient norms, by calling `torch.compile` on the function 🤔

+ some small changes / refactors.

## Motivation and Context

GRPO for sufficiently advanced tasks will likely require longer sequence lengths than we can fit on on a single GPU. Hence, let's add SP support.

Follow-ups:
- Add zigzag, stripe batch ring attn adapters + data splitting, gathering logic
- Upstream base GRPO trainer refactor (?)

## How has this been tested?

- Pytests (need additional coverage)
- Manual testing of:
  - SP + SFT (`pad_to_seq_len: false` vs. `pad_to_seq_len: true`, `sample_packing: false` vs. `sample_packing: true`, etc.)
  - SP + GRPO

Example GRPO + SP training run:

image

Note that the config differed a fair bit from the one in [our blog post](https://huggingface.co/blog/axolotl-ai-co/training-llms-w-interpreter-feedback-wasm), so they're not directly comparable.

## Social Handles (Optional)


src/axolotl/core/trainers/grpo/trainer.py
```diff
@@ -1,26 +1,281 @@
...
+    from peft import PeftConfig
 
 if is_deepspeed_available():
     import deepspeed
 
+if is_wandb_available():
+    import wandb
+
 
 class AxolotlGRPOTrainer(RngLoaderMixin, SchedulerMixin, GRPOTrainer):
```
winglian, line 59: Should this be a new `AxolotlGRPOSequenceParallelTrainer`? Do we want to introduce all the extra code for non-SP?

src/axolotl/train.py
```diff
@@ -199,15 +200,17 @@ def execute_training(
...
+            models=models,
             sequence_parallel_degree=cfg.sequence_parallel_degree,
             ring_attn_func=cfg.ring_attn_func,
         )
-        if cfg.sequence_parallel_degree > 1
-        else nullcontext()
-    )
 
     LOG.info("Starting trainer...")
     with flash_context, sequence_parallel_context:
```
winglian: This could probably be cleaned up from all the extra if/else nullcontexts using [`ExitStack`](https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack)

src/axolotl/utils/ctx_managers/sequence_parallel.py
```diff
@@ -0,0 +1,370 @@
...
+        self.apply_sequence_parallelism = functools.partial(
+            apply_sequence_parallelism,
+            local_rank=self.local_rank,
+            local_world_size=self.local_world_size,
+            ring_attn_func=self.ring_attn_func,
+        )
+
+    def __enter__(self):
+        # Forward pre-hook to apply sequence parallelism
+        def sequence_parallel_pre_hook(_, args, kwargs):
```
winglian, line 208: Any particular reason this isn't `*args, **kwargs`? Is it already passed in as a `list, dict`?

src/axolotl/core/trainers/grpo/trainer.py
winglian: Perhaps `AxolotlGRPOSequenceParallelTrainer(AxolotlGRPOTrainer)`?

src/axolotl/utils/ctx_managers/sequence_parallel.py
djsaunde: Yes, already passed as a `tuple`, `dict`.

src/axolotl/core/trainers/grpo/trainer.py
djsaunde: Yeah, this might be best to shield the vanilla GRPO trainer from breaking (at least, not as often) from upstream changes.

src/axolotl/common/datasets.py
```diff
@@ -133,7 +134,7 @@ def load_preference_datasets(
     total_num_steps: Optional[int] = int(
         math.ceil(len(train_dataset) * cfg.num_epochs / cfg.batch_size)
     )
-    if cfg.rl == "grpo":
+    if cfg.rl is RLType.GRPO:
```
salmanmohammadi, line 137: beaut

src/axolotl/core/trainer_builder.py
```diff
@@ -1140,67 +1147,69 @@ def build_training_arguments(self, total_num_steps):
...
 
-        dpo_trainer = self.hook_post_create_trainer(dpo_trainer)
-        for callback in self.get_post_trainer_create_callbacks(dpo_trainer):
-            dpo_trainer.add_callback(callback)
+        trainer = self.hook_post_create_trainer(trainer)
+        for callback in self.get_post_trainer_create_callbacks(trainer):
+            trainer.add_callback(callback)
 
-        return dpo_trainer
+        return trainer
```
salmanmohammadi, line 1230: This whole file also beaut

src/axolotl/core/trainers/grpo/sampler.py
```diff
@@ -0,0 +1,170 @@
...
+                * self.batch_size
+            )
+
+    def __iter__(self) -> Iterator[int]:
+        """Creates iterator over dataset indices.
+
+        Returns:
+            Iterator that yields indices into the dataset.
+        """
+        # Deterministically shuffle based on epoch and seed
```
salmanmohammadi, line 115: Do we support resuming from checkpoints between epoch boundaries? Would using the current step to seed be better if so?

src/axolotl/core/trainers/grpo/trainer.py
```diff
@@ -67,3 +119,618 @@ def _move_model_to_vllm(self):
...
+                description="training",
+            )
+
+        # Get sampler and create dataloader
+        sampler = self._get_train_sampler()
+        dataloader = self._prepare_dataloader(train_dataset, sampler, is_eval=False)
+
+        return dataloader
+
+    def _generate_and_score_completions(
```
salmanmohammadi, line 305: Sorry I'm being lazy here, could you TLDR the changes needed to this function?
We'll also need to gate on using Liger kernels (i.e. raise an error if SP + Liger), since we're landing #2648 in this release too, which relies on a modified `_generate_and_score_completions` upstream.

djsaunde: Wait, are SP + liger incompat? I think it works in the SFT case.

Changes:
- `ordered_set_of_prompts` assignment changes due to duplicating prompts across SP groups
- similarly for `completion_ids`
- similarly for `advantages`

winglian: Liger + GRPO are compatible, but adding PEFT to the mix currently isn't compatible.

src/axolotl/monkeypatch/attention/ring_attn/adapters/batch.py
```diff
@@ -16,24 +16,20 @@
...
 from transformers.modeling_utils import ALL_ATTENTION_FUNCTIONS
 
-from axolotl.monkeypatch.attention.ring_attn.patch import RingAttnFunc
+from axolotl.utils.schemas.enums import RingAttnFunc
 
 RING_ATTN_FUNC_MAPPING = {
-    RingAttnFunc.BATCH_RING: ring_flash_attn_func,
-    RingAttnFunc.BATCH_ZIGZAG: zigzag_ring_flash_attn_func,
-    RingAttnFunc.BATCH_STRIPE: stripe_flash_attn_func,
+    RingAttnFunc.BATCH_RING: torch.compile(ring_flash_attn_func),
```
salmanmohammadi, line 30: This is going to recompile every time `RING_ATTN_FUNC_MAPPING[RingAttnFunc.BATCH_RING]` is accessed, right? We should wrap this in a function and make sure it's only compiled and returned once.

src/axolotl/train.py
```diff
@@ -189,30 +190,42 @@ def execute_training(
...
+
+            stack.enter_context(
+                SequenceParallelContextManager(
+                    models=models,
+                    sequence_parallel_degree=cfg.sequence_parallel_degree,
+                    ring_attn_func=cfg.ring_attn_func,
+                )
+            )
+
+        torch.cuda.memory._record_memory_history(
```
salmanmohammadi: vestigial debugging code?

src/axolotl/train.py
```diff
@@ -189,30 +190,42 @@ def execute_training(
...
-    LOG.info("Starting trainer...")
-    with flash_context, sequence_parallel_context:
+        LOG.info("Starting trainer...")
         trainer.train(resume_from_checkpoint=resume_from_checkpoint)
 
+        import torch.distributed as dist
+        torch.cuda.memory._dump_snapshot(f"../memory_snapshot_{dist.get_rank()}.pickle")
+        # torch.cuda.memory._dump_snapshot(f"../memory_snapshot.pickle")
+        torch.cuda.memory._record_memory_history(enabled=None)
+
```
salmanmohammadi: As above

src/axolotl/utils/ctx_managers/sequence_parallel.py
```diff
@@ -0,0 +1,327 @@
...
+        self.sequence_parallel_degree = sequence_parallel_degree
+        self.ring_attn_func = ring_attn_func
+        self.process_group = get_ring_attn_group()
+
+        # Initialize sequence parallel group details
+        self.local_rank = dist.get_rank(self.process_group)
+        self.local_world_size = dist.get_world_size(self.process_group)
+
+        # Will store hook handles for removal
+        self.hook_handles: DefaultDict[int, list[RemovableHandle]] = defaultdict(list)
```
salmanmohammadi: noob q, why does this need to be a dict?

src/axolotl/utils/ctx_managers/sequence_parallel.py
```diff
@@ -0,0 +1,327 @@
...
+            self.hook_handles[key] = []
+
+    def gather_outputs(self, output: CausalLMOutputWithPast) -> CausalLMOutputWithPast:
+        """Gather sharded outputs from all ranks and reconstruct the full tensor."""
+        from torch.distributed.nn.functional import all_gather
+
+        for key, value in output.items():
+            if isinstance(value, torch.Tensor) and value.dim() > 1:
+                output[key] = AllGatherWithGrad.apply(value, self.process_group)
+                #     output[key] = torch.cat(gathered, dim=1)
```
salmanmohammadi: Do we need the import here? Also commented code

src/axolotl/utils/ctx_managers/sequence_parallel.py
```diff
@@ -0,0 +1,327 @@
...
+                concatenating along the sequence dimension.
+        """
+        ctx.group = group
+        ctx.rank = dist.get_rank(group)
+        ctx.world_size = dist.get_world_size(group)
+
+        # Gather shape metadata
+        local_shape = torch.tensor(list(input_tensor.shape), device=input_tensor.device)
+        world_size = ctx.world_size
+        all_shapes = [torch.zeros_like(local_shape) for _ in range(world_size)]
```
salmanmohammadi:
```suggestion
        all_shapes = [torch.zeros_like(local_shape) for _ in range(ctx.world_size)]
```
nit

src/axolotl/utils/data/rl.py
```diff
@@ -185,7 +186,7 @@ def load_split(dataset_cfgs, _cfg):
                     )
 
         combined_datasets = concatenate_datasets(split_datasets)
-        combined_datasets = combined_datasets.shuffle(seed=cfg.seed)
+        combined_datasets = combined_datasets.shuffle(seed=cfg.seed or 42)
```
SalmanMohammadi, line 189: nit: should we set `seed = cfg.seed or 42` once?

src/axolotl/utils/schemas/config.py
```diff
@@ -1444,8 +1472,9 @@ def check_min_torch_version(self):
         if self.env_capabilities and self.env_capabilities.torch_version:
             torch_version = self.env_capabilities.torch_version
             if version.parse(torch_version) < version.parse("2.5.1"):
-                LOG.warning(
-                    f"torch=={torch_version} may not be supported in future versions. Please consider upgrading to torch>=2.5.1."
-                )
+                if is_main_process(use_environ=True):
```
salmanmohammadi: FWIW this should be addressed properly in #2608

salmanmohammadi (commented):
Really really nice. A few nits, and we need to sync the `GRPOTrainer` changes with upstream, but this looks good to me.

src/axolotl/train.py
djsaunde: Hah, yeah, I should've moved this back to a draft PR.

src/axolotl/utils/ctx_managers/sequence_parallel.py
djsaunde: It doesn't 😄 I will change to a list.

src/axolotl/core/trainers/grpo/sampler.py
djsaunde: here, we don't. I basically took the `RepeatSampler` from the trl grpo implementation and modified it a bit, so we inherited this behavior from that.

src/axolotl/monkeypatch/attention/ring_attn/adapters/batch.py
djsaunde: I'm fairly sure the function gets compiled once, at dictionary definition time.

src/axolotl/utils/schemas/config.py
djsaunde: Cool! What's required to get that merged?

djsaunde:
Note: GRPO + SP + Liger results in exploding losses. I suggest we merge this now (once tests pass) and follow up with a fix for this case.

src/axolotl/utils/schemas/config.py
salmanmohammadi: I just need to spare a couple hours haha

djsaunde: haha, no worries. I've reverted these logging changes, we can use your branch for that

docs/sequence_parallelism.qmd
```diff
@@ -27,7 +25,7 @@ To enable sequence parallelism, add the following to your configuration file:
 sequence_parallel_degree: 4  # Split sequences across 4 GPUs
 # Optional; strides across the key dimension. Larger values use more memory but should make training faster.
 heads_k_stride: 1
-# Optional; one of "varlen_llama3", "batch_ring", "batch_zigzag", "batch_stripe". Defaults to
+# Optional; one of "varlen_llama3" or "batch_ring". Defaults to
```
salmanmohammadi, line 28: I don't think it's that critical but we might want to log a deprecation warning if a user tries to run with an old config with a now unsupported `rng_attn_func` otherwise they may be very confused why their config doesn't work anymore

salmanmohammadi:
> Note: GRPO + SP + Liger results in exploding losses. I suggest we merge this now (once tests pass) and follow up with a fix for this case.

This sounds good to me. Should we error out with a sensible message if a user tries to combine these features?

djsaunde:
Getting 429 errors from tests involving HF requests

tests/e2e/patched/test_sp.py
```diff
@@ -328,57 +365,58 @@ def test_batch_ring_rank1(self, sequence_parallel_batch):
...
-                [original_input_ids[:, :2], original_input_ids[:, 6:8]], dim=1
-            )
-
-            rank1_expected = torch.cat(
-                [original_input_ids[:, 2:4], original_input_ids[:, 4:6]], dim=1
-            )
-
-            assert torch.equal(result_rank0["input_ids"], rank0_expected)
-            assert torch.equal(result_rank1["input_ids"], rank1_expected)
+    # def test_batch_zigzag(self, sequence_parallel_batch):
```
salmanmohammadi, line 380: throw a TODO here too?

salmanmohammadi (approved):
looks good to go with green CI. awesome work as always

PR #2654, opened 2025-05-09
Fix for setting `adam_beta3` and `adam_epsilon2` for CAME Optimizer

xzuyn:
I misunderstood what `training_arguments_kwargs` was meant for initially, so previously you couldn't actually change them. This should let you modify the `adam_beta3` and `adam_epsilon2` correctly.

I also added the two to the config docs.

## Summary by CodeRabbit

- **New Features**
	- Added support for two new optimizer hyperparameters: `adam_beta3` and `adam_epsilon2`, which can now be configured and used with the CAME Optimizer.

- **Documentation**
	- Updated configuration documentation to include the new `adam_beta3` and `adam_epsilon2` options under optimizer settings.

### DIFF
--- a/docs/config.qmd
+++ b/docs/config.qmd
@@ -633,7 +633,9 @@ weight_decay:
 # adamw hyperparams
 adam_beta1:
 adam_beta2:
+adam_beta3:
 adam_epsilon:
+adam_epsilon2:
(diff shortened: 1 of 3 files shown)


NanoCode012 (approved):
Thanks for catching that!

PR #2655, opened 2025-05-09
fix: disable auto lora kernel if dropout nonzero

NanoCode012:
# Description

If user has dropout set, then, auto enabling it would just disable it later down the line. `Cannot patch layers - requires n
o dropout and no bias`

### DIFF
--- a/src/axolotl/utils/schemas/config.py
+++ b/src/axolotl/utils/schemas/config.py
@@ -1345,7 +1345,7 @@ class AxolotlConfigWCapabilities(AxolotlInputConfig):
             ):
                 return data
 
-            # Skip if dropout is not 0
+            # Skip if dropout is not 0, as auto enabling it would just disable it during runtime patch checks
             if data.get("lora_dropout") != 0:
                 return data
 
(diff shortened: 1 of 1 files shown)


src/axolotl/utils/schemas/config.py
```diff
@@ -1345,6 +1345,10 @@ def check_auto_enable_lora_kernels(cls, data):
             ):
                 return data
 
+            # Skip if dropout is not 0
+            if data.get("lora_dropout") != 0:
+                return data
```
salmanmohammadi, line 1350: Worth logging this so the user is aware?

salmanmohammadi: The reason you mentioned in the PR description would be good

salmanmohammadi (approved):

src/axolotl/utils/schemas/config.py
winglian: probably not needed for logging since this is a check on whether to auto enable it, so it's not necessarily changing any expected behavior

PR #2657, opened 2025-05-09
don't sort multipack sampler

djsaunde:
# Description

image
image
Comparisons vs the old legacy behavior 
## Social Handles (Optional)

We


winglian (approved):

NanoCode012:
Thanks for the fix!

PR #2661, opened 2025-05-10
Fixes 20250510

winglian:
- lean mistral ft tests, remove e2e torch 2.4.1 test
- make sure to pass save_only_model for RL


tcapelle:
can you split and merge the saves?

cicd/cleanup.sh
djsaunde, line 1: Why is this needed?

src/axolotl/utils/samplers/multipack.py
```diff
@@ -154,9 +157,33 @@ def pack_parallel(
...
-    with ProcessPoolExecutor(max_workers=num_processes) as executor:
-        for group_bins in executor.map(_process_group, tasks):
+
+    mp_ctx = None
+    if mp_start_method:
+        try:
+            mp_ctx = get_context(mp_start_method)
+        except ValueError:
+            LOG.warning(
+                f"Failed to get multiprocessing context '{mp_start_method}'. "
```
djsaunde, line 167: In what situation(s) would this happen?

djsaunde (approved):

src/axolotl/utils/samplers/multipack.py
winglian: somehow passed an incorrect multiprocessing method https://docs.python.org/3/library/multiprocessing.html#multiprocessing.get_context

cicd/cleanup.sh
winglian: Screenshot 2025-05-12 at 10 48 04 AM
over time if we don't clean this up, the disk usage starts to build up. We were at almost 400GB of these files on disk on the Modal volume

PR #2663, opened 2025-05-12
Activation checkpointing with offloading to disk with prefetch

winglian:
This PR adds an option to offload activations to disk and prefetch them on the backwards pass right before they're needed. This should really only be used in place of the regular CPU offloading on resource constrained systems that are low on system RAM like Colab. see https://colab.research.google.com/drive/1MBHGHiOtRORFdoHEMZ6jFriMkRqJJIDv?usp=sharing where we are able to train Qwen3-8B w QLoRA @ 28K context

### DIFF
--- a/src/axolotl/utils/gradient_checkpointing/offload_disk.py
+++ b/src/axolotl/utils/gradient_checkpointing/offload_disk.py
@@ -1,8 +1,12 @@
 """Disk offloaded checkpointing"""
 
 import os
+import queue
 import tempfile
+import threading
+import time
 import uuid
+from collections import deque
 
 import torch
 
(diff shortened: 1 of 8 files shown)


src/axolotl/utils/gradient_checkpointing/offload_disk.py
```diff
@@ -0,0 +1,535 @@
...
+            except queue.Empty:
+                time.sleep(0.01)  # Small sleep to prevent CPU spinning
+                continue
+
+    def save_tensor(self, tensor):
+        """Save tensor to disk asynchronously and return file path with thread-safe operations"""
+        # Generate unique file path
+        self.idx += 1
+        file_path = os.path.join(self.temp_dir, f"{self.idx:06d}-{uuid.uuid4()}.pt")
+
```
NanoCode012: Should we perhaps check if there's enough disk space free since these tensors could be quite large?

src/axolotl/utils/gradient_checkpointing/offload_disk.py
```diff
@@ -0,0 +1,535 @@
...
+            except queue.Empty:
+                time.sleep(0.01)  # Small sleep to prevent CPU spinning
+                continue
+
+    def save_tensor(self, tensor):
+        """Save tensor to disk asynchronously and return file path with thread-safe operations"""
+        # Generate unique file path
+        self.idx += 1
+        file_path = os.path.join(self.temp_dir, f"{self.idx:06d}-{uuid.uuid4()}.pt")
```
NanoCode012: Do we need idx in file name? uuid4 should be unique enough 
```suggestion
        file_path = os.path.join(self.temp_dir, f"{uuid.uuid4()}.pt")
```

src/axolotl/utils/gradient_checkpointing/offload_disk.py
```diff
@@ -0,0 +1,535 @@
...
+            # Clean up the file even on error
+            manager.cleanup_tensor(file_path)
+            raise
+
+    @staticmethod
+    def cleanup():
+        """Clean up the offload manager"""
+        if Disco._manager is not None:
+            Disco._manager.cleanup()
+            Disco._manager = None
```
NanoCode012: When is cleanup called? Should it be called at end of batch/epoch/training?

src/axolotl/utils/schemas/config.py
```diff
@@ -178,9 +178,9 @@ class AxolotlInputConfig(
 
     # torch_dtype: torch.dtype | None
 
-    gradient_checkpointing: Literal["unsloth", "offload"] | bool | None = Field(
-        default=False
-    )
+    gradient_checkpointing: (
+        Literal["unsloth", "offload", "offload_disk"] | bool | None
```
NanoCode012: Could we add to `config.qmd` as well?

NanoCode012 (commented):
Would this work for multi-node? Since tmp dir is not shared between, would there be any issues?

How about with deepspeed/fsdp activation offload? Would there be any conflicts?

src/axolotl/utils/gradient_checkpointing/offload_disk.py
winglian: Idx makes debugging easier. You can see the layers come through in the correct order for forward and backwards passes.

src/axolotl/utils/gradient_checkpointing/offload_disk.py
winglian: I don't think that's the concern of the trainer.

src/axolotl/utils/gradient_checkpointing/offload_disk.py
```diff
@@ -0,0 +1,531 @@
+"""
+DISCO - DIsk-based Storage and Checkpointing with Optimized prefetching
```
djsaunde, line 2: Is this your name for this method? A quick google search doesn't turn anything up

src/axolotl/utils/gradient_checkpointing/offload_disk.py
```diff
@@ -0,0 +1,531 @@
...
+
+class DiskOffloadManager:
+    """
+    Manages offloaded tensors and handles prefetching in a separate thread.
+    Includes synchronization to prevent race conditions.
+    """
+
+    def __init__(self, prefetch_size=3, prefetch_to_gpu=True, save_workers=4):
+        self.temp_dir = tempfile.mkdtemp(prefix="disco_")
+        # logger.debug(f"Created temporary directory at {self.temp_dir}")
```
djsaunde: old log

djsaunde: there's a few of these

src/axolotl/utils/gradient_checkpointing/offload_disk.py
```diff
@@ -0,0 +1,531 @@
...
+    """
+    DiskOffloadedGradientCheckpointer: Tensor Efficient Management with Prefetched Offloading
+    Advanced disk-based gradient checkpointer with prefetching.
+    """
+
+    # Shared manager instance across all checkpointing operations
+    _manager = None
+
+    @staticmethod
+    def get_manager(prefetch_size=1, prefetch_to_gpu=True, save_workers=4):
```
djsaunde: Could rename to `get_instance` to match our callback manager impl

src/axolotl/utils/gradient_checkpointing/offload_disk.py
djsaunde, line 1: nit: add typing, docstrings

src/axolotl/utils/gradient_checkpointing/offload_disk.py
```diff
@@ -0,0 +1,531 @@
...
+            self.tensor_paths.append(file_path)
+
+        # Acquire semaphore to limit concurrent save operations
+        self.save_semaphore.acquire()  # pylint: disable=consider-using-with
+        # Queue tensor for saving in background
+        self.save_queue.put((tensor.detach(), file_path))
+
+        return file_path
+
+    def wait_for_save(self, file_path, timeout=None):
```
djsaunde: I notice we return a success boolean, but it's not used anywhere. Should it be? What if this fails?

djsaunde (approved):
really interesting stuff! just some nits and comments

PR #2664, opened 2025-05-12
run codecov action at end of CI; only_pulls: true

djsaunde:
# Description

## Summary by CodeRabbit

- **Chores**
  - Updated CI workflow to aggregate and upload test coverage data in a single step after all tests complete, improving reliability of coverage reporting.
  - Modified scripts to retrieve and save coverage reports from test runs, ensuring coverage data is consistently collected.
  - Removed direct coverage uploads from individual test jobs to streamline the process.


.github/workflows/tests.yml
```diff
@@ -207,13 +207,12 @@ jobs:
...
+      - name: Upload coverage artifacts
+        uses: actions/upload-artifact@v4
         with:
-          token: ${{ secrets.CODECOV_TOKEN }}
-          files: ./coverage.xml
-          flags: unittests,pytorch-${{ matrix.pytorch_version }}
-          fail_ci_if_error: false
+          name: coverage-${{ matrix.pytorch_version }}
+          path: ./coverage.xml
+          retention-days: 1
```
NanoCode012, line 114: Not too clear about GH action, but are these artifacts shared across runs?

Do we need to make sure that the different PR CIs don't download the same artifacts if they're running simultaneously (maybe by adding pr or hash to name)?

.github/workflows/tests.yml
```diff
@@ -336,6 +335,14 @@ jobs:
...
           modal run cicd.e2e_tests
 
+      - name: Upload coverage artifacts
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: coverage-e2e-1st-${{ github.run_id }}
+          path: ./e2e-coverage.xml
+          retention-days: 1
+
```
winglian, line 243: Does removing this work as intended? Since this runs in Modal, it won't get uploaded to GitHub.

PR #2670, opened 2025-05-14
Add missing init file to liger plugin

BitPhinix:
Same as https://github.com/axolotl-ai-cloud/axolotl/pull/2642 but for the liger plugin


NanoCode012 (approved):
Thanks for the catch!

PR #2675, opened 2025-05-14
Fix: Make MLflow config artifact logging respect hf_mlflow_log_artifa…

C080:
# Description

Modified the `SaveAxolotlConfigtoMlflowCallback` class to check the `HF_MLFLOW_LOG_ARTIFACTS` environment variable before attempting to log artifacts. This ensures that the `hf_mlflow_log_artifacts: false` setting in the config file properly prevents all artifact logging attempts.

## Motivation and Context

The MLflow callback was attempting to log config file artifacts regardless of the `hf_mlflow_log_artifacts` setting. This caused errors for users without proper GCS permissions when using MLflow with a remote tracking server. 

When setting `hf_mlflow_log_artifacts: false` in the config file, users would expect all artifact logging to be disabled, but the callback was still trying to log the config file, resulting in permission errors.

## How has this been tested?

Tested with a remote MLflow server configuration where I didn't have GCS permissions. When setting `hf_mlflow_log_artifacts: false` in the config file, the training now proceeds without any permission errors.

Testing environment:
- Remote MLflow server with GCS backend
- User without storage.objects.create permissions
- QLora training on Llama-3.2-1B model

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## Summary by CodeRabbit

- **New Features**
  - Added the ability to control whether the Axolotl config file is logged as an MLflow artifact during training via the `HF_MLFLOW_LOG_ARTIFACTS` environment variable.

- **Chores**
  - Improved logging to inform users when artifact logging is skipped based on environment settings.

### DIFF
--- a/src/axolotl/utils/callbacks/mlflow_.py
+++ b/src/axolotl/utils/callbacks/mlflow_.py
@@ -1,5 +1,6 @@
 """MLFlow module for trainer callbacks"""
 
+import os
 import logging
 from shutil import copyfile
 from tempfile import NamedTemporaryFile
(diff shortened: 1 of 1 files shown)


src/axolotl/utils/callbacks/mlflow_.py
```diff
@@ -32,14 +33,17 @@ def on_train_begin(
...
             try:
-                with NamedTemporaryFile(
-                    mode="w", delete=False, suffix=".yml", prefix="axolotl_config_"
-                ) as temp_file:
-                    copyfile(self.axolotl_config_path, temp_file.name)
-                    mlflow.log_artifact(temp_file.name, artifact_path="")
-                    LOG.info(
-                        "The Axolotl config has been saved to the MLflow artifacts."
-                    )
+                if os.getenv("HF_MLFLOW_LOG_ARTIFACTS", "FALSE").upper() in ["TRUE", "1", "YES"]:
```
winglian: nit, might be useful to have a utility function for truth environment value checks.

winglian (approved):

NanoCode012:
Would it be better to set the condition here instead of within the callback?

https://github.com/axolotl-ai-cloud/axolotl/blob/c0a0c7534cc3a9a2b4c0ad6235fd5756f35ec17b/src/axolotl/core/trainer_builder.py#L180-L189

PR #2676, opened 2025-05-14
GRPO fixes (peft)

winglian:
Fixes the double-wrapping bug we were seeing with GRPO + PEFT. This should also fix the timeout issue we were seeing in CI with GRPO.

## Summary by CodeRabbit

- **Bug Fixes**
	- Re-enabled two previously skipped tests for improved test coverage.
	- Updated test environment variables for VLLM server processes.

- **Refactor**
	- Removed an internal method related to model weight transfer in the GRPO trainer.

- **Chores**
	- Increased CPU allocation for a CI/CD test function to enhance execution resources.

- **New Features**
	- Adjusted trainer configuration to conditionally exclude certain settings for specific reinforcement learning types.

### DIFF
--- a/src/axolotl/core/trainer_builder.py
+++ b/src/axolotl/core/trainer_builder.py
@@ -1170,7 +1170,7 @@ class HFRLTrainerBuilder(TrainerBuilderBase):
         if self.eval_dataset:
             trainer_kwargs["eval_dataset"] = self.eval_dataset
         if self.cfg.adapter and self.peft_config:
-            if self.cfg.rl != "grpo":
+            if self.cfg.rl is not RLType.GRPO:
                 trainer_kwargs["peft_config"] = self.peft_config
         if self.cfg.precompute_ref_log_probs is not None:
             trainer_kwargs["precompute_ref_log_probs"] = (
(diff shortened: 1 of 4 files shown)


src/axolotl/core/trainer_builder.py
```diff
@@ -1170,7 +1170,8 @@ def build(self, total_num_steps):
         if self.eval_dataset:
             trainer_kwargs["eval_dataset"] = self.eval_dataset
         if self.cfg.adapter and self.peft_config:
-            trainer_kwargs["peft_config"] = self.peft_config
+            if self.cfg.rl != "grpo":
```
mhenrichsen: Maybe use the enums from RLType?

src/axolotl/core/trainers/grpo/trainer.py
```diff
@@ -52,62 +49,12 @@
     # pylint: disable=unused-import
     from peft import PeftConfig
 
-if is_deepspeed_available():
-    import deepspeed
-
 
 class AxolotlGRPOTrainer(RngLoaderMixin, SchedulerMixin, GRPOTrainer):
```
salmanmohammadi, line 53: Wait do we need this class at all now?

djsaunde: The only thing added is the `_tag_names` now, which is what we do for some other RL trainers (not sure why we do this?).

winglian: it was an import from the code we removed.

winglian: the axolotl tag gets added to model cards so if they get uploaded, they show up under https://huggingface.co/models?library=axolotl

salmanmohammadi (approved):

PR #2686, opened 2025-05-16
SP dataloader patching + removing custom sampler / dataloader logic

djsaunde:
# Description

This PR adds a couple of patches for `accelerate` functions having to do with dataloader preparation for distributed training.

It also removes the custom logic for train / eval samplers and dataloaders that we previously needed to distribute data properly for sequence parallelism (SP).

Unfortunately this can't be applied to GRPO since the TRL impl has it's own custom sampler, but should be applicable for the other RL trainers.

## Motivation and Context

Less code to maintain (axolotl trainer, RL trainers custom logic). Also, puts us one step closer to upstreaming these changes by demo-ing how it could work.

## How has this been tested?

- [x] SFT manual testing
- [x] RL manual testing
- [x] pytests passing

## Summary by CodeRabbit

- **Refactor**
  - Removed sequence parallelism support from core training classes, mixins, and configuration for a simplified training workflow.
- **Bug Fixes**
  - Improved handling of positional and keyword arguments in sequence parallel context management.
- **Chores**
  - Removed temporary custom dataset preparation and evaluation logic in DPO trainer, reverting to standard behavior.
- **Documentation**
  - Updated sequence parallelism configuration examples and clarified ring attention function options.
  - Removed inline installation instructions and configuration snippets for sequence parallelism, directing users to dedicated guides.
- **New Features**
  - Added patches to enhance ring attention integration with data loader and device mesh setup.
  - Introduced ring attention monkeypatch module for managing distributed sequence parallelism groups and Flash Attention substitution.


djsaunde:
DPO (SP = 1 vs. SP = 2): image

ORPO (SP = 1 vs. SP = 2): image

KTO: tried `examples/llama-3/qlora-1b-kto.yaml` but backward pass is broken (?) even without SP enabled.

Other RL methods (SimPO, IPO) not tried since these are simple variants on already tried methods.

Diffs seem to be due to inter-run randomness? Not sure why it isn't exactly reproducible run to run (I checked this by running SP = 1 runs back to back; training metrics differ in this case).

docs/multi-gpu.qmd
```diff
@@ -94,10 +94,13 @@ Your Axolotl YAML config should contain the following lines:
 
 ```{.yaml}
 sequence_parallel_degree: 4  # Split each sequence into 4 parts, one per GPU
-flash_attention: true  # Required with sequence parallelism
```
NanoCode012, line 97: Do we not need FA anymore?

docs/multi-gpu.qmd
```diff
@@ -94,10 +94,13 @@ Your Axolotl YAML config should contain the following lines:
...
 
 # Optional; strides across the key dimension. Larger values use more memory but will make training faster.
 heads_k_stride: 1
+
+# Optional; one of "varlen_llama3" or "batch_ring". Defaults to
+# "varlen_llama3" when `sample_packing: true`, and "batch_ring" otherwise.
+ring_attn_func:
```
NanoCode012: Should we add a quick explanation on the difference?
```suggestion
# Optional; Choices are 
# - "varlen_llama3": ...
# - "batch_ring":  ...
# Defaults to "varlen_llama3" when `sample_packing: true`, otherwise "batch_ring"
ring_attn_func:
```

NanoCode012: Hm, maybe this should actually go in `config.qmd` instead. We seem to have three places (this, SP, config qmd) with this info. We may not want to duplicate this over several docs to prevent having stale docs. Should this just link to the SP guide?

src/axolotl/monkeypatch/attention/ring_attn/patch.py
```diff
@@ -129,3 +145,76 @@ def update_ring_attn_params(position_ids: torch.Tensor | None):
...
+    def _prepare_device_mesh(self):
+        """Prepare the device mesh for distributed training. The dataloader will
+        determine how to load data based on the device mesh.
+        """
+        if self.state.torch_tp_plugin:
+            return self.state.torch_tp_plugin.torch_device_mesh
+        if (
+            self.distributed_type == accelerate.accelerator.DistributedType.DEEPSPEED
+            and hasattr(self.state, "ds_device_mesh")
+        ):
```
NanoCode012: Not so familiar with this, but since there's a specific handling for DS. Do we need for FSDP?

src/axolotl/monkeypatch/attention/ring_attn/patch.py
```diff
@@ -129,3 +145,76 @@ def update_ring_attn_params(position_ids: torch.Tensor | None):
...
+        mesh_shape = (
+            world_size // sequence_parallel_degree,
+            sequence_parallel_degree,
+        )
+        device_ids = list(range(world_size))
+        return dist.DeviceMesh(
+            "cuda",
+            torch.tensor(device_ids).reshape(mesh_shape),
+            mesh_dim_names=["dp", "sp"],
+        )
```
NanoCode012: Would this be compatible with multi-node?

By reshaping it as such, does it mean, `mesh_shape[0] -> dp` and `mesh_shape[1] -> sp`?

src/axolotl/utils/ctx_managers/sequence_parallel.py
```diff
@@ -206,12 +207,25 @@ def __init__(
...
-                self.apply_sequence_parallelism(batch=kwargs)
+            # Get parameter names from the model's forward function
+            forward_params = list(
+                inspect.signature(self.models[0].forward).parameters.keys()
             )
 
-            return args, kwargs
+            updated_kwargs = kwargs.copy()
+            for i, arg in enumerate(args):
+                if i < len(forward_params):
```
NanoCode012, line 217: What does this check do? Is it for positional args to pass through?

NanoCode012 (commented):
This looks very clean! I just left a few comments on places I wasn't sure about.

docs/sequence_parallelism.qmd
```diff
@@ -67,9 +67,11 @@ sequence_len: 8192
 ...
 
 sequence_parallel_degree: 4  # Split each sequence into 4 parts, one per GPU
-flash_attention: true  # Required with sequence parallelism
 # Optional; strides across the key dimension. Larger values use more memory but should make training faster.
 heads_k_stride: 1
+# Optional; one of "varlen_llama3" or "batch_ring". Defaults to
```
salmanmohammadi, line 72: Is `batch_ring` compatible with sample packing?

src/axolotl/monkeypatch/attention/ring_attn/patch.py
```diff
@@ -1,11 +1,16 @@
-"""
```
salmanmohammadi, line 1: nit: we're already in the `monkeypatch` module, maybe just move this folder to `ring_attn.py`?

src/axolotl/monkeypatch/attention/ring_attn/patch.py
```diff
@@ -18,24 +23,39 @@
...
+            submesh_tp_size = 1
+            if "tp" in torch_device_mesh.mesh_dim_names:
+                submesh_tp_size = torch_device_mesh["tp"].size()
+            if "dp" in torch_device_mesh.mesh_dim_names:
+                submesh_dp_size = torch_device_mesh["dp"].size()
+            if "fsdp" in torch_device_mesh.mesh_dim_names:
+                submesh_fsdp_size = torch_device_mesh["fsdp"].size()
+            process_index = process_index // submesh_tp_size"""
+
+NEW_PREPARE_DATALOADER_CODE = """            submesh_fsdp_size = 1
```
salmanmohammadi: This seems like a much more convenient solution

salmanmohammadi: Would we also be able to re-use this for torch native SP? I think we'd need to use `cp` for the mesh dim in that case.

djsaunde: Yeah, I was thinking the next logical step was to add torch SP support. You're right about the `cp` naming.

docs/sequence_parallelism.qmd
djsaunde: No, varlen alone implies sample packing

src/axolotl/utils/ctx_managers/sequence_parallel.py
djsaunde: Some trainers pass e.g. `input_ids` as an arg (positional) rather than a kwarg. This logic handles that case.

src/axolotl/monkeypatch/attention/ring_attn/patch.py
djsaunde: > Would this be compatible with multi-node?

I don't really know the implications of multinode in this case, but my guess would be yes?

> By reshaping it as such, does it mean, mesh_shape[0] -> dp and mesh_shape[1] -> sp?

Yup.

src/axolotl/monkeypatch/attention/ring_attn/patch.py
djsaunde: Not too familiar with the requirements of FSDP, maybe @winglian can weigh in?

docs/multi-gpu.qmd
djsaunde: Yeah, I agree, I'll link to the SP guide.

docs/multi-gpu.qmd
djsaunde: Oops, that was a mistake.

djsaunde: Maybe aspirationally 😆 native torch SP support coming soon!

src/axolotl/monkeypatch/attention/ring_attn/patch.py
djsaunde: Anyways, this logic is just ripped from `accelerate` and updated to include the sequence parallel case.

Of the top of my head, I don't think FSDP requires any special data distribution apart from that done by DDP.

salmanmohammadi (approved):

PR #2691, opened 2025-05-19
add support for base image with uv

winglian:
We don't officially have a way to reproducibly support uv yet. Should be simple, but having a docker base image to test against will go a long way in making sure we have stable builds when using uv.

## Summary by CodeRabbit

- **New Features**
  - Introduced a new Docker image for a CUDA-enabled Python environment with PyTorch and additional Python packages, built using the "uv" toolchain.
  - Added a new installation method in the documentation for setting up the environment with the `uv` tool, including instructions for creating virtual environments and installing dependencies.

- **Chores**
  - Updated GitHub Actions workflow to explicitly specify Dockerfiles in the build matrix, improving clarity and maintainability for Docker image builds.
  - Updated default PyTorch and CUDA versions in CI/CD GPU test configurations.
  - Revised documentation to reflect updated Docker image tags and minimum PyTorch version requirements.
  - Fixed a typographical error in the welcome message displayed to users.

### DIFF
--- a/.github/workflows/base.yml
+++ b/.github/workflows/base.yml
@@ -28,42 +28,50 @@ jobs:
             python_version: "3.11"
             pytorch: 2.5.1
             torch_cuda_arch_list: "7.0 7.5 8.0 8.6 8.7 8.9 9.0+PTX"
+            dockerfile: "Dockerfile-base"
           - cuda: "124"
             cuda_version: 12.4.1
             cudnn_version: ""
             python_version: "3.11"
             pytorch: 2.6.0
             torch_cuda_arch_list: "7.0 7.5 8.0 8.6 8.7 8.9 9.0+PTX"
+            dockerfile: "Dockerfile-base"
           - cuda: "126"
             cuda_version: 12.6.3
             cudnn_version: ""
             python_version: "3.11"
             pytorch: 2.6.0
             torch_cuda_arch_list: "7.0 7.5 8.0 8.6 8.7 8.9 9.0+PTX"
+            dockerfile: "Dockerfile-base"
           - cuda: "126"
             cuda_version: 12.6.3
             cudnn_version: ""
             python_version: "3.11"
             pytorch: 2.7.0
             torch_cuda_arch_list: "7.0 7.5 8.0 8.6 8.7 8.9 9.0+PTX"
+            dockerfile: "Dockerfile-base"
           - cuda: "128"
             cuda_version: 12.6.3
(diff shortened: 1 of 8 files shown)


docs/installation.qmd
```diff
@@ -37,6 +37,36 @@ installed) in order not to clobber it, and so that we set the correct version of
...
+export UV_TORCH_BACKEND=cu126
+uv venv --no-project --relocatable
+source .venv/bin/activate
+```
+
+Install PyTorch
+```{.bash}
+uv pip install packaging setuptools wheel
+uv pip install torch==2.6.0
+uv pip install awscli pydantic
```
NanoCode012, line 63:
```suggestion
uv pip install packaging setuptools wheel
uv pip install torch==2.6.0
```

Do we need to add the third line? It should be installed with axolotl?

Edit: I saw that we include this in docker base, so this comment should be ignored.

docs/installation.qmd
```diff
@@ -37,6 +37,36 @@ installed) in order not to clobber it, and so that we set the correct version of
...
+```
+
+Install axolotl from PyPi
+```{.bash}
+uv pip install --no-build-isolation axolotl[deepspeed,flash-attn]
+# optionally install with vLLM if you're using torch==2.6.0 and want to train w/ GRPO
```
NanoCode012, line 70:
```suggestion

# optionally install with vLLM if you're using torch==2.6.0 and want to train w/ GRPO
```

Add spacing for easy reading

.github/workflows/base.yml
```diff
@@ -28,42 +28,96 @@ jobs:
...
+      matrix:
+        include:
+          - cuda: "126"
+            cuda_version: 12.6.3
             cudnn_version: ""
             python_version: "3.11"
-            pytorch: next
+            pytorch: 2.6.0
             torch_cuda_arch_list: "7.0 7.5 8.0 8.6 8.7 8.9 9.0+PTX"
+            dockerfile: "Dockerfile-uv-base"
```
NanoCode012: Could we not combine this with the above `build-base` (by adding another field `docker_image` in the matrix)? The only difference seems to the Docker image path to push to. We can also deprecate pushing to your personal org.

winglian: yeah, but the docker image tag to push to is already convoluted, and I didn't want to add more conditionals to that.

NanoCode012: That's fair point

PR #2696, opened 2025-05-20
feat: do not find turn indices if turn is not trainable

NanoCode012:
# Description

We were wasting compute trying to find the turns for indices we won't even train on. This reduces at least 1/2 computations done per sample (assuming assistant turns make up to 50% cases).

Secondly, the only case where we need the indices is when `train_on_eot` / `train_on_eos` is set to ALL, so we will do full computation.

Introduces a new warning if the last turn is not trainable (which is unusual):
```
                        "Last turn is not trainable, skipping having to find the turn indices. "
                        "This may cause incorrect last EOT/EOS token to be unmasked."
```

## How has this been tested?

Ran on dummy dataset with benchmark below: https://github.com/axolotl-ai-cloud/axolotl/pull/2696#issuecomment-2893971301

Need eyes to double check.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of non-trainable chat turns to prevent unnecessary processing and potential masking issues, with warnings logged if skipping may affect the last turn.

### DIFF
--- a/src/axolotl/prompt_strategies/chat_template.py
+++ b/src/axolotl/prompt_strategies/chat_template.py
@@ -425,7 +425,10 @@ class ChatTemplateStrategy(PromptTokenizingStrategy):
             LOG.debug(f"Should train: {should_train}")
 
             # turn not trainable, skip having to find the turn indices
-            if not should_train:
+            # unless last turn and train_on_eos/train_on_eot is all
+            if not should_train and (
+                self.train_on_eos != "all" and self.train_on_eot != "all"
+            ):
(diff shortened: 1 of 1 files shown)


salmanmohammadi (approved):
Awesome. Did you happen to verify the speedups?

NanoCode012:
On a 100k dummy tool dataset with 6 turns of short content, the time taken to tokenize went from a (three run) average of 83.3s -> 47.6s (decreased by about 42.8%). This change will be more impactful for those with even longer multi-turn conversation.

PR #2706, opened 2025-05-22
feat(doc): note lora kernel incompat with RLHF

NanoCode012:
# Description

>  Error 2: Model must be a PeftModelForCausalLM when enabling LoRa kernel patches. Since Gemma3 is not inherited from PeftModelForCausalLM, how can i use these monkey patches?

- from discord

Let's add to doc about this limitation for now. Should we add validation too?

## Summary by CodeRabbit

- **Documentation**
  - Added a note clarifying that LoRA kernels currently support only Supervised Fine-Tuning (SFT) and are not compatible with Reinforcement Learning with Human Feedback (RLHF) training.

### DIFF
--- a/docs/lora_optims.qmd
+++ b/docs/lora_optims.qmd
@@ -84,6 +84,10 @@ lora_qkv_kernel: true
 lora_o_kernel: true
 ```
 
+::: {.callout-note}
+Currently, LoRA kernels are not supported for RLHF training, only SFT.
+:::
+
 ## Requirements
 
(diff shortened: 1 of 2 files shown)


djsaunde:
Yeah, I'd recommend validating. I lean towards failing fast / being explicit about what's causing failures.

Do you know why gemma 3 can't be patched? I should look into this.

djsaunde (approved):
Approving but would appreciate validation :)

PR #2710, opened 2025-05-23
Fix Mistral chat template (mistral_v7_tekken)

mashdragon:
Per https://huggingface.co/mistralai/Mistral-Small-3.1-24B-Instruct-2503/commit/4b8dd8aae705887db5295fcbff4aedbb92d682eb#d2h-482763

# Description
The EOS token isn't set properly in the old unfixed chat template included in axolotl.

## Motivation and Context
You can't train Mistral instruction models properly without this change since the EOS token is not inserted after the assistant response.

## How has this been tested?

Ran on a dataset and observed the output of data preprocessing.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of system and message content types in chat, ensuring better support for both string and list formats.
  - Enhanced robustness for user, system, and assistant message processing, especially for messages containing multiple content blocks or images.


NanoCode012 (approved):
Thank you!

PR #2711, opened 2025-05-23
add support for Select Activation Checkpointing

winglian:
see https://pytorch.org/blog/activation-checkpointing-techniques/

sweeps in progress: https://wandb.ai/axolotl-ai/sac-sweeps?nw=nwuserwingaxolotl


salmanmohammadi:
The results are looking pretty good. I wonder if we need to investigate whether this interacts with flex compilation?

PR #2721, opened 2025-05-25
add two checks to handle legacy format interleaved multimodal ds

sumo43:
# Description

add two checks to handle legacy format interleaved ds

## Motivation and Context

if messages key exists but is None, the legacy conversion never happens and this breaks the processing later on.

## How has this been tested?

tried training a model with 2 interleaved datasets.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved handling of missing or null message and image data to prevent errors and ensure more reliable processing.
	- Added warnings when multiple images are detected per sample, guiding users on best practices for dataset formatting.


NanoCode012:
Thanks for the PR! Just letting CI to run :)

NanoCode012 (approved):

PR #2755, opened 2025-06-04
remove unused field for chat_template.default for DPO training

timofey:
# Description

DPO trainer expects only prompt+chosen+rejected columns present in the final dataset. However, the `chat_template.default` implementation passes an extra column `messages` which causes validations to fail.

## Motivation and Context

Fixes issue #2415 

## How has this been tested?

Tested using sample config:

```yaml
rl: dpo

datasets:
  - path: "..../dataset.jsonl"
    type: chat_template.default
    field_messages: "messages"
    field_chosen: "chosen"
    field_rejected: "rejected"
    message_property_mappings:
      role: role
      content: content
    roles:
      user: [ "user" ]
      assistant: [ "assistant" ]
      system: [ "system" ]
```

then by running `axolotl preprocess` and `axolotl train` commands.

Without the fix, it fails with the following errors:

```
KeyError: "Invalid keys in the example: {'messages', 'prompt', 'rejected', 'chosen'}"
```

## Summary by CodeRabbit

- **New Features**
  - The system now provides additional metadata indicating which columns should be removed during processing.
- **Bug Fixes**
  - Updated dataset tests to reflect correct feature presence and absence, improving data validation.

### DIFF
--- a/src/axolotl/prompt_strategies/dpo/chat_template.py
+++ b/src/axolotl/prompt_strategies/dpo/chat_template.py
@@ -91,4 +91,4 @@ def default(
 
         return result
 
-    return transform_fn, {"remove_columns":[field_messages]}
+    return transform_fn, {"remove_columns": [field_messages]}
--- a/src/axolotl/prompt_strategies/dpo/chat_template.py
+++ b/src/axolotl/prompt_strategies/dpo/chat_template.py
@@ -91,4 +91,4 @@ def default(
 
         return result
 
-    return transform_fn
+    return transform_fn, {"remove_columns":[field_messages]}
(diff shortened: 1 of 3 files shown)


winglian (approved):

NanoCode012 (approved):

winglian (changes requested):
The caller of the transform_fn is going to need to unpack the updated return (tuple)

winglian:
I think I fixed the linting issues and tests that use the updated return values

timofey:
oh thanks @winglian ! I'll take a stab at it tonight as well, looks like some of the tests are still failing

timofey:
@winglian can you run the tests on the pipeline again please? Adjusted the failed test, since the "conversation" field is removed now

winglian (approved):

PR #2760, opened 2025-06-05
FSDP1 -> FSDP2

salmanmohammadi:
Successfully loading Qwen 32B onto 2xA40 40GB
- [x] Enable instantiation of model on meta-device and load sharded state dict (for training models which would OOM on 1 GPU)
- [x] Fix FSDP2 FP32 upcast
- [x] Separetely wrap the LoRA modules when PEFT is enabled
- [x] Make sure QLoRA works
- [x] Disable DDP by default
- [ ] Enable resuming from sharded checkpoints for FSDP2 (moved to follow-up)
```bash
                                                                                                                                                                                                                        
[2025-06-07 01:29:13,357] [INFO] [axolotl.loaders.tokenizer.load_tokenizer:278] [PID:38522] [RANK:0] No Chat template selected. Consider adding a chat template for easier inference.                                   
[2025-06-07 01:29:13,357] [INFO] [axolotl.utils.data.sft.load_tokenized_prepared_datasets:270] [PID:38522] [RANK:0] Unable to find prepared dataset in last_run_prepared/ff73ba333522a3cc59befd3e37835d10               
[2025-06-07 01:29:13,357] [INFO] [axolotl.utils.data.sft.load_tokenized_prepared_datasets:271] [PID:38522] [RANK:0] Loading raw datasets...                                                                             
[2025-06-07 01:29:13,357] [WARNING] [axolotl.utils.data.sft.load_tokenized_prepared_datasets:273] [PID:38522] [RANK:0] Processing datasets during training can lead to VRAM instability. Please pre-process your dataset
.                                                                                                                                                                                                                       
[2025-06-07 01:29:13,357] [INFO] [axolotl.utils.data.sft.load_tokenized_prepared_datasets:280] [PID:38522] [RANK:0] No seed provided, using default seed of 42                                                          
[rank1]:[W607 01:29:14.144191148 ProcessGroupNCCL.cpp:4715] [PG ID 0 PG GUID 0 Rank 1]  using GPU 1 as device used by this process is currently unknown. This can potentially cause a hang if this rank to GPU mapping i
s incorrect. You can pecify device_id in init_process_group() to force use of a particular device.                                                                                                                      
[2025-06-07 01:29:14,245] [INFO] [axolotl.utils.data.sft.get_dataset_wrapper:485] [PID:38522] [RANK:0] Loading dataset: yahma/alpaca-cleaned with base_type: alpaca and prompt_style: None                              
[2025-06-07 01:29:14,623] [INFO] [axolotl.utils.data.utils.drop_long_seq_in_dataset:177] [PID:38522] [RANK:0] min_input_len: 32                                                                                         
[2025-06-07 01:29:14,623] [INFO] [axolotl.utils.data.utils.drop_long_seq_in_dataset:179] [PID:38522] [RANK:0] max_input_len: 1018                                                                                       
[2025-06-07 01:29:15,181] [WARNING] [axolotl.utils.data.utils.drop_long_seq_in_dataset:207] [PID:38522] [RANK:0] Dropped 765 long samples from dataset                                                                  
[2025-06-07 01:29:16,019] [INFO] [axolotl.utils.data.sft.load_tokenized_prepared_datasets:350] [PID:38522] [RANK:0] Saving merged prepared dataset to disk... last_run_prepared/ff73ba333522a3cc59befd3e37835d10        
Saving the dataset (1/1 shards): 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 50995/50995 [00:00<00:00, 338435.51 examples/s]
[rank0]:[W607 01:29:16.279282472 ProcessGroupNCCL.cpp:4715] [PG ID 0 PG GUID 0 Rank 0]  using GPU 0 as device used by this process is currently unknown. This can potentially cause a hang if this rank to GPU mapping i
s incorrect. You can pecify device_id in init_process_group() to force use of a particular device.                                                                                                                      
[2025-06-07 01:29:20,598] [INFO] [axolotl.utils.samplers.multipack.calc_min_len:428] [PID:38522] [RANK:0] gather_len_batches: [1165, 1165]                                                                              
[2025-06-07 01:29:20,609] [INFO] [axolotl.utils.data.sft.prepare_dataset:187] [PID:38522] [RANK:0] Maximum number of steps set at 1165                                                                                  
[2025-06-07 01:29:21,264] [INFO] [axolotl.loaders.tokenizer.load_tokenizer:278] [PID:38522] [RANK:0] No Chat template selected. Consider adding a chat template for easier inference.                                   
[2025-06-07 01:29:21,448] [INFO] [axolotl.integrations.liger.pre_model_load:87] [PID:38522] [RANK:0] Applying LIGER to qwen3 with kwargs: {'rope': True, 'cross_entropy': None, 'fused_linear_cross_entropy': True, 'rms
_norm': True, 'swiglu': True}                                                                                                                                                                                           
Loading checkpoint shards: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 17/17 [00:00<00:00, 281.16it/s]
Loading checkpoint shards: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 17/17 [00:00<00:00, 76.22it/s]
[2025-06-07 01:29:22,165] [INFO] [axolotl.loaders.model._configure_embedding_dtypes:309] [PID:38522] [RANK:0] Converting modules to torch.bfloat16                                                                      
[2025-06-07 01:29:22,173] [INFO] [axolotl.loaders.model.log_gpu_memory_usage:108] [PID:38522] [RANK:0] Memory usage after model load 0.000GB (+0.002GB cache, +0.984GB misc)                                            
[2025-06-07 01:29:23,957] [INFO] [axolotl.train.save_initial_configs:413] [PID:38522] [RANK:0] Pre-saving tokenizer to ./outputs/qat_out/...                                                                            
[2025-06-07 01:29:24,225] [INFO] [axolotl.train.save_initial_configs:416] [PID:38522] [RANK:0] Pre-saving model config to ./outputs/qat_out/...                                                                         
[2025-06-07 01:29:24,242] [INFO] [axolotl.train.execute_training:217] [PID:38522] [RANK:0] Starting trainer...                                                                                                          
[2025-06-07 01:29:25,280] [INFO] [axolotl.utils.samplers.multipack.calc_min_len:428] [PID:38522] [RANK:0] gather_len_batches: [1165, 1165]                                                                              
[2025-06-07 01:29:25,568] [INFO] [axolotl.monkeypatch.accelerate.fsdp2.fsdp2_load_full_state_dict:29] [PID:38522] [RANK:0] Broadcasting full state dict to all ranks...                                                 
[2025-06-07 01:29:43,111] [INFO] [axolotl.monkeypatch.accelerate.fsdp2.log_gpu_memory_usage:108] [PID:38522] [RANK:0] Memory usage after broadcasting full state dict 32.626GB (+0.501GB cache, +0.984GB misc)   
```

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Added support for Fully Sharded Data Parallel version 2 (FSDP2) with enhanced configuration options, migration guides, and validation.
  * Introduced comprehensive end-to-end tests for FSDP1 and FSDP2 multi-GPU training scenarios.
  * Improved model preparation and state dict loading for FSDP2, including LoRA adapter and CPU RAM efficient loading support.

* **Bug Fixes**
  * Strengthened validation and error handling for FSDP configurations to ensure compatibility with quantization and reinforcement learning methods.
  * Corrected configuration migration to rename and flatten FSDP config keys properly.

* **Documentation**
  * Clarified FSDP1 deprecation status, recommended FSDP2 usage, and added detailed migration instructions.
  * Updated configuration examples to showcase the new dictionary-based FSDP2 format and deprecated options.

* **Refactor**
  * Simplified and unified FSDP-related logic, configuration handling, and logging throughout the codebase.
  * Removed legacy PPO trainer code and associated imports.

* **Chores**
  * Added extensive tests for FSDP configuration migration and multi-GPU training workflows.
  * Enhanced logging consistency and added type annotations in utility functions.
  * Added a test utility decorator to require PyTorch 2.7.0 or higher for relevant tests.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


docs/multi-node.qmd
```diff
@@ -40,6 +40,8 @@ use_cpu: false
 
 Configure your model to use FSDP in the Axolotl yaml. For example:
 ```yaml
+
+asdasd
```
joecummings: well said

src/axolotl/monkeypatch/accelerate/fsdp2.py
```diff
@@ -191,16 +246,161 @@ def get_state_dict(self, model, unwrap=True):
...
+        from peft.tuners.lora import LoraLayer 
+    else:
+        LoraLayer = None
+    
+    auto_wrap_policy = fsdp2_prepare_auto_wrap_policy(fsdp2_plugin, auto_wrap_policy_type, model)
+    if auto_wrap_policy is not None:
+        for module in get_module_children_bottom_up(model)[:-1]:
+            ignored_params = []
+            if is_peft_model and isinstance(module, LoraLayer):
+                for active_adapter in module.active_adapters:
```
salmanmohammadi: Credit to @ebsmothers for this idea

src/axolotl/core/trainers/trl.py
```diff
@@ -16,64 +16,6 @@
 from axolotl.core.trainers.mixins.scheduler import SchedulerMixin
 
 
-class TRLPPOTrainer(PPOTrainer):
```
salmanmohammadi, line 19: This just isn't used right?

djsaunde: Don't think so

salmanmohammadi:
@coderabbitai summary

docs/multi-gpu.qmd
```diff
@@ -66,9 +66,69 @@ Start from Stage 1 -> Stage 2 -> Stage 3.
...
+
+```{.yaml}
+fsdp_version: 2
+fsdp_config:
+  offload_params: false
+  cpu_ram_efficient_loading: true
+  auto_wrap_policy: TRANSFORMER_BASED_WRAP
+  transformer_layer_cls_to_wrap: Qwen3DecoderLayer
+  state_dict_type: FULL_STATE_DICT
+  reshard_after_forward: true
```
winglian, line 125: do these need to have the `fsdp_` prefix? if not, we should update other parts below for consistency.

docs/multi-node.qmd
```diff
@@ -40,13 +40,13 @@ use_cpu: false
...
-fsdp:
-  - full_shard
-  - auto_wrap
 fsdp_config:
+  fsdp_version: 2
   fsdp_offload_params: true
   fsdp_state_dict_type: FULL_STATE_DICT
+  fsdp_auto_wrap_policy: TRANSFORMER_BASED_WRAP
   fsdp_transformer_layer_cls_to_wrap: LlamaDecoderLayer
+  fsdp_reshard_after_forward: true
```
winglian: `fsdp_` prefix?

src/axolotl/core/builders/causal.py
```diff
@@ -151,17 +151,13 @@ def build(self, total_num_steps):
...
             total_num_steps
         )
-
-        if self.cfg.fsdp:
-            training_arguments_kwargs["fsdp"] = self.cfg.fsdp
-            if self.cfg.fsdp_config:
-                training_arguments_kwargs["fsdp_config"] = {
-                    k.lstrip("fsdp_"): v for k, v in dict(self.cfg.fsdp_config).items()
-                }
-
```
winglian, line 161: seems we don't need the `fsdp_` prefix, but supported it for legacy reasons w accelerate. We should probably keep this section and warn in validation that the prefixes will be deprecated soon, otherwise all the existing FSDP1 configs will break.

src/axolotl/loaders/model.py
```diff
@@ -189,15 +194,15 @@ def _apply_post_model_load_setup(self):
...
+            and not self.is_qlora_and_fsdp_enabled
         ):
             self.model = self.model.merge_and_unload()
 
         self._resize_token_embeddings()
         self._adjust_model_config()
-        self._log_memory_usage()
         self._configure_embedding_dtypes()
         self._configure_qat()
+        log_gpu_memory_usage(LOG, "Memory usage after model load", 0)
```
winglian, line 205: by removing the device check, does this break when loading models into CPU for things like merging or other times when there is no gpu memory?

src/axolotl/loaders/model.py
```diff
@@ -604,8 +611,16 @@ def _configure_zero3_memory_efficient_loading(
...
+            if self.cfg.fsdp_config.fsdp_cpu_ram_efficient_loading:
+                skip_move_to_device = True
+                if "device_map" in self.model_kwargs:
+                    del self.model_kwargs["device_map"]
+            elif self.is_qlora_and_fsdp_enabled:
+                skip_move_to_device = True
+
         if (
-            self.qlora_fsdp
+            self.is_fsdp_enabled
```
winglian: I don't think this change is correct? the conditional body handles quantization, which may not apply to all fsdp use cases

docs/multi-node.qmd
```diff
@@ -40,13 +40,13 @@ use_cpu: false
 
 Configure your model to use FSDP in the Axolotl yaml. For example:
 ```yaml
-fsdp:
-  - full_shard
-  - auto_wrap
 fsdp_config:
+  fsdp_version: 2
```
winglian: this should get moved outside fsdp_config now?

tests/e2e/multigpu/test_fsdp2.py
```diff
@@ -0,0 +1,309 @@
+"""Test module for FSDP2 multi-GPU functionality."""
```
winglian, line 1: can we add some sort of test that validates that when we have `fsdp_config.fsdp_version` that it get's moved to cfg top level and handled correctly?

src/axolotl/monkeypatch/accelerate/fsdp2.py
```diff
@@ -191,17 +152,211 @@ def get_state_dict(self, model, unwrap=True):
     return state_dict
 
 
-def patch_accelerate_fsdp2():
-    import accelerate
-    from accelerate.utils import fsdp_utils
+def _process_lora_module_for_fsdp(module, fsdp2_kwargs):
```
winglian, line 157: Does this only support vanilla LoRA? What about DoRA?

salmanmohammadi: This supports DoRA by ignoring the magnitude params. I tested manually and I can also update with an E2E test.

src/axolotl/core/builders/causal.py
salmanmohammadi: I believe I have done this in this PR

src/axolotl/monkeypatch/accelerate/fsdp2.py
salmanmohammadi: *I added E2E tests and ended up having to also separately shard the DoRA params.

src/axolotl/loaders/model.py
salmanmohammadi: You are correct.

docs/multi-gpu.qmd
```diff
@@ -66,9 +66,69 @@ Start from Stage 1 -> Stage 2 -> Stage 3.
 
 :::
 
-## FSDP {#sec-fsdp}
+## Fully Sharded Data Parallel (FSDP) {#sec-fsdp}
 
-### Basic FSDP Configuration {#sec-fsdp-config}
+### FSDP2 (reccomended)
```
djsaunde:
```suggestion
### FSDP2 (recommended)
```

src/axolotl/loaders/model.py
```diff
@@ -251,22 +256,13 @@ def _adjust_model_config(self):
...
             self.model.config.eos_token_id = self.tokenizer.eos_token_id
 
-    def _log_memory_usage(self):
-        """Log device memory usage after model load."""
-        if hasattr(self.model, "device") and self.model.device.type in (
-            "cuda",
-            "mps",
-            "npu",
-        ):
-            log_gpu_memory_usage(LOG, "after model load", self.model.device)
```
djsaunde, line 261: Why remove this?

src/axolotl/loaders/model.py
```diff
@@ -430,7 +427,17 @@ def _set_device_map_config(self):
...
-        if not is_deepspeed_zero3_enabled():
+        is_ds_zero3 = is_deepspeed_zero3_enabled()
+
+        # FSDP requires control over device placement, so don't set device_map when FSDP is enabled
+        if self.is_fsdp_enabled:
+            # For QLoRA + FSDP, we still need to set device_map to "auto" for proper initialization
+            if self.is_qlora_and_fsdp_enabled:
+                self.model_kwargs["device_map"] = {
+                    "": int(os.environ.get("LOCAL_RANK", 0))
+                }
```
djsaunde, line 438: What does this do?

src/axolotl/monkeypatch/accelerate/fsdp2.py
```diff
@@ -191,17 +152,202 @@ def get_state_dict(self, model, unwrap=True):
...
+
+    from accelerate.utils import get_module_children_bottom_up, is_compiled_module
+    from accelerate.utils.fsdp_utils import fsdp2_prepare_auto_wrap_policy
+    from accelerate.utils.modeling import get_non_persistent_buffers
+    from torch.distributed.fsdp import (
+        CPUOffloadPolicy,
+        FSDPModule,
+        MixedPrecisionPolicy,
+        fully_shard,
+    )
```
djsaunde: Do we need to import all these in this scope? I think at least the stdlib imports probably don't need to be scoped here

src/axolotl/monkeypatch/accelerate/fsdp2.py
```diff
@@ -191,17 +152,202 @@ def get_state_dict(self, model, unwrap=True):
...
+            {k: v for k, v in model.named_buffers() if k in non_persistent_buffer_fqns}
+        )
+        # We move the model to meta device, as then sharding happens on meta device
+        model = model.to(torch.device("meta"))
+        # We need to re-tie the weights, not exactly sure why, but if we don't do this, reference to `lm_head/embed_tokens` stay hanging -> more VRAM usage
+        # We assume `transformers` models have a `tie_weights` method if they support it
+        if hasattr(model, "tie_weights"):
+            model.tie_weights()
+
+    is_peft_model = "Peft" in model.__class__.__name__
```
djsaunde: Does `isinstance(model, PeftModel)` not work here?

src/axolotl/monkeypatch/accelerate/fsdp2.py
```diff
@@ -191,17 +152,202 @@ def get_state_dict(self, model, unwrap=True):
...
+    if is_peft_model:
+        from peft.tuners.lora import LoraLayer
+
+        lora_layer = LoraLayer
+    else:
+        lora_layer = None
```
djsaunde: nit

```suggestion
    lora_layer = None
    if is_peft_model:
        from peft.tuners.lora import LoraLayer

        lora_layer = LoraLayer
```

src/axolotl/monkeypatch/accelerate/fsdp2.py
```diff
@@ -191,17 +152,202 @@ def get_state_dict(self, model, unwrap=True):
...
+    if log_bias_dtype_mismatch:
+        LOG.warning(
+            "Bias dtype mismatch detected in LoRA base linear layer, casting bias to weight dtype"
+        )
+
+    if fsdp2_plugin.cpu_ram_efficient_loading:
+        offload_to_cpu = isinstance(fsdp2_plugin.cpu_offload, CPUOffloadPolicy)
+        fsdp2_load_full_state_dict(
+            accelerator, model, original_sd, offload_to_cpu=offload_to_cpu
+        )
```
djsaunde, line 315: noob q: If this `if` evals False, do we still need to run `fsdp2_load_full_state_dict`?

src/axolotl/monkeypatch/accelerate/fsdp2.py
```diff
@@ -191,17 +152,202 @@ def get_state_dict(self, model, unwrap=True):
...
+            if "." in fqn:
+                parent_fqn, local_buffer_name = fqn.rsplit(".", 1)
+                parent_module = model.get_submodule(parent_fqn)
+            else:
+                local_buffer_name = fqn
+                parent_module = model
+
+            parent_module.register_buffer(
+                local_buffer_name, buffer_tensor, persistent=False
+            )
```
djsaunde, line 331: technically this could be nested under the preceding `if fsdp2_plugin.cpu_ram_efficient_loading` block with just the `if not model_has_params4bit` check

src/axolotl/loaders/model.py
djsaunde: Oh, I see it was moved elsewhere.

djsaunde (approved):
🚀

src/axolotl/monkeypatch/accelerate/fsdp2.py
salmanmohammadi: This was copied from upstream and I didn't spend a huge amount of time investigating its behaviour.

src/axolotl/monkeypatch/accelerate/fsdp2.py
salmanmohammadi: The state dict will already be on device and `fully_shard` will distribute the model instead.

docs/multi-gpu.qmd
```diff
@@ -66,9 +66,69 @@ Start from Stage 1 -> Stage 2 -> Stage 3.
...
-## FSDP {#sec-fsdp}
+## Fully Sharded Data Parallel (FSDP) {#sec-fsdp}
 
-### Basic FSDP Configuration {#sec-fsdp-config}
+### FSDP2 (recommended)
+
+::: {.callout-note}
+
+FSDP2 is recommended for new users. FSDP1 is deprecated and will be removed in an upcoming release of Axolotl.
+Currently, FSDP2 is not compatible with `load_in_8bit` or `load_in_4bit` and when using QLoRA. Please use DeepSpeed or set `fsdp_version` to 1 instead.
```
NanoCode012:
```suggestion
Currently, FSDP2 is not compatible with `load_in_8bit` or `load_in_4bit`. Please use DeepSpeed or set `fsdp_version: 1` instead.
```

I got a bit confused reading this. I think this simplifies it.

docs/multi-gpu.qmd
```diff
@@ -66,9 +66,69 @@ Start from Stage 1 -> Stage 2 -> Stage 3.
...
+fsdp_sync_module_states | **REMOVED**
+fsdp_cpu_ram_efficient_loading | cpu_ram_efficient_loading
+fsdp_state_dict_type | state_dict_type
+fsdp_use_orig_params | **REMOVED**
+
+
+For example, if you were using the following FSDP1 config:
+
+```{.yaml}
+fsdp_version: 1
```
NanoCode012, line 105: I recall version being nested under `fsdp_config`. If this is the new paradigm, could you also update the existing fsdp2 configs in `llama4` examples?

docs/multi-gpu.qmd
```diff
@@ -66,9 +66,69 @@ Start from Stage 1 -> Stage 2 -> Stage 3.
...
+  state_dict_type: FULL_STATE_DICT
+  reshard_after_forward: true
+```
+
+### FSDP1 {#sec-fsdp-config}
+
+::: {.callout-note}
+
+Using `fsdp` to configure FSDP is deprecated and will be removed in an upcoming release of Axolotl. Please use `fsdp_config` as above instead.
+
```
NanoCode012, line 133: How should `auto_wrap`/`full_shard` be migrated?

src/axolotl/cli/config.py
```diff
@@ -226,6 +227,7 @@ def load_cfg(
         },
     )
 
+    migrate_fsdp_config(cfg)
```
NanoCode012, line 230: Should this be done in the validator instead?

src/axolotl/core/builders/causal.py
```diff
@@ -151,17 +151,13 @@ def build(self, total_num_steps):
...
         if self.cfg.adapter == "qlora":
             training_arguments_kwargs["qlora"] = True
 
+        if self.cfg.fsdp_config or self.cfg.fsdp:
```
NanoCode012: Do we need to check for them individually or via `and`? If only fsdp is passed, then fsdp_config could be none?

src/axolotl/core/builders/rl.py
```diff
@@ -88,6 +88,10 @@ def _build_training_arguments(self, total_num_steps):
             total_num_steps=total_num_steps
         )
 
+        if self.cfg.fsdp_config:
+            training_args_kwargs["fsdp_config"] = self.cfg.fsdp_config
+            training_args_kwargs["fsdp"] = self.cfg.fsdp if self.cfg.fsdp else True
+
```
NanoCode012: I noticed it's the same in both. Could you move this to the base builder instead?

src/axolotl/loaders/adapter.py
```diff
@@ -152,7 +152,7 @@ def load_lora(
                 "Exception caught during model.print_trainable_parameters(): %s", exc
             )
     elif (
-        cfg.fsdp
+        cfg.fsdp_config
```
NanoCode012, line 155: L125 checks for both. Does this also need to check for fsdp?

I notice quite a few checks like this. Should we normalize it earlier during config validation?

src/axolotl/loaders/model.py
```diff
@@ -140,10 +140,15 @@ def has_flash_attn(self) -> bool:
         """Check if flash attention is installed."""
         return find_spec("flash_attn") is not None
 
+    @property
+    def is_fsdp_enabled(self):
+        """Property that determines if FSDP is enabled."""
+        return self.cfg.fsdp_config is not None or self.cfg.fsdp is not None
+
     @cached_property
```
NanoCode012: Does this need to be cached, otherwise, does is_fsdp_enabled need to be cached too?

src/axolotl/loaders/model.py
```diff
@@ -604,8 +611,16 @@ def _configure_zero3_memory_efficient_loading(
     def _build_model(self) -> bool:
         """Load model, with load strategy depending on config."""
         skip_move_to_device = False
+        if self.is_fsdp_enabled:
+            if self.cfg.fsdp_config.fsdp_cpu_ram_efficient_loading:
```
NanoCode012: Would this config need to have prefix `fsdp_` removed?

docs/multi-gpu.qmd
```diff
@@ -66,9 +66,69 @@ Start from Stage 1 -> Stage 2 -> Stage 3.
...
+FSDP1 | FSDP2
+-------- | --------
+fsdp_sharding_strategy | reshard_after_forward
+fsdp_backward_prefetch_policy | **REMOVED**
+fsdp_backward_prefetch | **REMOVED**
+fsdp_forward_prefetch | **REMOVED**
+fsdp_sync_module_states | **REMOVED**
+fsdp_cpu_ram_efficient_loading | cpu_ram_efficient_loading
+fsdp_state_dict_type | state_dict_type
+fsdp_use_orig_params | **REMOVED**
```
NanoCode012, line 99: If it's removed, could we provide an alternative of what users should do?

src/axolotl/loaders/model.py
```diff
@@ -691,33 +700,22 @@ def _build_model(self) -> bool:
                     trust_remote_code=self.cfg.trust_remote_code or False,
                     **self.model_kwargs,
                 )
+        elif self.cfg.gptq:
+            self.model = self.auto_model_loader.from_pretrained(
+                self.base_model,
+                config=self.model_config,
+                trust_remote_code=self.cfg.trust_remote_code or False,
```
NanoCode012, line 711: Off topic, but this `trust_remote_code` should probably be moved to the `model_kwargs` to prevent duplication.

src/axolotl/loaders/model.py
```diff
@@ -632,12 +647,6 @@ def _build_model(self) -> bool:
...
             and not self.cfg.gptq
         ):
-            # TODO: Do we need to open this up for all models?
-            if self.cfg.fsdp and self.cfg.fsdp_config.fsdp_cpu_ram_efficient_loading:
-                skip_move_to_device = True
-                if "device_map" in self.model_kwargs:
-                    del self.model_kwargs["device_map"]
-
             # Please don't remove underscore binding without reading the fn docstring.
             _ = self._configure_zero3_memory_efficient_loading()
```
NanoCode012, line 656: Out of scope a bit too, but we may want to move this up with fsdp, to prevent redundant call on L712 and simplify/remove this L631 model check

src/axolotl/monkeypatch/accelerate/fsdp2.py
```diff
@@ -5,99 +5,60 @@
...
+            sharded_param = sharded_param.cpu()
+
+        sharded_sd[param_name] = nn.Parameter(sharded_param)
+        del full_tensor
+        full_sd[param_name] = None
+    model.load_state_dict(sharded_sd, assign=True, strict=True)
+    end_time = time.time()
+    LOG.info(
+        f"Time taken to load full state dict: {(end_time - start_time):.2f} seconds"
+    )
```
NanoCode012: Should this be `debug`?

src/axolotl/monkeypatch/accelerate/fsdp2.py
```diff
@@ -191,17 +152,202 @@ def get_state_dict(self, model, unwrap=True):
...
+        )
+        original_non_persistent_buffers = copy.deepcopy(
+            {k: v for k, v in model.named_buffers() if k in non_persistent_buffer_fqns}
+        )
+        # We move the model to meta device, as then sharding happens on meta device
+        model = model.to(torch.device("meta"))
+        # We need to re-tie the weights, not exactly sure why, but if we don't do this, reference to `lm_head/embed_tokens` stay hanging -> more VRAM usage
+        # We assume `transformers` models have a `tie_weights` method if they support it
+        if hasattr(model, "tie_weights"):
+            model.tie_weights()
```
NanoCode012, line 288: We tie it once at `_resize_token_embeddings`. Do we need to again here?

src/axolotl/monkeypatch/accelerate/fsdp2.py
```diff
@@ -191,17 +152,202 @@ def get_state_dict(self, model, unwrap=True):
...
+                )
+                log_bias_dtype_mismatch |= module_log_bias_mismatch
+            if auto_wrap_policy(module) and not isinstance(module, FSDPModule):
+                fully_shard(module, **fsdp2_kwargs)
+
+    fully_shard(model, **fsdp2_kwargs)
+
+    if log_bias_dtype_mismatch:
+        LOG.warning(
+            "Bias dtype mismatch detected in LoRA base linear layer, casting bias to weight dtype"
```
NanoCode012:
```suggestion
            "Bias dtype mismatch detected in LoRA base linear layer, casted bias to weight dtype"
```
Since already been casted?

src/axolotl/monkeypatch/accelerate/fsdp2.py
```diff
@@ -191,17 +152,202 @@ def get_state_dict(self, model, unwrap=True):
...
+    log_bias_dtype_mismatch = False
+    if auto_wrap_policy is not None:
+        for module in get_module_children_bottom_up(model)[:-1]:
+            if is_peft_model and isinstance(module, lora_layer):
+                module_log_bias_mismatch = _process_lora_module_for_fsdp(
+                    module, fsdp2_kwargs
+                )
+                log_bias_dtype_mismatch |= module_log_bias_mismatch
+            if auto_wrap_policy(module) and not isinstance(module, FSDPModule):
+                fully_shard(module, **fsdp2_kwargs)
```
NanoCode012: Hm, when would the policy be None? The lora modules won't be moved then.

src/axolotl/monkeypatch/attention/flex_attn.py
```diff
@@ -6,6 +6,10 @@
 import torch
 import transformers
 
+from axolotl.utils.logging import get_logger
+
+LOG = get_logger(__name__)
+
 
 def patch_flex_wrapper(**flex_attn_compile_kwargs):
     # TODO remove this patch when transformers#37285 is merged and in a release
```
NanoCode012, line 15: This should be in release 4.53.1 https://github.com/huggingface/transformers/commit/794fde7b1c3d041519fc28ea3e1461b0cfcad4e7 . Do we need to remove?

src/axolotl/train.py
```diff
@@ -281,22 +268,14 @@ def save_trained_model(
...
-        elif state_dict_type == "FULL_STATE_DICT":
-            trainer.save_model(cfg.output_dir)
+    if trainer.is_fsdp_enabled:
+        if cfg.fsdp_config or cfg.fsdp:
+            if cfg.fsdp_config.final_state_dict_type:
+                state_dict_type = cfg.fsdp_config.final_state_dict_type
+            else:
+                state_dict_type = cfg.fsdp_config.state_dict_type
+            trainer.accelerator.state.fsdp_plugin.set_state_dict_type(state_dict_type)
+        trainer.save_model(cfg.output_dir)
```
NanoCode012, line 278: Previously, we would always save the full state dict for final. This may save sharded dict. Do we need to provide a `LOG.info` to merge it to users?

src/axolotl/utils/config/__init__.py
```diff
@@ -115,10 +115,11 @@ def normalize_config(cfg):
         "chrf",
     ]
     choose_device(cfg)
-    cfg.ddp = cfg.ddp if cfg.ddp is not None else cfg.world_size != 1
-    if cfg.ddp:
+    # cfg.ddp = cfg.ddp if cfg.ddp is not None else cfg.world_size != 1
+    if cfg.world_size != 1:
         cfg.device_map = {"": int(os.environ.get("LOCAL_RANK", 0))}
-        cfg.batch_size = cfg.batch_size * cfg.world_size
+        if cfg.fsdp or cfg.fsdp_config or cfg.ddp:
```
NanoCode012, line 120: How about deepspeed?

src/axolotl/utils/schemas/config.py
```diff
@@ -558,14 +560,24 @@ class AxolotlInputConfig(
         },
     )
     fsdp: list[str] | None = Field(
-        default=None, json_schema_extra={"description": "FSDP configuration"}
+        default=None,
+        json_schema_extra={"description": "FSDP configuration"},
+        deprecated="Configuring FSDP using `fsdp` is deprecated. Please use `fsdp_config` instead.",
```
NanoCode012: Let's add link to migration docs?

docs/multi-gpu.qmd
```diff
@@ -66,9 +66,69 @@ Start from Stage 1 -> Stage 2 -> Stage 3.
...
+### FSDP2 (recommended)
+
+::: {.callout-note}
+
+FSDP2 is recommended for new users. FSDP1 is deprecated and will be removed in an upcoming release of Axolotl.
+Currently, FSDP2 is not compatible with `load_in_8bit` or `load_in_4bit` and when using QLoRA. Please use DeepSpeed or set `fsdp_version` to 1 instead.
+
+:::
+
+#### Migrating from FSDP1 to FSDP2
```
NanoCode012:
```suggestion
#### Migrating from FSDP1 to FSDP2 {#sec-migrate-fsdp1-fsdp2}
```

Let's add a sec like this (feel free to adjust naming) so that we can refer to it elsewhere.

src/axolotl/utils/schemas/config.py
```diff
@@ -558,14 +560,24 @@ class AxolotlInputConfig(
...
     )
+    # TODO @SalmanMohammadi strongly type this as its own schema
     fsdp_config: dict[str, Any] | None = Field(
         default=None, json_schema_extra={"description": "FSDP configuration options"}
     )
```
NanoCode012, line 584: We should probably type this. But it would need to be `fsdp1options | fsdp2options` until the deprecation.

docs/multi-gpu.qmd
```diff
@@ -66,9 +66,69 @@ Start from Stage 1 -> Stage 2 -> Stage 3.
...
 
-## FSDP {#sec-fsdp}
+## Fully Sharded Data Parallel (FSDP) {#sec-fsdp}
 
-### Basic FSDP Configuration {#sec-fsdp-config}
+### FSDP2 (recommended)
+
+::: {.callout-note}
+
+FSDP2 is recommended for new users. FSDP1 is deprecated and will be removed in an upcoming release of Axolotl.
```
NanoCode012, line 79: Since we are recommending this, do we need to update the examples?

src/axolotl/utils/schemas/config.py
```diff
@@ -1115,6 +1118,95 @@ def check_qat_config(cls, data):
 
         return data
 
+    @model_validator(mode="before")
+    @classmethod
+    def check_fsdp_version(cls, data):
+        fsdp_config = data.get("fsdp_config", {})
+        if fsdp_config and str(data.get("fsdp_version")) != "2":
+            LOG.info(
+                "FSDP1 will be deprecated in an upcoming release of axolotl."
```
```suggestion
                "FSDP1 will be deprecated in an upcoming release of Axolotl."
```

src/axolotl/utils/schemas/config.py
```diff
@@ -1115,6 +1118,95 @@ def check_qat_config(cls, data):
...
+                )
+        return data
+
+    @model_validator(mode="before")
+    @classmethod
+    def check_fsdp2_base_model_quant_dpo(cls, data):
+        if data.get("fsdp_version") == 2 and data.get("rl") in [
+            RLType.DPO,
+            RLType.KTO,
+            RLType.ORPO,
```
NanoCode012, line 1179: There's also IPO (which is also DPO). Should we check for what works instead and error if it doesn't match?

src/axolotl/utils/schemas/config.py
```diff
@@ -1115,6 +1118,95 @@ def check_qat_config(cls, data):
...
+                    "Please configure `fsdp_version` as a top-level field."
+                )
+        return data
+
+    @model_validator(mode="before")
+    @classmethod
+    def check_fsdp_config_kwargs_prefix(cls, data):
+        if fsdp_config := data.get("fsdp_config"):
+            for key, _ in fsdp_config.items():
+                if key.startswith("fsdp_"):
```
NanoCode012, line 1205: This may trigger on `fsdp_version`.

src/axolotl/utils/schemas/config.py
```diff
@@ -1115,6 +1118,95 @@ def check_qat_config(cls, data):
...
+        return data
+
+    @model_validator(mode="before")
+    @classmethod
+    def check_fsdp2_lora_torch_2_7(cls, data):
+
+        if (
+            data.get("fsdp_version") == 2
+            and data.get("fsdp_config")
+            and data.get("adapter") == "lora"
```
NanoCode012: How about qlora?

src/axolotl/utils/trainer.py
```diff
@@ -648,11 +650,7 @@ def setup_trainer(
     """
     from axolotl.core.builders import HFCausalTrainerBuilder, HFRLTrainerBuilder
 
-    if (
-        cfg.torch_compile
-        and cfg.fsdp_config
-        and str(cfg.fsdp_config.fsdp_version) == "2"
-    ):
+    if cfg.torch_compile and cfg.fsdp_config and str(cfg.fsdp_version) == "2":
```
NanoCode012: Do we default fsdp_version? Why need str cast here? In case it's None?

tests/e2e/multigpu/test_fsdp1.py
```diff
@@ -0,0 +1,326 @@
...
+                "val_set_size": 0.01,
+                "datasets": [
+                    {
+                        "path": "tatsu-lab/alpaca",
+                        "type": "alpaca",
+                        "split": "train[:10%]",
+                    },
+                ],
+                "num_epochs": 1,
+                "max_steps": 2,
```
NanoCode012, line 72:
```suggestion
                "max_steps": 5,
```

Maybe increase a few steps or to 10 so we can check final loss better?

src/axolotl/utils/schemas/config.py
```diff
@@ -1115,6 +1118,95 @@ def check_qat_config(cls, data):
...
+    @classmethod
+    def check_fsdp2_base_model_quant_ram_efficient_loading(cls, data):
+        fsdp_config = data.get("fsdp_config")
+        if fsdp_config and data.get("fsdp_version") == 2:
+            if fsdp_config.get("fsdp_cpu_ram_efficient_loading") and (
+                data.get("load_in_8bit") or data.get("load_in_4bit")
+            ):
+                raise ValueError(
+                    "FSDP2 does not support load_in_8bit or load_in_4bit with cpu_ram_efficient_loading. Please use DeepSpeed or set `fsdp_version` to 1."
+                )
```
```suggestion
                raise ValueError(
                    "FSDP2 does not support load_in_8bit or load_in_4bit with cpu_ram_efficient_loading. Please turn off `cpu_ram_efficient_loading`, use DeepSpeed, or set `fsdp_version` to 1."
                )
```

NanoCode012 (commented):
Thanks for the huge work in adding better fsdp2. Left some comments

docs/multi-gpu.qmd
NanoCode012: After completing review, this is only the case with cpu efficient loading ram on. Do we need to mention that?

docs/multi-gpu.qmd
salmanmohammadi: I was thinking of updating and normalizing the configs in a follow up as this PR is already pretty beefy.

src/axolotl/cli/config.py
salmanmohammadi: In principle, yes, but we set accelerator environment variables using the FSDP config so it needs to happen before then.

src/axolotl/core/builders/causal.py
salmanmohammadi: This is mainly for backwards compatibility when a user has specified `fsdp` but not `fsdp_config`.

src/axolotl/loaders/model.py
salmanmohammadi: good catch!

docs/multi-gpu.qmd
salmanmohammadi: There just isn't analagous behaviour in FSDP2

src/axolotl/monkeypatch/accelerate/fsdp2.py
salmanmohammadi: This is copied from upstream.

src/axolotl/monkeypatch/accelerate/fsdp2.py
salmanmohammadi: Do you mean the auto wrap policy? This loop iterates from the bottom-up - from the child module to its parent module. Since the LoRA modules are child modules they will be wrapped before the auto wrap policy hits.

src/axolotl/utils/config/__init__.py
salmanmohammadi: I'm not sure what the behaviour should be with deepspeed.

docs/multi-gpu.qmd
salmanmohammadi: We can update the examples in a follow up.

src/axolotl/utils/schemas/config.py
salmanmohammadi: `fsdp_version` shouldn't be inside `fsdp_config` though

src/axolotl/utils/schemas/config.py
salmanmohammadi: I should actually just remove this check as I fixed it without needing torch 2.7

tests/e2e/multigpu/test_fsdp1.py
salmanmohammadi: I'll address this in a follow up, but I'm not currently checking the loss in the tests anyway.

src/axolotl/utils/schemas/config.py
salmanmohammadi: It was any RLType which uses the `DPOTrainer` - I'll add IPO here. Good catch.

src/axolotl/monkeypatch/attention/flex_attn.py
salmanmohammadi: I think this should be for another PR.

src/axolotl/loaders/model.py
salmanmohammadi: I'm not sure why it was cached as it is an inexpensive operation.

src/axolotl/loaders/adapter.py
salmanmohammadi: Commented above.

docs/multi-gpu.qmd
salmanmohammadi: Sorry, what do you mean this is the only case? What would you want to mention?

src/axolotl/monkeypatch/accelerate/fsdp2.py
salmanmohammadi: Yeah I suppose this whole block could be simplified since we always know peft will be installed with axolotl right?

src/axolotl/loaders/model.py
salmanmohammadi: Updated to guard on device type.

src/axolotl/utils/schemas/config.py
salmanmohammadi: There's a link in the validation function.

src/axolotl/loaders/model.py
salmanmohammadi: For QLoRA + FSDP accelerate requires that the model has been already moved to the device for its rank. I'm not sure why but this was the fix.

salmanmohammadi: Otherwise you get 
```bash
E           [rank1]: ValueError: You can't train a model that has been loaded in 8-bit or 4-bit precision on a different device than the one you're training on. Make sure you loaded the model on the correct device using for example `device_map={'':torch.cuda.current_device()}` or `device_map={'':torch.xpu.current_device()}`
```

docs/multi-gpu.qmd
salmanmohammadi: Updated and removed as I actually fixed this.

djsaunde:
QQ: any plans to move remaining fsdp configs to fsdp2?

salmanmohammadi:
> QQ: any plans to move remaining fsdp configs to fsdp2?

Definitely, by the time we officially remove FSDP1 at the latest.

PR #2766, opened 2025-06-06
Support --lora-on-cpu flag for DPO model merging

kallewoof:
The `lora_on_cpu` flag allows you to e.g. merge a LoRA and a model even when there isn't enough GPU VRAM by handling the LoRA part exclusively on CPU. This patch enables this for DPO based merging as well.

# Description

Check for the presence of the `lora_on_cpu` option in the DPO loader code, and use when set.

## Motivation and Context

Without this, you either need to have enough VRAM to fit everything, or not use the GPU at all for the merge process. This allows you to use the available VRAM while still spilling over into CPU RAM.

## How has this been tested?

I've used it to perform merges of DPO training runs using `axolotl merge-lora` and tested the resulting models.

## Summary by CodeRabbit

- **New Features**
  - Added an option to load the model entirely on CPU when a specific configuration flag is set, allowing for large memory allocation on CPU.

(note: the above summary is wrong -- we could always load the model entirely on CPU simply by doing `CUDA_VISIBLE_DEVICES= axolotl merge-lora [...]`)

### DIFF
--- a/src/axolotl/loaders/model.py
+++ b/src/axolotl/loaders/model.py
@@ -712,10 +712,6 @@ class ModelLoader:
                 # Please don't remove underscore binding without reading the fn docstring.
                 _ = self._configure_zero3_memory_efficient_loading()
 
-                if self.cfg.lora_on_cpu:
-                    self.model_kwargs["max_memory"] = {"cpu": "256GiB"}
-                    self.model_kwargs["device_map"] = {"": "cpu"}
-
                 self.model = self.auto_model_loader.from_pretrained(
                     self.base_model,
                     config=self.model_config,
@@ -780,6 +776,9 @@ class ModelLoader:
         dist_dtype: torch.dtype,
         before_kbit_train_or_finetune: bool,
     ):
+        dest = {"dtype": dist_dtype}
+        if self.cfg.lora_on_cpu:
+            dest["device"] = "cpu"
         for name, module in self.model.named_modules():
             if "norm" in name:
                 module.to(dist_dtype)
@@ -790,4 +789,4 @@ class ModelLoader:
                     # don't upcast lm_head for btlm
                     continue
             if any(m in name for m in embedding_modules) and hasattr(module, "weight"):
-                module.to(dist_dtype)
+                module.to(**dest)
--- a/src/axolotl/loaders/model.py
+++ b/src/axolotl/loaders/model.py
@@ -712,6 +712,10 @@ class ModelLoader:
                 # Please don't remove underscore binding without reading the fn docstring.
                 _ = self._configure_zero3_memory_efficient_loading()
 
+                if self.cfg.lora_on_cpu:
+                    self.model_kwargs["max_memory"] = {"cpu": "256GiB"}
+                    self.model_kwargs["device_map"] = {"": "cpu"}
+
                 self.model = self.auto_model_loader.from_pretrained(
                     self.base_model,
                     config=self.model_config,


winglian (approved):
lgtm, @NanoCode012 ?

src/axolotl/loaders/model.py
```diff
@@ -706,6 +706,10 @@ def _build_model(self) -> bool:
                 # Please don't remove underscore binding without reading the fn docstring.
                 _ = self._configure_zero3_memory_efficient_loading()
 
+                if self.cfg.lora_on_cpu:
```
salmanmohammadi: Could this not be included in the check below for deepspeed where we set `skip_move_to_device=True`?

NanoCode012:
I'm a bit confused on this. We already seem to set this on this function specifically for merge? Is this section not being hit?

https://github.com/axolotl-ai-cloud/axolotl/blob/09c685fd2c80f30e0d4a8ee937f8cc9058fce15f/src/axolotl/loaders/adapter.py#L132-L143

winglian:
> I'm a bit confused on this. We already seem to set this on this function specifically for merge? Is this section not being hit?
> 
> https://github.com/axolotl-ai-cloud/axolotl/blob/09c685fd2c80f30e0d4a8ee937f8cc9058fce15f/src/axolotl/loaders/adapter.py#L132-L143

Isn't that only when `lora_model_dir` is explicitly set? for merging, don't we automatically infer that, so we don't necessarily have that during model load?

src/axolotl/loaders/model.py
kallewoof: No, the OOM happens during the `self.model = self.auto_model_loader.from_pretrained()` call when `--lora-on-cpu=false`.

Edit: correction, it happens later (see trace below), but I doln't think `skip_move_to_device` is the correct patch, since we want to load as much as we can onto the GPU and then do the rest on CPU. I could be wrong on this though. Will test.

kallewoof:
> I'm a bit confused on this. We already seem to set this on this function specifically for merge? Is this section not being hit?

Right, it's not being hit in this case.

I can show stack traces pre-patch if that is necessary.

NanoCode012:
> No, the OOM happens during the `self.model = self.auto_model_loader.from_pretrained()` call when `--lora-on-cpu=false`.

Could you give us info on how to repro this issue? Did you pass a lora path to `base_model` ? This sounds less like  use case for `lora_on_cpu` ?

kallewoof:
@NanoCode012 Sure, here's main trying to merge a DPO trained model with `--lora-on-cpu` set:
```
$ axolotl merge-lora dpo_config.yml  --lora-model-dir ./dpo_model --lora-on-cpu true
[2025-06-10 08:53:52,209] [INFO] [axolotl.utils.config.log_gpu_memory_usage:107] [PID:2815328] [RANK:0] cuda memory usage baseline: 0.000GB (+0.422GB misc)

     #@@ #@@      @@# @@#
    @@  @@          @@  @@           =@@#                               @@                 #@    =@@#.
    @@    #@@@@@@@@@    @@           #@#@=                              @@                 #@     .=@@
      #@@@@@@@@@@@@@@@@@            =@# @#     ##=     ##    =####=+    @@      =#####+  =#@@###.   @@
    @@@@@@@@@@/  +@@/  +@@          #@  =@=     #@=   @@   =@#+  +#@#   @@    =@#+  +#@#   #@.      @@
    @@@@@@@@@@  ##@@  ##@@         =@#   @#      =@# @#    @@      @@   @@    @@      #@   #@       @@
     @@@@@@@@@@@@@@@@@@@@          #@=+++#@=      =@@#     @@      @@   @@    @@      #@   #@       @@
                                  =@#=====@@     =@# @#    @@      @@   @@    @@      #@   #@       @@
    @@@@@@@@@@@@@@@@  @@@@        #@      #@=   #@=  +@@   #@#    =@#   @@.   =@#    =@#   #@.      @@
                                 =@#       @#  #@=     #@   =#@@@@#=    +#@@=  +#@@@@#=    .##@@+   @@
    @@@@  @@@@@@@@@@@@@@@@

[2025-06-10 08:53:52,243] [INFO] [axolotl.cli.utils.load_model_and_tokenizer:318] [PID:2815328] [RANK:0] loading tokenizer... mistral-small-24b-instruct
[2025-06-10 08:53:52,681] [INFO] [axolotl.cli.utils.load_model_and_tokenizer:321] [PID:2815328] [RANK:0] loading model...
Loading checkpoint shards: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 10/10 [07:16<00:00, 43.63s/it]
[2025-06-10 09:01:09,703] [INFO] [axolotl.loaders.model.log_gpu_memory_usage:107] [PID:2815328] [RANK:0] cuda memory usage after model load: 21.954GB (+0.004GB cache, +0.681GB misc)
  File "/home/user/ws/miniconda3/envs/ai/bin/axolotl", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/user/ws/axolotl/src/axolotl/cli/main.py", line 362, in main
    cli()
  File "/home/user/ws/miniconda3/envs/ai/lib/python3.12/site-packages/click/core.py", line 1161, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/ws/miniconda3/envs/ai/lib/python3.12/site-packages/click/core.py", line 1082, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/home/user/ws/miniconda3/envs/ai/lib/python3.12/site-packages/click/core.py", line 1697, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/ws/miniconda3/envs/ai/lib/python3.12/site-packages/click/core.py", line 1443, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/ws/miniconda3/envs/ai/lib/python3.12/site-packages/click/core.py", line 788, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/ws/axolotl/src/axolotl/cli/utils.py", line 65, in wrapper
    return func(*args, **filtered_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/ws/axolotl/src/axolotl/cli/main.py", line 308, in merge_lora
    do_cli(config=config, **kwargs)
  File "/home/user/ws/axolotl/src/axolotl/cli/merge_lora.py", line 90, in do_cli
    do_merge_lora(cfg=parsed_cfg)
  File "/home/user/ws/axolotl/src/axolotl/cli/merge_lora.py", line 28, in do_merge_lora
    model, tokenizer, processor = load_model_and_tokenizer(cfg=cfg)
  File "/home/user/ws/axolotl/src/axolotl/cli/utils.py", line 323, in load_model_and_tokenizer
    model, _ = model_loader.load()
               ^^^^^^^^^^^^^^^^^^^
  File "/home/user/ws/axolotl/src/axolotl/loaders/model.py", line 164, in load
    self._apply_post_model_load_setup()
  File "/home/user/ws/axolotl/src/axolotl/loaders/model.py", line 199, in _apply_post_model_load_setup
    self._configure_embedding_dtypes()
  File "/home/user/ws/axolotl/src/axolotl/loaders/model.py", line 274, in _configure_embedding_dtypes
    self._convert_embedding_modules_dtype(
  File "/home/user/ws/axolotl/src/axolotl/loaders/model.py", line 783, in _convert_embedding_modules_dtype
    module.to(dist_dtype)
  File "/home/user/ws/miniconda3/envs/ai/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1343, in to
    return self._apply(convert)
           ^^^^^^^^^^^^^^^^^^^^
  File "/home/user/ws/miniconda3/envs/ai/lib/python3.12/site-packages/torch/nn/modules/module.py", line 930, in _apply
    param_applied = fn(param)
                    ^^^^^^^^^
  File "/home/user/ws/miniconda3/envs/ai/lib/python3.12/site-packages/torch/nn/modules/module.py", line 1329, in convert
    return t.to(
           ^^^^^
torch.OutOfMemoryError: CUDA out of memory. Tried to allocate 2.50 GiB. GPU 0 has a total capacity of 23.58 GiB of which 1.35 GiB is free. Including non-PyTorch memory, this process has 22.21 GiB memory in use. Of the allocated memory 21.95 GiB is allocated by PyTorch, and 3.61 MiB is reserved by PyTorch but unallocated. If reserved but unallocated memory is large try setting PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True to avoid fragmentation.  See documentation for Memory Management  (https://pytorch.org/docs/stable/notes/cuda.html#environment-variables)
```

I was wrong about the point of the crash though, but I think the patch position is correct. It's not about whether the model should be loaded onto the device or not (I could simply do CUDA_VISIBLE_DEVICES= if I wanted to avoid using the GPU, but I do want to use the GPU as well).

src/axolotl/loaders/model.py
kallewoof: I tested and the OOM crash occurs even with `skip_move_to_device` set.

kallewoof:
I'm noticing that this isn't always working as it should. I'm seeing only minimal GPU use and most ops done on CPU RAM in some runs. Investigating.

kallewoof:
Apologies. After investigating, it looks like the original approach was not properly making use of the GPU VRAM available. I was sure I verified that, but when retesting it's not doing so anymore, so I must have been imagining things.

I've rewritten the patch to now use `device="cpu"` in `_convert_embedding_modules_dtype` when `lora_on_cpu` is set. This does properly make use of the GPU up to the amount available without running out of RAM when doing the LoRA merge/save.

It's possible that this will hit OOM's in certain cases, but I don't want to be too aggressive with moving things around unneccessarily.

kallewoof:
@NanoCode012 Is there a blocker here? I tried to do a merge using `CUDA_VISIBLE_DEVICES=` to do CPU only and this failed with a ton of 
```
torch/nn/modules/module.py:2397: UserWarning: for base_model.model.model.layers.33.self_attn.q_proj.lora_A.default.weight: copying from a non-meta parameter in the checkpoint to a meta parameter in the current model, which is a no-op. (Did you mean to pass `assign=True` to assign items in the state dictionary to their corresponding key in the module instead of copying them in place?)
```
errors after spending a very long time loading the model into CPU RAM (and swap because it wouldn't fit).

If what I'm proposing is already doable, please provide me with a command on how to do this. To reiterate, all I want is to make use of the GPU for partially loading model + LoRA stuff when doing a merge.

NanoCode012:
@kallewoof , sorry for slow response. Team was mostly away last week.

I'm just a bit unclear on the side effects for this, as this seems to be moving embeddings onto cpu now. Would it be better to create a new config called, say, `embeddings_on_cpu` TBD? And if we would like to narrow it down for merge only, we can do a check for `lora_model_dir` during schema validation?

kallewoof:
> I'm just a bit unclear on the side effects for this, as this seems to be moving embeddings onto cpu now. Would it be better to create a new config called, say, `embeddings_on_cpu` TBD? And if we would like to narrow it down for merge only, we can do a check for `lora_model_dir` during schema validation?

So, when you do a merge and you don't have enough VRAM to load the entire model and LoRA, you end up with partially loading the model into VRAM and then spilling the remainder into CPU RAM. That works perfectly fine. However, when you do the merge with LoRA part, this ends up trying to use the GPU VRAM, which we have already filled to the max with (parts of) the model. The result is an OOM crash.

`lora_on_cpu` addresses this by doing the LoRA part specifically on CPU. The model is (fully or partially) loaded into VRAM but the LoRA is loaded/merged in CPU, hence the option name.

However, trying to use this flag when merging a DPO trained model does nothing. The code still has an OOM crash, which happens during the `_convert_embedding_modules_dtype` call (hence why this is where the `cpu` device tweak is made).

Thus, the answer to your question about side effects is: `lora_on_cpu` is, AFAIK, never used outside of merging operations, so the side effects should be zero. If `lora_on_cpu` has evolved since I proposed it originally, then I'm outdated and need to do my homework. Let me know if so!

winglian (approved):
this all seems sane to me. Thanks @kallewoof !

PR #2774, opened 2025-06-09
Feat: add tool calling support via tools column

NanoCode012:
# Description

Note: We had prior support for loading manually preprocessed `tool` datasets by parsing and placing into the assistant/other role contents. However, this PR streamlines the process by being template independent.

---

Add 

```yaml
datasets:
    ....
    field_tools: tools
```

to load the `tools` column in datasets and pass to `apply_chat_template`.

Added docs on the structure of `tools` dataset and added test with sample data format for reference.

### DIFF
--- a/docs/config.qmd
+++ b/docs/config.qmd
@@ -173,6 +173,10 @@ datasets:
     # Key containing the messages (default: "messages")
     field_messages: messages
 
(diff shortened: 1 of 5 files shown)


winglian (approved):
Amazing. Thanks! 🙌

PR #2780, opened 2025-06-11
Feat: Add Magistral and mistral-common tokenizer support

NanoCode012:
# Description

Add Magistral and mistral-common package support.

## How has this been tested?

Preprocess ran locally and manual training runs on cloud.

## Summary by CodeRabbit

- **New Features**
  - Added support for the mistral-common tokenizer with new configuration options and dedicated tokenizing strategies.
  - Introduced a detailed fine-tuning configuration file and README for the Magistral Small model using QLoRA.
  - Added optional "mistral" dependency for easy installation of the mistral-common package.

- **Documentation**
  - Enhanced docs with new configuration references and a comprehensive Magistral Small fine-tuning guide.

- **Bug Fixes**
  - Disabled multiprocessing automatically when the tokenizer does not support it for improved stability.

- **Refactor**
  - Improved configuration validation to enforce compatibility rules for mistral-common tokenizer usage.
  - Updated prompt strategy selection logic to accommodate new tokenizer options.


src/axolotl/utils/mistral_tokenizer.py
```diff
@@ -0,0 +1,489 @@
...
+        )
+
+        tokens: list[int] = self._mistral.encode_chat_completion(chat_completion).tokens
+
+        if tokenize:
+            return tokens
+
+        return self._mistral.decode(tokens)
+
+    def pad(
```
NanoCode012, line 353: Would love eyes on this section. I think I'm duplicating code / not as efficient padding.

src/axolotl/utils/schemas/config.py
```diff
@@ -1265,6 +1265,49 @@ def check_muon_deepspeed_fsdp(cls, data):
...
+    @field_validator("tokenizer_use_mistral_common")
+    @classmethod
+    def check_mistral_common_import(cls, tokenizer_use_mistral_common):
+        if tokenizer_use_mistral_common:
+            try:
+                import mistral_common  # noqa: F401 # pylint:disable=unused-import
+            except ImportError as exception:
+                raise ImportError(
+                    "mistral-common is required for mistral models. Please install it with `pip install mistral-common`."
+                ) from exception
```
djsaunde: Do we need this guard? I see you added `mistral_common` to our `requirements.txt` so this shouldn't get hit.

docs/config.qmd
```diff
@@ -27,6 +27,8 @@ trust_remote_code:
 tokenizer_use_fast:
 # Whether to use the legacy tokenizer setting, defaults to True
 tokenizer_legacy:
+# Whether to use mistral-common tokenizer. If set to True, it will use the mistral-common tokenizer.
+tokenizer_use_mistral_common:
```
djsaunde, line 31: I think this is fine for now, but it might be better to auto-detect needing the mistral common tokenizer as this can be a stumbling point for users.

src/axolotl/loaders/tokenizer.py
```diff
@@ -121,6 +121,23 @@ def modify_tokenizer_files(
...
+
+        from axolotl.utils.mistral_tokenizer import HFMistralTokenizer
+
+        # Load the HF-compatible wrapper around MistralTokenizer
+        tokenizer = HFMistralTokenizer.from_pretrained(cfg.tokenizer_config)
+
+        return tokenizer
+
+    if cfg.tokenizer_use_mistral_common:
+        return _load_mistral_common_tokenizer(cfg)
```
djsaunde: The assert shouldn't be needed if you're already checking `if cfg.tokenizer_use_mistral_common:` here.

src/axolotl/prompt_strategies/chat_template.py
```diff
@@ -15,6 +17,9 @@
 from axolotl.utils.logging import get_logger
 from axolotl.utils.schemas.datasets import DatasetConfig
 
+if TYPE_CHECKING:
+    from axolotl.utils.mistral_tokenizer import HFMistralTokenizer
```
djsaunde, line 21: 💯

src/axolotl/prompt_strategies/chat_template.py
```diff
@@ -650,9 +661,9 @@ def find_turn(
             return -1, -1
 
         LOG.debug(f"Content boundaries: {start_idx}, {end_idx}")
-        LOG.debug(
-            f"Content tokens: {self.tokenizer.convert_ids_to_tokens(full_ids[start_idx:end_idx])}"
-        )
+        # LOG.debug(
+        #     f"Content tokens: {self.tokenizer.convert_ids_to_tokens(full_ids[start_idx:end_idx])}"
+        # )
```
djsaunde: Should be removed / added back?

src/axolotl/prompt_strategies/chat_template.py
```diff
@@ -796,14 +807,105 @@ def _get_messages(self, prompt):
...
+        sequence_len: int,
+        roles_to_train: list[str] | None = None,
+        train_on_eos: str | None = None,
+        train_on_eot: str | None = None,
+        eot_tokens: list[str] | None = None,
+        split_thinking: bool | None = False,
+    ):
+        # Call the parent's parent __init__ (PromptTokenizingStrategy) to skip ChatTemplateStrategy's validation
+        # pylint: disable=non-parent-init-called
+        # pylint: disable=super-init-not-called
```
djsaunde: nit

```suggestion
        # pylint: disable=non-parent-init-called,super-init-not-called
```

src/axolotl/utils/mistral_tokenizer.py
```diff
@@ -0,0 +1,489 @@
...
+    """
+
+    def __init__(self, mistral: _MistralTokenizer, path_or_repo_id: str):
+        """
+        Args:
+            tokenizer: The tokenizer to wrap.
+            path: The path to the tokenizer files.
+        """
+        self._mistral = mistral
+        self._padding_side = "right"
```
djsaunde, line 48: stupid question: doesn't the mistral tokenizer pad from the left?

src/axolotl/utils/mistral_tokenizer.py
```diff
@@ -0,0 +1,489 @@
...
+        self.chat_template = None
+        self._tokenizer_path = _get_file_path(path_or_repo_id, "tekken.json")
+
+        # Try to load system prompt if available
+        try:
+            self._system_prompt = self._load_system_prompt(
+                path_or_repo_id=path_or_repo_id
+            )
+        except FileNotFoundError:
+            pass
```
djsaunde: You should initialize `self._system_prompt` as `None` here; otherwise, trying to access it later will cause an error.

Or, maybe it's not used anywhere?

I would recommend using `contextlib.suppress` instead of the `try` / `expect` block.

src/axolotl/utils/mistral_tokenizer.py
```diff
@@ -0,0 +1,489 @@
...
+import math
+import os
+from shutil import copyfile
+from typing import TYPE_CHECKING, Optional
+
+import numpy as np
+from huggingface_hub import hf_hub_download
+from mistral_common.tokens.tokenizers.mistral import (
+    MistralTokenizer as _MistralTokenizer,
+)
```
djsaunde: Why change the name?

src/axolotl/utils/mistral_tokenizer.py
```diff
@@ -0,0 +1,489 @@
...
+    ) -> "HFMistralTokenizer":
+        """
+        Download a mistral tokenizer from HF Hub and wrap it.
+        """
+
+        if revision:
+            raise NotImplementedError("Revision not supported yet")
+
+        # check if tokenizer_config is a valid local path
+        base = _MistralTokenizer.from_file(
```
djsaunde: This class supports `.from_hf_hub` as well now

src/axolotl/utils/mistral_tokenizer.py
djsaunde: stupid question: can't we call `transformers` code here instead of writing this ourselves? I admit I'm not super familiar with this area

djsaunde (commented):
Can we add a smoke test?

Unit tests would also be nice but that could come later.

src/axolotl/utils/mistral_tokenizer.py
NanoCode012: I would love to if there's one. Do you know whether there's one?

src/axolotl/utils/schemas/config.py
NanoCode012: Yes, I moved it to optional dependency like you've previously done.

docs/config.qmd
NanoCode012: Yup, I agree here. I'll add a simple name check for now

src/axolotl/utils/mistral_tokenizer.py
djsaunde: `transformers.PreTrainedTokenizer.pad` exists but I'm not sure if / how much of the functionality overlaps with yours.

src/axolotl/utils/mistral_tokenizer.py
NanoCode012: It's not used anywhere. I'm deleting this block but keeping the utility function for now in case we want to revisit this. I think we don't want to default add system blocks without user's knowledge

src/axolotl/utils/mistral_tokenizer.py
NanoCode012: fixed :)

src/axolotl/utils/mistral_tokenizer.py
NanoCode012: Yep. For now, I'll keep this for loading tekken only. I think we need a function for `find_tokenizer_in_folder` like they have in `from_hf_hub`. We need the local tokenizer path to copy over the tokenizer during `save_pretrained`.

src/axolotl/datasets.py
```diff
@@ -48,6 +48,10 @@ def process(self, dataset):
         features = dataset.features.keys()
         num_proc = min(64, self.process_count if self.process_count else os.cpu_count())
 
+        # Disable multiprocessing if the tokenizer doesn't support it (e.g., mistral_common)
+        if not getattr(self.prompt_tokenizer, "supports_multiprocessing", True):
```
winglian, line 52: Might be good to add a LOG.info here

src/axolotl/utils/mistral_tokenizer.py
NanoCode012: I'm not sure about this too. For fine-tuning, it makes sense to be `right`? Anyways, I think the below would override it either way. Not sure if this block still needs it.

https://github.com/axolotl-ai-cloud/axolotl/blob/00cda8cc70ca6a2f501cef7e843ae87931faece7/src/axolotl/loaders/tokenizer.py#L174-L175

NanoCode012:
TODOs left:
- [ ] Check `pad` handling
- [ ] Unit test for Magistral tokenizer (mistral-common)
- [ ] Smoke test for Magistral training

winglian (approved):
Good to ship for release, but let's add a follow-on for adding some unit tests.

NanoCode012:
Added tests for the mistral common tokenizer. 

The model itself is just a Mistral model so no need duplicate smoke tests.

PR #2786, opened 2025-06-12
update loss value for flakey e2e test

winglian:
## Summary by CodeRabbit

- **Tests**
  - Updated the Llama pretraining test to allow a slightly higher baseline loss threshold, improving flexibility in test validation.

### DIFF
--- a/tests/e2e/test_llama_pretrain.py
+++ b/tests/e2e/test_llama_pretrain.py
@@ -14,19 +14,14 @@ class TestPretrainLlama:
     """Test case for Llama models w pretraining"""
 
     @pytest.mark.parametrize(
-        "sample_packing",
-        [True, False],
-    )
-    @pytest.mark.parametrize(
-        "pretrain_multipack_attn",
-        [True, False],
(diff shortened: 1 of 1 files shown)


tests/e2e/test_llama_pretrain.py
```diff
@@ -23,7 +23,9 @@ class TestPretrainLlama:
     )
     def test_pretrain(self, temp_dir, sample_packing, pretrain_multipack_attn):
         if not sample_packing and pretrain_multipack_attn:
-            return
+            pytest.skip(
+                "Combination not supported (multipack attention without sample packing)"
+            )
```
salmanmohammadi: I think you could just parameterize with a list of tuples rather than using the matrix, which means you wouldn't need this skip.

salmanmohammadi (approved):

tests/e2e/test_llama_pretrain.py
winglian: thanks, updated

PR #2796, opened 2025-06-17
Eval on training begin instead of after step 1

mhenrichsen:
Current implementation evaluates after step 1, which makes doesn't make sense, because the first eval is to establish a baseline.

Changed to run before training starts instead of step 1.

## Summary by CodeRabbit

- **Bug Fixes**
  - Evaluation now runs before training begins instead of after the first step, enhancing the accuracy of initial model assessment.

- **Documentation**
  - Clarified when evaluation is triggered during the training process.


NanoCode012:
Superseded by #2797 as discussed on discord

PR #2804, opened 2025-06-18
Enable Memory Efficient Loading when using Deepspeed 3 for Mistral

benHeid:
# Description

When using Mistral the model type `MistralForCausalLM` and during initialisation it will run in this elif block:

```python
        elif (
            self.model_type
            and self.model_type != "AutoModelForCausalLM"
            and not self.cfg.trust_remote_code
        )
```

this block does not configure deepspeed. Thus, I added the code for it.

## Motivation and Context

It enables deepspeed with mistral models as well as a significant faster loading of the models when using deepspeed.

## How has this been tested?

I compared the load times using mistral small 24B on a setup with 4 A100s. The change significantly improved the speed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved memory-efficient loading for certain model types to ensure consistent performance across all configurations.


NanoCode012 (changes requested):
Hey, thanks for PR. Do you have the numbers before/after?

Could I propose we change this a bit? Instead of adding this call here. Could you add `mistral` to this list https://github.com/axolotl-ai-cloud/axolotl/blob/eb3a57eb17053da3b4437566579d12dbc580e7e3/src/axolotl/loaders/model.py#L625

That section would enable efficient loading like you've done. We will be refactoring this later, so it would apply to more models.

winglian (approved):

NanoCode012:
Follow up here, this refactor in the PR https://github.com/axolotl-ai-cloud/axolotl/pull/3067/files#diff-63b03c4c6c36921a7f697424a2d347e2fc26fee5a23b990aa697cb61557cbf54 solves this.

PR #2808, opened 2025-06-18
get_logger use_environ fix

djsaunde:
# Description

## Summary by CodeRabbit

- **Refactor**
  - Simplified logging configuration by removing environment variable dependencies across multiple modules.
  - Centralized logger utility imports for consistency.
  - Enhanced type annotations and clarity in distributed utilities.
  - Added a new function to access distributed state.

- **Style**
  - Improved docstring formatting and code readability in several modules.

### DIFF
--- a/src/axolotl/cli/config.py
+++ b/src/axolotl/cli/config.py
@@ -26,7 +26,7 @@ from axolotl.utils.mlflow_ import setup_mlflow_env_vars
 from axolotl.utils.trainer import prepare_opinionated_env, prepare_optim_env
 from axolotl.utils.wandb_ import setup_wandb_env_vars
 
-LOG = get_logger(__name__, use_environ=True)
+LOG = get_logger(__name__)
 
 
 def check_remote_config(config: Union[str, Path]) -> Union[str, Path]:
--- a/src/axolotl/integrations/base.py
+++ b/src/axolotl/integrations/base.py
@@ -33,7 +33,7 @@ from transformers import PreTrainedModel, Trainer
 from axolotl.utils.dict import DictDefault
 from axolotl.utils.logging import get_logger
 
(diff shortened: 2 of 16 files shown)


NanoCode012:
Could an alternative to this fix be, to adjust the `should_log` function to check for whether a distributed state has been enabled first before calling `is_main_process`?

https://github.com/axolotl-ai-cloud/axolotl/blob/eb3a57eb17053da3b4437566579d12dbc580e7e3/src/axolotl/utils/logging.py#L24-L28

Otherwise, would adjusting `is_distributed` be the better approach?

https://github.com/axolotl-ai-cloud/axolotl/blob/eb3a57eb17053da3b4437566579d12dbc580e7e3/src/axolotl/utils/distributed.py#L51-L58

I'm not clear why a bool check would start a distributed state if False.

djsaunde:
> Could an alternative to this fix be, to adjust the `should_log` function to check for whether a distributed state has been enabled first before calling `is_main_process`?
> 
> https://github.com/axolotl-ai-cloud/axolotl/blob/eb3a57eb17053da3b4437566579d12dbc580e7e3/src/axolotl/utils/logging.py#L24-L28
> 
> Otherwise, would adjusting `is_distributed` be the better approach?
> 
> https://github.com/axolotl-ai-cloud/axolotl/blob/eb3a57eb17053da3b4437566579d12dbc580e7e3/src/axolotl/utils/distributed.py#L51-L58
> 
> I'm not clear why a bool check would start a distributed state if False.

Yeah, that's cleaner.

Yeah, not sure why we do that 🤔

winglian (approved):

djsaunde:
Per @NanoCode012's comment, there was a simpler way of doing this, by checking whether the global state was initialized when attempting to do a distributed-aware log.

PR #2823, opened 2025-06-23
feat: add sageattention

NanoCode012:
# Description

Adds SageAttention https://github.com/thu-ml/SageAttention/

Since it has similar interface as sdpa_attention, I used that implementation and flash attention in transformers to cross check.

## How has this been tested?

No test yet!

## Summary by CodeRabbit

- **New Features**
  - Added support for SageAttention as a selectable attention mechanism.
  - Introduced a configuration option to enable SageAttention.
- **Bug Fixes**
  - Added validation to prevent enabling SageAttention alongside unsupported sample packing.
  - Added validation to enforce hardware compatibility for SageAttention.
- **Documentation**
  - Updated configuration descriptions with references to SageAttention for user guidance.


winglian:
@NanoCode012 what's the sage vs flash attn VRAM usage?

NanoCode012:
@winglian , weirdly not getting vram savings as in benchmarks.

Current **early** wandb result show that: about 20% faster with same vram usage. However, kernel benchmarking showed it using less vram (when <32k context at least).

More runs needs to be done still.

NanoCode012:
Updated PR from main and added more validation/docs on attention. It is a bit faster than FA for adapter mode.

I added warning that this is not recommended for FFT due to unstable loss. I did not add test as I didn't want to install another module by default.

Edit Feb 2026: We will merge this PR as it is an external dependency to let users try it out. We understand that the metrics are not stable and are always open for help fixing it.

PR #2824, opened 2025-06-23
fix: remove unnecessary movement of eval logits to cpu

NanoCode012:
# Description

Edit: See https://github.com/axolotl-ai-cloud/axolotl/pull/2824#issuecomment-3002851516

---

There has been reports about eval taking more vram than training. I suspect it's this config which we set to grad accu. Would love some tests for this!

This config does not seem to be running mini-batches as we think?

https://huggingface.co/docs/transformers/main_classes/trainer#transformers.TrainingArguments.eval_accumulation_steps

## Summary by CodeRabbit

- **Refactor**
  - Removed explicit setting of the "eval_accumulation_steps" parameter in training configuration.

### DIFF
--- a/src/axolotl/core/builders/causal.py
+++ b/src/axolotl/core/builders/causal.py
@@ -215,10 +215,6 @@ class HFCausalTrainerBuilder(TrainerBuilderBase):
                 self.cfg.auto_find_batch_size
             )
 
-        training_arguments_kwargs["eval_accumulation_steps"] = (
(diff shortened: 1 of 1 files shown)


NanoCode012:
From reading the trainer evaluation loop, I don't think we need this config. The VRAM consumed by just an eval pass is very small compared to training, so this config would just cause unnecessary gpu->cpu movement.

However, it should not be the cause for increased vram usage as initially stated.

PR #2829, opened 2025-06-25
gating _gather_outputs (causes increased vram usage)

djsaunde:
# Description

Title. This gives back the original VRAM savings that we saw in the original implementation.

Possible follow-on items:
- Add checks for VRAM usage in smoke tests
- Add smoke tests for other, untested SP + trainer combos 

## Summary by CodeRabbit

- **New Features**
  - Added an option to control whether outputs are gathered during sequence parallel training, based on the selected reinforcement learning type.

- **Documentation**
  - Updated descriptions to reflect the new option for gathering outputs during training.

### DIFF
--- a/src/axolotl/train.py
+++ b/src/axolotl/train.py
@@ -219,6 +219,7 @@ def execute_training(
                     gradient_accumulation_steps=cfg.gradient_accumulation_steps,
                     ring_attn_func=cfg.ring_attn_func,
                     heads_k_stride=cfg.heads_k_stride,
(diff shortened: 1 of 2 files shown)


NanoCode012 (approved):
Would this save vram for all except for GRPO?

djsaunde:
> Would this save vram for all except for GRPO?

Yes. I don't have a fix yet for GRPO; we need the original tensors in that case.

winglian (approved):

PR #2840, opened 2025-06-27
add assertion for packing patch to _get_unpad_data

winglian:
# Description

Setting ourselves to catch changes earlier in case upstream removes this. The patch would normally silently pass even if this attribute didn't previously exist.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling with a clear message if a required attribute is missing due to upstream changes.


salmanmohammadi (approved):
LGTM, but is there a test that will pick this up?

winglian:
> LGTM, but is there a test that will pick this up?

yeah, basically any test that uses packing, which we have plenty of

PR #2842, opened 2025-06-27
Ensure device mesh patching is applied

djsaunde:
# Description

Our accelerate patching to enable SP seems to have broken recently; this fixes it. I also am more explicit about using FSDP when enabled.

## How has this been tested?

Confirmed working with FSDP x SP 2 x 2 device mesh on 4x H100 SXM. Also works with various optimizations (Liger optims, CCE).

Example config (modified from user's): https://gist.github.com/djsaunde/aca6285273cf9d476e69baa1cdcab6c7. This uses ~29GB VRAM per GPU.

I'll repro training losses and post here as well.

## Summary by CodeRabbit

* **New Features**
  * Improved support for sequence parallelism by automatically applying relevant patches when enabled in the configuration.

* **Bug Fixes**
  * Corrected a typo in documentation for data loader patching.

* **Refactor**
  * Simplified and updated the handling of device mesh dimensions for better alignment with PyTorch naming conventions.
  * Relocated sequence parallel patch application logic to ensure it is consistently applied during model loading.
  * Removed redundant patching during ring attention registration for cleaner integration.

* **Tests**
  * Updated tests to load configuration from disk files, enhancing test reliability and realism.


winglian (approved):
:shipit:

winglian:
Looks like you need a guard in the  _apply_sequence_parallel_patches function in case sp degree is none

salmanmohammadi (approved):

src/axolotl/loaders/patch_manager.py
```diff
@@ -234,7 +234,7 @@ def _apply_gemma3_conditional_generation_forward_patch(self):
 
     def _apply_sequence_parallel_patches(self):
         """Apply sequence parallelism patches."""
-        if self.cfg.sequence_parallel_degree > 1:
+        if self.cfg.sequence_parallel_degree and self.cfg.sequence_parallel_degree > 1:
```
djsaunde, line 237: Technically we should fix our tests since this doesn't happen in practice :) but it's okay for now I suppose.

PR #2851, opened 2025-07-01
feat(doc): update docker tag examples

NanoCode012:
# Description

We encouraged users to use cu128 for Blackwell but didn't provide example tags. This PR adds that.

## Summary by CodeRabbit

* **Documentation**
  * Updated Docker documentation to include new and alternative image tags for base and main images.
  * Revised tagged release version information to reflect the latest release.


NanoCode012:
@coderabbitai , can you check where the ordering needs to be? I think it's already ordered across sections by cuda then torch.

salmanmohammadi (approved):

PR #2853, opened 2025-07-01
fix:add sample_packing_sequentially to trainer args

v-dicicco:
# Description

The `MultipackBatchSampler` of the base trainer uses the argument `sample_packing_sequentially` (see [here](https://github.com/axolotl-ai-cloud/axolotl/blob/35fdbce1022547158a5e451a11350513f4ac4385/src/axolotl/core/trainers/base.py#L116)), however its value is not updated [during the creation](https://github.com/axolotl-ai-cloud/axolotl/blob/main/src/axolotl/core/builders/causal.py#L247-L267) of the Trainer arguments (see [here](https://github.com/axolotl-ai-cloud/axolotl/blob/main/src/axolotl/core/builders/causal.py#L247-L267)). This results in the argument always using its [default value](https://github.com/v-dicicco/axolotl/blob/main/src/axolotl/core/training_args_base.py#L35-L39) of `False`.

This PR adds the `sample_packing_sequentially` option to the the Training Arguments.

## Summary by CodeRabbit

* **New Features**
  * Added support for the `sample_packing_sequentially` configuration option in training settings.

### DIFF
--- a/src/axolotl/core/builders/causal.py
+++ b/src/axolotl/core/builders/causal.py
@@ -253,6 +253,10 @@ class HFCausalTrainerBuilder(TrainerBuilderBase):
         training_arguments_kwargs["eval_sample_packing"] = bool(
             self.cfg.eval_sample_packing
         )
+        if self.cfg.sample_packing_sequentially is not None:
+            training_arguments_kwargs["sample_packing_sequentially"] = (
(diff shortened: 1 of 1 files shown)


djsaunde (approved):
Weird, could have sworn I was using this feature a few weeks ago. Thanks for catching!

PR #2859, opened 2025-07-03
fix: set add_generation_prompt to False when apply chat template for multimodal

NanoCode012:
# Description

For some reason, `add_generation_prompt` was set to true which can lead to unexpected effects in multimodal.

## Summary by CodeRabbit

* **Bug Fixes**
  * Adjusted chat processing to no longer add a generation prompt during message handling.

### DIFF
--- a/src/axolotl/utils/collators/mm_chat.py
+++ b/src/axolotl/utils/collators/mm_chat.py
@@ -50,7 +50,7 @@ class MultiModalChatDataCollator(DataCollatorMixin):
             # This method requires transformers>=4.49.0
             result = self.processing_strategy.processor.apply_chat_template(
                 example["messages"],
-                add_generation_prompt=True,
+                add_generation_prompt=False,
                 tokenize=True,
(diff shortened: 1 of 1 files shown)


src/axolotl/utils/collators/mm_chat.py
```diff
@@ -50,7 +50,7 @@ def process_rows(
             # This method requires transformers>=4.49.0
             result = self.processing_strategy.processor.apply_chat_template(
                 example["messages"],
-                add_generation_prompt=True,
+                add_generation_prompt=False,
```
salmanmohammadi, line 53: @NanoCode012 do you know why setting it to `True` would cause issues? I saw [here](https://huggingface.co/docs/transformers/en/chat_templating_multimodal#image-inputs) that it was set to `True`

NanoCode012: Sorry, I missed this. It's for use with generation, not training.

PR #2862, opened 2025-07-03
don't use tokenizer parallelism when using packing

winglian:
When using packing, the stderr gets spammed with

```
stderr: To disable this warning, you can either:                                                       
stderr:         - Avoid using `tokenizers` before the fork if possible                                 
stderr:         - Explicitly set the environment variable TOKENIZERS_PARALLELISM=(true | false)         
stderr: huggingface/tokenizers: The current process just got forked, after parallelism has already been used. Disabling parallelism to avoid deadlocks
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Disabled tokenizer parallelism when sample packing is enabled to prevent potential conflicts during multiprocessing.


djsaunde (approved):

salmanmohammadi (approved):
tyvm

PR #2866, opened 2025-07-05
respect shuffle_merged_datasets for single dataset too

winglian:
# Description
#2845 forced shuffling, but can have unexpected side effects when comparing newer to older experiments where we need a stable shuffle/split of train/test data. Fixes shuffle to respect shuffle_merged_datasets so we can disable the shuffle to repeat old experiments.

Trying to re-run a YAML with a single dataset results in different results because the eval split is now different.
Screenshot 2025-07-04 at 11 18 47 PM

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dataset merging logic to prevent shuffling when curriculum sampling is enabled or when shuffling of merged datasets is disabled in the configuration.

### DIFF
--- a/src/axolotl/utils/data/shared.py
+++ b/src/axolotl/utils/data/shared.py
@@ -526,7 +526,8 @@ def merge_datasets(datasets: list[Dataset], cfg: DictDefault) -> Dataset:
     if len(datasets) == 1:
         ds = datasets[0]
 
-        # Do not shuffle if curriculum sampling is enabled
+        # Do not shuffle if curriculum sampling is enabled or
+        # shuffle_merged_datasets is disabled
         if cfg.curriculum_sampling or not cfg.shuffle_merged_datasets:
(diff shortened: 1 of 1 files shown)


src/axolotl/utils/data/shared.py
```diff
@@ -527,7 +527,7 @@ def merge_datasets(datasets: list[Dataset], cfg: DictDefault) -> Dataset:
         ds = datasets[0]
 
         # Do not shuffle if curriculum sampling is enabled
```
NanoCode012:
```suggestion
        # Do not shuffle if curriculum sampling is enabled or
        # shuffle_merged_datasets is disabled
```

NanoCode012 (approved):
Good catch! I wasn't sure whether to add this check earlier as it was for merged datasets. Should we also adjust the Pydantic model field description?

PR #2879, opened 2025-07-07
Fix link in FSDP + QLoRA docs.

float-trip:
Old link led to README.md, but this part was refactored out a few months ago.

## Summary by CodeRabbit

* **Documentation**
  * Updated the documentation link for enabling FSDP in the Axolotl config to reference an internal section for easier navigation.


NanoCode012 (approved):
Thanks for the catch

PR #2885, opened 2025-07-08
add 2.7.0 torch images back to support vlllm

winglian:
vllm is pinned to torch==2.7.0, so we can't ship them with our torch 2.7.1 images. 

## Summary by CodeRabbit

* **Chores**
  * Expanded build configurations to include support for PyTorch 2.7.0 with CUDA 12.6.3 and Python 3.11 in automated workflows.
  * Updated package requirements to use `xformers==0.0.31.post1` for PyTorch 2.7+ and set `vllm` to require version 0.9.0 or higher.

### DIFF
--- a/.github/workflows/base.yml
+++ b/.github/workflows/base.yml
@@ -50,6 +50,13 @@ jobs:
             cuda_version: 12.6.3
             cudnn_version: ""
             python_version: "3.11"
+            pytorch: 2.7.0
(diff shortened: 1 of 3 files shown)


NanoCode012:
I don't think I see vllm images being built, only normal version @winglian 

https://hub.docker.com/r/axolotlai/axolotl-cloud/tags

Issue #2740, opened 2025-05-29
for grpo , TypeError: None is not a callable object


sankexin:
### Please check that this issue hasn't been reported before.

- [x] I searched previous [Bug Reports](https://github.com/axolotl-ai-cloud/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

train grpo or dapo , qwen3

### Current behaviour

```
[2025-05-29 08:49:00,092] [WARNING] [axolotl.prompt_strategies.base.load:33] [PID:4859] [RANK:0] unable to load strategy .gsm8k_grpo
[rank0]: Traceback (most recent call last):
[rank0]:   File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
[rank0]:     return _run_code(code, main_globals, None,
[rank0]:   File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
[rank0]:     exec(code, run_globals)
[rank0]:   File "/home/axolotl/src/axolotl/cli/train.py", line 125, in <module>
[rank0]:     fire.Fire(do_cli)
[rank0]:   File "/usr/local/lib/python3.10/dist-packages/fire/core.py", line 135, in Fire
[rank0]:     component_trace = _Fire(component, args, parsed_flag_args, context, name)
[rank0]:   File "/usr/local/lib/python3.10/dist-packages/fire/core.py", line 468, in _Fire
[rank0]:     component, remaining_args = _CallAndUpdateTrace(
[rank0]:   File "/usr/local/lib/python3.10/dist-packages/fire/core.py", line 684, in _CallAndUpdateTrace
[rank0]:     component = fn(*varargs, **kwargs)
[rank0]:   File "/home/axolotl/src/axolotl/cli/train.py", line 99, in do_cli
[rank0]:     return do_train(parsed_cfg, parsed_cli_args)
[rank0]:   File "/home/axolotl/src/axolotl/cli/train.py", line 51, in do_train
[rank0]:     dataset_meta = load_preference_datasets(cfg=cfg, cli_args=cli_args)
[rank0]:   File "/home/axolotl/src/axolotl/common/datasets.py", line 133, in load_preference_datasets
[rank0]:     train_dataset, eval_dataset = load_prepare_preference_datasets(cfg)
[rank0]:   File "/home/axolotl/src/axolotl/utils/data/rl.py", line 201, in load_prepare_preference_datasets
[...]

PR #2894, opened 2025-07-10
fix: customized dataset with simpo

ganler:
# Description

Current documentation and code for running SimPO is broken. This PR fixes it. 

## Motivation and Context

Similar to error reported in https://github.com/axolotl-ai-cloud/axolotl/issues/2740#issuecomment-2923595647

If we run custom datasets for KTO and DPO following the docu, `type` is not a dictionary.
and for dpo custom datasets, the prompt field is not applied.

## How has this been tested?

Tested using local runs.

works after the fix

image

## Summary by CodeRabbit

* **Documentation**
  * Updated RLHF dataset configuration documentation to use a nested mapping for custom dataset field mappings, improving clarity in DPO and KTO method examples.

* **Bug Fixes**
  * Corrected prompt formatting to use the configured prompt field name instead of a fixed key, ensuring compatibility with custom dataset configurations.

* **Refactor**
  * Expanded configuration validation to support both "dpo" and "simpo" values for RL datasets, improving flexibility in dataset handling.

### DIFF
--- a/docs/rlhf.qmd
+++ b/docs/rlhf.qmd
@@ -275,15 +275,14 @@ rl: dpo
 datasets:
   - path: ...
     split: train
-    type: user_defined.default
-
-    field_prompt: "prompt"
-    field_system: "system"
-    field_chosen: "chosen"
-    field_rejected: "rejected"
-    prompt_format: "{prompt}"
-    chosen_format: "{chosen}"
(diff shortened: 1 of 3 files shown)


winglian (approved):
lgtm, thanks!

PR #2897, opened 2025-07-10
use warmup_ratio as a better default than warmup steps since it's data dependent

winglian:
using hardcoded warmup steps can lead to too much warmup depending on the dataset. 

Screenshot 2025-07-10 at 12 03 33 PM

## Summary by CodeRabbit

* **Chores**
  * Updated training configuration files to specify warmup duration as a ratio of total training steps (`warmup_ratio: 0.1`) instead of a fixed number of steps. This change affects learning rate warmup scheduling across multiple example setups. No other user-facing changes were made.

### DIFF
--- a/examples/archived/cerebras/btlm-ft.yml
+++ b/examples/archived/cerebras/btlm-ft.yml
@@ -66,7 +66,7 @@ flash_optimum:
 gptq_groupsize:
 gptq_model_v1:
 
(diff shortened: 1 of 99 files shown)


NanoCode012 (approved):
Nice change. The defaults might be bad for those unaware.

PR #2900, opened 2025-07-10
Activation Offloading w CUDA Streams

winglian:
trl copied over the activation offloading context manager from torchtune, so we're going to use that to improve the speed when using activation offloading by ~5%.

Rather than overloading the `gradient_checkpointing` arg, we've separated this into a new `activation_offloading` arg as well. The `activation_offloading` arg also supports the prior method (`legacy`) in case there are edge cases with this, as well as `disk` for offloading activations to disk that was also supported. 

Although trl uses the standard HF gradient checkpointing layers and combines them with the new context manager, it doesn't seem to play nicely with other fused optimizations, so we're going to use the standard pytorch wrapping utility and wrap the models ourselves for this. 

## Summary by CodeRabbit

* **New Features**
  * Added support for activation offloading during training with multiple modes: default CPU offload with CUDA streams, legacy CPU offload, and disk offload for very large context lengths.
  * Introduced a new configuration option to enable activation offloading alongside gradient checkpointing.
  * Added a detailed user guide on gradient checkpointing and activation offloading in the documentation.

* **Improvements**
  * Enhanced garbage collection controls to better manage memory cleanup during training, including triggering GC before and after evaluation steps.
  * Updated configuration schema and validation to support activation offloading options and provide warnings for deprecated settings.

* **Bug Fixes**
  * Added validation to prevent enabling activation offloading without gradient checkpointing, reducing configuration errors.


src/axolotl/core/trainers/mixins/activation_checkpointing.py
```diff
@@ -0,0 +1,37 @@
...
+
+import contextlib
+
+from torch import nn
+from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import (
+    apply_activation_checkpointing,
+)
+from torch.distributed.fsdp.wrap import ModuleWrapPolicy
+from transformers import GradientCheckpointingLayer, Trainer
+from trl.models.activation_offloading import get_act_offloading_ctx_manager
```
djsaunde, line 13: noob q: why do we need both `apply_activation_checkpointing` and `get_act_offloading_ctx_manager`? Your PR description mentions just using the torch utilities, which is preferable if possible.

src/axolotl/utils/schemas/config.py
```diff
@@ -349,7 +354,7 @@ class AxolotlInputConfig(
     gradient_checkpointing: Literal["offload", "offload_disk"] | bool | None = Field(
         default=False,
         json_schema_extra={
-            "description": "Whether to use gradient checkpointing. Available options are: true, false, 'offload', 'offload_disk'. https://huggingface.co/docs/transformers/v4.18.0/en/performance#gradient-checkpointing"
+            "description": "Whether to use gradient checkpointing. Available options are: true, false, 'streams', 'offload', 'offload_disk'. https://huggingface.co/docs/transformers/v4.18.0/en/performance#gradient-checkpointing"
```
djsaunde: I think this needs to be removed now? "streams" isn't referenced anywhere it seems

src/axolotl/utils/schemas/validation.py
```diff
@@ -941,6 +941,28 @@ def check_offload_grad_checkpointing(self):
...
+    def check_gradient_checkpointing_w_offload(self):
+        if self.gradient_checkpointing == "offload":
+            LOG.warning(
+                "`offload` is deprecated for gradient_checkpointing, use `activation_offloading: true`"
+            )
+            self.gradient_checkpointing = True
+            self.activation_offloading = True
+        if self.gradient_checkpointing == "offload_disk":
+            LOG.warning(
+                "`offload_disk` is deprecated for gradient_checkpointing, use `activation_offloading: disk`"
```
djsaunde, line 945: This doesn't seem to be supported by the current schema, which just allows bools.

src/axolotl/utils/schemas/config.py
djsaunde: Also, I find mixing types (bool vs. str) to be a little confusing tbh.

src/axolotl/core/training_args_base.py
```diff
@@ -18,6 +18,10 @@ class AxolotlTrainingMixins:
     model_type: Optional[str] = field(
         default=None, metadata={"help": "HF model configuration model_type."}
     )
+    activation_offloading: bool | None = field(
```
djsaunde: Should this allow `Literal["disk", "legacy"]` as well?

djsaunde (approved):
Approved with notes.

src/axolotl/core/training_args_base.py
winglian: this is just to inform the trainer, but isn't necessary for those options.

src/axolotl/utils/schemas/validation.py
winglian: Screenshot 2025-07-10 at 4 27 22 PM
I think I might have fixed the schema between the time you wrote this and hit submit.

src/axolotl/core/trainers/mixins/activation_checkpointing.py
winglian: so apply activation/gradient checkpointing store the activations; usually in GPU RAM, and the get_act_offloading_ctx_manager is for offloading those to CPU RAM, so both are needed together.

src/axolotl/utils/schemas/config.py
winglian: I'm not sure the best way to offer flexibility and simple defaults. Feels like just setting it to true is a good default to the newer stream impl. Open to a better way of exposing this. Thoughts?

winglian:
@coderabbitai why would this PR break the KD plugin?

```
  File "/workspace/axolotl/src/axolotl/cli/train.py", line 49, in do_train
  File "/workspace/axolotl/src/axolotl/train.py", line 542, in train
    ) = setup_model_and_trainer(cfg, dataset_meta)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/axolotl/src/axolotl/train.py", line 501, in setup_model_and_trainer
    trainer = setup_trainer(
              ^^^^^^^^^^^^^^
  File "/workspace/axolotl/src/axolotl/utils/trainer.py", line 674, in setup_trainer
    return trainer_builder.build(total_num_steps)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/axolotl/src/axolotl/core/builders/causal.py", line 337, in build
    training_args = training_args_cls(  # pylint: disable=unexpected-keyword-arg
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: AxolotlTrainingArguments.__init__() got an unexpected keyword argument 'kd_ce_alpha'
```

fizzAI:
This PR seems to have somewhat broken behavior when activation checkpointing is disabled. On a model that breaks with it for some reason (i.e. Granite 3.3), even when I don't have `activation_checkpointing` in my config, the context manager for each train step is still assigned to the TRL manager which then crashes the train. If I patch out the conditional and make it always use a null context manager, it works again. I didn't have the time to debug any further than that, but it seems like somehow the activation checkpointing is getting set to true somewhere between the config being read and the trainer being inited.

See https://discord.com/channels/1104757954588196865/1111279858136383509/1394862889797681245 and below for more info here

PR #2903, opened 2025-07-11
move unmaintained examples to archive

winglian:
# Description

## Summary by CodeRabbit

* **Documentation**
  * Added a README to clarify that examples in the archived directory are no longer maintained and may not be functional, but are kept for reference.


salmanmohammadi (approved):
Maybe we should move these to their own repo?

PR #2904, opened 2025-07-11
Remove extra torch.compile call

djsaunde:
# Description

Title. Passing `torch_compile: true` to transformers allows the model to be compiled later in the `trainer.train()` codepath.

## Motivation and Context

This is needed to use `torch.compile` with fp8 training, as doing torch.compile at the point we currently do causes the HF trainer to skip `accelelerator.prepare` (! I think this probably breaks other torch compiled things silently?).

transformers PR for that is [placeholder].

## How has this been tested?

Manually.

## Summary by CodeRabbit

* **Refactor**
  * Updated accelerator settings and cache limits for improved training configuration.
  * Removed conditional model compilation during training to simplify behavior.
  * Improved consistency in handling accelerator configuration within training arguments.


src/axolotl/core/trainers/base.py
```diff
@@ -75,18 +75,6 @@ def __init__(
         if self.args.orpo_alpha:
             self.loss_fct = torch.nn.CrossEntropyLoss(reduction="none")
 
-    def _wrap_model(self, model, training=True, dataloader=None):
-        if self.args.torch_compile:
-            torch._dynamo.config.accumulated_cache_size_limit = (  # pylint: disable=protected-access
```
djsaunde, line 80: We can probably keep this.

djsaunde: I noticed that flash attention tends to exhaust this quickly due to variable length sequences needing a recompile (?) for each of the flash attn varlen calls. So even increasing the cache size like this won't be good.

winglian: We should probably figure out how to wrap the flash attention call with a torch.compile.disable decorator in a seperate pr

winglian (commented):
Changes so far lgtm

NanoCode012 (commented):
I think this should work. Like you've said, we already pass to the `training_args_kwargs` in base builder.

winglian (approved):
lgtm. thanks!

PR #2906, opened 2025-07-11
checkpoint model on first step callback

djsaunde:
# Description

Title. This is a good sanity check and ensures users don't waste a long time training, only for their model checkpoint to fail at the end of their run.

Note that we default the value for the config `save_first_step` to `True`. It's up for debate as to whether this is a good default.
- Pros: Strong sanity check, large time savings in the end for broken config combos
- Cons: Wastes a bit of time at the start of training

## Motivation and Context

As reported on our discord, some user runs end with errors on the final model checkpoint. This should limit the amount of frustration in these cases by failing faster.

## How has this been tested?

Manually with a few configs.

## Summary by CodeRabbit

* **New Features**
  * Added an optional configuration to control saving a model checkpoint after the first training step.
  * Introduced a callback to enable saving the model at the first step when configured.
  * Updated example configurations with commented hints for enabling first-step checkpoint saving.

* **Tests**
  * Added end-to-end tests verifying the "save first step" checkpoint behavior.
  * Updated numerous existing tests to include the new configuration parameter.


src/axolotl/utils/callbacks/__init__.py
```diff
@@ -141,6 +140,24 @@ def on_step_end(
...
+
+    def on_step_end(
+        self,
+        args: TrainingArguments,  # pylint: disable=unused-argument
+        state: TrainerState,
+        control: TrainerControl,
+        **_kwargs,
+    ) -> TrainerControl:
+        print("in SaveModelOnFirstStepCallback.on_step_end")
+        print(state.global_step)
```
winglian: Debugging

src/axolotl/utils/callbacks/__init__.py
```diff
@@ -141,6 +140,24 @@ def on_step_end(
...
+        if state.global_step == 1:
+            print("state.global_step == 0")
```
winglian: Another one

winglian (approved):
Good to go once prints from testing are removed.

tests/e2e/solo/test_save_first_step.py
```diff
@@ -0,0 +1,102 @@
...
+
+from axolotl.common.datasets import load_datasets
+from axolotl.train import train
+from axolotl.utils.config import normalize_config, validate_config
+from axolotl.utils.dict import DictDefault
+
+from ..utils import check_model_output_exists, with_temp_dir
+
+
+class TestSaveFirstStepCallback(unittest.TestCase):
```
NanoCode012: Does this need to be a `solo` test? It doesn't seem to patch anything.

NanoCode012 (approved):
I like this idea, but we may now get people complaining the training takes a bit longer to start (depending on model size).

I however would prefer this in as it would save people doing large runs time+resources.

tests/e2e/solo/test_save_first_step.py
NanoCode012: We may consider adding a LOG.DEBUG / info that we're doing a save on first step and how to disable?

salmanmohammadi:
I would suggest we add `# save_first_step: False` to all of our example configs to increase visibility - or even `# save_first_step: True # uncomment this to validate a test run with your config` as an alternative to auto-enabling.

tests/e2e/solo/test_save_first_step.py
djsaunde: good catch! it doesn't need to be in solo/

djsaunde:
> I would suggest we add `# save_first_step: False` to all of our example configs to increase visibility - or even `# save_first_step: True # uncomment this to validate a test run with your config` as an alternative to auto-enabling.

I think the latter is good alternative.

Or, we could raise a warning when it's not explicitly set? I'd prefer not to add warnings when possible, though, since we already have a lot of logs clutter.

winglian:
I rebased and pushed as there were some merge conflicts from merging other PRs.

salmanmohammadi (approved):

PR #2907, opened 2025-07-12
move the plugin post trainer create to the setup trainer

winglian:
# Description
Moving the plugin setup to setup trainer makes it so calling trainer.train() is all that needs to be done after. (e.g. colab)

## Summary by CodeRabbit

* **Refactor**
  * Improved the training workflow by reorganizing internal plugin-related operations for better process management. No changes to user-facing functionality.


NanoCode012 (approved):
Do we need to move, `post_train` to inside `execute_training` too?

PR #2908, opened 2025-07-12
Apply generic fused liger ce, cce, and tiledmlp for arbitrary models

winglian:
# Description

While Liger doesn't have patches for all models, we know most models have model forwards very similar to llama, so let's patch models with a generic flce if possible and log.

This is useful for experimenting if it works out of the box for newer models like LFM2

## Summary by CodeRabbit

* **New Features**
  * Added support for flexible, memory-efficient forward passes and loss computation for causal language models, including integration with PEFT and FSDP.
  * Introduced a configuration option to control the use of the original MLP in tiled MLP setups.
  * Expanded compatibility for cut cross-entropy and tiled MLP patches to additional model architectures.
  * Added a utility for consistent model class name handling across integrations.

* **Improvements**
  * Enhanced plugin mechanisms for LIGER and cut cross-entropy integrations, enabling more robust and dynamic patching based on model type.
  * Streamlined internal logic for dynamic model patching and class prefix derivation, improving maintainability and reducing errors.

* **Bug Fixes**
  * Improved error handling and logging for unsupported or misconfigured model types during plugin patching.


src/axolotl/integrations/liger/models/base.py
```diff
@@ -0,0 +1,191 @@
...
+        # Dynamically import the module and MLP class
+        module_path = f"transformers.models.{model_type}.modeling_{model_type}"
+        model_cls_prefix = "".join(
+            [part.capitalize() for part in model_type.split("_")]
+        )
+        module = __import__(module_path, fromlist=[f"{model_cls_prefix}ForCausalLM"])
+        model_cls = getattr(module, f"{model_cls_prefix}ForCausalLM")
+
+        model_cls.forward = lce_forward
+    # pylint: disable=duplicate-code
```
NanoCode012: Instead of doing this, could we just load from the auto class ? https://github.com/huggingface/transformers/blob/37f8b0b53512e6aae0cfd15746c133c101783178/src/transformers/models/auto/modeling_auto.py#L551

Same for CCE

I'm not sure if loading a generic one is always right as some models may have some slight differences.

winglian: Loading from the auto model list doesn't work as expected due to some weird python monkey patching quirks since those modules are loaded within a different scope 🤷

winglian: Yeah, agreed that it won't always work, but it allows users to see if it does if a model comes out before we can get to it. Most models use something similar to llama anyways

src/axolotl/integrations/liger/__init__.py
```diff
@@ -18,170 +18,5 @@
...
-                cross_entropy=cfg.liger_cross_entropy,
-                fused_linear_cross_entropy=cfg.liger_fused_linear_cross_entropy,
-                rms_norm=cfg.liger_rms_norm,
-                swiglu=cfg.liger_glu_activation,
-            )
-        else:
-            LOG.warning(
-                f"Unsupported model config type: {cfg.model_config_type}. Liger not applied."
-            )
+from .plugin import LigerPlugin  # pylint: disable=unused-import. # noqa: F401
```
djsaunde: If you add `LigerPlugin` to the `__all__` attribute, you can get rid of the linter disables. This also explicitly tells the import system that it's part of the public API

src/axolotl/integrations/liger/models/base.py
```diff
@@ -0,0 +1,191 @@
...
+            (masked), the loss is only computed for the tokens with labels in `[0, ..., config.vocab_size]`.
+
+        logits_to_keep (`int` or `torch.Tensor`, *optional*):
+            If an `int`, compute logits for the last `logits_to_keep` tokens. If `0`, calculate logits for all
+            `input_ids` (special case). Only last token logits are needed for generation, and calculating them only for that
+            token can save memory, which becomes pretty significant for long sequences or large vocabulary size.
+            If a `torch.Tensor`, must be 1D corresponding to the indices to keep in the sequence length dimension.
+            This is useful when using packed tensor format (single dimension for batch and sequence length).
+
+    Returns:
```
djsaunde: Nit: missing return

src/axolotl/integrations/liger/models/base.py
```diff
@@ -0,0 +1,191 @@
...
+from liger_kernel.transformers.trainer.orpo_trainer import _FSDPForwardRedirection
+from liger_kernel.utils import PEFT_AVAILABLE
+from peft.utils import ModulesToSaveWrapper
+from torch.distributed.fsdp import FullyShardedDataParallel
+from transformers.modeling_outputs import CausalLMOutputWithPast
+
+
+def lce_forward(
+    self,
+    *args,
```
djsaunde, line 20: Do we need to support possibly variable-length args here?

src/axolotl/utils/schemas/config.py
```diff
@@ -563,6 +563,13 @@ class AxolotlInputConfig(
         },
     )
 
+    tiled_mlp_use_original_mlp: bool | None = Field(
```
djsaunde, line 579: I see this defaults as `None`; are there any implications of not using the original MLP? I'm guessing that certain models' implementation could be sensitive to this?

src/axolotl/monkeypatch/tiled_mlp.py
```diff
@@ -45,11 +45,12 @@ def tiled_mlp_forward(self, x):
...
-                self.down_proj.weight,
-                self.gate_proj.weight,
-                self.up_proj.weight,
-            ]
+            if not self._compute_params:  # pylint: disable=protected-access
+                self._compute_params = [  # pylint: disable=protected-access
+                    p for p in self.parameters() if p.requires_grad
+                ]
+
+            compute_params = self._compute_params  # pylint: disable=protected-access
```
djsaunde, line 53: Are these pylint disables needed? We're accessing class members.

djsaunde (approved):
Cool stuff!

Some questions / comments but otherwise looks good.

src/axolotl/utils/schemas/config.py
winglian: yeah, I forget which arch i was looking at previously, but there are some MLPs with just two parameters or are fused implementations.

src/axolotl/monkeypatch/tiled_mlp.py
winglian: yeah, it was giving me some errors when I was linting

src/axolotl/integrations/liger/models/base.py
winglian: I was trying to seperate what's required  from not, so generally all non-kwargs were required.

src/axolotl/integrations/liger/models/base.py
winglian: going to drop this, the annotation lists the types, and upstream in liger, they just have a code sample there 🤷

src/axolotl/integrations/liger/__init__.py
winglian: thanks, TIL

PR #2911, opened 2025-07-12
fix for upstream refactor of KwargsForCausalLM

winglian:
# Description

changes for upstream refactor https://github.com/huggingface/transformers/pull/39120 

## Summary by CodeRabbit

* **Refactor**
  * Improved compatibility with different versions of external libraries through conditional imports. No changes to user-facing functionality.


src/axolotl/integrations/kd/kernels/models.py
```diff
@@ -6,15 +6,21 @@
...
-    """
+    class TransformersKwargs(FlashAttentionKwargs, LossKwargs):
+        """
+        placeholder kwargs for hf model classes
+        """
+
+except ImportError:
+    from transformers.utils.generic import (  # type: ignore[no-redef]
+        TransformersKwargs,
+    )
```
NanoCode012: Should this not be the reversed? try to import upstream generic then fallback to making it ourselves

Alternatively, just remove the import typing?

winglian: I don't think it matters on order since one is in the current t release and the other is in main, so I have no preference

winglian:
closing.., although UI shows this was merged already yesterday? 🤷 

Screenshot 2025-07-16 at 5 06 59 PM

PR #2915, opened 2025-07-14
fix ddp for incorrect steps

winglian:
#2760 caused a bug in the batch size logic, resulting in any non-fsdp training that used multiple (N) gpu's to take N-times more steps because it didn't account for that anymore. This should go out as an immediate hotfix

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automatic configuration of distributed data parallel (DDP) settings based on environment size, ensuring more reliable initialization for distributed setups.  
* **Tests**
  * Added tests to verify correct batch size calculation across different multi-GPU training configurations.

### DIFF
--- a/src/axolotl/utils/config/__init__.py
+++ b/src/axolotl/utils/config/__init__.py
@@ -115,6 +115,7 @@ def normalize_config(cfg):
         "chrf",
     ]
     choose_device(cfg)
+    cfg.ddp = cfg.ddp if cfg.ddp is not None else cfg.world_size != 1
     if cfg.world_size != 1:
         cfg.device_map = {"": int(os.environ.get("LOCAL_RANK", 0))}
         if cfg.fsdp or cfg.fsdp_config or cfg.ddp:
--- a/tests/test_train.py
+++ b/tests/test_train.py
@@ -1,0 +1,44 @@
+"""Test for batch size calculation for multi-gpu training."""
+
+import pytest
+
+from axolotl.utils.config import normalize_config, validate_config
+from axolotl.utils.dict import DictDefault
+
+
+@pytest.fixture(name="train_base_cfg")
+def fixture_train_base_cfg():
+    return DictDefault(
+        base_model="gpt2",
+        learning_rate=1e-3,
+        datasets=[
+            {
+                "path": "mhenrichsen/alpaca_2k_test",
+                "type": "alpaca",
+            },
+        ],
+        micro_batch_size=2,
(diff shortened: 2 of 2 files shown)


src/axolotl/utils/config/__init__.py
```diff
@@ -115,6 +115,7 @@ def normalize_config(cfg):
         "chrf",
     ]
     choose_device(cfg)
+    cfg.ddp = cfg.ddp if cfg.ddp is not None else cfg.world_size != 1
     if cfg.world_size != 1:
         cfg.device_map = {"": int(os.environ.get("LOCAL_RANK", 0))}
         if cfg.fsdp or cfg.fsdp_config or cfg.ddp:
```
NanoCode012, line 121: I think I left one question on the linked PR whether we need to check cfg.deepspeed also for this condition? If yes, do we even need this condition? We previously always set this even on DS.

src/axolotl/utils/config/__init__.py
```diff
@@ -115,6 +115,7 @@ def normalize_config(cfg):
         "chrf",
     ]
     choose_device(cfg)
+    cfg.ddp = cfg.ddp if cfg.ddp is not None else cfg.world_size != 1
```
salmanmohammadi, line 118: I was confused by this line, do you know why removing it causes the bug?

NanoCode012: I think, users never set this -> model is not moved to correct device & batch size is not set per following condition.

salmanmohammadi: Shouldn't users specify `cfg.ddp` if they wish to use DDP? Maybe we should error out if they haven't specified a distributed config but `world_size !=1`? I think I wanted to add this behaviour in my original PR but missed it.

winglian: I think the intent of cfg.ddp is that it covers all data parallel. It is possible to specify multiple gpus with model parallel, but in that case it's single process so this still covers it. Right now this is a major issue since if anyone is doing an 8-gpu run that normally takes 4 hours, it would currently you take 32 hours  💸💸💸

salmanmohammadi: Oh man. OK stamping to unblock this but I think the semantics of `cfg.ddp` vs `cfg.fsdp_config` need to be clearer.

salmanmohammadi (approved):

PR #2916, opened 2025-07-14
fix xformers version for python 2.6

winglian:
Since we're only supporting cu126 for torch 2.6.0, we need to use the post3 release and manually specify the index url to install from.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the required version of the `xformers` package for compatibility with PyTorch 2.6.
  * Added support for CUDA 12.6 wheels.

* **Documentation**
  * Clarified comments to indicate support for PyTorch 2.6.0+cu126 only.


NanoCode012:
What's wrong with CU124? Is it due to FA2 being broken?

Issue #2914, opened 2025-07-13
Remove Limiter on Multiprocessing during Dataset Tokenization


VarunGumma:
### ⚠️ Please check that this feature request hasn't been suggested before.

- [x] I searched previous [Ideas in Discussions](https://github.com/axolotl-ai-cloud/axolotl/discussions/categories/ideas) didn't find any similar feature requests.
- [x] I searched previous [Issues](https://github.com/axolotl-ai-cloud/axolotl/labels/enhancement) didn't find any similar feature requests.

### 🔖 Feature description

Remove the limiter on the multiprocessing during dataset tokenization.

### ✔️ Solution

In line 49 of `datasets.py` (in the `process` method), there is a limiter on the multiprocessing during tokenization in the form of:

```python 
num_proc = min(64, self.process_count if self.process_count else os.cpu_count())
``` 

Please remove the `min` function from it and make it as:

```python
num_proc = self.process_count if self.process_count else os.cpu_count()
```

This enables all the designated processes to take part in the tokenization. For a CPU with 96 cores, only 64 were being used initially, and the speedup achieved with this modification was very significant for large scale datasets with long-context inputs. 

### ❓ Alternatives

NA

### 📝 Additional Context

NA

### Acknowledgements

- [x] My issue title is concise, descriptive, and in title casing.
- [x] I have searched the existing issues to make sure this feature has not been requested yet.
- [x] I have provided enough information for the maintainers to understand and evaluate this request.

NanoCode012:
Weird, I thought we already opened it up in a recent PR. Would you be interested in making a PR for this?

You may also need to adjust `dataset_processes` in `AxolotlInputConfig` schema. Seems like we have a limit there too?

VarunGumma:
Hi @NanoCode012 , I have opened a PR with the required changes (#2918 ).

Please take a look and merge it if you find it appropriate. Thank you!

Issue #2913, opened 2025-07-13
Improve Processed Dataset Saving Speed and Shards


VarunGumma:
### ⚠️ Please check that this feature request hasn't been suggested before.

- [x] I searched previous [Ideas in Discussions](https://github.com/axolotl-ai-cloud/axolotl/discussions/categories/ideas) didn't find any similar feature requests.
- [x] I searched previous [Issues](https://github.com/axolotl-ai-cloud/axolotl/labels/enhancement) didn't find any similar feature requests.

### 🔖 Feature description

For some very big datasets, the number of shards while saving is way too high and the saving takes too long: 

- Add multiprocessing while saving processed dataset to the disk.
- Provide an interface/config key to save the processed dataset in a particular number of shards. 

### ✔️ Solution

In the `save_preprocessed_dataset` function in `shared.py`, please set `num_proc` and `num_shards` arguments in the save to disk function as shown in the example below. (I have hard-coded it as 5 in my usecase, but please make it flexible and part of the `cfg`). These arguments have given me tremendous speedups while saving the dataset. 

```python
def save_preprocessed_dataset(
    cfg: DictDefault,
    dataset: Dataset,
    dataset_hash: str,
    split: str,
) -> None:
    """Save preprocessed dataset to disk and optionally push to the HF Hub."""
    num_workers = cfg.dataset_processes ## TODO: Move this here
    prepared_ds_path = get_prepared_dataset_path(cfg, dataset_hash)

    if isinstance(dataset, IterableDataset):
        ds_from_iter = Dataset.from_generator(
            functools.partial(_generate_from_iterable_dataset, dataset),
            features=dataset.features,
            num_proc=num_workers,
            split=split,
            gen_kwargs={
                "worker_id": list(range(num_workers)),
                "num_workers": [num_workers] * num_workers,
            },
        )
        ds_from_iter.save_to_disk(
            str(prepared_ds_path), num_proc=num_workers, num_shards=5 
        ) ## TODO: Add arguments here
    else:
        os.makedirs(prepared_ds_path, exist_ok=True)
        dataset.save_to_disk(
            str(prepared_ds_path), num_proc=num_workers, num_shards=5
        ) ## TODO: Add arguments here
    if cfg.push_dataset_to_hub:
        LOG.info(
            "Pushing merged prepared dataset to Huggingface hub at "
            f"{cfg.push_dataset_to_hub} (version {dataset_hash})...",
            main_process_only=False,
        )
[...]

Issue #2912, opened 2025-07-13
Qwen Tokenizer Missing Attribute


VarunGumma:
### Please check that this issue hasn't been reported before.

- [x] I searched previous [Bug Reports](https://github.com/axolotl-ai-cloud/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

The data preprocessing should run successfully, and set the required tokens. 

### Current behaviour

The preprocessing fails with the following error:

``` python
raise AttributeError(f"{self.__class__.__name__} has no attribute {key}")
AttributeError: Qwen2TokenizerFast has no attribute eod_id
```

### Steps to reproduce

set `is_qwen_derived_model` as True

### Config yaml

```yaml
Just set `set `is_qwen_derived_model` as True in any qwen SFT yaml examples.
```

### Possible solution

Line 195 in `load_tokenizer` function of `tokenizer.py` seems to have a typo or bug which states `tokenizer.eod_id`. However, Qwen tokenizer does not have such an attribute. For more context, I am providing the entire `if` block where the bug occurs. I think it should have been `eos_token_id` in place of `eod_id`, but again, I am not sure given the logic in the block. 

```python
    if cfg.is_qwen_derived_model:
        token_ids = ["bos_token_id", "eos_token_id", "pad_token_id", "unk_token_id"]
        for attr_name in token_ids:
            if getattr(tokenizer, attr_name) is None:
                setattr(tokenizer, attr_name, tokenizer.eod_id)

        token_names = ["bos_token", "eos_token", "pad_token", "unk_token"]
        for attr_name in token_names:
            if getattr(tokenizer, attr_name) is None:
                setattr(tokenizer, attr_name, "<|endoftext|>")
```

### Which Operating Systems are you using?

- [x] Linux
- [ ] macOS
- [ ] Windows

### Python Version

3.12

### axolotl branch-commit

main/d6e4a611e5eaced05cf295f90095ffb71e2448a5

### Acknowledgements

- [x] My issue title is concise, descriptive, and in title casing.
- [x] I have searched the existing issues to make sure this bug has not been reported yet.
- [x] I am using the latest version of axolotl.
- [x] I have provided enough information for the maintainers to reproduce and diagnose the issue.

NanoCode012:
Thanks for the Issue. I believe I added this check for their original model Qwen(1). I don't think we need this section for newer models. See an old issue for ref: https://github.com/QwenLM/Qwen3/issues/48

[...]

Issue #2753, opened 2025-06-03
Improve default dataset_processes behavior


djsaunde:
### ⚠️ Please check that this feature request hasn't been suggested before.

- [x] I searched previous [Ideas in Discussions](https://github.com/axolotl-ai-cloud/axolotl/discussions/categories/ideas) didn't find any similar feature requests.
- [x] I searched previous [Issues](https://github.com/axolotl-ai-cloud/axolotl/labels/enhancement) didn't find any similar feature requests.

### 🔖 Feature description

Currently, we use something like `min(64, self.process_count if self.process_count else os.cpu_count())` in a few places in the codebase. I think we can use `os.cpu_count()` or some function of it to tokenize / save / load / etc. the dataset faster. It's worth doing some testing on this to see what time savings we can have.

### ✔️ Solution

Experiment with different settings on difference instances. My gut feeling is that week can make the number of CPUs the default process count.

### Acknowledgements

- [x] My issue title is concise, descriptive, and in title casing.
- [x] I have searched the existing issues to make sure this feature has not been requested yet.
- [x] I have provided enough information for the maintainers to understand and evaluate this request.

PR #2918, opened 2025-07-14
Improve Dataset Processing Multiprocessing, Sharding, and Qwen Tokenizer Bug Fix.

VarunGumma:
This pull request introduces the following improvements and fixes:

# Description

- Removes the hard limiter on multiprocessing during dataset tokenization, allowing full utilization of available CPU cores or user-specified process count. 

- Improves the saving speed of processed datasets by enabling configurable multiprocessing and sharding, making these parameters flexible and part of the configuration. 

- Fixes a bug in the Qwen tokenizer integration where a missing or incorrect attribute (`eod_id`) caused preprocessing failures when using Qwen-derived models. 

## Motivation and Context

- **Multiprocessing Limiter Removal**: Previously, tokenization was limited to a maximum of 64 processes, even on machines with more cores. This change removes the min(64, ...) limiter, enabling all available or user-specified CPU cores to be used, resulting in significant speedups for large-scale, long-context datasets.  (#2914) 

- **Configurable Sharding and Multiprocessing on Save**: Saving large processed datasets was slow due to excessive sharding and non-configurable multiprocessing. By exposing num_proc and num_shards as configuration options, users can now optimize saving speed for their hardware and dataset size. (#2913)

- **Qwen Tokenizer Attribute Bug**: The Qwen tokenizer integration tried to access a non-existent `eod_id` attribute, causing failures during preprocessing. The logic has been corrected to use the appropriate token attributes, ensuring compatibility and successful preprocessing for Qwen-derived models. (#2912)

## How has this been tested?

- Unit and integration tests were run for dataset tokenization and saving, verifying that:
    - The correct number of processes are spawned during tokenization, matching the user configuration or system CPU count.
    - Datasets are saved with the specified number of shards and processes, and the saved data is consistent and readable.

- Manual testing was performed on:
    - A machine with more than 64 CPU cores to confirm all cores are utilized during tokenization.
    - Qwen SFT YAML configurations to ensure preprocessing completes without attribute errors and the correct tokens are set.

## Types of changes

- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

## Social Handles (Optional)

- [Personal Website](https://varungumma.github.io/)
- [Twitter/X](https://x.com/VarunGumma23)
- [GitHub](https://github.com/VarunGumma/)

## Summary by CodeRabbit

* **New Features**
  * Added support for specifying the number of shards when saving prepared datasets.

* **Improvements**
  * Enhanced dataset saving to allow parallel processing and sharding for improved performance.
  * Updated configuration options for dataset processing to provide more flexibility in resource usage.
  * Improved handling of special token IDs for certain tokenizers, ensuring more robust error messaging.
  * Removed upper limit on process count for dataset processing to better utilize available CPU resources.
  * Refined default process count determination to better adapt to different runtime environments.


winglian (approved):
There changes all seem sane enough. iirc, we had the hard cap in place because we had noticed that because a lot of modern CPUs use hyperthreading, using os.count actually was worse/slower than simply capping. One thing we could consider is using `os.cpu_count // 2` instead

src/axolotl/utils/schemas/config.py
```diff
@@ -203,9 +207,7 @@ class AxolotlInputConfig(
         },
     )
     dataset_processes: int | None = Field(
-        default=min(
-            int(os.environ.get("AXOLOTL_DATASET_PROCESSES", 32)), os.cpu_count()
-        ),  # type: ignore[type-var]
+        default=int(os.environ.get("AXOLOTL_DATASET_PROCESSES", os.cpu_count())),  # type: ignore[type-var]
```
NanoCode012: Hm, back to this. I notice we try to catch this (due to old code) in few places.
1. Here `config.py`
2. `normalize_config` 
3. Within the ds `process` `src/axolotl/datasets.py`
4. Chat ds parser `src/axolotl/core/datasets/chat.py` (need to also remove limiter here)

I think we can remove all except first as it'll run first in our validator.

src/axolotl/utils/schemas/config.py
```diff
@@ -193,6 +193,10 @@ class AxolotlInputConfig(
         json_schema_extra={"description": "Index of shard to use for whole dataset"},
     )
     skip_prepare_dataset: bool | None = False
+    num_save_shards: int | None = Field(
```
NanoCode012: I'm thinking whether we want `ds` in the name to signify it's for datasets only.

NanoCode012: Since this defaults to None, does `dataset.save_to_disk` support None if we pass it? Just to be sure we're not overriding their defaults.

src/axolotl/loaders/tokenizer.py
```diff
@@ -189,10 +189,15 @@ def _load_mistral_common_tokenizer(cfg: DictDefault):
...
+            "Please check the tokenizer configuration."
+        )
         token_ids = ["bos_token_id", "eos_token_id", "pad_token_id", "unk_token_id"]
         for attr_name in token_ids:
             if getattr(tokenizer, attr_name) is None:
-                setattr(tokenizer, attr_name, tokenizer.eod_id)
+                setattr(tokenizer, attr_name, default_tok_id)
 
         token_names = ["bos_token", "eos_token", "pad_token", "unk_token"]
         for attr_name in token_names:
```
NanoCode012: I think we may need to change the `"<|endoftext|>"` below to the `eos_token` value?

However, now that I step back, I'm wondering if we even need this section. The user should be setting it under `special_tokens: ` instead of us handling it like this.

src/axolotl/utils/schemas/config.py
VarunGumma: yes, I think the name `num_save_ds_shards` is better and to be more specific `prepared_dataset_num_shards`. I would go with the second, as it is more explicit as well. What's your opinion?

src/axolotl/loaders/tokenizer.py
VarunGumma: Actually true, I feel the whole block can be removed as the newer Qwen models don't use it at all, and I am not happy hardcoding all the unavailable tokens as `eos`/`eos_id`. Again, older Qwen models might need this block.

src/axolotl/utils/schemas/config.py
VarunGumma: I just noticed that another Issue (#2753) also discusses this issue. Tagging it here. 
Btw, my fix already addresses (3) in `process` `src/axolotl/datasets.py`.  
I made the modifications for (2) and (4). Pushing it now.

src/axolotl/loaders/tokenizer.py
NanoCode012: Hm, could we change this check to, `hasattr(tokenizer, "eod_id"): <- sign of qwen base` before the 2 for loops. Then, this section would only run for the old qwen base. We can then keep the original implementation without having to fallback to eos.

VarunGumma: yes, that seems like a good idea. 

The best `if` case here could be `if cfg.is_qwen_derived_model and hasattr(tokenizer, "eod_id")`, and revert the code block to what we had earlier. But maybe the `"<|endoftext|>"` should also be changed to `tokenizer.eos_token` below

src/axolotl/utils/schemas/config.py
VarunGumma: @NanoCode012 , Here are some numbers around the `num_shards` and `num_proc` argument on a 24 core CPU. Thanks for the reminder, I realized, we have to set the `max_shard_size=None` to avoid issues later. 

```python
Saving the dataset (15/15 shards): 100%|██████████████████████████████████████████████████████████████████| 349317/349317 [00:06<00:00, 56775.48 examples/s]
 | > Dataset saved in 6.16 seconds. (num_shards=None, num_proc=None, max_shard_size=None)
Saving the dataset (24/24 shards): 100%|█████████████████████████████████████████████████████████████████| 349317/349317 [00:00<00:00, 490461.43 examples/s]
 | > Dataset saved in 0.74 seconds. (num_proc=24, num_shards=None, max_shard_size=None)
Saving the dataset (5/5 shards): 100%|███████████████████████████████████████████████████████████████████| 349317/349317 [00:01<00:00, 240478.69 examples/s]
 | > Dataset saved in 1.47 seconds. (num_proc=24, num_shards=5, max_shard_size=None)
 ```

Comment (NanoCode012) [reply src/axolotl/utils/schemas/config.py]: Thanks for the numbers, By default, how many shards would it attempt to save?

File: src/axolotl/loaders/tokenizer.py (hunk 3 above)
Comment (NanoCode012) [reply src/axolotl/loaders/tokenizer.py]: > But maybe the "<|endoftext|>" should also be changed to tokenizer.eos_token below

In this block, we are checking and maybe setting `eos_token` too (so  `tokenizer.eos_token` may be empty). I'll say to just keep it for legacy unless another reviewer think we can just remove this.

File: src/axolotl/utils/schemas/config.py (hunk 2 above)
Comment (VarunGumma) [reply src/axolotl/utils/schemas/config.py]: That depends. I think the `max_shard_size` is default at "500Mb". So, for a very long context dataset with ~9M data points for me, it creates 1000+ shards. In the example shown before, it saves 15 by default.

File: src/axolotl/utils/schemas/config.py (hunk 1 above)
Comment (VarunGumma) [reply src/axolotl/utils/schemas/config.py]: @NanoCode012, I have made the all discussed fixes and pushed them along in a new commit. Please review it soon, as this feature will really help me and lot others as well.

File: src/axolotl/loaders/tokenizer.py (hunk 3 above)
Comment (VarunGumma) [reply src/axolotl/loaders/tokenizer.py]: yes, you are absolutely correct! I have kept it as is in the latest commit, and just converted the check to `if cfg.is_qwen_derived_model and hasattr(tokenizer, "eod_id")`. This avoids the error if someone adds `is_qwen_derived_model` to their config for the newer qwen models.

File: src/axolotl/core/datasets/chat.py [hunk 4]
```diff
@@ -41,14 +41,11 @@ def map_fn(ex):
                 )
             return ex.tokenized(model_transform)
 
-        process_or_cpu_count: int = (
-            process_count or os.cpu_count()  # type: ignore[assignment]
-        )
-        num_proc = min(32, process_or_cpu_count)
+        process_or_cpu_count = process_count if process_count else os.cpu_count()
```
NanoCode012: We pass `cfg.dataset_processes` to process_count already. 

https://github.com/axolotl-ai-cloud/axolotl/blob/2c408b5c5eb2cc152e310ca22928eefaa91c3ee2/src/axolotl/utils/data/wrappers.py#L84

`cfg.dataset_processes` would default to the os cpu count as well in the validator stage, so we don't need to have this check here. Just pass `process_count` to the `num_proc` below.

src/axolotl/datasets.py
```diff
@@ -46,7 +46,7 @@ def __init__(  # pylint: disable=super-init-not-called
 
     def process(self, dataset):
         features = dataset.features.keys()
-        num_proc = min(64, self.process_count if self.process_count else os.cpu_count())
+        num_proc = self.process_count if self.process_count else os.cpu_count()
```
NanoCode012: Same as above.

src/axolotl/loaders/tokenizer.py
```diff
@@ -188,7 +188,8 @@ def _load_mistral_common_tokenizer(cfg: DictDefault):
         tokenizer.padding_side = "left"
 
     # Qwen base only has single token, so we need to set the special tokens
-    if cfg.is_qwen_derived_model:
+    # the following check is for Qwen1 series models of base models
```
NanoCode012:
```suggestion
    # the following check is for Qwen1 base models
```

NanoCode012 (changes requested):
Thanks, this looks almost good to go. Could you revert your changes to Colab too and check the comment below?

src/axolotl/core/datasets/chat.py
VarunGumma: Sure, I will make this change.

src/axolotl/datasets.py
VarunGumma: Sorry, I did not get this. Do you mean I just pass `self.process_count` instead of having an `if` case in line 49?

NanoCode012: Yup!

src/axolotl/loaders/tokenizer.py
VarunGumma: fixed! Please take a look at the final commit.

src/axolotl/datasets.py
VarunGumma: fixed!

src/axolotl/core/datasets/chat.py
VarunGumma: fixed!

VarunGumma:
@winglian and @NanoCode012 , can you please take a look at the changes and please let me know if you need any more information. Please approve it if you all is good.

NanoCode012:
Let's wait for CI a bit and I'll run some manual checks on my end as well

NanoCode012 (approved):
Fixed lint and updated handling of default `dataset_processes`

src/axolotl/utils/schemas/config.py
```diff
@@ -1199,3 +1206,16 @@ def default_dataloader_opts(cls, data):
...
 
         return data
+
+    @model_validator(mode="before")
+    @classmethod
+    def default_dataset_processes(cls, data):
+        if data.get("dataset_processes") is None:
+            if axolotl_dataset_processes := os.environ.get("AXOLOTL_DATASET_PROCESSES"):
+                data["dataset_processes"] = int(axolotl_dataset_processes)
+            elif runpod_cpu_count := os.environ.get("RUNPOD_CPU_COUNT"):
```
winglian, line 1216: 🔥 🙌

PR #2926, opened 2025-07-15
basic torchao fp8 mixed precision training

djsaunde:
# Description

This PR updates the existing logic gated by the `fp8` config to work correctly with the current state of accelerate torchao fp8 support. It also adds the `fp8_enable_fsdp_float8_all_gather` config which is passed through to the `Float8LinearConfig`, and can speed up runs with FSDP enabled by up to ~50% depending on the configuration.

## Motivation and Context

`torchao.float8.convert_to_float8_training` converts `Linear` layers in the given model to `Float8Linear` layers with a cast config (e.g., `cast_configs=i:dyn_ten_e4m3,w:dyn_ten_e4m3,go:dyn_ten_e5m2"`). Model weights remain in the dtype specified in the user's config (e.g., bfloat16), but ultimately get converted to fp8 at matmul time (forward and backward); see [here](https://github.com/pytorch/ao/blob/main/torchao/float8/float8_linear.py).

This kind of fp8 training should be faster than bfloat16 training with `torch.compile` applied. `torch.compile` currently breaks (multi-GPU only + DDP) with:

```
[rank0]: RuntimeError: 
[rank0]: During the backward, we encountered a tensor subclass where we guessed its
[rank0]: metadata incorrectly.

[rank0]: Expected metadata: {'_orig_dtype': torch.bfloat16, '_linear_mm_config': LinearMMConfig(output=ScaledMMConfig(emulate=False, use_fast_accum=True, fp8_output=False, pad_inner_dim=False), grad_input=ScaledMMConfig(emulate=False, use_fast_accum=False, fp8_output=False, pad_inner_dim=False), grad_weight=ScaledMMConfig(emulate=False, use_fast_accum=False, fp8_output=False, pad_inner_dim=False)), '_gemm_input_role': <GemmInputRole.WEIGHT: 'weight'>, '_axiswise_dim': None}, expected type: <class 'torchao.float8.float8_tensor.Float8Tensor'>
```

## How has this been tested?

- [ ] Manual testing
  - [x] Single GPU
  - [x] Multi GPU (DDP)
  - [x] Multi GPU (FSDP)
  - [x] Single GPU + torch.compile
  - [ ] Multi GPU (DDP) + torch.compile
  - [x] Multi GPU (FSDP) + torch.compile
- [x] Add smoke tests

Not 100% sure, but some of the marked tested settings seem flaky depending on the run / hardware?

## TODO:

- [ ] Fix torch.compile metadata error for multi-GPU settings
- [x] Enumerate all other common features this is compatible with
- [x] Understand performance gain / degradation depending on features / hyperparams
- [x] Add docs

## Summary by CodeRabbit

* **New Features**
  * Added comprehensive documentation on mixed precision training, including FP16, BF16, and experimental FP8 formats.
  * Introduced support for FP8 mixed precision training with new configuration options, including FSDP float8 all-gather.
  * Provided an example configuration for training LLaMA-3 3B with FP8 and FSDP2.
* **Bug Fixes**
  * Added validation and warnings for optimal FP8 training setup.
* **Tests**
  * Implemented new end-to-end tests for FP8 training on both single and multi-GPU setups, including FSDP2 and LoRA configurations.
  * Added unit tests to ensure compatibility of FP8-related monkeypatches with upstream code.
* **Documentation**
  * Updated website sidebar to include new mixed precision and gradient accumulation guides.


src/axolotl/loaders/patch_manager.py
```diff
@@ -150,11 +150,17 @@ def _apply_model_specific_patches(self):
     def _apply_fp8_patches(self):
         """Apply patches for FP8 support."""
         if self.cfg.fp8:
+            from axolotl.monkeypatch.accelerate_torchao_fsdp_check import (
```
winglian: since this is FSDP specific, should we gate `if self.cfg.fp8 and self.cfg.fsdp_config:`?

djsaunde: It ~should work~ works outside of FSDP ~IIUC~

src/axolotl/monkeypatch/accelerate_torchao_fsdp_check.py
```diff
@@ -0,0 +1,77 @@
...
+from axolotl.monkeypatch.utils import detab_code
+from axolotl.utils.logging import get_logger
+
+LOG = get_logger(__name__)
+
+ORIGINAL_ACCELERATE_CODE = """
+    if self.fp8_backend == "AO" and self.state.fsdp_plugin.cpu_ram_efficient_loading:
+"""
+
+PATCHED_ACCELERATE_CODE = """
```
S1ro1: This should only exist in one version of accelerate (1.8) so maybe there's a bit smarted way to do this? Not exactly sure if possible downstream, though it's merged rn and 1.9 is gonna be released soon.

src/axolotl/loaders/patch_manager.py
winglian: Should we remove fsdp from the config name on that case?

src/axolotl/monkeypatch/accelerate_torchao_fsdp_check.py
winglian: Yeah. Merged this morning? https://github.com/huggingface/accelerate/pull/3677

S1ro1: Yes, this PR exactly, new release should be out this week, so only 1.8 is affected.

src/axolotl/loaders/patch_manager.py
djsaunde: I'm removing the check for FSDP here, so I think the naming makes sense

src/axolotl/monkeypatch/accelerate_torchao_fsdp_check.py
djsaunde: Given that the change will be in the next release, and this works, I'm happy to keep it and just remove when we update to 1.9.

djsaunde:
Quick notes from today:

Testing llama 3.1 8B, for large enough sequence length, we see speed improvements (~10-30% more iter/s) using torchao fp8 + torch.compile + either SDPA or flex attention on a single GPU, over several different settings of sequence length, batch size. Flash attention cannot be torch compiled and results in slower overall training times when used in conjuction with torchao fp8.

Multi-GPU is another story.  FSDP2 + torchao fp8 + torch.compile results in significantly (~25%) fewer iter/s for 2x H100s, while DDP just fails with [this error](https://gist.github.com/djsaunde/7225c0afe9bd96b94e67158cc75f2fac).

S1ro1:
> Testing llama 3.1 8B, for large enough sequence length, we see speed improvements (~10-30% more iter/s) using torchao fp8 + torch.compile + either SDPA or flex attention on a single GPU, over several different settings of sequence length, batch size. Flash attention cannot be torch compiled and results in slower overall training times when used in conjuction with torchao fp8.
> 
> Multi-GPU is another story. FSDP2 + torchao fp8 + torch.compile results in significantly (~25%) fewer iter/s for 2x H100s, while DDP just fails with [this error](https://gist.github.com/djsaunde/7225c0afe9bd96b94e67158cc75f2fac).

So there is this very helpful [chart](https://github.com/pytorch/ao/blob/main/torchao/float8/README.md#performance) in torchao docs that should tell a lot about where to expect what speedup. Regarding distributed, I haven't really tested more configurations, but on a single node with large enough seq-len (8192) you can see quite a nice improvement. Have a script to reproduce that in Accelerate: [here](https://github.com/huggingface/accelerate/tree/main/examples/fsdp2)

djsaunde:
> > Testing llama 3.1 8B, for large enough sequence length, we see speed improvements (~10-30% more iter/s) using torchao fp8 + torch.compile + either SDPA or flex attention on a single GPU, over several different settings of sequence length, batch size. Flash attention cannot be torch compiled and results in slower overall training times when used in conjuction with torchao fp8.
> > Multi-GPU is another story. FSDP2 + torchao fp8 + torch.compile results in significantly (~25%) fewer iter/s for 2x H100s, while DDP just fails with [this error](https://gist.github.com/djsaunde/7225c0afe9bd96b94e67158cc75f2fac).
> 
> So there is this very helpful [chart](https://github.com/pytorch/ao/blob/main/torchao/float8/README.md#performance) in torchao docs that should tell a lot about where to expect what speedup. Regarding distributed, I haven't really tested more configurations, but on a single node with large enough seq-len (8192) you can see quite a nice improvement. Have a script to reproduce that in Accelerate: [here](https://github.com/huggingface/accelerate/tree/main/examples/fsdp2)

Thx! Yeah, that chart is nice. I'm seeing speedups in line with what they state for single GPU at least. Something appears to be bugged on our side for DDP and, to a lesser extent, FSDP2. Appreciate the script!

winglian:
What if you disable checkpointing/offloading?

djsaunde:
> What if you disable checkpointing/offloading?

Same result unfortunately.

src/axolotl/monkeypatch/accelerate_torchao_fsdp_check.py
djsaunde: removed :)

djsaunde:
> > What if you disable checkpointing/offloading?
> 
> Same result unfortunately.

I'm dumb, disabling _fsdp gradient checkpointing_ allows fp8 training to go ~10-15% faster than bf16 for the models / hyperparams I tested. @SalmanMohammadi pointed out that torchtune's FFT uses gradient checkpointing by default, and their [benchmarks](https://github.com/pytorch/torchtune/pull/2546#issue-2964713405) show a ~15% improvement, so something is fishy here.

Forgive the long run names, thought I'd place all the relevant config info in them to tell the full story:

W B Chart 7_17_2025, 3_36_56 PM

(I can't get wandb to not cut off the green run's name, but that's fp8 + fsdp2 + no activation checkpointing)

docs/mixed_precision.qmd
```diff
@@ -0,0 +1,149 @@
...
+
+::: {.callout-important}
+**torch.compile is critical for FP8 performance**
+
+FP8 training requires `torch_compile: true` to see meaningful speedups. Without compilation, FP8 may actually be slower and use more memory than FP16/BF16.
+:::
+
+### Advanced FP8 Configs {#sec-fp8-advanced}
+
+For FSDP (Fully Sharded Data Parallel) training:
```
salmanmohammadi: Let's link to the FSDP section in multi_gpu.qmd

src/axolotl/utils/schemas/config.py
```diff
@@ -344,6 +344,7 @@ class AxolotlInputConfig(
         default=None, json_schema_extra={"description": "Use CUDA fp16"}
     )
     fp8: bool | None = None
+    fp8_enable_fsdp_float8_all_gather: bool | None = None
```
salmanmohammadi: Should this be gated behind FSDP2?

winglian (commented):
Can we add a quick test case like `assert check_create_accelerate_code_is_patchable()` so we get a test failure if the patched code changes upstream?

src/axolotl/utils/schemas/validation.py
```diff
@@ -360,6 +360,35 @@ def check_fft_possible_bad_config(self):
...
+                "see speed improvements. Please consider setting `torch_compile: "
+                "true` in your config."
+            )
+        if data.get("fp8") and (
+            data.get("fsdp_config", {}).get("activation_checkpointing", False) is True
+            or data.get("fsdp_config", {}).get("fsdp_activation_checkpointing", False)
+            is True
+        ):
+            LOG.warning(
+                "FP8 + FSDP2 + activation checkpointing may be slower than BF16 "
```
winglian, line 378: Us the recommendation here to avoid offloading?

winglian (approved):
Just a minor test addition and nit about informing next step on validation. Good to go otherwise once those are addressed

src/axolotl/utils/schemas/validation.py
djsaunde: For now, yes. We need to figure out why this is happening; torchtune uses activation checkpointing in this setting and they still get speedups, and torchtune vs. accelerate's implementation doesn't seem to be meaningfully different

djsaunde:
Going to merge and hope to address a few follow-ups in additional PRs.

PR #2937, opened 2025-07-17
misc fixes 202507

winglian:
## Summary by CodeRabbit

* **Chores**
  * Updated coverage reporting configuration to provide informational status updates.

* **Refactor**
  * Adjusted logging behavior so that certain dataset loading messages are now shown only in the main process, reducing duplicate log output in multi-process environments.

* **New Features**
  * Added support for the "llama4" model type to enhance compatibility with new attention mechanisms.

### DIFF
--- a/src/axolotl/utils/data/shared.py
+++ b/src/axolotl/utils/data/shared.py
@@ -460,13 +460,13 @@ def load_preprocessed_dataset(cfg: DictDefault, dataset_hash: str) -> Dataset | 
     ):
         LOG.info(
             f"Loading prepared dataset from disk at {prepared_ds_path}...",
-            main_process_only=False,
+            main_process_only=True,
         )
         return load_from_disk(str(prepared_ds_path))
 
     LOG.info(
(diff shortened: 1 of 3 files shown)


src/axolotl/utils/data/shared.py
```diff
@@ -460,13 +460,13 @@ def load_preprocessed_dataset(cfg: DictDefault, dataset_hash: str) -> Dataset |
     ):
         LOG.info(
             f"Loading prepared dataset from disk at {prepared_ds_path}...",
-            main_process_only=False,
+            main_process_only=True,
```
salmanmohammadi, line 463: I think we log on rank zero by default so we could omit this?

salmanmohammadi (approved):

src/axolotl/utils/data/shared.py
winglian: we only have `LOG.warning_once`, otherwise we have to use `main_process_only=...`

PR #2941, opened 2025-07-17
make pad_to_sequence_len default to the same value as sample_packing

winglian:
# Description

Sane default to reduce number of config options a user needs to worry about.

## Summary by CodeRabbit

* **Chores**
  * Removed the sequence padding option from numerous configuration files, impacting how input sequences are handled during processing and training.
  * Clarified default behavior for sequence padding in configuration schema descriptions.
* **Tests**
  * Added tests to ensure sequence padding defaults align with sample packing settings.

### DIFF
--- a/src/axolotl/utils/schemas/config.py
+++ b/src/axolotl/utils/schemas/config.py
@@ -435,7 +435,7 @@ class AxolotlInputConfig(
     pad_to_sequence_len: bool | None = Field(
         default=None,
         json_schema_extra={
-            "description": "Pad inputs so each step uses constant sized buffers. This will reduce memory fragmentation and may prevent OOMs, by re-using memory more efficiently"
+            "description": "Pad inputs so each step uses constant sized buffers. This will reduce memory fragmentation and may prevent OOMs, by re-using memory more efficiently. Defaults to the value of `sample_packing`"
         },
     )
     curriculum_sampling: bool | None = Field(
--- a/src/axolotl/utils/schemas/config.py
+++ b/src/axolotl/utils/schemas/config.py
@@ -435,7 +435,7 @@ class AxolotlInputConfig(
     pad_to_sequence_len: bool | None = Field(
         default=None,
         json_schema_extra={
-            "description": "Pad inputs so each step uses constant sized buffers. This will reduce memory fragmentation and may prevent OOMs, by re-using memory more efficiently. Defaults to the value of `sample_packing`"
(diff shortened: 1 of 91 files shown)


src/axolotl/utils/schemas/validation.py
```diff
@@ -97,6 +97,16 @@ def check_test_datasets_bench(cls, data):
             data["test_datasets"] = [{"path": "axolotl-ai-co/empty-test-ds"}]
         return data
 
+    @model_validator(mode="before")
+    @classmethod
+    def default_pad_to_sequence_len_value(cls, data):
```
djsaunde: Do we / should we raise a warning if sample packing is enabled and pad to sequence len is not?

djsaunde (approved):

src/axolotl/utils/schemas/validation.py
```diff
@@ -97,6 +97,16 @@ def check_test_datasets_bench(cls, data):
...
+        if (
+            data.get("sample_packing") is not None
+            and data.get("pad_to_sequence_len") is None
+        ):
+            data["pad_to_sequence_len"] = data["sample_packing"]
```
NanoCode012: fyi, we have another validator doing almost the same thing https://github.com/axolotl-ai-cloud/axolotl/blob/170322a1f0e099bd8e184d2d352cf2f60437527c/src/axolotl/utils/schemas/validation.py#L206-L220

src/axolotl/utils/schemas/validation.py
winglian: Nano pointed out below we already have something that does, so removing this. This PR is almost exclusively for removing the `pad_to_sequence_len` option from all the example YAMLs to reduce cognitive load for users.

src/axolotl/utils/schemas/config.py
```diff
@@ -435,7 +435,7 @@ class AxolotlInputConfig(
     pad_to_sequence_len: bool | None = Field(
         default=None,
         json_schema_extra={
-            "description": "Pad inputs so each step uses constant sized buffers. This will reduce memory fragmentation and may prevent OOMs, by re-using memory more efficiently"
+            "description": "Pad inputs so each step uses constant sized buffers. This will reduce memory fragmentation and may prevent OOMs, by re-using memory more efficiently. Defaults to the value of `sample_packing`"
```
```suggestion
            "description": "Pad inputs so each step uses constant sized buffers. This will reduce memory fragmentation and may prevent OOMs, by re-using memory more efficiently. Defaults to True if `sample_packing` enabled"
```

NanoCode012 (approved):

PR #2942, opened 2025-07-17
set torchao quant config on config.json of saved model

winglian:
# Description
Add the torch ao quantization config to the config.json so it can be loaded without knowledge of the quantization settings.

## Summary by CodeRabbit

* **New Features**
  * Improved quantization support by introducing enhanced configuration for embedding quantization, allowing better tracking and inclusion of embedding layers during quantization.

* **Chores**
  * Updated command-line interface to use a different argument key for specifying the model path.

### DIFF
--- a/src/axolotl/utils/quantization.py
+++ b/src/axolotl/utils/quantization.py
@@ -20,6 +20,7 @@ from torchao.quantization.quant_api import (
     UIntXWeightOnlyConfig,
     _is_linear,
 )
+from transformers import TorchAoConfig
 
 from axolotl.utils.schemas.enums import TorchIntDType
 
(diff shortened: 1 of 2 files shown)


salmanmohammadi:
Did you check whether the outputs/weights were the same as for a model loaded without the config?

src/axolotl/utils/quantization.py
```diff
@@ -160,6 +163,7 @@ def quantize_model_for_ptq(
             embedding_quantize_config,
             filter_fn=lambda m, _: isinstance(m, torch.nn.Embedding),
         )
+        model.config.quantization_config = quantization_config
```
salmanmohammadi, line 166: Do we need to set this kwarg https://github.com/huggingface/transformers/blob/967045082faaaaf3d653bfe665080fd746b2bb60/src/transformers/quantizers/base.py#L61 ?

salmanmohammadi:
Closing in favour of #3107

PR #2946, opened 2025-07-19
make the initial call to tokenizer.pad not spam the console

winglian:
# Description

When training starts, we pad but the tokenizer usually gets called in fork/multiprocess so it spams the console with

```
Please note that with a fast tokenizer, using the `__call__` method is faster than using a method to encode the text followed by a call to the `pad` method to get a padded encoding.
```

This PR sets a variable so the check doesn't emit the mesage

## Summary by CodeRabbit

* **Bug Fixes**
  * Suppressed deprecation warnings related to padding on fast tokenizers, reducing unnecessary warning messages for users.
* **New Features**
  * Added validation to prevent incompatible configuration options from being enabled simultaneously.
* **Improvements**
  * Enhanced logging to provide clearer informational messages while limiting output to the main process in distributed environments.

### DIFF
--- a/src/axolotl/utils/schemas/validation.py
+++ b/src/axolotl/utils/schemas/validation.py
@@ -3,7 +3,6 @@
 # pylint: disable=too-many-boolean-expressions
 
 import json
-import logging
 import tempfile
 from pathlib import Path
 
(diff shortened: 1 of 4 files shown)


src/axolotl/utils/schemas/validation.py
```diff
@@ -116,7 +116,8 @@ def check_eval_packing(cls, data):
             and not data.get("eval_table_size")
         ):
             LOG.info(
-                "explicitly setting `eval_sample_packing` to match `sample_packing`"
+                "explicitly setting `eval_sample_packing` to match `sample_packing`",
+                main_process_only=True,
```
NanoCode012, line 120: I'm not sure the regular logger supports this arg?

`LOG = logging.getLogger(__name__)`

winglian: I'll change the logger.

PR #2947, opened 2025-07-19
[prototype] N-D distributed parallel support

winglian:
# Description

based on @SalmanMohammadi's upstream accelerate PR (https://github.com/huggingface/accelerate/pull/3682/files), I was able to hack together an implementation that needs no upstream changes so we can experiment on what the actual needed changes should be.

One caveat is that accelerate's `DataLoaderDispatcher` does not support process & iterate on rank0 and dispatch to other ranks. e.g. `accelerator_config.dispatch_batches: true`. We should also test that this works for various combinations of w/o dispatching as well as split batches, etc.

Combinations to test: 

- [x] HSDP
- [x] HSDP + TP
- [x] HSDP + SP
- [x] TP + SP
- [x] FSDP2 + TP + SP
- [ ] HSDP + TP + SP (needs multinode - 16xGPUs)
- [ ] DPO + HSDP
- [ ] DPO + FSDP2 + TP + SP
- [ ] GRPO + HSDP
- [ ] GRPO + FSDP2 + TP + SP

TP support doesn't seem to work properly in transformers w/o Deepspeed AutoTP

related PRs:
- https://github.com/huggingface/transformers/pull/39538
- https://github.com/huggingface/accelerate/pull/3682
- https://github.com/huggingface/accelerate/pull/3498
- https://github.com/axolotl-ai-cloud/ml-cross-entropy/pull/10
- https://github.com/huggingface/accelerate/pull/3687
- https://github.com/huggingface/transformers/pull/39562
- https://github.com/huggingface/transformers/pull/38949


src/axolotl/loaders/tokenizer.py
```diff
@@ -295,4 +295,8 @@ def _load_mistral_common_tokenizer(cfg: DictDefault):
         LOG.info(
             "No Chat template selected. Consider adding a chat template for easier inference."
         )
+
+    # make the tokenizer.pad call quieter 🤐
+    tokenizer.deprecation_warnings["Asking-to-pad-a-fast-tokenizer"] = True
```
NanoCode012: Can we do, `hasattr(tokenizer, "deprecation_warnings")` else it'll break for mistral tokenizer

src/axolotl/core/builders/base.py
```diff
@@ -528,3 +530,20 @@ def _set_base_training_args(
         self._configure_accelerator_config(training_args_kwargs)
 
         return training_args_kwargs, trainer_kwargs
+
+    def _configure_dist_parallel(self, training_args_kwargs: dict):
+        if (
+            self.cfg.tensor_parallel_size > 1
+            or (self.cfg.dp_shard_size and self.cfg.dp_shard_size > 1)
+            or self.cfg.sequence_parallel_degree > 1
```
djsaunde: aside: we should rename to `context_parallel_size` to better match the ecosystem and deprecate `sequence_parallel_size`.

src/axolotl/loaders/model.py
```diff
@@ -620,6 +621,20 @@ def _configure_zero3_memory_efficient_loading(
 
     def _build_model(self) -> bool:
         """Load model, with load strategy depending on config."""
+
+        if self.cfg.tensor_parallel_size and self.cfg.tensor_parallel_size > 1:
+            # from pretrained only needs to know about Tensor Parallelism for sharding weights
+            dist_parallel = DistParallel.build(
```
djsaunde, line 628: I notice we build this three separate times. Is it possible to just have a singleton instance that we build once and refer to thereafter?

src/axolotl/monkeypatch/dist_parallel/patch.py
```diff
@@ -0,0 +1,87 @@
...
+            if "cp" in torch_device_mesh.mesh_dim_names:
+                submesh_cp_size = torch_device_mesh["cp"].size()
+            if "dp" in torch_device_mesh.mesh_dim_names:
+                submesh_dp_size = torch_device_mesh["dp"].size()
+            if "fsdp" in torch_device_mesh.mesh_dim_names:
+                submesh_fsdp_size = torch_device_mesh["fsdp"].size()
+            try:
+                submesh_dp_fsdp_size = torch_device_mesh["dp_fsdp"].size()
+            except KeyError:
+                pass
```
djsaunde: Maybe use `if "dp_fsdp" in torch_device_mesh` here instead

djsaunde: Actually, is this even set anywhere?

djsaunde (commented):
Just leaving a few comments after a quick glance.

Question: can we add support for the HYBRID_SHARD fsdp2 strategy as part of this PR (i.e., replicating across nodes while sharding intra node)? Or maybe it doesn't quite fit here?

src/axolotl/loaders/model.py
winglian: I thought about doing a singleton, but then didn't want to deal with the complexity of singletons in CI. Happy to give it a whirl though if we're confident that this shouldn't change ever during runtime

src/axolotl/monkeypatch/dist_parallel/patch.py
winglian: thanks, I think I need to go back and fix this whole method.

winglian:
Screenshot 2025-07-19 at 1 44 13 PM
`HYBRID_SHARD` is accomplished using the 2d-mesh now

src/axolotl/core/builders/base.py
winglian: migrated `sequence_parallel_degree` -> `context_parallel_size`

src/axolotl/utils/schemas/validation.py
```diff
@@ -878,36 +878,50 @@ def check_fsdp_sharded_state_dict_w_safetensors(self):
...
+    @classmethod
+    def check_dp_shard_size_fsdp(cls, data):
+        if data.get("fsdp_config", {}):
+            dp_shard_size = data.get("dp_shard_size", None)
+            if dp_shard_size and dp_shard_size <= 1:
+                raise ValueError("dp_shard_size must be greater than 1 when using FSDP")
+        if not data.get("fsdp_config", {}) and data.get("dp_shard_size", 1) > 1:
+            raise ValueError("FSDP should be enabled when using dp_shard_size > 1")
+        return data
+
```
NanoCode012, line 921:
```suggestion
    @model_validator(mode="before")
    @classmethod
    def check_dp_shard_size_fsdp(cls, data):
        fsdp_config = data.get("fsdp_config", {})
        dp_shard_size = data.get("dp_shard_size", None)

        has_fsdp = bool(fsdp_config)
        dp_sharding_enabled = dp_shard_size and dp_shard_size > 1

        if has_fsdp and dp_shard_size and dp_shard_size <= 1:
            raise ValueError("dp_shard_size must be greater than 1 when using FSDP")

        if not has_fsdp and dp_sharding_enabled:
            raise ValueError("FSDP should be enabled when using dp_shard_size > 1")

        return data

```
Small refactor

src/axolotl/utils/schemas/validation.py
```diff
@@ -878,36 +878,50 @@ def check_fsdp_sharded_state_dict_w_safetensors(self):
...
+                    ):
+                        ds_config["zero_optimization"][
+                            "gather_16bit_weights_on_model_save"
+                        ] = True
+                        should_save = True
+                    if should_save:
+                        temp_dir = tempfile.mkdtemp()
+                        with open(
+                            Path(temp_dir) / "autotp_ds.json", "w", encoding="utf-8"
+                        ) as ds_fout:
```
NanoCode012, line 948: Would it be safe to save it in a temp directory? It could get cleaned up. Maybe save in `data.output_dir`?

src/axolotl/integrations/liger/args.py
```diff
@@ -62,3 +60,13 @@ def check_tiled_mlp_conflict(cls, data):
...
         return data
+
+    @model_validator(mode="before")
+    @classmethod
+    def check_liger_rms_norm_tensor_parallel(cls, data):
+        if data.get("liger_rms_norm") and data.get("tensor_parallel_size", 1) <= 1:
+            raise ValueError(
+                "`liger_rms_norm` is incompatible with tensor parallelism, "
+                "see https://github.com/linkedin/Liger-Kernel/issues/826"
+            )
```
NanoCode012, line 71:
```suggestion
        if data.get("liger_rms_norm") and data.get("tensor_parallel_size", 1) > 1:
            raise ValueError(
                "`liger_rms_norm` is incompatible with tensor parallelism, "
                "see https://github.com/linkedin/Liger-Kernel/issues/826"
            )
```

Do you mean this condition?

src/axolotl/loaders/model.py
NanoCode012: Alternatively, could we have a utility function that accepts a `cfg` and return the device mesh / distparallel?

src/axolotl/monkeypatch/accelerate/axolotl.py
```diff
@@ -0,0 +1,32 @@
...
+    def _prepare_device_mesh(self):
+        if (
+            not (
+                self.distributed_type
+                == accelerate.accelerator.DistributedType.DEEPSPEED
+                and hasattr(self.state, "ds_device_mesh")
+            )
+            and self.world_device_mesh is not None
+        ):
+            return self.world_device_mesh
```
NanoCode012, line 31: Would it be better to check, if distributed_type == fsdp, to return world_device_mesh?

src/axolotl/loaders/model.py
```diff
@@ -620,6 +621,24 @@ def _configure_zero3_memory_efficient_loading(
...
+                dp_shard_size=self.cfg.dp_shard_size,
+                tp_size=self.cfg.tensor_parallel_size,
+                cp_size=self.cfg.context_parallel_size,
+                fsdp=bool(self.cfg.fsdp_config),
+                world_size=None,
+            )
+            device_mesh = dist_parallel.get_device_mesh()
+            self.model_kwargs["tp_size"] = self.cfg.tensor_parallel_size
+            self.model_kwargs["tp_plan"] = "auto"
+            self.model_kwargs["device_mesh"] = device_mesh[("tp",)]
```
salmanmohammadi, line 638: real nice

src/axolotl/loaders/model.py
S1ro1: Do you actually need this on your side? It feels to me like you can reuse `ParallelismConfig` from upstream for this and not repeat code that much. If you need extra functionality, open to suggestions.

salmanmohammadi: I think we can swap this out once upstream changes land.

src/axolotl/utils/schemas/validation.py
winglian: this as existed for a while already. It shouldn't matter since it gets re-generated every run and logged into wandb

PR #2955, opened 2025-07-20
Autocomplete axolotl CLI

winglian:
# Description

I had tried using `click`'s recommendation for autocompletion @ https://click.palletsprojects.com/en/stable/shell-completion/, but the responsiveness is too slow because it executes the CLI and ends up loading up torch. This whole step is high latency and feels terrible to use so I decided to simply statically setup the autocompletion with the current commands. 

Also, only for preprocess and train, it will autocomplete any directories and yaml/yml files.

## Summary by CodeRabbit

* **New Features**
  * Introduced Bash tab completion for the Axolotl CLI, providing context-sensitive command and file suggestions.
  * Bash completion is now automatically enabled when using the Axolotl CLI inside the Docker container.

### DIFF
--- a/.axolotl-complete.bash
+++ b/.axolotl-complete.bash
@@ -10,8 +10,11 @@ _axolotl_completions() {
         return 0
     fi
 
-    # If the previous word was "train" or "preprocess"
-    if [[ "$prev" == "train" || "$prev" == "preprocess" ]]; then
+    # Commands that should complete with directories and YAML files
+    local yaml_commands=("merge-sharded-fsdp-weights" "quantize" "vllm-serve" "evaluate" "inference" "merge-lora" "preprocess" "train")
+
+    # Check if previous word is in our list
+    if [[ " ${yaml_commands[*]} " =~ " $prev " ]]; then
         # Complete with directories and YAML files
         COMPREPLY=($(compgen -d -- "$cur"))  # directories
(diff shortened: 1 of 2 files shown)


.axolotl-complete.bash
```diff
@@ -0,0 +1,26 @@
...
+    prev="${COMP_WORDS[COMP_CWORD-1]}"
+
+    # If we're completing the first argument (the command)
+    if [[ $COMP_CWORD -eq 1 ]]; then
+        COMPREPLY=($(compgen -W "delinearize-llama4 fetch lm-eval merge-sharded-fsdp-weights quantize vllm-serve evaluate inference merge-lora preprocess train" -- "$cur"))
+        return 0
+    fi
+
+    # If the previous word was "train" or "preprocess"
+    if [[ "$prev" == "train" || "$prev" == "preprocess" ]]; then
```
djsaunde: Why not also for `evaluate`, `inference`, etc.?

.axolotl-complete.bash
djsaunde, line 1: noob question: does this script need to get added to MANIFEST.in for pypi installs? not sure how it works in this situation

winglian: It doesn't, and from a quick search I don't think pypi can install it to the user's home directory if using pip install. I'm okay with this being only for Docker image users.

.axolotl-complete.bash
winglian: updated.

PR #2956, opened 2025-07-20
fix deprecate deepspeed stage3_gather_16bit_weights_on_model_save arg

winglian:
# Description

The latest version of Deepspeed would give the error `Extra inputs are not permitted`, changing this arg name seems to fix the issue

## Summary by CodeRabbit

* **Chores**
  * Updated configuration files by renaming keys related to optimization settings for clarity. No changes to functionality or configuration values.

### DIFF
--- a/deepspeed_configs/zero3.json
+++ b/deepspeed_configs/zero3.json
@@ -9,7 +9,7 @@
     "stage3_param_persistence_threshold": "auto",
     "stage3_max_live_parameters": 0,
     "stage3_max_reuse_distance": 0,
-    "stage3_gather_16bit_weights_on_model_save": true
+    "gather_16bit_weights_on_model_save": true
   },
   "bf16": {
     "enabled": "auto"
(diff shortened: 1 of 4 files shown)


NanoCode012 (commented):
Looks good. It's weird as they alias it in deepspeed

https://github.com/deepspeedai/DeepSpeed/blob/c2bb53f20fa32d6cbf472c08a42959a287dd9049/deepspeed/runtime/zero/config.py#L243

But they deprecate in another https://github.com/deepspeedai/DeepSpeed/blob/c2bb53f20fa32d6cbf472c08a42959a287dd9049/deepspeed/runtime/zero/config.py#L265-L269

Should we also just rename the rest?

- prefetch_bucket_size
- param_persistence_threshold
- model_persistence_threshold
- max_live_parameters
- max_reuse_distance
- module_granularity_threshold (but we don't use this)

winglian:
> Should we also just rename the rest?

updated!

NanoCode012 (approved):

PR #2959, opened 2025-07-21
improve handling of file lock when content is empty

winglian:
# Description
 Fixes #2951 

I think what's happening is the file contents are empty, so parsing as an int gives a similar error:

```
>>> int("")
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 10: ''
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of empty or invalid process counter files to prevent errors.
  * Enhanced cleanup logic to ensure proper execution even when the process counter is zero or negative.

### DIFF
--- a/src/axolotl/utils/data/lock.py
+++ b/src/axolotl/utils/data/lock.py
@@ -46,7 +46,8 @@ class FileLockLoader:
     def _increment_counter(self):
         """Safely increment the process counter."""
         if self.counter_path.exists():
-            count = int(self.counter_path.read_text().strip())
+            counter_content = self.counter_path.read_text().strip()
+            count = int(counter_content) if counter_content else 0
         else:
             count = 0
         self.counter_path.write_text(str(count + 1))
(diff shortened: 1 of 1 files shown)


src/axolotl/utils/data/lock.py
```diff
@@ -46,18 +46,20 @@ def load(self, load_fn: Callable[[], Any]) -> Any:
     def _increment_counter(self):
         """Safely increment the process counter."""
         if self.counter_path.exists():
-            count = int(self.counter_path.read_text().strip())
+            counter_content = self.counter_path.read_text().strip()
+            count = int(counter_content) if counter_content else 0
         else:
             count = 0
```
NanoCode012, line 52: Should we make a simple `_get_counter` which handles this check if exists? Then we can re-use it in the `cleanup` as well.

winglian: I don't think so. One is within a `FileLock` context and the other is simply a file exists check. I think 2 lines of duplicated code is fine.

NanoCode012 (approved):

src/axolotl/utils/data/lock.py
```diff
@@ -46,18 +46,20 @@ def load(self, load_fn: Callable[[], Any]) -> Any:
     def _increment_counter(self):
         """Safely increment the process counter."""
         if self.counter_path.exists():
-            count = int(self.counter_path.read_text().strip())
+            counter_content = self.counter_path.read_text().strip()
+            count = int(counter_content) if counter_content else 0
```
djsaunde, line 50: I don't know why `counter_content` would be something other than an int if it exists. If it exists, it should have an int written to it. Probably the file was mid-write when this was originally hit. Are you able to reproduce the original error?

But hey, if this works, we should go with it for now. We may need to be more clever with the counter file handling.

djsaunde (approved):

src/axolotl/utils/data/lock.py
winglian: I haven't been able to replicate, but it's likely because it's on a shared network drive between nodes.

PR #2962, opened 2025-07-21
we don't need to call check_dataset_labels when skip_prepare_dataset is set

winglian:
# Description

Report of error, but when using `skip_prepare_dataset: true`:
```
File "/workspace/axolotl/src/axolotl/cli/preprocess.py", line 101, in do_cli

    do_preprocess(parsed_cfg, parsed_cli_args)

  File "/workspace/axolotl/src/axolotl/cli/preprocess.py", line 57, in do_preprocess

    load_datasets(cfg=cfg, cli_args=cli_args)

  File "/workspace/axolotl/src/axolotl/common/datasets.py", line 96, in load_datasets

    check_dataset_labels(

  File "/workspace/axolotl/src/axolotl/utils/tokenization.py", line 20, in check_dataset_labels

    check_example_labels(dataset[idx], tokenizer, text_only=text_only)

  File "/workspace/axolotl/src/axolotl/utils/tokenization.py", line 27, in check_example_labels

    input_ids = example["input_ids"]

                ~~~~~~~^^^^^^^^^^^^^

KeyError: 'input_ids'
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved an issue where certain configuration options were not correctly detected during preprocessing.
  * Improved handling by logging an error and skipping preprocessing when specific settings are present, instead of raising an error.

### DIFF
--- a/src/axolotl/cli/preprocess.py
+++ b/src/axolotl/cli/preprocess.py
@@ -38,7 +38,7 @@ def do_preprocess(cfg: DictDefault, cli_args: PreprocessCliArgs) -> None:
 
     for key in ["skip_prepare_dataset", "pretraining_dataset"]:
         if cfg.get(key):
-            LOG.warning(
+            LOG.error(
                 f"You have set `{key}:`. `preprocess` is not needed. Run the `axolotl train` CLI directly instead."
             )
             return
--- a/src/axolotl/cli/preprocess.py
+++ b/src/axolotl/cli/preprocess.py
@@ -37,7 +37,7 @@ def do_preprocess(cfg: DictDefault, cli_args: PreprocessCliArgs) -> None:
     check_user_token()
 
     for key in ["skip_prepare_dataset", "pretraining_dataset"]:
-        if cfg.get("key"):
+        if cfg.get(key):
             raise ValueError(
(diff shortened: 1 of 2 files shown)


NanoCode012:
Hm, a bit confused. Is this not called first? https://github.com/axolotl-ai-cloud/axolotl/blob/bc1076d8a2fe567c64be3c040df3619fe9000109/src/axolotl/cli/preprocess.py#L39-L43

winglian:
No. The cfg.get("key") is looking for the literal string "key" (bug) rather than the variable name set by key

NanoCode012:
> No. The cfg.get("key") is looking for the literal string "key" (bug) rather than the variable name set by key

Oof. My bad. Should we just fix that instead?

winglian:
> Should we just fix that instead?

fixed!

NanoCode012:
I think users will miss the LOG.warning. Should we keep the error? Do we call `do_preprocess` when `skip_prepare_dataset: true`?

winglian:
Screenshot 2025-07-22 at 5 45 33 AM

I'll mark it as an error so it stands out in red. But at the end of the day, it doesn't matter if they miss it and move on to the train step it doesn't matter that the preprocess isn't going to run anyways.  It's just that erroring for an no-op is bad.

NanoCode012 (approved):

PR #2971, opened 2025-07-23
garbage collect on the end of the step if we're going to save a checkpoint

winglian:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory management by performing additional cleanup during save intervals and at the end of training when using step-based saving strategies. This helps ensure more efficient use of system resources during training.

### DIFF
--- a/src/axolotl/utils/callbacks/__init__.py
+++ b/src/axolotl/utils/callbacks/__init__.py
@@ -27,7 +27,11 @@ from transformers import (
     TrainerState,
     TrainingArguments,
 )
-from transformers.trainer_utils import PREFIX_CHECKPOINT_DIR, IntervalStrategy
+from transformers.trainer_utils import (
+    PREFIX_CHECKPOINT_DIR,
+    IntervalStrategy,
+    SaveStrategy,
+)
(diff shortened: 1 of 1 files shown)


src/axolotl/utils/callbacks/__init__.py
```diff
@@ -879,6 +883,17 @@ def on_step_end(
...
+            args.save_strategy == SaveStrategy.STEPS
+            and state.save_steps > 0
+            and state.global_step % state.save_steps == 0
+        ):
+            # gc on save steps in case anything is loaded to CPU RAM like offloaded tensors
+            self._gc()
+        elif state.global_step >= state.max_steps:
+            if args.save_strategy == SaveStrategy.STEPS:
+                # gc on save steps in case anything is loaded to CPU RAM like offloaded tensors
+                self._gc()
```
NanoCode012, line 896: Would this work if save strategy == epoch? Do we need to handle that too?

winglian: I'm just using what the transformers callback uses to determine when to save

Issue #2978, opened 2025-07-24
KD (distillation) chat template error when using pre-collected logprobs


zxnie:
### Please check that this issue hasn't been reported before.

- [x] I searched previous [Bug Reports](https://github.com/axolotl-ai-cloud/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

Seems like the `ChatTemplateStrategyWithKDv2` requires:
- "conversation field"
- "logprobs_field"
- "target_token_ids" (this is not configurable in YAML)

In this class:
`ChatTemplateStrategyWithKDv2`
this function:
```
def _tokenize_single_prompt(self, prompt):
        logprobs = prompt.pop(self.logprobs_field)
        target_token_ids = prompt.pop("target_token_ids")
        tokenized_prompt = super()._tokenize_single_prompt(prompt)
        tokenized_prompt[self.logprobs_field] = logprobs
        tokenized_prompt["target_token_ids"] = target_token_ids
        tokenized_prompt = self.transform_logprobs(tokenized_prompt)

        return tokenized_prompt
```
`self.logprobs_field` is already poped, then in this line, inside the function:
`tokenized_prompt = super()._tokenize_single_prompt(prompt)`
It tried to pop it again, which lead to the error.

I do not know how to fix it, but please look into it.

### Current behaviour

I put the error trace here:
```
Tokenizing Prompts:   0%|          | 0/4766 [00:00<?, ? examples/s]
Type: <class 'datasets.formatting.formatting.LazyRow'>
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/zinie/Projects/train_ax/.venv/lib/python3.12/site-packages/axolotl/cli/train.py", line 119, in <module>
    fire.Fire(do_cli)
  File "/home/zinie/Projects/train_ax/.venv/lib/python3.12/site-packages/fire/core.py", line 135, in Fire
    component_trace = _Fire(component, args, parsed_flag_args, context, name)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/zinie/Projects/train_ax/.venv/lib/python3.12/site-packages/fire/core.py", line 468, in _Fire
    component, remaining_args = _CallAndUpdateTrace(
                                ^^^^^^^^^^^^^^^^^^^^
[...]

PR #2990, opened 2025-07-30
fix:kd_distillation key_error logprobs

ved1beta:
key error log_probs 
# Description
log_probs getting popped twice once in child ``` ChatTemplateStrategyWithKDv2 ``` class and second in parent class 
same with target_token_ids
parent class ``` ChatTemplateStrategyWithKD ``` calling ``` transform_logprobs ```
of child expecting dict

added a bypass to avoid calling ``` transform_logprobs ``` of child 
and saving log_probs and target_token_ids if not none 

## Motivation and Context
#2978 

## How has this been tested?
similar config file as presented in the issue  was able to reproduce the issue 
with dataset 
```
{"conversation": [{"role": "user", "content": "Hello, how are you?"}, {"role": "assistant", "content": "I'm doing great, thank you for asking!"}], "logprobs": [[-0.1, -0.5], [-0.2, -0.8]], "target_token_ids": [[123, 456], [789, 101]]}
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of optional fields in chat prompt tokenization to prevent issues when certain data is missing.

### DIFF
--- a/src/axolotl/integrations/kd/chat_template.py
+++ b/src/axolotl/integrations/kd/chat_template.py
@@ -286,7 +286,9 @@ class ChatTemplateStrategyWithKDv2(ChatTemplateStrategyWithKD):
     def _tokenize_single_prompt(self, prompt):
         logprobs = prompt.pop(self.logprobs_field)
         target_token_ids = prompt.pop("target_token_ids")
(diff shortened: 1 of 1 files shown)


NanoCode012 (approved):
Thanks for catching this.

NanoCode012:
I simplified the handling a bit to pop logprobs to the parent class, so we don't try to do the same thing in both classes.

PR #2991, opened 2025-07-30
actually call the register method on plugins

winglian:
# Description

The plugins' individual register method wasn't being called.
## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured that all plugins are properly registered with the updated configuration after changes are applied.

### DIFF
--- a/src/axolotl/cli/config.py
+++ b/src/axolotl/cli/config.py
@@ -159,6 +159,9 @@ def plugin_set_cfg(cfg: DictDefault):
     if cfg.get("plugins"):
         plugin_manager = PluginManager.get_instance()
         plugin_manager.cfg = cfg
+        # now that we have the finalized cfg, register the plugins individually
+        for plugin in plugin_manager.plugins.values():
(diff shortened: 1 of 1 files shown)


src/axolotl/cli/config.py
```diff
@@ -159,6 +159,9 @@ def plugin_set_cfg(cfg: DictDefault):
     if cfg.get("plugins"):
         plugin_manager = PluginManager.get_instance()
         plugin_manager.cfg = cfg
+        # now that we have the finalized cfg, register the plugins individually
+        for plugin in plugin_manager.plugins.values():
+            plugin.register(cfg)
```
djsaunde, line 164: We already do this [here](https://github.com/axolotl-ai-cloud/axolotl/blob/22810c97b72a397d877e27fe9b75b3be10920e27/src/axolotl/cli/config.py#L145)

NanoCode012 (approved):
Nice catch. Seems like non of our plugins implemented `register(`, so it slipped through.

PR #2996, opened 2025-08-01
fix: move memory usage log to trainer.log

NanoCode012:
# Description

With current callback, there is a desync that occurs after an eval.

```
Tried to log to step 199 that is less than the current step 200. Steps must be monotonically increasing, so this data will be ignored. See https://wandb.me/define-metric to log data out of order.
```

In this PR, the logs are appended before they are propagated to the callback handlers. An alternative solution tried was modifying the `logs` in a callback but that callback is likely executed after a metric handler like wandb is called, so it has almost no effect.

Side effect of this PR, the logs are now printed:
```
{'loss': 1.6104, 'grad_norm': 2.421875, 'learning_rate': 0.00013268753144081652, 'memory/max_memory_active': 12.16070032119751, 'memory/max_memory_allocated': 12.16070032119751, 'memory/device_memory_reserved': 12.32421875, 'epoch': 0.46}
```
image

Wandb also rewrites names of the log to include `train/`. https://github.com/huggingface/transformers/blob/a5923d4de7df2fbd1f373dfcfe983216b79b6937/src/transformers/integrations/integration_utils.py#L631

## Summary by CodeRabbit

* **New Features**
  * Added GPU memory usage logging directly to training logs, allowing users to monitor memory usage during training.

* **Removals**
  * Removed the GPU memory stats callback; GPU memory statistics are now integrated into the main training logs.

### DIFF
--- a/src/axolotl/core/builders/base.py
+++ b/src/axolotl/core/builders/base.py
@@ -36,7 +36,6 @@ from axolotl.monkeypatch.trainer.lr import patch_trainer_get_lr
 from axolotl.utils import is_comet_available, is_mlflow_available
 from axolotl.utils.callbacks import (
     GCCallback,
-    GPUStatsCallback,
     SaveAxolotlConfigtoWandBCallback,
     SaveModelOnFirstStepCallback,
 )
@@ -140,8 +139,6 @@ class TrainerBuilderBase(abc.ABC):
             )
         if self.cfg.save_first_step:
             callbacks.append(SaveModelOnFirstStepCallback())
-
-        callbacks.append(GPUStatsCallback(cfg=self.cfg))
 
         if self.cfg.profiler_steps:
             callbacks.append(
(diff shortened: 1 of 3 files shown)


NanoCode012:
Rethinking this, I don't think we need to show the max memory in the logs as it's redundant (not really changing). However, adding it in the callbacks cause the desync issue. Alternative is to only log every X steps of total steps.

salmanmohammadi:
> Rethinking this, I don't think we need to show the max memory in the logs as it's redundant (not really changing). However, adding it in the callbacks cause the desync issue. Alternative is to only log every X steps of total steps.

Which max memory do you mean?

winglian:
> Which max memory do you mean?

@SalmanMohammadi I think he means that it shows up in the stdout logs:

> ```
> {'loss': 1.6104, 'grad_norm': 2.421875, 'learning_rate': 0.00013268753144081652, 'memory/max_memory_active': 12.16070032119751, 'memory/max_memory_allocated': 12.16070032119751, 'memory/device_memory_reserved': 12.32421875, 'epoch': 0.46}
> ```

winglian (approved):
I think if we're going to log gpu stats to wandb, it's all or nothing.

PR #3007, opened 2025-08-02
drop old patches and code that are no longer needed

winglian:
# Description

## Summary by CodeRabbit

* **Refactor**
  * Removed support and configuration options for fused QKV flash attention across all relevant components and documentation.
  * Simplified flash attention patching logic for supported models.
  * Updated validation checks to only consider MLP fusion flags.
* **Documentation**
  * Updated configuration documentation to reflect the removal of fused QKV options.
* **Tests**
  * Updated test configurations to remove fused QKV flash attention settings.


NanoCode012 (approved):
Changes look fine. CI fail from train loss being high which is not related to this.

PR #3012, opened 2025-08-04
fix: deepcopy lr in RexLR scheduler.

nyxkrage:
This fixes a problem where when the lr is a scalar tensor, the base_lrs in the get_lr function end up being references to the current learning rate, rather than the correct initial learning rate.

See also related pytorch PR https://github.com/pytorch/pytorch/pull/127190/

I discovered this when the during a run with the learning_rate set to 5e-5, the reported learning rate after the warmup ended up being 4.7e-7 instead.
I narrowed it down to being due to the lr being a tensor, and thus found the above linked pytorch PR, and after applying the change to the rexlr constructor, the learning rate correctly peaks at the given learning rate.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of learning rate values in the scheduler to prevent issues when the learning rate is a tensor.

### DIFF
--- a/src/axolotl/utils/schedulers.py
+++ b/src/axolotl/utils/schedulers.py
@@ -45,8 +45,10 @@ class RexLR(LRScheduler):
 
         # Ensure each parameter group has an "initial_lr" key to avoid issues when resuming.
         for group in optimizer.param_groups:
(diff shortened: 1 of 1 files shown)


winglian (approved):
great find! thanks!

PR #3024, opened 2025-08-06
allow custom trainer_cls to be defined as a module reference in the YAML

winglian:
# Description

This allows users to extend the `AxolotlTrainer` in file/module without the extra complexity of a custom plugin

for example, create a `my_trainer.py` with:
```python
from axolotl.core.trainers.base import AxolotlTrainer

class MyTrainer(AxolotlTrainer):
    def training_step(self, model, inputs):
        # TODO do something before the training step
        return super().training_step(model, inputs)
```

and then in your yaml, you can use:
```yaml
trainer_cls: my_trainer.MyTrainer
```

## Summary by CodeRabbit

* **New Features**
  * Added support for specifying a custom trainer class via configuration, allowing advanced users to override the default trainer by providing a module path string.
* **Documentation**
  * Updated configuration schema with a new optional field for specifying a custom trainer class.
* **Tests**
  * Added tests to verify dynamic import functionality for custom trainer classes.

### DIFF
--- a/src/axolotl/core/builders/causal.py
+++ b/src/axolotl/core/builders/causal.py
@@ -43,6 +43,7 @@ from axolotl.utils.collators import (
     V2BatchSamplerDataCollatorForSeq2Seq,
 )
 from axolotl.utils.collators.mm_chat import MultiModalChatDataCollator
+from axolotl.utils.import_helper import get_cls_from_module_str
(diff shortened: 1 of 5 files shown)


djsaunde (approved):
Coderabbit error handling comments seem reasonable.

Would be nice to have a simple test.

PR #3044, opened 2025-08-08
use exec instead of subprocess to make ctrl+c nicer for cli

winglian:
right now, when training from the cli, if you ctrl+c to stop the training, it just ends up dumping a call stack forever requiring pkill to stop the processes. This PR swaps the subprocess with an exec to replace the running process with the training process

## Summary by CodeRabbit

* **New Features**
  * Improved training command to better handle multiple configuration sets.
  * Added an option to replace the current process when launching training for more efficient execution.

* **Refactor**
  * Enhanced configuration generation and training launch processes to support the new execution mode.

* **Tests**
  * Updated CLI tests to reflect changes in training process execution method.

### DIFF
--- a/src/axolotl/cli/utils/train.py
+++ b/src/axolotl/cli/utils/train.py
@@ -64,10 +64,10 @@ def build_command(base_cmd: list[str], options: dict[str, Any]) -> list[str]:
     return cmd
 
 
-def generate_config_files(config: str, sweep: str | None) -> Iterator[str]:
+def generate_config_files(config: str, sweep: str | None) -> Iterator[tuple[str, int]]:
     """Generate list of configuration files to process."""
     if not sweep:
-        yield config
+        yield config, 1
         return
 
     # Load sweep and base configurations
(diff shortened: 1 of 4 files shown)


src/axolotl/cli/utils/train.py
```diff
@@ -64,10 +64,10 @@ def build_command(base_cmd: list[str], options: dict[str, Any]) -> list[str]:
     return cmd
 
 
-def generate_config_files(config: str, sweep: str | None) -> Iterator[str]:
+def generate_config_files(config: str, sweep: str | None) -> Iterator[tuple[str, int]]:
     """Generate list of configuration files to process."""
     if not sweep:
-        yield config
+        yield config, 1
```
djsaunde: Technically you could just yield a bool, but not really important.

djsaunde (approved):

src/axolotl/cli/utils/train.py
winglian: Thanks, changed it to a bool and added a docstring for clarification.

PR #3062, opened 2025-08-13
feat: add peft_trainable_token_indices

NanoCode012:
# Description

Adds https://huggingface.co/docs/peft/v0.17.0/en/developer_guides/lora#efficiently-train-tokens-alongside-lora

Users can now specify specific tokens to train on. For ex:
```yaml
peft_trainable_token_indices: [30000, 30001, 30002]
```

This UX is a bit annoying though if a user wants to pass in a long list which they need to manually type in.

Secondly, one use case for this can be when users add new tokens, should we support a new (for ex:, `peft_trainable_added_tokens: bool`) to auto detect this?

## How has this been tested?

Nope, but just pass through to Peft.

## Summary by CodeRabbit

- New Features
  - Added support to fine-tune only selected tokens during LoRA training. You can now specify trainable token indices globally or per embedding layer.
  - New configuration option accepts either a list of token IDs or a mapping of embedding layer names to token ID lists, enabling more targeted and memory-efficient training.

### DIFF
--- a/src/axolotl/loaders/adapter.py
+++ b/src/axolotl/loaders/adapter.py
@@ -98,6 +98,8 @@ def load_lora(
         lora_config_kwargs["use_rslora"] = cfg.peft_use_rslora
     if cfg.peft_layer_replication:
         lora_config_kwargs["layer_replication"] = cfg.peft_layer_replication
+    if cfg.peft_trainable_token_indices:
+        lora_config_kwargs["trainable_token_indices"] = cfg.peft_trainable_token_indices
 
     lora_config = LoraConfig(
         r=cfg.lora_r,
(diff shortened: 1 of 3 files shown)


winglian:
we already accept `fix_untrained_tokens` as a `list[int]`, so might want to reuse part of that behavior

NanoCode012:
> we already accept `fix_untrained_tokens` as a `list[int]`, so might want to reuse part of that behavior

Could you clarify what you mean here? How would we know whether the user would want to train those tokens only (instead of the whole layer)?

Does that function still work also?

winglian (approved):
Maybe we should warn that a user should add token ids in fix_untrained_tokens if they don't exist in this list too.

NanoCode012:
> Maybe we should warn that a user should add token ids in fix_untrained_tokens if they don't exist in this list too.

Done :)

Issue #775, opened 2023-10-23
Option to drop vs truncate on rows longer than context length


winglian:
### ⚠️ Please check that this feature request hasn't been suggested before.

- [X] I searched previous [Ideas in Discussions](https://github.com/OpenAccess-AI-Collective/axolotl/discussions/categories/ideas) didn't find any similar feature requests.
- [X] I searched previous [Issues](https://github.com/OpenAccess-AI-Collective/axolotl/labels/enhancement) didn't find any similar feature requests.

### 🔖 Feature description

Currently for most prompt types, rows with tokens longer than the context currently get dropped. Let's make this customizable. 

### ✔️ Solution

Add an option like excess_len with drop vs truncate as options 

### Acknowledgements

- [X] My issue title is concise, descriptive, and in title casing.
- [X] I have searched the existing issues to make sure this feature has not been requested yet.
- [X] I have provided enough information for the maintainers to understand and evaluate this request.

Palmik:
I would like this feature for completion strategy. Currently the completion strategy splits long inputs into multiple examples, which may be surprising: https://github.com/OpenAccess-AI-Collective/axolotl/blob/964d858da007ff5a06e91aa8bae9a22e3ff8c043/src/axolotl/prompt_strategies/completion.py#L56-L57

Having the option to truncate (or skip) instead would be appreciated.

ved1beta:
close as in favour or 3068

PR #2662, opened 2025-05-12
Implement configurable handling of excess tokens in datasets

mhenrichsen:
- Added `excess_token_handling` option to the configuration, allowing users to choose between "drop" and "truncate" for handling tokens exceeding the maximum sequence length.
- Introduced `truncate_or_drop_long_seq` function to manage both single and batched samples based on the selected handling method.
- Updated relevant dataset processing functions to utilize the new handling option, ensuring backward compatibility with existing "drop" behavior.
- Enhanced logging to reflect truncation actions in dataset processing.

This change improves flexibility in managing sequence lengths during training and evaluation.

# Description
Adds `excess_token_handling`. Either drops or truncates longer sequences.

## How has this been tested?
Ran the following config:
```
base_model: NousResearch/Llama-3.2-1B
# Automatically upload checkpoint and final model to HF
# hub_model_id: username/custom_model_name

datasets:
  - path: teknium/GPT4-LLM-Cleaned
    type: alpaca
dataset_prepared_path: last_run_prepared
val_set_size: 0.1
output_dir: ./outputs/lora-out

adapter: lora
lora_model_dir:

sequence_len: 256
sample_packing: true
eval_sample_packing: true
pad_to_sequence_len: true
excess_token_handling: truncate

lora_r: 16
lora_alpha: 32
lora_dropout: 0.05
lora_target_modules:
  - gate_proj
  - down_proj
  - up_proj
  - q_proj
  - v_proj
  - k_proj
  - o_proj

wandb_project:
wandb_entity:
wandb_watch:
wandb_name:
wandb_log_model:

gradient_accumulation_steps: 2
micro_batch_size: 2
num_epochs: 1
optimizer: adamw_8bit
lr_scheduler: cosine
learning_rate: 0.0002

bf16: auto
tf32: false

gradient_checkpointing: true
resume_from_checkpoint:
logging_steps: 1
flash_attention: true

loss_watchdog_threshold: 5.0
loss_watchdog_patience: 3

warmup_steps: 10
evals_per_epoch: 4
saves_per_epoch: 1
weight_decay: 0.0
special_tokens:
  pad_token: "<|end_of_text|>"
  ```
  
  With `excess_token_handling: truncate` it gives:
  ```
  [2025-05-12 12:32:25,508] [DEBUG] [axolotl.calculate_total_num_steps:624] [PID:5957] [RANK:0] total_num_steps: 805
[...]

PR #3068, opened 2025-08-14
[feat] truncation support with excess_length_strategy

ved1beta:
# Description
added excess_len with drop = default 

## Motivation and Context
#775

# tested 
following config runs fine  

```

base_model: microsoft/DialoGPT-small  
model_type: AutoModelForCausalLM
tokenizer_type: AutoTokenizer


datasets:
  - path: tatsu-lab/alpaca
    type: alpaca
 
    test_datasets:
      - path: tatsu-lab/alpaca
        type: alpaca
        ds_type: json
        data_files:
          - alpaca_data.json


sequence_len: 512  
excess_length_strategy: truncate  # this 
min_sample_len: 10   # and this 


micro_batch_size: 2
eval_batch_size: 2
num_epochs: 1
max_steps: 50 


optimizer: adamw_torch
lr_scheduler: cosine
learning_rate: 5e-5
train_on_inputs: false
group_by_length: false
bf16: auto
fp16: false
tf32: false


logging_steps: 10
eval_steps: 25
eval_strategy: steps
save_steps: 25
save_strategy: steps
output_dir: ./outputs/test-truncation

sample_packing: false  
pad_to_sequence_len: true
dataset_processes: 4


load_in_8bit: false
load_in_4bit: false
strict: false

special_tokens:
  pad_token: '[PAD]'
```

## Summary by CodeRabbit

* **New Features**
  * Added config option to choose handling of over-length tokenized sequences: "drop" (default) or "truncate".
  * "Truncate" mode slices long sequences to the configured length, updates batch tensors accordingly, and filters out too-short samples. Logs indicate number of removed/truncated samples.

* **Chores**
  * Public API and data-loading flow updated to use the new long-sequence handler and mode-driven behavior.

* **Tests**
  * Tests updated to reflect the new handler and configuration-driven behavior.

### DIFF
--- a/src/axolotl/utils/data/utils.py
+++ b/src/axolotl/utils/data/utils.py
@@ -152,8 +152,6 @@ def truncate_long_seq(sample, sequence_len=2048, min_sequence_len=2):
     """
     Truncate samples whose sequence length is too long (> sequence_len)
     or drop those too short (< min_sequence_len).
-
-    Works for both single-example (list[int]) or batched (list[list[int]]).
     """
     min_sequence_len = min_sequence_len or 2
 
--- a/src/axolotl/utils/data/utils.py
+++ b/src/axolotl/utils/data/utils.py
@@ -192,15 +192,59 @@ def drop_long_seq_in_dataset(
     if filter_map_kwargs:
         drop_long_kwargs["desc"] = f"Dropping Long Sequences (>{sequence_len})"
 
+    behave = (cfg.excess_len or "drop").lower()
+    if behave not in {"drop", "truncate"}:
+        behave = "drop"
+
+    if behave == "drop":
+        dataset = dataset.filter(
+            drop_long,
+            batched=True,
+            **filter_map_kwargs,
+            **drop_long_kwargs,
+        )
+        if prior_len:
+            dropped = prior_len - len(dataset)
+            if dropped:
+                LOG.warning(f"Dropped {dropped} long samples from dataset")
+        return dataset
+
+    # Truncation: keep rows but slice to sequence_len
+    def _truncate_batch(batch: dict) -> dict:
+        if "input_ids" in batch:
+            batch["input_ids"] = [seq[:sequence_len] for seq in batch["input_ids"]]
+        if "attention_mask" in batch:
+            batch["attention_mask"] = [
+                seq[:sequence_len] for seq in batch["attention_mask"]
+            ]
+        if "labels" in batch:
+            batch["labels"] = [seq[:sequence_len] for seq in batch["labels"]]
+        if "position_ids" in batch:
+            batch["position_ids"] = [
+                seq[:sequence_len] for seq in batch["position_ids"]
+            ]
+        return batch
+
+    map_kwargs = dict(
+        num_proc=cfg.dataset_processes, load_from_cache_file=not cfg.is_preprocess
+    )
+    dataset = dataset.map(
+        _truncate_batch,
+        batched=True,
+        desc=f"Truncating Sequences to {sequence_len}",
+        **({} if isinstance(dataset, IterableDataset) else map_kwargs),
+    )
+
+    # drop samples shorter than cfg.min_sample_len
     dataset = dataset.filter(
-        drop_long,
+        functools.partial(
+            drop_long_seq,
+            sequence_len=sequence_len,
+            min_sequence_len=cfg.min_sample_len,
+        ),
         batched=True,
         **filter_map_kwargs,
-        **drop_long_kwargs,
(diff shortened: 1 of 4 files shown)


src/axolotl/utils/schemas/config.py
```diff
@@ -414,6 +414,12 @@ class AxolotlInputConfig(
             "description": "The maximum length of an input to train with, this should typically be less than 2048 as most models have a token/context limit of 2048"
         },
     )
+    excess_len: Literal["drop", "truncate"] | None = Field(
```
djsaunde: Maybe `excess_length_strategy`? Then can use that instead of renaming to `behave` elsewhere.

src/axolotl/utils/data/utils.py
djsaunde, line 1: consider renaming `drop_long_seq_in_dataset` to `handle_long_seq_in_dataset` or similar

src/axolotl/utils/data/utils.py
```diff
@@ -192,15 +192,60 @@ def drop_long_seq_in_dataset(
...
+            batch["attention_mask"] = [
+                seq[:sequence_len] for seq in batch["attention_mask"]
+            ]
+        if "labels" in batch:
+            batch["labels"] = [seq[:sequence_len] for seq in batch["labels"]]
+        if "position_ids" in batch:
+            batch["position_ids"] = [
+                seq[:sequence_len] for seq in batch["position_ids"]
+            ]
+        return batch
```
djsaunde: Since this doesn't depend on anything in the local scope you could move it outside the `drop_long_seq_in_dataset` function

src/axolotl/utils/data/utils.py
```diff
@@ -192,15 +192,60 @@ def drop_long_seq_in_dataset(
     if filter_map_kwargs:
         drop_long_kwargs["desc"] = f"Dropping Long Sequences (>{sequence_len})"
 
+    excess_length_strategy = (cfg.excess_length_strategy or "drop").lower()
+    if excess_length_strategy not in {"drop", "truncate"}:
+        excess_length_strategy = "drop"
```
djsaunde: I think this is handled by your `Literal` pydantic field (i.e., invalid values can't be passed).

src/axolotl/utils/data/utils.py
djsaunde: Oh, I see it depends on sequence length, but in that case you should consider doing the same `functools.apply_partial` pattern as with `drop_long`.

src/axolotl/utils/data/utils.py
```diff
@@ -192,15 +192,60 @@ def drop_long_seq_in_dataset(
...
+        functools.partial(
+            drop_long_seq,
+            sequence_len=sequence_len,
+            min_sequence_len=cfg.min_sample_len,
+        ),
         batched=True,
         **filter_map_kwargs,
-        **drop_long_kwargs,
+        desc=f"Dropping Too-Short Sequences (<{cfg.min_sample_len or 2})",
     )
```
djsaunde: Eh, not a huge fan of this, maybe just include the minimum sequence length logic in your new method as well.

src/axolotl/utils/data/utils.py
```diff
@@ -192,15 +192,60 @@ def drop_long_seq_in_dataset(
...
+    map_kwargs = {
+        "num_proc": cfg.dataset_processes,
+        "load_from_cache_file": not cfg.is_preprocess,
+    }
+    dataset = dataset.map(
+        _truncate_batch,
+        batched=True,
+        desc=f"Truncating Sequences to {sequence_len}",
+        **({} if isinstance(dataset, IterableDataset) else map_kwargs),
+    )
```
djsaunde: Can we use `filter` like with `drop_long` and re-use the `filter_map_kwargs` we created above?

djsaunde: If you took this approach you could do something like:

```python
     drop_fn = drop_long if excess_length_strategy == "drop" else truncate_long
     dataset = dataset.filter(
            drop_fn,
            batched=True,
            **filter_map_kwargs,
            **drop_long_kwargs,
        )
        if prior_len:
            dropped = prior_len - len(dataset)
            if dropped:
                LOG.warning(f"Dropped {dropped} long samples from dataset")
        return dataset
```

djsaunde:
If you have any example truncations to paste here I'd be curious to see them!

ved1beta:
>If you have any example truncations to paste here I'd be curious to see them!

examples truncations as in ? for these changes ?

winglian:
is this a duplicate of #2662 ?

ved1beta:
Yess

djsaunde (approved):
Thanks!

PR #3063, opened 2025-08-13
Various fixes for VLMs

winglian:
report that when doing VLM training, where the image is a remote url, they get the error:

```
  File "/workspace/axolotl/src/axolotl/utils/data/wrappers.py", line 224, in _handle_loaded_strategy
    dataset_wrapper = wrap_dataset_for_tokenized_prompt(
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/axolotl/src/axolotl/datasets.py", line 88, in wrap_dataset_for_tokenized_prompt
    return TokenizedPromptDataset(prompt_tokenizer, dataset, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/axolotl/src/axolotl/datasets.py", line 41, in __init__
    self.process(dataset).data,
    ^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/axolotl/src/axolotl/datasets.py", line 63, in process
    return dataset.map(
           ^^^^^^^^^^^^
  File "/root/miniconda3/envs/py3.11/lib/python3.11/site-packages/datasets/arrow_dataset.py", line 560, in wrapper
    out: Union["Dataset", "DatasetDict"] = func(self, *args, **kwargs)
                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/miniconda3/envs/py3.11/lib/python3.11/site-packages/datasets/arrow_dataset.py", line 3309, in map
    for rank, done, content in iflatmap_unordered(
  File "/root/miniconda3/envs/py3.11/lib/python3.11/site-packages/datasets/utils/py_utils.py", line 626, in iflatmap_unordered
    [async_result.get(timeout=0.05) for async_result in async_results]
  File "/root/miniconda3/envs/py3.11/lib/python3.11/site-packages/datasets/utils/py_utils.py", line 626, in <listcomp>
    [async_result.get(timeout=0.05) for async_result in async_results]
     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/miniconda3/envs/py3.11/lib/python3.11/site-packages/multiprocess/pool.py", line 774, in get
    raise self._value
KeyError: 'Indexing with integers is not available when using Python based feature extractors'
```

## Summary by CodeRabbit

- New Features
[...]

PR #3070, opened 2025-08-15
fix: unify default for conversations_field [skip-e2e]

NanoCode012:
# Description

Addresses https://github.com/axolotl-ai-cloud/axolotl/pull/3063#discussion_r2278508684

We now use `messages` as default. This is the last place using `conversations`.

## Summary by CodeRabbit

* **Refactor**
  * Changed default input field for chat message transforms from "conversations" to "messages"; behavior and validations unchanged.
* **Examples**
  * Removed explicit message-field mappings from multiple example configs and updated a few examples to reference prepared dataset, validation split, and output paths.
* **Migration Note**
  * If your data uses "conversations", configure the transform explicitly; otherwise ensure a "messages" field is available.

### DIFF
--- a/docs/multimodal.qmd
+++ b/docs/multimodal.qmd
@@ -42,7 +42,6 @@ datasets:
   - path: HuggingFaceH4/llava-instruct-mix-vsft
     type: chat_template
     split: train[:1%]
-    field_messages: messages
 
 # (optional) if doing lora, only finetune the Language model,
 # leave the vision model and vision tower frozen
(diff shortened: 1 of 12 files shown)


winglian (changes requested):
we can probably remove any 
```yaml
    field_messages: messages
```
from examples, and track down examples that don't currently use `field_messages` and have datasets with conversations since those defaults would no longer work.

winglian:
We also should run e2e tests since this affects tests that might load datasets for our smoke tests

PR #3072, opened 2025-08-15
Tokens per second logging [skip-e2e]

salmanmohammadi:
image
It's worth noting in the above example that the TP graphs are processing the same examples, whereas the FSDP graph has 2x the number of examples per gradient accm step.

## Summary by CodeRabbit

- New Features
  - Optional tokens-per-second (tkps) metric during training via include_tkps; per-GPU throughput reported each logging interval.

- Telemetry/Logging
  - Memory metric labels simplified (max_active, max_allocated, device_reserved in GiB).
  - GPU/MPS memory reporting adjusted and peak GPU stats reset to improve accuracy.
  - Logs include tokens_per_second_per_gpu when enabled.

- Documentation
  - Updated config/help text and added include_tkps field describing behavior and limits.


src/axolotl/core/trainers/base.py
```diff
@@ -581,12 +588,19 @@ def log(self, logs: dict[str, float], start_time: float | None = None) -> None:
...
-                logs["memory/max_mem_allocated(gib)"] = round(allocated, 2)
-                logs["memory/device_mem_reserved(gib)"] = round(reserved, 2)
+                logs["memory/max_active (GiB)"] = round(active, 2)
+                logs["memory/max_allocated (GiB)"] = round(allocated, 2)
+                logs["memory/device_reserved (GiB)"] = round(reserved, 2)
             except (ValueError, TypeError, FileNotFoundError):
                 pass
 
+        if self.args.include_tkps and train_eval == "train":
+            # each rank will log its own tokens per second
```
winglian, line 598: Is there some sort of all gather for this that needs to happen? Wandb reporting only happens on rank0

src/axolotl/monkeypatch/accelerate/fsdp2.py
```diff
@@ -282,7 +282,7 @@ def fsdp2_prepare_model(accelerator, model: torch.nn.Module) -> torch.nn.Module:
         # `fully_shard` doesn't accept `None` in case of `MixedPrecisionPolicy`
         "mp_policy": fsdp2_plugin.mixed_precision_policy or MixedPrecisionPolicy(),
         "mesh": (
-            mesh[tuple(accelerator.state.parallelism_config.fsdp_dim_names)]
+            mesh[tuple(accelerator.parallelism_config.fsdp_dim_names)]
```
winglian: Lets put this in a separate PR

src/axolotl/core/trainers/base.py
salmanmohammadi: If we gather then we're reporting average tkps across all ranks. It's a different metric - but I was mirroring torchtitan/torchtune which both report per-rank tkps. Non-wandb logging will still report per-rank

winglian: all reduce would average across ranks, right? but all gather would keep each ranks metric. Also, did some digging into transformers implementation and upstream, tokens per second is a "final" metric that only gets output at the end of training. I'm thinking we should leave that particular argument option alone and just have our own new argument in the config for this.

salmanmohammadi: Sounds good. 
> all reduce would average across ranks, right? but all gather would keep each ranks metric.

I'm not sure I follow here, would we then have a metric for tracking each rank reported on rank 0?

src/axolotl/core/training_args_base.py
```diff
@@ -50,6 +50,12 @@ class AxolotlTrainingMixins:
         default=False,
         metadata={"help": "Use real batches for efficient training."},
     )
+    include_tkps: bool = field(
```
djsaunde, line 53: Do we need to make this configurable? If yes, can it be on by default?

djsaunde (approved):
Approve with Q

winglian (approved):

src/axolotl/core/training_args_base.py
salmanmohammadi: I'm happy to, but it may pollute the logs as it also logs in stdout.

djsaunde: I'm okay with this personally

src/axolotl/core/trainers/base.py
djsaunde: btw you can specify the reduction for the all reduce: avg, sum, etc. I think what we want is the all reduce sum in this case

src/axolotl/core/trainers/base.py
```diff
@@ -530,9 +540,6 @@ def create_accelerator_and_postprocess(self):
 
         super().create_accelerator_and_postprocess()
 
-        # now we need to put parallelism_config back on the PartialState since we rely on that info in other places
-        # PartialState().parallelism_config = self.accelerator.state.parallelism_config
-
```
NanoCode012, line 535: Is this intended?

PR #3076, opened 2025-08-17
handle batch size correchtly when using split and dispatch batches

winglian:
The batch size handling when using split batches isn't quite right. This sets the correct batch size based on how we handle batches when pretraining and needing to scale it properly for dispatch_batches/split_batches

## Summary by CodeRabbit

- Bug Fixes
  - Corrected batch size computation during pretraining when both split and dispatch batching are enabled, with sensible handling across training and evaluation and with/without sample packing. Improves stability and throughput by avoiding unintended per-process sizing.
  - Made pretraining sample concatenation handling more robust to falsy configuration values, ensuring the appropriate collator is selected during pretraining. Prevents misconfiguration that could impact batching and data processing behavior.

### DIFF
--- a/src/axolotl/core/trainers/base.py
+++ b/src/axolotl/core/trainers/base.py
@@ -272,6 +272,20 @@ class AxolotlTrainer(
                     num_workers=self.args.dataloader_num_workers,
                     rank=self.args.process_index,
                 )
+
+        if (
+            self.args.accelerator_config is not None
+            and self.args.accelerator_config.split_batches
+            and self.args.accelerator_config.dispatch_batches
+        ):
+            if self.args.sample_packing and self.args.pretraining:
+                if not self.args.eval_sample_packing and not is_training:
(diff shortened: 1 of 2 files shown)


src/axolotl/core/trainers/base.py
```diff
@@ -272,6 +272,20 @@ def _get_dataloader(
...
                 )
+
+        if (
+            self.args.accelerator_config is not None
+            and self.args.accelerator_config.split_batches
+            and self.args.accelerator_config.dispatch_batches
+        ):
+            if self.args.sample_packing and self.args.pretraining:
+                if not self.args.eval_sample_packing and not is_training:
+                    dataloader_params["batch_size"] *= self.accelerator.num_processes
```
salmanmohammadi, line 283: Should this only be scaling by the total dp size?

winglian: accelerator handles the actual dp sharding, so I don't think so? I'll test.

Issue #3069, opened 2025-08-14
Dataloader OOM during pretraining


Blyzi:
### Please check that this issue hasn't been reported before.

- [x] I searched previous [Bug Reports](https://github.com/axolotl-ai-cloud/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

Should be able to continue pretraining by streaming a dataset with the axolotl docker container v12.1.

### Current behaviour

My setup:
- Singularity container from the docker one: axolotlai/axolotl:0.12.1 [224c81330c25](https://hub.docker.com/layers/axolotlai/axolotl/0.12.1/images/sha256-224c81330c25be54a6aec4a6e8490b34962dfbd132031572ab283ec8c1aedf27)
- 4 H100 80G
- 512Go RAM
- 256 parquet files (50Go total) with samples length averaging 30/50 k tokens with a max at multiple millions.

During training or sometimes at startup, the dataloader is killed because of a RAM OOM.
```
  0%|          | 0/1891 [00:00<?, ?it/s]
  0%|          | 1/1891 [10:17<324:05:32, 617.32s/it]
  0%|          | 2/1891 [10:37<139:34:34, 266.00s/it]
  0%|          | 3/1891 [10:56<80:26:57, 153.40s/it] 
  0%|          | 4/1891 [11:16<52:47:20, 100.71s/it]
  0%|          | 5/1891 [11:36<37:29:15, 71.56s/it] 
  0%|          | 6/1891 [11:56<28:20:37, 54.13s/it]
  0%|          | 7/1891 [12:16<22:24:31, 42.82s/it]
  0%|          | 8/1891 [12:35<18:30:37, 35.39s/it]
  0%|          | 9/1891 [12:55<15:56:27, 30.49s/it]
                                                   

  0%|          | 9/1891 [12:55<15:56:27, 30.49s/it]
  1%|          | 10/1891 [13:15<14:12:51, 27.20s/it]
  1%|          | 11/1891 [13:35<13:06:06, 25.09s/it]
  1%|          | 12/1891 [13:55<12:14:16, 23.45s/it]
  1%|          | 13/1891 [14:15<11:42:52, 22.46s/it]
  1%|          | 14/1891 [14:35<11:15:03, 21.58s/it]
  1%|          | 15/1891 [14:55<11:00:59, 21.14s/it]
  1%|          | 16/1891 [15:15<10:47:18, 20.71s/it]
[...]

PR #3081, opened 2025-08-19
Feature/pretraining-auto-truncate

Blyzi:
# Description

This PR adds auto-truncation support for pretraining_datasets.

## Motivation and Context

While investigating an issue in the [dataloader](#3069), I noticed an opportunity to improve the handling of pretraining_datasets.
Although this PR is not a direct fix for that issue, adding auto-truncation was straightforward.

One important thing to notice is that it will become harder for the user to know the `max_steps` value accordingly to its datasets because it's no longer the number of samples.

## How has this been tested?

I tested the changes locally on my setup with multiple GPUs.
The modification is scoped only to pretraining_datasets with or without concatenation, and only when sample packing is disabled.
While I did not perform extensive testing, I validated the tokenisation process.

## Types of changes

- [x] Enhancement (non-breaking change that adds functionality)

## Social Handles

[Blyzi](https://github.com/Blyzi)

## Summary by CodeRabbit

- Refactor
  - Pretraining data encoding now produces fixed-size token chunks without truncating inputs or adding EOS/PAD tokens; optional concatenation merges samples before chunking.
  - Padding and attention masks are applied consistently; labels ignore padded positions.

- Documentation
  - Removed callout about automatic splitting of long completion samples and updated related examples in dataset format docs.

### DIFF
--- a/src/axolotl/utils/data/pretraining.py
+++ b/src/axolotl/utils/data/pretraining.py
@@ -38,6 +38,18 @@ def encode_pretraining(
         for sample in full_inputs["attention_mask"]
     ]
 
+    # Resolve a safe pad token id for chunk padding
+    pad_id = (
+        tokenizer.pad_token_id
+        if tokenizer.pad_token_id is not None
+        else (tokenizer.eos_token_id or 0)
+    )
+
+    if tokenizer.pad_token_id is None:
+        LOG.warning(
+            "tokenizer.pad_token_id is None; falling back to %s for padding", pad_id
+        )
+
     inputs_ids, target_ids, attention_mask = [], [], []
 
     # Concatenate if specified all input_ids and attention masks into one tensor when concatenate is True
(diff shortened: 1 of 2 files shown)


src/axolotl/utils/data/pretraining.py
```diff
@@ -24,157 +24,81 @@ def encode_pretraining(
...
-                        tokenizer.pad_token_id,
-                        dtype=torch.long,
-                    ),
-                ),
-                dim=0,
+    inputs_ids, target_ids, attention_mask = [], [], []
+
+    # Concatenate if specified all input_ids and attention masks into one tensor when concatenate is True
+    if concatenate:
+        full_inputs["input_ids"] = [torch.cat(full_inputs["input_ids"], dim=0)]
```
djsaunde, line 57: Dumb question, not too familiar with this codepath: is this not the same as sample packing?

Blyzi: "Dumb question", don't worry IA is moving way too fast 😄 

The main objective of sample packing is to avoid attention noise between samples while filling the sequence length. It has been proven to improve model performance while maintaining computational speed. Historically, during pre-training, we didn't care about noise between samples, so we simply concatenated everything.

winglian: in pretraining, this can already be accomplished with `sample_packing: true` and `pretrain_multipack_attn: false`

winglian: it's an underdocumented feature, but should be equivalent to no sample packing with concatenate, but with more efficient packing

Blyzi: Do you guys need me to do anything before merging it ?

Issue #2402, opened 2025-03-11
FutureWarning: FSDP.state_dict_type() and FSDP.set_state_dict_type() are being deprecated. Please use APIs, get_state_dict() and set_state_dict(), which can support different parallelisms, FSDP1, FSDP2, DDP.


NanoCode012:
### ⚠️ Please check that this feature request hasn't been suggested before.

- [x] I searched previous [Ideas in Discussions](https://github.com/axolotl-ai-cloud/axolotl/discussions/categories/ideas) didn't find any similar feature requests.
- [x] I searched previous [Issues](https://github.com/axolotl-ai-cloud/axolotl/labels/enhancement) didn't find any similar feature requests.

### 🔖 Feature description

```
FutureWarning: FSDP.state_dict_type() and FSDP.set_state_dict_type() are being deprecated. Please use APIs, get_state_dict() and set_state_dict(), which can support different parallelisms, FSDP1, FSDP2, DDP. API doc: https://pytorch.org/docs/stable/distributed.checkpoint.html#torch.distributed.checkpoint.state_dict.get_state_dict .Tutorial: https://pytorch.org/tutorials/recipes/distributed_checkpoint_recipe.html .
```

### ✔️ Solution

https://pytorch.org/tutorials/recipes/distributed_checkpoint_recipe.html 
https://pytorch.org/docs/stable/distributed.checkpoint.html#torch.distributed.checkpoint.state_dict.get_state_dict

### Acknowledgements

- [x] My issue title is concise, descriptive, and in title casing.
- [x] I have searched the existing issues to make sure this feature has not been requested yet.
- [x] I have provided enough information for the maintainers to understand and evaluate this request.

NanoCode012:
Forgot where this was thrown, but likely here https://github.com/axolotl-ai-cloud/axolotl/blob/80304c26a70e21ed8522fdbd53bcb290f9c6b7d3/src/axolotl/train.py#L246

ved1beta:
[consider closing ](https://github.com/axolotl-ai-cloud/axolotl/pull/3083#discussion_r2288453806)

NanoCode012:
Sure, thanks for the followup

PR #3083, opened 2025-08-19
removed: deprecated state_dict_type() set_state_dict_type()

ved1beta:
# Description
removed: deprecated state_dict_type() set_state_dict_type()

## Motivation and Context

#2402 

## Summary by CodeRabbit

* **Refactor**
  * Moved model state collection onto PyTorch’s centralized distributed checkpoint API for FSDP workflows, standardizing save behavior.
* **Bug Fixes**
  * More reliable model saves in distributed/FSDP setups; honors full-state-dict options (CPU offload, rank0-only) and reduces save-time failures.
* **Chores**
  * Simplified internal state-dict handling and removed redundant configuration paths.

### DIFF
--- a/src/axolotl/monkeypatch/accelerate/fsdp2.py
+++ b/src/axolotl/monkeypatch/accelerate/fsdp2.py
@@ -161,16 +161,13 @@ def get_state_dict(self, model, unwrap=True):
             torch.distributed.barrier()
     elif self.distributed_type == DistributedType.FSDP:
         from torch.distributed.fsdp import FullStateDictConfig
-        from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
-        from torch.distributed.fsdp import StateDictType
 
         full_state_dict_config = FullStateDictConfig(
             offload_to_cpu=True, rank0_only=True
         )
-        with FSDP.state_dict_type(
-            model, StateDictType.FULL_STATE_DICT, full_state_dict_config
-        ):
-            state_dict = model.state_dict()
+        state_dict = torch.distributed.checkpoint.state_dict.get_state_dict(
+            model, options={"fsdp_config": full_state_dict_config}
+        )
     else:
         if unwrap:
             model = self.unwrap_model(model)
--- a/src/axolotl/train.py
+++ b/src/axolotl/train.py
@@ -281,7 +281,6 @@ def save_trained_model(
                 state_dict_type = cfg.fsdp_config.final_state_dict_type
             else:
                 state_dict_type = cfg.fsdp_config.state_dict_type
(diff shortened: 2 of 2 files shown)


src/axolotl/train.py
```diff
@@ -259,8 +259,7 @@ def save_trained_model(
                 state_dict_type = cfg.fsdp_config.final_state_dict_type
             else:
                 state_dict_type = cfg.fsdp_config.state_dict_type
-            trainer.accelerator.state.fsdp_plugin.set_state_dict_type(state_dict_type)
```
salmanmohammadi, line 262: Why was this removed?

src/axolotl/monkeypatch/accelerate/fsdp2.py
```diff
@@ -161,16 +161,13 @@ def get_state_dict(self, model, unwrap=True):
...
-        from torch.distributed.fsdp import StateDictType
 
         full_state_dict_config = FullStateDictConfig(
             offload_to_cpu=True, rank0_only=True
         )
-        with FSDP.state_dict_type(
-            model, StateDictType.FULL_STATE_DICT, full_state_dict_config
-        ):
-            state_dict = model.state_dict()
+        state_dict = torch.distributed.checkpoint.state_dict.get_state_dict(
```
salmanmohammadi, line 168: I think we should leave this as is and close the parent issue as unplanned for a couple reasons:
1) Upstream accelerate has a different implementation https://github.com/huggingface/accelerate/blob/main/src/accelerate/utils/fsdp_utils.py
2) We are upstreaming this file so we don't have to maintain these changes

Thank you for the quick fix, but unless this is blocking or breaking something I don't think we should land this atm.

src/axolotl/train.py
ved1beta: set_state_dict_type() is also deprecated  mentioned in the issue

src/axolotl/monkeypatch/accelerate/fsdp2.py
ved1beta: thanks for the review closing :sob:

src/axolotl/train.py
salmanmohammadi: FWIW `set_state_dict_type` is a method on Accelerator's FSDP plugin, which is unrelated to the deprecated pytorch API in the parent issue.

salmanmohammadi: You can see more here https://github.com/huggingface/accelerate/blob/16bf1d89016e03f5b0d8545e9883df7fe9ab9b5f/src/accelerate/utils/dataclasses.py#L1886

PR #3084, opened 2025-08-20
make always skip_move_to_device default as true

winglian:
# Description

We should default to never moving the model to device until we're ready to train/shard as doing so prematurely will lead to OOMs.

## Summary by CodeRabbit

* **Refactor**
  * Changed default behavior to skip moving models to device before sharding. This is now enabled by default; set the flag to false to restore legacy behavior.

* **Documentation**
  * Updated description to clarify the new default and how to revert to legacy behavior.

### DIFF
--- a/src/axolotl/utils/schemas/model.py
+++ b/src/axolotl/utils/schemas/model.py
@@ -65,10 +65,9 @@ class ModelInputConfig(BaseModel):
     )
 
     experimental_skip_move_to_device: bool | None = Field(
-        default=None,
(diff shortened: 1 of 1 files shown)


salmanmohammadi (approved):
Should we just remove this config field?

winglian:
> Should we just remove this config field?

My thought was to have this in main for a bit with the fallback option of setting this to None/false in case something breaks. We can strip it all out in the next release.

PR #3086, opened 2025-08-20
deploy training jobs to baseten w truss in axolotl cli

winglian:
allows remote training on Baseten from the axolotl cli, for example

```
axolotl train config.yaml --cloud examples/cloud/baseten.yaml
```

## Summary by CodeRabbit

- **New Features**
  - Added Baseten as a selectable cloud training provider via the CLI with explicit error for unsupported providers.
  - Example cloud config for Baseten (H100 GPU, gpu_count: 8, node_count: 1) and required secrets (HF_TOKEN, WANDB_API_KEY).
  - Preprocessing is handled during the train flow.

- **Chores**
  - Added Baseten training templates and scripts to automate preprocess + train runs.

### DIFF
--- a/examples/cloud/baseten.yaml
+++ b/examples/cloud/baseten.yaml
@@ -1,0 +1,13 @@
+provider: baseten
+project_name:
+volumes:
+  - name: axolotl-artifacts
+    mount: /workspace/artifacts
+
+secrets:
+  - HF_TOKEN
+  - WANDB_API_KEY
+
+gpu: h100
+gpu_count: 8
+node_count: 1
--- a/examples/cloud/baseten.yaml
+++ b/examples/cloud/baseten.yaml
@@ -1,8 +1,5 @@
 provider: baseten
 project_name:
-volumes:
-  - name: axolotl-artifacts
-    mount: /workspace/artifacts
 
 secrets:
   - HF_TOKEN
--- a/src/axolotl/cli/cloud/baseten/template/train_sft.py
+++ b/src/axolotl/cli/cloud/baseten/template/train_sft.py
@@ -1,0 +1,79 @@
+"""
+Baseten Training Script for Axolotl
+"""
+
+# pylint: skip-file
+import yaml
+from truss.base import truss_config
+
+# Import necessary classes from the Baseten Training SDK
+from truss_train import definitions
+
+cloud_config = yaml.safe_load(open("cloud.yaml", "r"))
+gpu = cloud_config.get("gpu", "h100")
+gpu_count = int(cloud_config.get("gpu_count", 1))
+node_count = int(cloud_config.get("node_count", 1))
+project_name = cloud_config.get("project_name", "axolotl-project") or "axolotl-project"
+volume_mounts = cloud_config.get("volumes", [])
+checkpoint_mount_dir = (
+    volume_mounts[0]["mount"] if volume_mounts else "/workspace/artifacts"
+)
+secrets = cloud_config.get("secrets", [])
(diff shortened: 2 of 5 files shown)


src/axolotl/cli/cloud/baseten/template/train_sft.py
```diff
@@ -0,0 +1,79 @@
...
+    enable_cache=True,
+)
+
+# 3. Define the Compute Resources for the Training Job
+training_compute = definitions.Compute(
+    node_count=2,
+    accelerator=truss_config.AcceleratorSpec(
+        accelerator=truss_config.Accelerator.H100,
+        count=8,
+    ),
```
bolasim: I think we need to updated based on the gpu and counts passed in

bolasim (approved):

src/axolotl/cli/cloud/baseten/template/train_sft.py
rcano-baseten: +1

src/axolotl/cli/cloud/baseten/template/train_sft.py
```diff
@@ -0,0 +1,79 @@
...
+for secret_name in secrets:
+    env_vars[secret_name] = definitions.SecretReference(name=secret_name)
+
+training_runtime = definitions.Runtime(
+    start_commands=[  # Example: list of commands to run your training script
+        f"/bin/sh -c 'chmod +x ./{script_name} && ./{script_name}'"
+    ],
+    environment_variables=env_vars,
+    # checkpointing_config=checkpointing,
+    enable_cache=True,
```
rcano-baseten: This should be 

```
cache_config=definitions.CacheConfig(enabled=True)
```

examples/cloud/baseten.yaml
```diff
@@ -0,0 +1,13 @@
+provider: baseten
+project_name:
+volumes:
```
rcano-baseten: Because the training cache currently has a hardcoded mount path, do we want to provide a special name for the baseten cache? e.g. 
```
- name: baseten-training-cache
```

and then when we iterate through the volumes, we can toggle the cache_config.enabled based on the presence of this name?

src/axolotl/cli/cloud/baseten/template/train_sft.py
```diff
@@ -0,0 +1,79 @@
...
+# Import necessary classes from the Baseten Training SDK
+from truss_train import definitions
+
+cloud_config = yaml.safe_load(open("cloud.yaml", "r"))
+gpu = cloud_config.get("gpu", "h100")
+gpu_count = int(cloud_config.get("gpu_count", 1))
+node_count = int(cloud_config.get("node_count", 1))
+project_name = cloud_config.get("project_name", "axolotl-project") or "axolotl-project"
+volume_mounts = cloud_config.get("volumes", [])
+checkpoint_mount_dir = (
```
rcano-baseten: Based on the way we've built checkpointing, I don't think this will work out of the box as expected. I would suggest removing this logic for checkpointing

examples/cloud/baseten.yaml
rcano-baseten: For checkpointing, we might want to do something like 
```
- name: baseten-checkpointing
  mount: <some path> 
```

I think this helps indicate that there is behavior that is different from typical volume

rcano-baseten (commented):
Left some thoughts! Would love get your thoughts on the `volumes` pieces since the cache and checkpointing don't quite fit in the mold of the existing schema

rcano-baseten (approved):
Looks great! 🚢

PR #3092, opened 2025-08-21
Add ruff, remove black, isort, flake8, pylint

djsaunde:
# Description

Title.

## Motivation and Context

Speed, reduced deps.

Downside is there's a bunch of little changes (due to ruff having slight diffs from black) and clutters the git blame.

One annoyance is that ruff doesn't support all the pylint rules, so that's where a lot of the little changes are coming from (but this is being worked on; see https://github.com/astral-sh/ruff/issues/970).

Bonus: I can pad my contribution stats 🤔 

## How has this been tested?

N/A


winglian:
> One annoyance is that ruff doesn't support all the pylint rules,

Are the uncovered rules serious enough to consider waiting?

winglian (approved):

djsaunde:
> Are the uncovered rules serious enough to consider waiting?

IMO no, I only became aware of them because we ignore them, and those ignore statements got removed in the diff.

PR #3093, opened 2025-08-21
add tokenizer_save_jinja_files to keep legacy behavior of including chat template in tokenizer_config.json

winglian:
a change a while back in transformers defaulted to always saving a separate chat template jinja file which. This makes it possible to use the legacy behavior

## Summary by CodeRabbit

- New Features
  - Added a config option to control saving tokenizer Jinja template files; applied across training, quantization, and merge flows.
  - Tokenizer artifacts are saved both before and after training to ensure expected files are included.

- Bug Fixes
  - Improved precision handling on Apple Silicon (MPS): fp16 enabling now respects bf16 and explicit fp16 disablement.

- Tests
  - Added end-to-end tests validating tokenizer artifact saving, including when template saving is disabled.

### DIFF
--- a/src/axolotl/cli/merge_lora.py
+++ b/src/axolotl/cli/merge_lora.py
@@ -43,7 +43,10 @@ def do_merge_lora(*, cfg: DictDefault) -> None:
             safe_serialization=safe_serialization,
             progressbar=True,
         )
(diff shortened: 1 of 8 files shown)


salmanmohammadi (approved):
🫡

PR #3102, opened 2025-08-26
patch for ds_grads_remaining in deepspeed

ved1beta:
## Description
patch for ds_grads_remaining in deepspeed

## Motivation and Context
deepspeed issue [ds_grads_remaining]( https://github.com/deepspeedai/DeepSpeed/issues/7203)

## Summary by CodeRabbit

- **New Features**
  - Automatic detection and application of compatibility patches when running with DeepSpeed ZeRO‑3; patches applied conditionally and logged.

- **Bug Fixes**
  - Improved stability when using gradient checkpointing with DeepSpeed ZeRO‑3 by forwarding DeepSpeed-related attributes to wrapped modules, reducing runtime errors and warnings when DeepSpeed is unavailable.

### DIFF
--- a/src/axolotl/loaders/model.py
+++ b/src/axolotl/loaders/model.py
@@ -214,7 +214,6 @@ class ModelLoader:
         self._set_quantization_config()
         self._set_attention_config()
         self._check_model_requirements()
-        self.patch_manager._apply_patch_deepspeed_zero3()
 
     def _apply_post_model_load_setup(self):
         """Configure the model after it has been loaded."""
--- a/src/axolotl/loaders/model.py
+++ b/src/axolotl/loaders/model.py
@@ -217,6 +217,7 @@ class ModelLoader:
         self._set_quantization_config()
         self._set_attention_config()
         self._check_model_requirements()
+        self.patch_manager._apply_patch_deepspeed_zero3()
 
         try:
             if (
--- a/src/axolotl/loaders/model.py
+++ b/src/axolotl/loaders/model.py
@@ -219,10 +219,14 @@ class ModelLoader:
         self._check_model_requirements()
 
         try:
-            if is_deepspeed_zero3_enabled() or os.getenv("ACCELERATE_DEEPSPEED_ZERO_STAGE") == "3":
+            if (
+                is_deepspeed_zero3_enabled()
+                or os.getenv("ACCELERATE_DEEPSPEED_ZERO_STAGE") == "3"
+            ):
                 patch_deepspeed_zero3_missing_attributes()
         except Exception:
             LOG.warning("DeepSpeed ZeRO Stage 3 missing attributes patch not applied")
+
     def _apply_post_model_load_setup(self):
         """Configure the model after it has been loaded."""
         # Handle PeftModel if needed
--- a/src/axolotl/loaders/model.py
+++ b/src/axolotl/loaders/model.py
@@ -59,9 +59,6 @@ from axolotl.utils.distributed import (
 from axolotl.utils.logging import get_logger
 from axolotl.utils.model_shard_quant import load_sharded_model_quant
 from axolotl.utils.schemas.enums import RLType
-from axolotl.monkeypatch.utils import (
-    patch_deepspeed_zero3_missing_attributes,
-)
 
 LOG = get_logger(__name__)
 PLUGIN_MANAGER = PluginManager.get_instance()
@@ -218,15 +215,6 @@ class ModelLoader:
         self._set_attention_config()
         self._check_model_requirements()
         self.patch_manager._apply_patch_deepspeed_zero3()
-
-        try:
-            if (
-                is_deepspeed_zero3_enabled()
-                or os.getenv("ACCELERATE_DEEPSPEED_ZERO_STAGE") == "3"
-            ):
-                patch_deepspeed_zero3_missing_attributes()
-        except Exception:
-            LOG.warning("DeepSpeed ZeRO Stage 3 missing attributes patch not applied")
 
     def _apply_post_model_load_setup(self):
         """Configure the model after it has been loaded."""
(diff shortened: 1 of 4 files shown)


src/axolotl/monkeypatch/utils.py
```diff
@@ -12,6 +12,9 @@
     _prepare_4d_causal_attention_mask_for_sdpa,
 )
 from transformers.utils import is_torch_bf16_gpu_available
+from axolotl.utils.logging import get_logger
+
+LOG = get_logger(__name__)
```
NanoCode012:
```suggestion
```

Maybe leftover that was missed

ved1beta:
tested with
```  
base_model: TinyLlama/TinyLlama-1.1B-Chat-v1.0
tokenizer_config: TinyLlama/TinyLlama-1.1B-Chat-v1.0
model_config_type: llama

trust_remote_code: false

# Enable the cut cross entropy plugin that was causing issues
plugins:
  - axolotl.integrations.cut_cross_entropy.CutCrossEntropyPlugin

# Test dataset - small for quick verification
datasets:
  - path: tatsu-lab/alpaca
    type: alpaca
    data_files:
      - data/train-00000-of-00001-a09b74b3ef9c3b56.parquet
    
dataset_prepared_path: ./test_prepared_data
val_set_size: 0.1
output_dir: ./outputs/deepspeed_zero3_test

# Reduce sequence length for memory efficiency
sequence_len: 2048
# Disable sample packing for simpler test
sample_packing: false
pad_to_sequence_len: true

adapter: qlora
lora_r: 8
lora_alpha: 16
lora_dropout: 0.05
lora_target_linear: true

load_in_4bit: true
bnb_4bit_quant_type: nf4
bnb_4bit_compute_dtype: bfloat16
bnb_4bit_use_double_quant: true

micro_batch_size: 1
num_epochs: 0.1
optimizer: paged_adamw_8bit
lr_scheduler: constant
learning_rate: 1e-5

bf16: true
tf32: true
gradient_checkpointing: true
  use_reentrant: false

# Disable flash attention to avoid compatibility issues
# flash_attention: true
low_cpu_mem_usage: true

logging_steps: 1
save_steps: 50
eval_steps: 50

deepspeed: deepspeed_configs/zero3_bf16_cpuoffload_all.json

special_tokens:
  pad_token: <|end_of_text|>

debug: true

# Disable wandb to avoid VLLM imports
use_wandb: false 

```

src/axolotl/loaders/model.py
```diff
@@ -214,6 +214,7 @@ def _apply_pre_model_load_setup(self):
         self._set_quantization_config()
         self._set_attention_config()
         self._check_model_requirements()
+        self.patch_manager._apply_patch_deepspeed_zero3()
```
NanoCode012: You don't need to call it here. Patches from patch manager are called in `self.patch_manager.apply_pre_model_load_patches()` above.

You only need to call your `_apply_patch_deepspeed_zero3` in the `patch_manager.apply_pre_model_load_patches`.

src/axolotl/loaders/patch_manager.py
```diff
@@ -471,3 +472,20 @@ def _apply_lora_kernel_patch(self, model):
...
+                patch_deepspeed_zero3_missing_attributes,
+            )
+            from transformers.integrations.deepspeed import is_deepspeed_zero3_enabled
+
+            if (
+                is_deepspeed_zero3_enabled()
+                or os.getenv("ACCELERATE_DEEPSPEED_ZERO_STAGE") == "3"
+            ):
+                patch_deepspeed_zero3_missing_attributes()
+        except Exception as e:
```
NanoCode012: This exception is quite broad. Could you narrow it down to the potential error that would be raised? 

Btw, inside the function, you already have a try catch though, so I'm not sure this is needed.

src/axolotl/monkeypatch/deepspeed_utils.py
```diff
@@ -0,0 +1,57 @@
...
+            """
+            Patched version that initializes ds_grads_remaining before DeepSpeed
+            tries to use it in its hooks.
+            """
+            if count is None:
+                count = [0]
+            if not hasattr(module, "ds_grads_remaining"):
+                module.ds_grads_remaining = 0
+                LOG.debug(f"Initialized ds_grads_remaining for {type(module).__name__}")
+            return original_register_module(self, module, count)
```
NanoCode012: I'm not sure this is the right approach.

While the error is that `ds_grads_remaining` is not found, by just setting it to zero, this would result in `post_bwd_fn` always being set in the module (and possibly running `PostBackwardFunctionModule` multiple times).

https://github.com/deepspeedai/DeepSpeed/blob/38d1a9eb64c9e01e32eccc50b25ba18925287441/deepspeed/runtime/zero/parameter_offload.py#L424-L458

The fix is to adjust the `if not hasattr(module, "post_bwd_fn"):` to `if "post_bwd_fn" not in module.__dict__`. You can achieve this by replacing the source. You can refer to how it's done in LoRA Kernels https://github.com/axolotl-ai-cloud/axolotl/blob/79ddaebe9a6af7efefebdbb54772d11d09561786/src/axolotl/monkeypatch/lora_kernels.py#L167-L228

winglian:
FYI, the specific configuration that I get this is when I use qlora + deepspeed zero3 with:
```yaml
gradient_checkpointing: true
activation_offloading: true
```

EDIT: https://gist.github.com/winglian/54e2a57d3d0bf2c7771e8ed91f06f1e0
single GPU
`WORLD_SIZE=1 LOCAL_RANK=0 MASTER_ADDR=0.0.0.0 MASTER_PORT=29500 axolotl train config.yaml`

winglian:
I think something like this is sufficient as it's an issue with the checkpoint wrapper

```python
from torch.distributed.algorithms._checkpoint.checkpoint_wrapper import CheckpointWrapper

original_setattr = CheckpointWrapper.__setattr__

def new_setattr(self, name: str, value) -> None:
    if name.startswith('ds_') and hasattr(self, '_checkpoint_wrapped_module'):
        setattr(self._checkpoint_wrapped_module, name, value)
    else:
        original_setattr(self, name, value)

CheckpointWrapper.__setattr__ = new_setattr
```

PR #3106, opened 2025-08-26
Add support for PyTorch 2.8.0

winglian:
# Description

## Summary by CodeRabbit

* **Tests**
  * Expanded test matrices to include PyTorch 2.8.0 and added GPU end-to-end variants for CUDA 12.8.1 with Python 3.11.
* **Chores**
  * Added CUDA 12.8 (128) build variants across CI build jobs (additive); final variant marked as latest where applicable.
  * Relaxed triton version constraint in requirements to allow newer 3.4+ releases.
  * Adjusted CI setup logic to recognize Torch 2.8+ (no behavioral changes).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -55,7 +55,7 @@ jobs:
       fail-fast: false
       matrix:
         python_version: ["3.11"]
-        pytorch_version: ["2.6.0", "2.7.0", "2.7.1", "2.8.0"]
+        pytorch_version: ["2.6.0", "2.7.1", "2.8.0"]
     timeout-minutes: 20
 
     steps:
@@ -130,7 +130,7 @@ jobs:
       fail-fast: false
       matrix:
         python_version: ["3.11"]
-        pytorch_version: ["2.6.0", "2.7.0", "2.7.1", "2.8.0"]
+        pytorch_version: ["2.6.0", "2.7.1", "2.8.0"]
     timeout-minutes: 20
 
     steps:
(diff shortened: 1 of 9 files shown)


.github/workflows/tests.yml
```diff
@@ -55,7 +55,7 @@ jobs:
       fail-fast: false
       matrix:
         python_version: ["3.11"]
-        pytorch_version: ["2.6.0", "2.7.0", "2.7.1"]
```
djsaunde, line 58: any reason not to drop 2.7.0 in favor of 2.7.1?

requirements.txt
```diff
@@ -3,7 +3,7 @@
 # START section of dependencies that don't install on Darwin/MacOS
 bitsandbytes==0.47.0
 # triton 3.4.0 is not compatible with CCE
-triton>=3.0.0,<3.4.0
+triton>=3.0.0
```
djsaunde, line 5: Doesn't torch 2.8.0 require triton 3.4.0, and this ends up with an incompatibility? I think we need to fix our CCE, right?

djsaunde (approved):

.github/workflows/tests.yml
winglian: I think we had originally kept 2.7.0 due to vllm. let me double check and you're most likely spot on that we can drop 2.7.0

requirements.txt
winglian: I think this gets addressed with https://github.com/axolotl-ai-cloud/ml-cross-entropy/pull/19

salmanmohammadi (approved):

requirements.txt
```diff
@@ -3,7 +3,7 @@
 # START section of dependencies that don't install on Darwin/MacOS
 bitsandbytes==0.47.0
 # triton 3.4.0 is not compatible with CCE
```
NanoCode012:
```suggestion
```
Remove comment

PR #3107, opened 2025-08-26
Migrate QAT API; fix `axolotl quantize` for QAT-ed models; add NVFP4

salmanmohammadi:
The QAT API in torchao has recently been updated https://github.com/pytorch/ao/pull/2629. 

## Summary by CodeRabbit

- New Features
  - Unified quantization workflow with expanded dtypes (int4, int8, FP8, NVFP4) and streamlined QAT/PTQ APIs.
  - New CLI flag to push quantized models to the Hub with an auto-suffixed model name.
  - Baseten training template now auto-selects GPU type (H100/B200).

- Documentation
  - Added note explaining Hub name suffixing for quantized models.

- Examples
  - New and updated Llama 3 QAT configs (e.g., longer context, flash attention).

- Behavior Changes
  - Tokens-per-second metric disabled by default.

- Chores
  - Dependency bump to torchao 0.13.0; optional fbgemm-gpu extra and CI usage.

- Tests
  - E2E tests migrated to new quantization APIs and dtypes; added environment guards.


src/axolotl/core/training_args_base.py
```diff
@@ -51,7 +51,7 @@ class AxolotlTrainingMixins:
     )
     eval_sample_packing: Optional[bool] = field(
         default=None,
-        metadata={"help": "Use sample packing for efficient evals."},
+        metadata={"helps": "Use sample packing for efficient evals."},
```
winglian: Accidental edit?

src/axolotl/utils/quantization.py
```diff
@@ -45,50 +39,58 @@ def get_ptq_config(
...
+            "Use Int8DynamicActivationInt4WeightConfig with activation_dtype=TorchAOQuantDType.int8 and weight_dtype=TorchAOQuantDType.int4 instead."
+        )
+    if (
+        activation_dtype == TorchAOQuantDType.int8
+        and weight_dtype == TorchAOQuantDType.int8
+    ):
+        raise ValueError(
+            "Int8DynamicActivationInt8WeightConfig is not supported by torchao QAT. "
+            "Use Int8DynamicActivationInt4WeightConfig with activation_dtype=TorchAOQuantDType.int8 and weight_dtype=TorchAOQuantDType.int4 instead."
+        )
```
winglian: Should these live in the schema validation so it gets caught before train runtime?

examples/llama-3/3b-qat-fsdp2.yaml
```diff
@@ -6,24 +6,16 @@ load_in_8bit: false
...
 strict: false
 
-plugins:
-  - axolotl.integrations.liger.LigerPlugin
-
-liger_rope: true
-liger_rms_norm: true
-liger_glu_activation: true
-liger_layer_norm: true
-liger_fused_linear_cross_entropy: true
```
winglian: Does this no longer work?

src/axolotl/utils/quantization.py
```diff
@@ -3,30 +3,24 @@
...
-    IntXQuantizationAwareTrainingConfig,
+    QATConfig,
 )
 from torchao.quantization.quant_api import (
-    Int4DynamicActivationInt4WeightConfig,
+    Float8DynamicActivationFloat8WeightConfig,
+    Float8DynamicActivationInt4WeightConfig,
     Int4WeightOnlyConfig,
     Int8DynamicActivationInt4WeightConfig,
-    Int8DynamicActivationInt8WeightConfig,
```
winglian, line 18: Is w8a8 no linger supported? Would be nice to align support with downstream inference support. https://docs.vllm.ai/en/stable/features/quantization/int8.html

salmanmohammadi: cc @andrewor14

src/axolotl/utils/quantization.py
andrewor14: These are technically still supported in torchao. I'm just not sure if we have corresponding kernels in vllm for them. Maybe @jerryzh168 can comment on this

jerryzh168: int8 + int8 has corresponding kernels in torchao, either `_scaled_mm`, or the triton kernel that people can autotune

examples/llama-3/3b-qat-fsdp2-nvfp4.yaml
```diff
@@ -0,0 +1,64 @@
...
+strict: false
+
+plugins:
+  - axolotl.integrations.liger.LigerPlugin
+
+liger_rope: true
+liger_rms_norm: true
+liger_glu_activation: true
+liger_layer_norm: true
+liger_fused_linear_cross_entropy: true
```
andrewor14, line 16: remove these to be consistent?

requirements.txt
```diff
@@ -64,7 +64,7 @@ langdetect==1.0.9
 immutabledict==4.2.0
 antlr4-python3-runtime==4.13.2
 
-torchao==0.12.0
+torchao @ git+https://github.com/pytorch/torchao.git@13029fb6855bc19ceb8215b6dab204146908464b
```
andrewor14: we're about to release 0.13.0, so you can change it to that instead (in maybe 2 days)

src/axolotl/utils/schemas/quantization.py
```diff
@@ -35,26 +35,32 @@ class QATConfig(BaseModel):
...
-    def validate_dtype(cls, v: Any) -> TorchIntDType | None:
+    def validate_dtype(cls, v: Any) -> TorchAOQuantDType | None:
         if v == "int4":
-            return TorchIntDType.int4
+            return TorchAOQuantDType.int4
         if v == "int8":
-            return TorchIntDType.int8
-        raise ValueError(f"Invalid dtype: '{v}'. Must be one of: ['int4', 'int8']")
+            return TorchAOQuantDType.int8
+        if v == "float8_e4m3fn":
```
andrewor14: should we add an alias for "fp8" and "float8" as well to be more user friendly?

src/axolotl/utils/quantization.py
```diff
@@ -45,50 +53,59 @@ def get_ptq_config(
...
+    ):
+        raise ValueError(
+            "Int4DynamicActivationInt4WeightConfig is not supported by torchao QAT."
+        )
+    if (
+        activation_dtype == TorchAOQuantDType.int8
+        and weight_dtype == TorchAOQuantDType.int8
+    ):
+        raise ValueError(
+            "Int8DynamicActivationInt8WeightConfig is not supported by torchao QAT."
```
andrewor14, line 86: I think we can keep these configs? Any reason why we're removing them? Also the current error message is technically wrong, since we do still support them

src/axolotl/utils/quantization.py
```diff
@@ -45,50 +53,59 @@ def get_ptq_config(
...
+    if (
+        activation_dtype == TorchAOQuantDType.int8
+        and weight_dtype == TorchAOQuantDType.int4
+    ):
+        return Int8DynamicActivationInt4WeightConfig(group_size=group_size or -1)
+    if (
+        activation_dtype == TorchAOQuantDType.float8_e4m3fn
+        and weight_dtype == TorchAOQuantDType.float8_e4m3fn
+    ):
+        return Float8DynamicActivationFloat8WeightConfig(version=2)
```
andrewor14: maybe just leave out the version since it defaults to 2 (and will default to the highest version in the future)

src/axolotl/utils/quantization.py
```diff
@@ -45,50 +53,59 @@ def get_ptq_config(
...
+        and weight_dtype == TorchAOQuantDType.int8
+    ):
+        raise ValueError(
+            "Int8DynamicActivationInt8WeightConfig is not supported by torchao QAT."
+        )
+    if (
+        activation_dtype == TorchAOQuantDType.int8
+        and weight_dtype == TorchAOQuantDType.int4
+    ):
+        return Int8DynamicActivationInt4WeightConfig(group_size=group_size or -1)
```
andrewor14: hmm I'm not sure what -1 will do here, maybe just leave this out?

src/axolotl/utils/quantization.py
```diff
@@ -45,50 +53,59 @@ def get_ptq_config(
...
-    if activation_dtype == TorchIntDType.int4 and weight_dtype == TorchIntDType.int4:
-        return Int4DynamicActivationInt4WeightConfig()
-    if activation_dtype == TorchIntDType.int8 and weight_dtype == TorchIntDType.int8:
-        return Int8DynamicActivationInt8WeightConfig()
-    if activation_dtype == TorchIntDType.int8 and weight_dtype == TorchIntDType.int4:
-        return Int8DynamicActivationInt4WeightConfig()
+        if weight_dtype == TorchAOQuantDType.int8:
+            raise ValueError("Int8WeightOnlyConfig is not supported by torchao QAT.")
+        if weight_dtype == TorchAOQuantDType.int4:
+            return Int4WeightOnlyConfig(group_size=group_size or -1, version=2)
```
andrewor14: note that if you want BC you should use version=1 (that was the one you had before this PR, which uses the tinygemm kernel). If you want to default to the fbgemm kernel, then you should use version 2

src/axolotl/utils/quantization.py
```diff
@@ -45,50 +53,59 @@ def get_ptq_config(
...
+        activation_dtype == TorchAOQuantDType.float8_e4m3fn
+        and weight_dtype == TorchAOQuantDType.int4
+    ):
+        return Float8DynamicActivationInt4WeightConfig()
+    if weight_dtype == TorchAOQuantDType.nvfp4:
+        from torchao.prototype.mx_formats import NVFP4InferenceConfig
+
+        if group_size is not None and group_size != 16:
+            raise ValueError("NVFP4 quantization must use a group_size of 16")
+        return NVFP4InferenceConfig()
```
andrewor14, line 111: note that this one actually only does 1 level of scaling (unlike the specification in NVIDIA, which does 2 levels). This is because the PTQ only supports 1 level of scaling currently, but we plan to change this in the future (i.e. add dynamic per tensor scale). QAT supports both but currently uses 1 level of scaling to match the PTQ config

src/axolotl/utils/quantization.py
```diff
@@ -3,30 +3,38 @@
...
 )
 
-from axolotl.utils.schemas.enums import TorchIntDType
+from axolotl.utils.schemas.enums import TorchAOQuantDType
 
+quantization_config_to_str = {
+    Int4WeightOnlyConfig: "int4w",
+    Int8DynamicActivationInt4WeightConfig: "int8daint4w",
+    Float8DynamicActivationFloat8WeightConfig: "float8dafloat8w",
+    Float8DynamicActivationInt4WeightConfig: "float8daint4w",
```
andrewor14: nit: I feel we can leave out "da" from the name since it's mostly implied (static activations not super common right now). What do you think about the following naming:

```
int4: single dtype implies weight-only
nvfp4
int8int4: two dtypes imply dynamic activation + weights
fp8fp8
fp8int4
```

andrewor14 (commented):
Thanks, torchao changes look great!

src/axolotl/utils/quantization.py
salmanmohammadi: I was going by the supported configs [here](https://github.com/pytorch/ao/blob/379010f883f95d846e9b57d73cfb0d208c2a6664/torchao/quantization/qat/fake_quantize_config.py#L323). For example, consider the following snippet
```bash
> python
Python 3.11.11 (main, Jan  5 2025, 06:40:04) [Clang 19.1.6 ] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from torchao.quantization.quant_api import Int8DynamicActivationInt8WeightConfig
Skipping import of cpp extensions: Could not find libtorchao_ops_aten library in any of the provided paths
W0904 10:32:19.920000 16459 torch/distributed/elastic/multiprocessing/redirects.py:29] NOTE: Redirects are currently not supported in Windows or MacOs.
>>> conf = Int8DynamicActivationInt8WeightConfig()
>>> from torchao.quantization.qat import QATConfig
>>> QATConfig(conf)
QATConfig(base_config=Int8DynamicActivationInt8WeightConfig(layout=PlainLayout(), act_mapping_type=<MappingType.SYMMETRIC: 1>, weight_only_decode=False, set_inductor_config=True), activation_config=None, weight_config=None, step='prepare')
>>> import torch.nn as nn
>>> model = nn.Linear(100, 100)
>>> qat_config =  QATConfig(conf)
>>> quantize_(model, qat_config)
  File "<stdin>", line 1, in <module>
  File "/Users/salmanmohammadi/projects/axolotl/.venv/lib/python3.11/site-packages/torchao/quantization/quant_api.py", line 539, in quantize_
    _replace_with_custom_fn_if_matches_filter(
  File "/Users/salmanmohammadi/projects/axolotl/.venv/lib/python3.11/site-packages/torchao/quantization/quant_api.py", line 209, in _replace_with_custom_fn_if_matches_filter
    model = replacement_fn(model, *extra_args)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/salmanmohammadi/projects/axolotl/.venv/lib/python3.11/site-packages/torchao/quantization/qat/api.py", line 205, in _qat_config_transform
    (act_config, weight_config) = _infer_fake_quantize_configs(base_config)
  File "/Users/salmanmohammadi/projects/axolotl/.venv/lib/python3.11/site-packages/torchao/quantization/qat/fake_quantize_config.py", line 424, in _infer_fake_quantize_configs
    raise ValueError("Unexpected base config: %s" % base_config)
ValueError: Unexpected base config: Int8DynamicActivationInt8WeightConfig(layout=PlainLayout(), act_mapping_type=<MappingType.SYMMETRIC: 1>, weight_only_decode=False, set_inductor_config=True)
```
Let me know if I'm missing something here

examples/llama-3/3b-qat-fsdp2-nvfp4.yaml
salmanmohammadi: Yeah we probably don't need these when users are running on such a beefy GPU.

src/axolotl/utils/quantization.py
salmanmohammadi: Great suggestion : )

src/axolotl/utils/quantization.py
andrewor14: oh I see, I should add them to the new API as well, thought this was for PTQ only

salmanmohammadi: Will this make it into 0.13?

tests/e2e/test_quantization.py
```diff
@@ -206,6 +190,114 @@ def test_quantize_model_for_ptq(
...
+        self,
+        model,
+    ):
+        quantize_model(model, TorchAOQuantDType.nvfp4, TorchAOQuantDType.nvfp4)
+        for child in list(model.children()):
+            if isinstance(child, torch.nn.Linear):
+                assert isinstance(child.weight, LinearActivationQuantizedTensor), (
+                    "Linear weight should be quantized with activation quantization"
+                )
+                assert isinstance(child.weight, AffineQuantizedTensor), (
```
andrewor14: actually we don't use AQT for NVFP4 anymore (or in any of the newer quantization schemes). These use the new dtype specific tensor subclasses: https://github.com/pytorch/ao/blob/e368b6147f233274a1757f8a31157afdd424b97c/torchao/prototype/mx_formats/nvfp4_tensor.py#L64

If you want to check that activation is also nvfp4 here, you can check:
```
child.weight.act_quant_kwargs is not None and isinstance(child.weight.act_quant_kwargs, QuantizeTensorToNVFP4Kwargs)
```

salmanmohammadi: tysm

winglian:
@SalmanMohammadi does this supersede #3119 ?

.github/workflows/tests.yml
```diff
@@ -304,7 +304,7 @@ jobs:
             pytorch: 2.8.0
             num_gpus: 1
             gpu_type: "B200"
-            axolotl_extras:
+            axolotl_extras: fbgemm-gpu
```
winglian, line 307: We should probably include these in the docker artifacts.

https://github.com/axolotl-ai-cloud/axolotl/blob/main/.github/workflows/main.yml#L39-L43
and
https://github.com/axolotl-ai-cloud/axolotl/blob/main/.github/workflows/main.yml#L118-L122

andrewor14:
> @SalmanMohammadi does this supersede https://github.com/axolotl-ai-cloud/axolotl/pull/3119 ?

Yep just closed #3119

.github/workflows/tests.yml
salmanmohammadi: This dependency is only required for the int4 kernels - I figured we can include it in the docker images by default if there's interest from users and otherwise we can hold off on the addition dependency, as I found some corner cases when installing it. We have helpful error messages if users attempt to use these kernels without fbgemm installed.

src/axolotl/core/training_args_base.py
```diff
@@ -50,7 +50,7 @@ class AxolotlTrainingMixins:
         metadata={"help": "Use real batches for efficient training."},
     )
     include_tkps: bool = field(
-        default=True,
+        default=False,
```
winglian: I think this reverts a recent change?

src/axolotl/utils/quantization.py
winglian: Did we get a resolution on this?

src/axolotl/utils/quantization.py
winglian: Any updates on this or will it need to be addressed in an ao patch release?

winglian (approved):
I mostly have a bunch of follow ups on previous comments to make sure they are addressed. Otherwise seems good to me.

src/axolotl/utils/quantization.py
salmanmohammadi: It'll need to be in another ao release.

winglian: Is there a PR in ao we can track for this getting added back?

winglian:
2.8.0 still failing on:
```
>       to_quant = torch.split(x.to(torch.float), group_size, dim=-1)
                               ^^^^^^^^^^^^^^^^^
E       torch.AcceleratorError: CUDA error: no kernel image is available for execution on the device
E       CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
E       For debugging consider passing CUDA_LAUNCH_BLOCKING=1
E       Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.
```

salmanmohammadi:
> 2.8.0 still failing on:
> 
> ```
> >       to_quant = torch.split(x.to(torch.float), group_size, dim=-1)
>                                ^^^^^^^^^^^^^^^^^
> E       torch.AcceleratorError: CUDA error: no kernel image is available for execution on the device
> E       CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
> E       For debugging consider passing CUDA_LAUNCH_BLOCKING=1
> E       Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.
> ```

I'm able to manually run this on a B200 instance. I'm going to exclude this test for now and come back to it.
```bash
root@9a09ddf52b98:/workspace/axolotl# python
Python 3.11.13 (main, Jun  5 2025, 13:12:00) [GCC 11.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from axolotl.utils.schemas.enums import TorchAOQuantDType
>>> import torch
>>> weight_dtype = TorchAOQuantDType.int4
>>> act_dtype = TorchAOQuantDType.float8_e4m3fn
>>> from axolotl.utils.quantization import quantize_model
TMA benchmarks will be running without grid constant TMA descriptor.
>>> model = torch.nn.Linear(2048, 2048)
>>> quantize_model(model, weight_dtype, None, act_dtype)
```

PR #3113, opened 2025-08-29
fix(rl): pass max_prompt_len to training args as max_prompt_length

alex-ht:
# Description

This PR fixes a bug where `max_prompt_len` in GRPO config was not being passed into the training arguments.  
Now, if `self.cfg.max_prompt_len` is set, it will correctly be mapped to `training_args_kwargs["max_prompt_length"]`.

## Motivation and Context

- Previously, setting `max_prompt_len` in the config had no effect.  
- This fix ensures that the config parameter is properly respected by the training loop.  
- Helps prevent unexpected behavior when users rely on `max_prompt_len` to control context size.

## How has this been tested?

- Modified GRPO code to include `max_prompt_length` when `self.cfg.max_prompt_len` is provided.  
- Verified via logging that the value is correctly propagated into `training_args_kwargs`.  
- Ran a training job with `max_prompt_len=1024` and confirmed that the effective prompt length was limited.  

Environment:
- Python 3.12  
- PyTorch 2.7.1
- Axolotl 0.13.0.dev0

N/A

## Types of changes

- [ ] Documentation update

## Summary by CodeRabbit

* **New Features**
  * GRPO training now respects the max_prompt_len setting by passing it into training as max_prompt_length, making prompt length handling consistent across RL strategies.
  * Improves control over prompt truncation, memory usage, and training stability for GRPO workflows.
  * No other training behavior or public interfaces were changed.

### DIFF
--- a/src/axolotl/core/builders/rl.py
+++ b/src/axolotl/core/builders/rl.py
@@ -151,6 +151,9 @@ class HFRLTrainerBuilder(TrainerBuilderBase):
             training_args_cls = GRPOStrategy.get_training_args_class()
             training_args_kwargs.update(GRPOStrategy.set_training_args_kwargs(self.cfg))
             blocklist_args_kwargs = GRPOStrategy.get_blocklist_args_kwargs()
+            
+            if self.cfg.max_prompt_len:
+                training_args_kwargs["max_prompt_length"] = self.cfg.max_prompt_len
 
         elif self.cfg.rl in [RLType.DPO, RLType.IPO]:
             training_args_cls = AxolotlDPOConfig
--- a/src/axolotl/core/builders/rl.py
+++ b/src/axolotl/core/builders/rl.py
@@ -134,6 +134,14 @@ class HFRLTrainerBuilder(TrainerBuilderBase):
             if self.cfg.cpo_alpha is not None:
                 training_args_kwargs["cpo_alpha"] = self.cfg.cpo_alpha
 
+            # Handle when max_prompt_length == max_length from defaults
+            # CPOTrainer requires strictly less than
+            if (
+                training_args_kwargs["max_prompt_length"]
+                == training_args_kwargs["max_length"]
+            ):
+                training_args_kwargs["max_prompt_length"] -= 1
+
         elif self.cfg.rl is RLType.ORPO:
             training_args_cls = AxolotlORPOConfig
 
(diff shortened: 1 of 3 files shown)


alex-ht:
@coderabbitai review

NanoCode012:
Thanks for catching this! I see that we do set it for all except for CPO.

Could you remove the other settings https://github.com/axolotl-ai-cloud/axolotl/blob/0094a2d744553fb89e4874e2d79ac309f26cae77/src/axolotl/core/builders/rl.py#L134-L135 (including DPO's in `src/axolotl/core/trainers/dpo/__init__.py`) and set it to apply for all RL methods instead like below here https://github.com/axolotl-ai-cloud/axolotl/blob/0094a2d744553fb89e4874e2d79ac309f26cae77/src/axolotl/core/builders/rl.py#L120-L122

alex-ht:
> Thanks for catching this! I see that we do set it for all except for CPO.
> 
> Could you remove the other settings
> 
> https://github.com/axolotl-ai-cloud/axolotl/blob/0094a2d744553fb89e4874e2d79ac309f26cae77/src/axolotl/core/builders/rl.py#L134-L135
> 
> (including DPO's in `src/axolotl/core/trainers/dpo/__init__.py`) and set it to apply for all RL methods instead like below here
> https://github.com/axolotl-ai-cloud/axolotl/blob/0094a2d744553fb89e4874e2d79ac309f26cae77/src/axolotl/core/builders/rl.py#L120-L122

done, please review them

NanoCode012 (approved):
Thank you!

winglian (commented):
@NanoCode012 I don't know that we needed to pull it up out of the explicit places. `max_prompt_length` is not necessarily defined across all `*POConfig`'s, and could lead to object constructor failures.

NanoCode012:
@winglian , I saw that all except 2 of our *po configs do have this set. The ones missing were GRPO and CPO which both have this arg in their config too.

Is there a *PO config that doesn't have this that we support?

Another point is: should we be defaulting this to `cfg.sequence_len` if `max_prompt_len` is not defined?

Another user with this issue reported truncated prompts because of the lack of this setting: https://discord.com/channels/1104757954588196865/1111279858136383509/1415930062485585960

NanoCode012 (approved):
I've changed this to default `max_prompt_length` to `sequence_len` if not provided. Previously, the default `512` for all the trainers may unknowingly trim inputs. Ideally, we would want users to set this themselves.

PR #3120, opened 2025-08-30
go uv-first

### LINKED ISSUE #3121: uv-first

djsaunde:
# Description

Title. Move to `uv` over `pip` commands for installing from PyPI (`uv add`) and from source (`uv sync`), update Docker images and CI to use `uv` everywhere.

torch is now installed as part of the regular installation process. However, `uv` will respect the pre-installed torch version in the current environment (so long as it's within the pyproject.toml bounds.

I've also made flash-attn and deepspeed platform-dependent, non-optional dependencies, since it seems we _always_ recommend users to install (and seems like we only did this for macos incompatibility reasons?), but of course this is up for debate.

## Motivation and Context

- `uv` is quite fast --> translates to users' install times, CI build times
- `uv` has nice features (better resolution -> fewer dependency conflicts, torch autoselect (technically no need to preinstall torch unless you want an older supported version), automatic wheel selection, deterministic resolution)
- multiple requirements.txt -> pyproject.toml
- `setuptools_scm` auto-versioning 😌
- `auto-gptq` -> `gptqmodel` (caused uv locking issues)

## How has this been tested?

- [x] manually
- [x] CI (in progress)
- [x] docker image builds (in progress)
- [x] benchmarks
- [x] auto-versioning

### Tests CI comparison

`pip`: 
image

`uv`:
image

### Source install comparison

System info:
  - OS: Ubuntu 24.04.1 LTS
  - CPU: AMD Ryzen 7 3700X 8-Core Processor (16 threads)
  - Memory: 16GB RAM

#### Warm start

To compare apples to apples, I started with no torch installed and benchmarked the from-source install runtimes for both main (`pip install -e . ...`) and this branch (`uv sync`).

`pip`: 2m 45s ([script](https://gist.github.com/djsaunde/18a6bc94ab5180daa00b219f74e8ea13))
`uv`: 2s ([script](https://gist.github.com/djsaunde/ca527f1646907758b73a6bff73634fad))

Thanks to pre-resolving dependencies (via `uv.lock`), `uv` is super fast.

The real-world runtime is not quite right: in both cases, the needed packages are already cached locally so there's a warm start.

#### Cold start

Note that this benchmark will depends on my download speed (peak ~500MiB/s).

`pip`: 4m 34s ([script](https://gist.github.com/djsaunde/d587e5a65ebbbd6163faac5993be2249))
`uv`: 1m 24s ([script](https://gist.github.com/djsaunde/2266bdeccfd14e61e1914b445a0a153d))

`pip` is ~3.25x slower than `uv`.

In real-world use, I expect the average user will have some packages cached, but not all, so the runtimes should be somewhere between the warm and cold start cases. Anyways, we still save ~2-3m in either case.

## Summary by CodeRabbit

- Documentation
  - Updated installation guides to recommend uv with quick-start and advanced paths; simplified README requirements and examples; integration docs now use uv commands.
- Chores
  - Migrated packaging to pyproject.toml with explicit dependencies; removed setup.py and legacy requirements files.
  - Switched CI and Docker workflows to uv-based installs; base images/tags now -uv; extended CUDA arch list.
  - Adjusted manifests and coverage config to align with new packaging.
- Tests
  - Updated CI triggers and nightly/e2e pipelines; modal invocations now use -m filtering.


djsaunde:
Opening early for CI.

djsaunde:
This is pretty much good to go, save for migrating all the docker images to build on the uv base image. The old base images should be dropped (I think?). Frankly, I'm a bit bad at docker so hopefully I didn't break anything 😅

PR #3125, opened 2025-09-02
feat: add arg to enable dft in liger

NanoCode012:
# Description

Adds support for https://github.com/linkedin/Liger-Kernel/pull/860

Enable via `liger_use_token_scaling: true`

- [x] Requires upstream release from Liger, `>0.6.2`

## Summary by CodeRabbit

* New Features
  * Added optional setting to enable token scaling for fused linear cross-entropy. When turned on with the existing FLCE option, each token’s loss is scaled by its predicted probability (detached), and this behavior is enforced automatically during training.

* Documentation
  * Updated usage example to include the new FLCE-specific option, demonstrating how to enable token scaling alongside fused linear cross-entropy.

### DIFF
--- a/src/axolotl/integrations/liger/README.md
+++ b/src/axolotl/integrations/liger/README.md
@@ -18,6 +18,9 @@ liger_rms_norm: true
 liger_glu_activation: true
 liger_layer_norm: true
 liger_fused_linear_cross_entropy: true
+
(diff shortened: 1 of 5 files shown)


winglian:
This should be a draft, right? Since it needs a new Liger release

PR #3135, opened 2025-09-05
sleep model support

ved1beta:
# Description
sleep model support

## Motivation and Context

[https://docs.vllm.ai/en/latest/features/sleep_mode.html
](https://docs.vllm.ai/en/latest/features/sleep_mode.html)

works with
``` axolotl vllm-serve config.yml ```
``` 
vllm:
  enable_sleep_mode: true
  sleep_timeout: 60
```

## Summary by CodeRabbit

- New Features
  - Added optional sleep mode for vLLM to offload VRAM when idle.
  - Added configurable sleep timeout (in seconds) before entering sleep.
  - Options are available via both CLI and configuration; disabled by default.

### CODE REVIEW

**Comment by winglian** [comment]
I think this requires upstream changes to trl.scripts.vllm_serve.py

**Comment by qgallouedec** [comment]
We suport a sleep mode for collocated vLLM. I'm not sure why you'd need the same with the server mode, can you please elaborate?

File: src/axolotl/core/trainers/grpo/__init__.py [hunk 1]
```diff
@@ -52,6 +52,7 @@ def set_training_args_kwargs(cls, cfg: DictDefault) -> dict[str, Any]:
             if trl.vllm_mode:
                 grpo_args_kwargs["vllm_mode"] = trl.vllm_mode
             if trl.vllm_mode == "colocate":
+                grpo_args_kwargs["enable_sleep_mode"] = True
```
salmanmohammadi: Shouldn't this use the config field?

CodeMan62: done

src/axolotl/cli/args.py
```diff
@@ -103,6 +103,14 @@ class VllmServeCliArgs:
     reasoning_parser: Optional[str] = field(
         default=None,
     )
+    enable_sleep_mode: Optional[bool] = field(
```
salmanmohammadi: These are CLI args for `axolotl vllm serve`, and sleep mode isn't needed for vLLM server mode - we don't need these here.

src/axolotl/cli/vllm_serve.py
```diff
@@ -20,6 +20,9 @@ class AxolotlScriptArguments(ScriptArguments):
     reasoning_parser: str = field(default="", kw_only=True)
     enable_reasoning: bool | None = field(default=None, kw_only=True)
 
+    enable_sleep_mode: bool | None = field(default=None, kw_only=True)
```
salmanmohammadi: See above comment

src/axolotl/utils/schemas/vllm.py
```diff
@@ -57,3 +57,15 @@ class VllmConfig(BaseModel):
         default=None,
         json_schema_extra={"description": "Reasoning parser for VLLM"},
     )
+    enable_sleep_mode: bool | None = Field(
+        default=None,
+        json_schema_extra={
+            "description": "Enable sleep mode for vLLM to offload VRAM when idle"
+        },
+    )
+    sleep_timeout: int | None = Field(
```
salmanmohammadi: What is this being used for?

salmanmohammadi (changes requested):
Sorry, I just realized that `VLLMConfig` is the schema which is just used for vLLM server mode. Could you please move this to `TRLConfig` as `vllm_enable_sleep_mode`?

salmanmohammadi (approved):

PR #3136, opened 2025-09-06
feat: support training with JSON string tool arguments

gamersover:
## Description
Fix PyArrow data type inconsistency errors when datasets library loads data with dict-format tool arguments.

## Motivation and Context
When using hf datasets to load training data with dict-format arguments, PyArrow encounters data type inconsistency errors. The issue occurs when the same field has different data types across records. For example:

```json
"tool_calls": [
  {
    "function": {
      "name": "func1",
      "arguments": {
        "time": "2025-08-01"        // string type
      }
    }
  },
  {
    "function": {
      "name": "func2", 
      "arguments": {
        "time": 1690876800          // number type
      }
    }
  }
]
```

PyArrow fails during schema inference because the `time` field has inconsistent types (string vs number). Converting arguments to JSON string format ensures all argument fields are consistently typed as strings.

## How has this been tested?
- **Data Type Consistency**: Tested with datasets containing mixed argument types (time as string/number, id as string/integer) to verify consistent loading
- **Dataset Loading**: Verified that datasets.load_dataset() works correctly with JSON string arguments

**Testing Environment**:
* python3.11
* torch2.6
* cuda 12.8

## Types of changes
- [x] Documentation update

## Summary by CodeRabbit

- **New Features**
  - Chat templates now accept tool-call arguments encoded as JSON strings; stringified arguments are parsed into objects and invalid JSON produces clearer errors.

- **Documentation**
  - Conversation dataset docs include a warning advising to save identically named tool arguments with differing types as JSON strings and show an example.

- **Tests**
  - Added tests for dict vs JSON-string argument handling and mixed-type time values; tokenizer fixtures adjusted.


winglian:
If we could add another test case for this functionality to `tests/prompt_strategies/test_chat_template_ds_schema_unification.py`, that would be perfect.

NanoCode012:
Would the current case ever be an issue? I don't think someone would set different types for the same field.

I think my earlier comment was vague. What I meant was:
```json
    "function": {
      "name": "func1",
      "min": 0
    }

// second sample

    "function": {
      "name": "func2",
      // would there be a null min?
    }
```

gamersover:
To ensure broader compatibility, I believe it is necessary to add support for this case. I have encountered a situation where two different tools defined the same parameter name but with different types, which led to conflicts.

docs/dataset-formats/conversation.qmd
```diff
@@ -191,9 +191,7 @@ Instead of passing `tools` via the system prompt, an alternative method would be
                     "type": "function",
                     "function": {
                         "name": "...",
-                        "arguments": {
-                            "...": "...",
-                        }
+                        "arguments": "{\"...\": \"...\"}" // should be json string
```
NanoCode012: This seems to be an edge case, so we can revert this and put a NOTE / TIP at the bottom that for those with exact name keys but diff types to use json string instead.

Otherwise, users with working datasets may also try to convert it when unneeded.

NanoCode012 (commented):
Like wing have said, could you also add a test for `arguments: dict` and one for `json string`? We want to ensure this doesn't break backwards compatibility.

docs/dataset-formats/conversation.qmd
gamersover: Yes, that makes sense. I’ll prepare and share a test later to help verify the fix.

gamersover:
> Like wing have said, could you also add a test for `arguments: dict` and one for `json string`? We want to ensure this doesn't break backwards compatibility.

Unfortunately, we're still constrained by the datasets library's data loading mechanism. If we mix str and dict formats, it will directly throw an error. So users need to choose either all str or all dict - they can't be mixed.

I'd recommend we provide clear guidance to users suggesting they use the str format, as this can help avoid unnecessary compatibility issues and potential confusion.

NanoCode012:
> > Like wing have said, could you also add a test for `arguments: dict` and one for `json string`? We want to ensure this doesn't break backwards compatibility.
> 
> Unfortunately, we're still constrained by the datasets library's data loading mechanism. If we mix str and dict formats, it will directly throw an error. So users need to choose either all str or all dict - they can't be mixed.
> 
> I'd recommend we provide clear guidance to users suggesting they use the str format, as this can help avoid unnecessary compatibility issues and potential confusion.

I think there may be some misunderstanding. What I meant was: to add test for regular dict (our current default) and a separate one for json load to ensure both works. Not suggesting users to use both at same time :)

gamersover:
I've added comprehensive test coverage for the tool call arguments handling:

### Changes
- Added new test file: `tests/prompt_strategies/test_chat_templates_tool_call_string_arguments.py`

### Test Coverage
- `test_dict_and_str_args_produce_identical_output`: Verifies that both dict and JSON string formats of arguments produce identical tokenization results
- `test_str_args_with_mixed_time_types_no_error`: Ensures string format arguments handle mixed field types correctly (string vs number for the same field across different tool calls)

The tests pass with Qwen3-0.6B model. Let me know if you need any adjustments or additional test cases.

NanoCode012 (approved):

NanoCode012:
Thank you for the PR. I touched up a few things and merged it.

PR #3159, opened 2025-09-13
Debug log, logging improvements

djsaunde:
# Description

- debug log
  - all terminal output + debug-level logs and higher (regardless of `LOG_LEVEL` / `AXOLOTL_LOG_LEVEL`) get written to `f"{cfg.output_dir}/debug.log"`
  - resume from checkpoint / auto resume aware (continue log)
- fix slightly borked ruff config
- cleanup and simplifications

## Motivation and Context

Users can now just paste in their debug.log whenever opening an issue! 

## How has this been tested?

- [x] manually
- [x] pytest
- [x] multiprocess

For multiprocess, we might be duplicating logs, I'll have to check / fix if so.

E.g.:

```bash
 AXOLOTL_LOG_LEVEL=critical axolotl train ../configs/smollm.yaml --max-steps 5
```

[output](https://gist.github.com/djsaunde/8ba3b528b9334b23c3f9b5b89e8b0f5a)

```bash
$ cat model-out/debug.log
```

[output](https://gist.github.com/djsaunde/e9c665206f1be5dc8b2b605089b66b85)

## Summary by CodeRabbit

- New Features
  - Dedicated debug log file in the output directory, with optional stdout/stderr mirroring.
  - File captures DEBUG-level detail; console uses concise, colorized formatting with rank-aware context.
  - Python warnings are routed into logs; third‑party noise is filtered.

- Improvements
  - Clearer resume-from-checkpoint behavior with explicit logging.
  - Respects pre-set environment variables and reduces console verbosity for non-critical messages.

- Chores
  - Updated linting configuration and pre-commit settings.

- Tests
  - Added comprehensive tests for logging configuration, file capture, stdout teeing, and append/truncate behavior.


src/axolotl/cli/train.py
```diff
@@ -17,6 +17,9 @@
 from axolotl.train import train
 from axolotl.utils.config import normalize_config, resolve_dtype
 from axolotl.utils.dict import DictDefault
+from axolotl.utils.logging import get_logger
```
salmanmohammadi: Unused?

src/axolotl/utils/logging.py
```diff
@@ -40,10 +39,6 @@ def warning_once(self, *args, **kwargs):
 
 
 def get_logger(name: str, log_level: str | None = None) -> MultiProcessAdapter:
-    if log_level is None:
```
salmanmohammadi, line 43: How is this configured now?

src/axolotl/utils/tee.py
```diff
@@ -0,0 +1,166 @@
...
+
+        return _log_path
+
+
+def close_debug_log() -> None:
+    """Flush and close the debug log and uninstall the stdout/stderr tee.
+
+    Safe to call even if not initialized.
+    """
+    global _file_handle, _log_path, _tee_installed, _orig_stdout, _orig_stderr
```
salmanmohammadi, line 145: What's going on with `_tee_installed`? It's not immediately obvious to me.

src/axolotl/utils/logging.py
djsaunde: log level is configured in `configure_logging` via `DEFAULT_LOGGING_CONFIG`

src/axolotl/utils/tee.py
djsaunde: in case `prepare_debug_log` gets called multiple times, this guards against rewrapping stdout / stderr. This happens in the tests

src/axolotl/cli/__init__.py
```diff
@@ -4,5 +4,7 @@
 
 from axolotl.logging_config import configure_logging
 
-os.environ["HF_HUB_ENABLE_HF_TRANSFER"] = "1"
+os.environ.setdefault("TOKENIZERS_PARALLELISM", "false")
```
NanoCode012, line 7: What would be a side effect of this (like cases where no packing)?

src/axolotl/cli/config.py
```diff
@@ -227,8 +228,11 @@ def load_cfg(
         },
     )
 
+    # NOTE(djsaunde): We start outputting to output_dir/debug.log at this point since we
+    # have to wait for cfg.output to be resolved. We could call this earlier if we write
+    # to a temporary file, and then move it later.
+    prepare_debug_log(cfg)
     prepare_optim_env(cfg)
-    prepare_opinionated_env(cfg)
```
NanoCode012, line 231: Maybe I missed sth but why are we moving the pytorch alloc call from here to cli main?

src/axolotl/train.py
```diff
@@ -196,7 +196,18 @@ def execute_training(
...
+        auto_resume = bool(cfg.auto_resume_from_checkpoints)
+        if resume_from_checkpoint or auto_resume:
+            if resume_from_checkpoint:
+                LOG.info(
+                    "Resuming training from checkpoint: %s", resume_from_checkpoint
+                )
+            else:
+                LOG.info("Resuming training (auto-resume enabled)")
+
         LOG.info("Starting trainer...")
```
NanoCode012: Is this a new addition or duplicate of check in `utils/train.py`?

NanoCode012 (commented):
Could we test logging is as expected in Colab as well? It's called directly via python instead of via cli, so we should check if we need to modify the notebook

src/axolotl/train.py
djsaunde: Ah, duplicate, but we don't log when we resume, only when we autoresume. I think it'd be helpful to have both.

src/axolotl/cli/__init__.py
djsaunde: very very minor performance hit

src/axolotl/cli/config.py
djsaunde: `prepare_opinionated_env` didn't do the pytorch alloc

.pre-commit-config.yaml
```diff
@@ -14,7 +14,7 @@ repos:
     rev: v0.12.12
     hooks:
     -   id: ruff
-        args: [--fix, --select, I]
+        args: [--fix]
```
winglian, line 17: does this revert the isort you added recently?

djsaunde:
> Could we test logging is as expected in Colab as well? It's called directly via python instead of via cli, so we should check if we need to modify the notebook

@NanoCode012 great callout, turns out colab's logging and dictConfig don't play nicely 😢 will need to disable or fix somehow

src/axolotl/logging_config.py
```diff
@@ -144,12 +141,8 @@ def format(self, record):
 
 def configure_logging():
     """Configure with default logging"""
-    init()  # Initialize colorama
```
winglian, line 125: is colorama's `init` optional or does it need to be setup elsewhere?

djsaunde: Oops, didn't mean to remove

winglian (approved):

PR #3167, opened 2025-09-18
Add FSDP v2 swap memory support + QLoRA compatibility fixes

gholmes829:
# Description

  This PR enables training larger models than would normally be possible on resource constrained systems. This is accomplished by two main changes:
1. Adding opt-in config option to disable memory pinning for FSDP v2 which allows swap memory fallback in cases where CPU offloading exhausts available RAM.
2. Removes outdated validation prohibiting combination of FSDP v2 + QLoRA + cpu_ram_efficient_loading. This also indirectly unlocks enabling `qlora_sharded_model_loading` with FSDP v2. Previously, I had to fallback to FSDP v1 which is clearly undesirable.

Key changes:
- Adds swap memory support by disabling memory pinning when `cpu_offload_pin_memory: false` is set
- Passes configuration via env var to FSDP v2 monkeypatch, which configures `CPUOffloadPolicy(pin_memory=False)`
- Removes outdated blocker preventing FSDP v2 + QLoRA + `cpu_ram_efficient_loading` combination
- Updated validation / unit tests
  - FSDP v1 correctly rejects the swap memory feature
  - FSDP v2 requires `offload_params: true` when disabling pin_memory
  - FSDP v2 + QLoRA + `cpu_ram_efficient_loading` combination now works
- Documentation in the FSDP & QLoRA guide

Now viable:
```yaml
...
adapter: qlora
qlora_sharded_model_loading: true

fsdp_version: 2
fsdp_config:
  offload_params: true
  cpu_offload_pin_memory: false  # New field enables swap memory usage
  cpu_ram_efficient_loading: true  # Now works with FSDP v2 + QLoRA!
...
```

## Motivation and Context

  I have been messing around with several different popular fine tuning frameworks. One thing that drew me to Axolotl in particular, was its mention of special FSDP + QLoRa support, touted as potentially enabling training 70B model on 2x3090s / 48GB VRAM. I was sold and wasted no time trying it out.

...only to realize I hadn't stopped to read some of the fine print. The training feat in question was indeed conducted by only 2 consumer grade GPUs -- with the help of 128GB RAM. I only had 32 GB RAM to spare, and I eventually ran out of experimental optimization flags to hopelessly try out.

But then I realized, why should I be stopped from training a larger model? I had plenty of disk space and wasn't in a great hurry to fine tune my models, so I configured my system to provide 128GB swap memory to the container housing Axolotl.

[...]

PR #3170, opened 2025-09-22
`FSDPConfig`

salmanmohammadi:
## Summary by CodeRabbit

- New Features
  - Introduced a dedicated, typed FSDP configuration with clear options (e.g., activation checkpointing, parameter offloading, state dict types, auto-wrap policies, mixed precision).
  - Improved validation by using structured fields for safer, clearer config handling.

- Bug Fixes
  - Normalization/migration now consistently renames FSDP keys and removes legacy/conflicting entries, improving reliability across versioned configs.

- Tests
  - Expanded coverage to validate FSDP config migration, renamed fields, and preservation of non-versioned settings.


src/axolotl/utils/schemas/config.py
```diff
@@ -667,8 +668,7 @@ class AxolotlInputConfig(
         json_schema_extra={"description": "FSDP configuration"},
         deprecated="Configuring FSDP using `fsdp` is deprecated. Please use `fsdp_config` instead. ",
     )
-    # TODO @SalmanMohammadi strongly type this as its own schema
-    fsdp_config: dict[str, Any] | None = Field(
+    fsdp_config: FSDPConfig | None = Field(
```
NanoCode012, line 671: Would this mean we're now dropping fsdp1 support? Given that the fsdp_config schema only has fsdp2 configs, any prior fsdp1 configs would be dropped.

salmanmohammadi: `fsdp_config` should also include FSDP1 fields, am I missing some?

salmanmohammadi: There is also the old way of configuring FSDP1 through `cfg.fsdp` which remains.

NanoCode012: For ex:
```
fsdp_config:
  fsdp_transformer_layer_cls_to_wrap: MistralDecoderLayer
  fsdp_activation_checkpointing: true
```
With the new pydantic schema, would these be ignored?

salmanmohammadi: No, we strip `fsdp_` from any fields which have it and emit a deprecation warning to encourage users to just use `state_dict_type` etc. But those fields are common to both FSDP1 and FSDP2.

salmanmohammadi: There are tests which cover this in `test_normalize_config.py`.

NanoCode012: Thanks for clarifying. Forgot it gets stripped before the pydantic

NanoCode012 (approved):

salmanmohammadi:
Please hold off on merging until #3167 lands.

PR #3177, opened 2025-09-23
torch_dtype -> dtype

ved1beta:
# Description
torch_dtype -> dtype
``` `torch_dtype` is deprecated! Use `dtype` instead!   ```

## Motivation and Context
#3146 
## How has this been tested?
CI tests should be enough 

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensures models load correctly with the latest libraries by aligning dtype handling with updated APIs (no change to precision defaults).
* **Refactor**
  * Standardized dtype parameter usage across model loading and adapter paths for improved compatibility and consistency.
* **Tests**
  * Updated end-to-end tests to reflect the new dtype parameter usage, preserving existing test coverage and behavior.

### DIFF
--- a/src/axolotl/loaders/adapter.py
+++ b/src/axolotl/loaders/adapter.py
@@ -206,7 +206,7 @@ def load_llama_adapter(
         peft_model = PeftModel.from_pretrained(
             model,
             cfg.lora_model_dir,
-            torch_dtype=torch.float16,
+            dtype=torch.float16,
         )
     else:
         peft_model = get_peft_model(model, peft_config)
--- a/src/axolotl/loaders/adapter.py
+++ b/src/axolotl/loaders/adapter.py
@@ -206,7 +206,7 @@ def load_llama_adapter(
         peft_model = PeftModel.from_pretrained(
             model,
             cfg.lora_model_dir,
-            dtype=torch.float16,
+            torch_dtype=torch.float16,
         )
     else:
         peft_model = get_peft_model(model, peft_config)
--- a/src/axolotl/cli/delinearize_llama4.py
+++ b/src/axolotl/cli/delinearize_llama4.py
@@ -85,9 +85,7 @@ def do_cli(model: Union[Path, str], output: Union[Path, str]) -> None:
     unpatch_llama4 = patch_llama4_linearized_modeling()
     from transformers import Llama4ForConditionalGeneration
 
-    model_ = Llama4ForConditionalGeneration.from_pretrained(
-        model, torch_dtype=torch.bfloat16
-    )
+    model_ = Llama4ForConditionalGeneration.from_pretrained(model, dtype=torch.bfloat16)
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: src/axolotl/loaders/adapter.py [hunk 1]
```diff
@@ -206,7 +206,7 @@ def load_llama_adapter(
         peft_model = PeftModel.from_pretrained(
             model,
             cfg.lora_model_dir,
-            torch_dtype=torch.float16,
+            dtype=torch.float16,
```
NanoCode012: Can you double check if this error is raised in PeftModel? I only see it in PretrainedModel.from_pretrained . https://github.com/huggingface/transformers/blob/f64354e89a55a456e024b301468da2dbb10e11b3/src/transformers/modeling_utils.py#L2320-L2321

For PeftModel, I see that it already loads the proper `dtype` attribute but not clear if there's BC handling for torch_dtype https://github.com/huggingface/peft/blob/50329a713899cc4f963e26142b1ca688a6166882/src/peft/peft_model.py#L1464-L1465

djsaunde:
thanks for doing this. I wonder if we should move our `cfg.torch_dtype` -> `cfg.dtype` and alias the former to the latter with a deprecation warning

salmanmohammadi:
> `torch_dtype`

We use `bf16`, `fp16` instead no?

djsaunde:
> > `torch_dtype`
> 
> We use `bf16`, `fp16` instead no?

My understanding that `bf16` / `fp16` is used for mixed precision, and `torch_type` for master weights. I'm not sure we're actually honoring this everywhere in our codebase. CC @winglian for a sanity check

djsaunde:
Oh, we autoset here: https://github.com/axolotl-ai-cloud/axolotl/blame/b3b92687c4ba8792d343b6b1a616f541840db8b3/src/axolotl/utils/config/__init__.py#L89

tbh to me this seems a little backward. the user should be able specify arbitrary master weights / mixed precision settings

winglian:
> Oh, we autoset here: https://github.com/axolotl-ai-cloud/axolotl/blame/b3b92687c4ba8792d343b6b1a616f541840db8b3/src/axolotl/utils/config/__init__.py#L89
> 
> tbh to me this seems a little backward. the user should be able specify arbitrary master weights / mixed precision settings

honestly, everyone uses the bf16 master weights because of the memory overhead of fp32 master weights.

djsaunde:
> > Oh, we autoset here: https://github.com/axolotl-ai-cloud/axolotl/blame/b3b92687c4ba8792d343b6b1a616f541840db8b3/src/axolotl/utils/config/__init__.py#L89
> > tbh to me this seems a little backward. the user should be able specify arbitrary master weights / mixed precision settings
> 
> honestly, everyone uses the bf16 master weights because of the memory overhead of fp32 master weights.

yeah, not disputing that

salmanmohammadi (approved):

PR #3186, opened 2025-09-25
update lora kernels docs

djsaunde:
# Description


djsaunde:
@NanoCode012 related to our conversation. also, apparently I never ungated LoRA with bias in the validation, so I added that here + a test.

djsaunde:
ignore, not quite right

PR #3191, opened 2025-09-29
Diffusion trainer fix: shift logits to align with input tokens

djsaunde:
# Description

Title.

## Motivation and Context

Pretrained autoregressive models treat the output logits as right-shifted by one. By doing this, we should be able to use pretrained AR models effectively for diffusion model fine-tuning!

## Summary by CodeRabbit

- Bug Fixes
  - Corrects token alignment during diffusion generation and training so predictions match the corresponding input positions. This improves sampling stability and reduces artifacts in outputs.
  - Enhances loss calculation accuracy by aligning model predictions with inputs, leading to more consistent training behavior and better-quality results during inference.


src/axolotl/integrations/diffusion/utils.py
```diff
@@ -157,3 +157,10 @@ def create_bidirectional_attention_mask(
 
     # Add head dimension: [batch_size, 1, seq_len, seq_len]
     return bidirectional_mask.unsqueeze(1)
+
+
+def shift_logits_to_input_positions(logits: torch.Tensor) -> torch.Tensor:
+    """Align next-token logits with their input token positions for diffusion."""
+    if logits.size(1) <= 1:
+        return logits
+    return torch.cat([logits[:, :1], logits[:, :-1]], dim=1)
```
NanoCode012, line 166: What is this trying to do? Concat logit's first column and 1..N column together?

NanoCode012 (commented):
Test fail unrelated: out of space

src/axolotl/integrations/diffusion/utils.py
djsaunde: It's a bit of a hack to use pretrained causal LMs for diffusion fine-tuning. we're shifting logits to the right by one position so we align the input logits with the output logits

Unfortunately we're duplicating the first token, but I couldn't think of a better way to do it. open to ideas here

djsaunde: unfortunately we're duplicating the first token, but I can't think of a better way to accommodate pretrained causal LMs for diffusion fine-tuning. open to ideas here

winglian:
just as a data point, I had messed around with an early version of Dan's diffusion trainer a while back and here's the change I made to support next-token prediction https://github.com/axolotl-ai-cloud/axolotl/commit/cf8c93e2ee0920d2e223788db1c480fc8c6f13fb. my changes may be unnecessary, but wanted to make sure we didn't miss anything.

NanoCode012 (approved):

PR #3205, opened 2025-10-08
fix: transformers deprecate load_in_Xbit in model_kwargs

NanoCode012:
# Description

Upstream plans to deprecate these configs. We already used their new methods but was also using their legacy setting.

Related, 
https://github.com/huggingface/transformers/pull/41287
https://github.com/huggingface/transformers/pull/41283

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed inconsistencies where 4‑bit/8‑bit loading could be unintentionally overridden, ensuring explicit configuration flags are always honored.
  * QLoRA and LoRA paths now reliably follow the user’s selected quantization settings.

* **Refactor**
  * Simplified quantization handling by removing hidden fallbacks and cleanup steps, leading to more predictable and transparent behavior when configuring model loading.

### DIFF
--- a/src/axolotl/loaders/model.py
+++ b/src/axolotl/loaders/model.py
@@ -515,9 +515,6 @@ class ModelLoader:
             if self.cfg.model_quantization_config_kwargs:
                 mxfp4_kwargs = self.cfg.model_quantization_config_kwargs
             self.model_kwargs["quantization_config"] = Mxfp4Config(**mxfp4_kwargs)
-        else:
-            self.model_kwargs["load_in_8bit"] = self.cfg.load_in_8bit
-            self.model_kwargs["load_in_4bit"] = self.cfg.load_in_4bit
 
         if self.cfg.gptq:
             if not hasattr(self.model_config, "quantization_config"):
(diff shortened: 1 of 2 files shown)


src/axolotl/loaders/model.py
```diff
@@ -580,9 +575,7 @@ def _set_quantization_config(self):
             self.model_kwargs["quantization_config"] = BitsAndBytesConfig(
                 **bnb_config,
             )
-        elif self.cfg.adapter == "lora" and self.model_kwargs.get(
-            "load_in_8bit", False
-        ):
+        elif self.cfg.adapter == "lora" and self.cfg.load_in_8bit:
             bnb_config = {
                 "load_in_8bit": True,
```
NanoCode012, line 580: As we can do LoRA with bf16 weights, I think this may need to not check for 8bit in if condition and moved to the bnb config

salmanmohammadi (approved):

src/axolotl/loaders/model.py
```diff
@@ -596,11 +589,6 @@ def _set_quantization_config(self):
                 **bnb_config,
             )
 
-        # no longer needed per https://github.com/huggingface/transformers/pull/26610
-        if "quantization_config" in self.model_kwargs or self.cfg.gptq:
-            self.model_kwargs.pop("load_in_8bit", None)
-            self.model_kwargs.pop("load_in_4bit", None)
-
```
winglian, line 603: Should this be removed? It's a backup to remove them being set too, right?

NanoCode012: We set it up above, then remove it here. The former is removed with this PR. We can keep it but seems redundant?

NanoCode012: I double checked on this and we have tests checking if those args exist (which they don't) otherwise, the test would fail.

PR #3219, opened 2025-10-16
Fix trainer dataloader slow loading issue

qywu:
The current design removes "length" column before sampler in src/axolotl/core/trainers/base.py. However, it results in slow multipack sampler processing. 

The root cause is from https://github.com/axolotl-ai-cloud/axolotl/blob/main/src/axolotl/utils/samplers/utils.py#L13. When the dataset is large `get_dataset_length` will be extremely slow since it needs to iterate through the entire dataset.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,7 +49,7 @@ def parse_requirements(extras_require_map):
             try:
                 torch_version = version("torch")
             except PackageNotFoundError:
-                torch_version = "2.6.0"  # default to torch 2.6
+                torch_version = "2.8.0"  # default to torch 2.6
             _install_requires.append(f"torch=={torch_version}")
 
             version_match = re.match(r"^(\d+)\.(\d+)(?:\.(\d+))?", torch_version)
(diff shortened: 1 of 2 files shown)


NanoCode012:
Hey, thanks for finding this. Do you have some benchmarks for before/after change?

~Could you also remove the accidental `setup.py` commit?~ Edit: Seems like Wing reviewed and made 2.8 default below.

setup.py
```diff
@@ -49,7 +49,7 @@ def parse_requirements(extras_require_map):
             try:
                 torch_version = version("torch")
             except PackageNotFoundError:
-                torch_version = "2.6.0"  # default to torch 2.6
+                torch_version = "2.8.0"  # default to torch 2.6
```
winglian:
```suggestion
                torch_version = "2.8.0"  # default to torch 2.8.0
```

src/axolotl/core/trainers/base.py
winglian, line 245: Would `length` also get dropped here too?

qywu: yeah, it is fine to drop length after sampler since sampler has used that info already.

winglian (approved):
did some quick tests and lgtm. thanks!

Issue #3217, opened 2025-10-16
DPO Training Fails with Tool Role Messages - KeyError: 'tool'


Ofir408:
### Please check that this issue hasn't been reported before.

- [x] I searched previous [Bug Reports](https://github.com/axolotl-ai-cloud/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

# DPO Training Fails with Tool Role Messages - KeyError: 'tool'

## Bug Description

DPO training fails when dataset contains `role: "tool"` messages, even though Llama 3.1 natively supports tool roles.

## Error

```
KeyError: 'tool'
File "/axolotl/prompt_strategies/dpo/chat_template.py", line 57
"role": role_map[m[message_property_mappings["role"]]],
```

## Reproduction

**Data format:**
```json
{
  "messages": [
    {"role": "user", "content": "Search for AI info"},
    {"role": "assistant", "tool_calls": [{"id": "call_123", "function": {"name": "search"}}]},
    {"role": "tool", "name": "search", "tool_call_id": "call_123", "content": "Results..."}
  ],
  "chosen": {"role": "assistant", "content": "Based on results..."},
  "rejected": {"role": "assistant", "content": "I don't know."}
}
```

**Config:**
```yaml
base_model: meta-llama/Llama-3.1-8B-Instruct
rl: dpo
chat_template: llama3
datasets:
  - path: data.jsonl
    type: chat_template
    field_messages: "messages"
    field_chosen: "chosen" 
    field_rejected: "rejected"
    message_property_mappings:
      role: role
      content: content
    roles:
      user: ["user"]
      assistant: ["assistant"]
      system: ["system"]
      tool: ["tool"]
    roles_to_train: ["assistant"]

```

**Command:** `axolotl preprocess config.yaml`

## Expected Behavior

Should work since:
- Llama 3.1 supports tool roles natively
- SFT training works fine with same data/config
- `llama3` chat template supports tools

## Root Cause

DPO chat template processor missing `tool` role in `role_map`.

## Request

Add native `tool` role support to DPO training to match SFT capabilities.

### Current behaviour

I wrote above

### Steps to reproduce

I wrote above

### Config yaml

```yaml

```

### Which Operating Systems are you using?

- [ ] Linux
- [x] macOS
- [ ] Windows

### Python Version

3.10

### axolotl branch-commit

main

### Acknowledgements

- [x] My issue title is concise, descriptive, and in title casing.
[...]

PR #3228, opened 2025-10-23
include tool in default message_property_mappings

winglian:
# Description

adds `tool` to `message_property_mappings` alongside `role` and `content`

## Motivation and Context

#3217 

## Summary by CodeRabbit

* **New Features**
  * Chat templates now include "tool" as a default message property mapping. Previously, only role and content were included in the default mappings. This enhancement provides first-class support for tool properties, enabling better configuration of message templates for tool and function calling scenarios.

### DIFF
--- a/src/axolotl/prompt_strategies/chat_template.py
+++ b/src/axolotl/prompt_strategies/chat_template.py
@@ -48,9 +48,9 @@ class ChatTemplatePrompter(Prompter):
     ):
         # check if message_property_mappings is None or empty dict
         if message_property_mappings is None or (not message_property_mappings):
+            default_message_property_mappings_keys = ["role", "content", "tool"]
             message_property_mappings = {
-                "role": "role",
-                "content": "content",
+                prop: prop for prop in default_message_property_mappings_keys
             }
             if template_thinking_key and field_thinking:
                 message_property_mappings[template_thinking_key] = field_thinking


NanoCode012:
I don't think this solves the linked issue.

First, DPO uses a separate `chat_template` https://github.com/axolotl-ai-cloud/axolotl/blob/main/src/axolotl/prompt_strategies/dpo/chat_template.py

`tool` is actually not at the message level as seen in your changes. The one at that level would be `tool_calls` which we need to add.

I think the required changes (as also posted in Discord to the author) is:
1. Make messages load other props too https://github.com/axolotl-ai-cloud/axolotl/blob/bb33fda44d8cc889230698539b8df5a7ba114b67/src/axolotl/prompt_strategies/dpo/chat_template.py#L55-L61
2. Adjust chosen and rejected to take other props (like name for mistral) https://github.com/axolotl-ai-cloud/axolotl/blob/bb33fda44d8cc889230698539b8df5a7ba114b67/src/axolotl/prompt_strategies/dpo/chat_template.py#L73-L76
3. Carefully evaluate whether there can be edge cases when we use this method of detecting boundaries https://github.com/axolotl-ai-cloud/axolotl/blob/bb33fda44d8cc889230698539b8df5a7ba114b67/src/axolotl/prompt_strategies/dpo/chat_template.py#L92-L108

Issue #3169, opened 2025-09-20
Cannot save checkpoints after training already started


seungduk-yanolja:
### ⚠️ Please check that this feature request hasn't been suggested before.

- [x] I searched previous [Ideas in Discussions](https://github.com/axolotl-ai-cloud/axolotl/discussions/categories/ideas) didn't find any similar feature requests.
- [x] I searched previous [Issues](https://github.com/axolotl-ai-cloud/axolotl/labels/enhancement) didn't find any similar feature requests.

### 🔖 Feature description

I forgot to modify the number of saves in the config file, and the training has already started and is running.
It would be great if we added a feature that saves a checkpoint when there is a key input. Or, Axolotl can offer an API endpoint that can communicate even during training.

### ✔️ Solution

1. Add a hotkey that triggers checkpoint saving
2. Run a simple API server

### Acknowledgements

- [x] My issue title is concise, descriptive, and in title casing.
- [x] I have searched the existing issues to make sure this feature has not been requested yet.
- [x] I have provided enough information for the maintainers to understand and evaluate this request.

djsaunde:
Thanks for your suggestion! I like the idea of having a simple API server that allows users to interact with live training runs -- this is something I've been thinking about lately, and it could have a lot of use cases (live inference, training stats, etc.). We'll chat about this internally and see if we can't implement something that solves this.

waveplate:
It would be nice to be able to send axolotl `SIGHUP` or `SIGUSR1` to trigger a checkpoint save. I doubt many people even know about the file-based trigger before they need it.

NanoCode012:
@waveplate , we have this https://github.com/axolotl-ai-cloud/axolotl/blob/66a3de36292e6a6afa5e0916c8392e7db41ba822/src/axolotl/train.py#L163-L166 . If you sent SIGINT (or control+c), it should start the save procedure

PR #3233, opened 2025-10-26
feat: save checkpoint after training started 

ved1beta:
# Description
feat: save checkpoint after training started

## Motivation and Context
#3169

## How has this been tested?
unit tests

## Release Notes

* **New Features**
  * Added dynamic checkpoint triggering: save checkpoints on-demand during training using file-based triggers (marker file) or Unix signals (SIGUSR1).
  * Seamless support for single-process and distributed training environments with configurable check intervals.

### DIFF
--- a/src/axolotl/utils/callbacks/dynamic_checkpoint.py
+++ b/src/axolotl/utils/callbacks/dynamic_checkpoint.py
@@ -99,7 +99,6 @@ class DynamicCheckpointCallback(TrainerCallback):
                     trigger_detected = True
                     self.should_save_checkpoint = False  # Reset flag
 
-            # In distributed mode, synchronize the trigger decision across all ranks
             if is_distributed():
                 import torch
                 import torch.distributed as dist
@@ -110,18 +109,11 @@ class DynamicCheckpointCallback(TrainerCallback):
                     torch.device("cuda" if torch.cuda.is_available() else "cpu"),
                 )
 
-                if is_main_process():
-                    trigger_tensor = torch.tensor(
-                        1 if trigger_detected else 0,
-                        dtype=torch.long,
-                        device=device,
-                    )
-                else:
-                    trigger_tensor = torch.tensor(
-                        0,
-                        dtype=torch.long,
-                        device=device,
-                    )
+                trigger_tensor = torch.tensor(
+                    1 if trigger_detected else 0,
+                    dtype=torch.long,
+                    device=device,
+                )
 
                 dist.broadcast(trigger_tensor, src=0)
 
--- a/src/axolotl/utils/callbacks/dynamic_checkpoint.py
+++ b/src/axolotl/utils/callbacks/dynamic_checkpoint.py
@@ -1,0 +1,139 @@
+import signal
+import os
+from pathlib import Path
+from typing import Optional
+from transformers import (
+    TrainerCallback,
+    TrainerControl,
+    TrainerState,
+    TrainingArguments,
+)
+from axolotl.utils.distributed import (
+    is_main_process,
+    barrier,
+    is_distributed,
+)
+from axolotl.utils.logging import get_logger
+
+LOG = get_logger(__name__)
+
+DEFAULT_TRIGGER_FILENAME = ".axolotl_save_checkpoint"
+
+
+class DynamicCheckpointCallback(TrainerCallback):
+    """
+    Callback to save checkpoints on-demand during training via:
+    1. File-based trigger (works everywhere, rank 0 checks file)
+    2. Signal-based trigger (Unix/Linux only, rank 0 sets flag)
+
+    Thread-safe for multi-GPU distributed training.
+    Addresses issue: https://github.com/axolotl-ai-cloud/axolotl/issues/3169
+
+    Usage:
+        # File-based:
+        touch /path/to/output_dir/.axolotl_save_checkpoint
+        # Signal-based (Unix/Linux):
+        kill -SIGUSR1 <training_pid>
+    """
+
+    def __init__(self, cfg):
+        self.cfg = cfg
+        self.enabled = cfg.dynamic_checkpoint.enabled
+
+        if not self.enabled:
+            return
+
+        self.trigger_filename = (
+            cfg.dynamic_checkpoint.trigger_file_path or DEFAULT_TRIGGER_FILENAME
+        )
+
+        self.check_interval = cfg.dynamic_checkpoint.check_interval
+        self.enable_signal = cfg.dynamic_checkpoint.enable_signal
+
+        self.should_save_checkpoint = False
+
+        if self.enable_signal and hasattr(signal, "SIGUSR1") and is_main_process():
+            signal.signal(signal.SIGUSR1, self._signal_handler)
+            LOG.info(
+                f"Dynamic checkpoint: Signal handler registered on rank 0 (SIGUSR1). "
+                f"Trigger with: kill -SIGUSR1 {os.getpid()}",
+                main_process_only=True,
+            )
+
+        LOG.info(
+            f"Dynamic checkpoint enabled. To trigger checkpoint save:\n"
+            f"  • File: touch {cfg.output_dir}/{self.trigger_filename}\n"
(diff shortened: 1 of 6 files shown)


src/axolotl/utils/callbacks/dynamic_checkpoint.py
```diff
@@ -0,0 +1,156 @@
...
+
+                trigger_tensor = torch.tensor(
+                    1 if trigger_detected else 0,
+                    dtype=torch.long,
+                    device=f"cuda:{torch.cuda.current_device()}",
+                )
+
+                dist.broadcast(trigger_tensor, src=0)
+
+                trigger_detected = bool(trigger_tensor.item())
```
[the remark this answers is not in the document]
ved1beta: resolved

src/axolotl/utils/callbacks/dynamic_checkpoint.py
```diff
@@ -0,0 +1,159 @@
...
+                    device = (
+                        torch.device("cuda", torch.cuda.current_device())
+                        if torch.cuda.is_available()
+                        else torch.device("cpu")
+                    )
+                trigger_tensor = torch.tensor(
+                    1 if trigger_detected else 0,
+                    dtype=torch.long,
+                    device=device,
+                )
```
NanoCode012, line 118: I'm not so clear what this does?

src/axolotl/utils/callbacks/dynamic_checkpoint.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+class DynamicCheckpointCallback(TrainerCallback):
+    """
+    Callback to save checkpoints on-demand during training via:
+    1. File-based trigger (works everywhere, rank 0 checks file)
+    2. Signal-based trigger (Unix/Linux only, rank 0 sets flag)
+
+    Thread-safe for multi-GPU distributed training.
+    Addresses issue: https://github.com/axolotl-ai-cloud/axolotl/issues/3169
```
NanoCode012:
```suggestion
```

src/axolotl/utils/callbacks/dynamic_checkpoint.py
```diff
@@ -0,0 +1,159 @@
...
+
+                trigger_detected = bool(trigger_tensor.item())
+
+                barrier()
+
+        if trigger_detected:
+            control.should_save = True
+            LOG.info(
+                f"Saving dynamic checkpoint at step {state.global_step}",
+                main_process_only=False,  # Log on all ranks for debugging
```
NanoCode012:
```suggestion
                main_process_only=True,
```

I'm not sure we need to log on all ranks

src/axolotl/utils/schemas/dynamic_checkpoint.py
```diff
@@ -0,0 +1,37 @@
...
+
+class DynamicCheckpointConfig(BaseModel):
+    """Configuration for dynamic checkpoint triggering during training."""
+
+    enabled: bool = Field(
+        default=False,
+        json_schema_extra={
+            "description": "Enable dynamic checkpoint triggering during training. "
+            "Create a file '.axolotl_save_checkpoint' in output_dir to trigger. "
+            "Addresses: https://github.com/axolotl-ai-cloud/axolotl/issues/3169"
```
NanoCode012:
```suggestion
```

src/axolotl/utils/callbacks/dynamic_checkpoint.py
ved1beta: its for multi gpu only first rank should log. simplified it a little bit

salmanmohammadi: Yeah this doesn't make sense to me...

NanoCode012: I think you can do a `is_main_process` and compute inside? It would work for both single/multi-gpu

src/axolotl/utils/callbacks/dynamic_checkpoint.py
```diff
@@ -0,0 +1,138 @@
...
+                import torch
+                import torch.distributed as dist
+
+                device = getattr(
+                    args,
+                    "device",
+                    torch.device("cuda" if torch.cuda.is_available() else "cpu"),
+                )
+
+                if is_main_process():
```
salmanmohammadi: Can you explain why you need this?

ved1beta: ok ig nested is_main_process() check is unnecessary

salmanmohammadi: I'm not sure I follow!

src/axolotl/utils/schemas/dynamic_checkpoint.py
```diff
@@ -0,0 +1,23 @@
...
+
+    enabled: bool = Field(
+        default=False,
+        json_schema_extra={
+            "description": "Enable dynamic checkpoint triggering during training. "
+            "Create a file '.axolotl_save_checkpoint' in output_dir to trigger. "
+        },
+    )
+    check_interval: int = Field(
+        default=100,
```
winglian: I think a default of 100 negates the benefit of this feature depending on what is already configured as save_steps.
```suggestion
        default=10,
```

src/axolotl/utils/schemas/dynamic_checkpoint.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class DynamicCheckpointConfig(BaseModel):
+    """Configuration for dynamic checkpoint triggering during training."""
+
+    enabled: bool = Field(
+        default=False,
+        json_schema_extra={
+            "description": "Enable dynamic checkpoint triggering during training. "
+            "Create a file '.axolotl_save_checkpoint' in output_dir to trigger. "
```
winglian:
```suggestion
            "Create a file '.axolotl_save_checkpoint' in the configured `output_dir` to trigger. "
```

src/axolotl/utils/callbacks/dynamic_checkpoint.py
```diff
@@ -0,0 +1,130 @@
...
+from axolotl.utils.distributed import (
+    barrier,
+    is_distributed,
+    is_main_process,
+)
+from axolotl.utils.logging import get_logger
+
+LOG = get_logger(__name__)
+
+DEFAULT_TRIGGER_FILENAME = ".axolotl_save_checkpoint"
```
winglian: using "hidden files" is a bad pattern. Requires extra cognitive overhead to remember to use `ls -la` to look for it when touching it manually. Something like `axolotl_checkpoint.save` is probably good too.

src/axolotl/utils/schemas/dynamic_checkpoint.py
```diff
@@ -0,0 +1,23 @@
+"""Schema for dynamic checkpoint configuration."""
+
+from pydantic import BaseModel, Field
+
+
+class DynamicCheckpointConfig(BaseModel):
```
winglian, line 6: It might be good to allow the user to override the trigger filename in this configuration

src/axolotl/utils/callbacks/dynamic_checkpoint.py
ved1beta: noted :saluting_face:

NanoCode012 (approved):
I think the changes are isolated, so this can be interesting feature to have in.

I think we can see user feedback if there's any issue.

src/axolotl/utils/callbacks/dynamic_checkpoint.py
winglian: I think this is correct. So you have to create the tensor on all ranks, and on rank 0, it will correctly set based on the value of `trigger_detected` and then below, on `dist.broadcast` with `src=0`, all ranks get the tensor updated to what ever the main process found.

winglian: @coderabbitai, what do you think?

winglian (approved):

PR #3238, opened 2025-10-30
_get_tools in ChatTemplateStrategy : function "parameters" can be dict or string

eduardzl:
# Description
When training of function calls, "tools" elements of a dataset can contain same parameter name but with different types. Datasets fails to load such training set. This fix allows "parameters" element of function call to be string( by running "json.dumps" in preparation of training data set). The _get_tools function will iterate over tool definitions, if "parameters" element is dict, it will keep that way, it is a string, it will be converted to dict by invoking "json.loads" on string value.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of tool parameters to properly parse string-based parameters, ensuring consistency across tool parameter processing.
  * Enhanced error logging and handling when parameter parsing encounters issues, providing better visibility into failures.

### DIFF
--- a/docs/dataset-formats/conversation.qmd
+++ b/docs/dataset-formats/conversation.qmd
@@ -218,6 +218,13 @@ If you have tool arguments with same name but different dtypes (like `"time": st
 ```
 "arguments": "{\"...\": \"...\"}"
 ```
+
+The same is applicable for tool parameters.
+
+```
+"parameters": "{\"...\": \"...\"}"
+```
+
 :::
 
 Example config for Llama4:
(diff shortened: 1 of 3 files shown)


NanoCode012:
Thank you for the PR and confirming it worked! 

Would you be able to:
1. Add test similar to how it's done on tool_call https://github.com/axolotl-ai-cloud/axolotl/blob/main/tests/prompt_strategies/test_chat_template_ds_schema_unification.py
2. Add this info to docs as a warning tooltip for the tool use section?

Discord ref for context behind issue: https://discord.com/channels/1104757954588196865/1111279858136383509/1433100330949677147

NanoCode012 (approved):

Issue #3184, opened 2025-09-25
Support muon with deepspeed


zinccat:
### ⚠️ Please check that this feature request hasn't been suggested before.

- [x] I searched previous [Ideas in Discussions](https://github.com/axolotl-ai-cloud/axolotl/discussions/categories/ideas) didn't find any similar feature requests.
- [x] I searched previous [Issues](https://github.com/axolotl-ai-cloud/axolotl/labels/enhancement) didn't find any similar feature requests.

### 🔖 Feature description

Current code checks and blocks using muon with deepspeed, now with a new pr for deepspeed, it should enable this possibility 

### ✔️ Solution

https://github.com/deepspeedai/DeepSpeed/pull/7509

### Acknowledgements

- [x] My issue title is concise, descriptive, and in title casing.
- [x] I have searched the existing issues to make sure this feature has not been requested yet.
- [x] I have provided enough information for the maintainers to understand and evaluate this request.

NanoCode012:
We currently use a custom implementation for muon https://github.com/axolotl-ai-cloud/axolotl-contribs-mit/blob/main/src/axolotl/contribs/mit/muon.py

We're open to a PR for this if you would like to give it a try.

PR #3258, opened 2025-11-11
Allow muon optimizer with DeepSpeed Zero 1-2

lhl:
DeepSpeed has muon support merged for DSZ1/2: https://github.com/deepspeedai/DeepSpeed/pull/7509

I made a simple modification to support this: https://github.com/axolotl-ai-cloud/axolotl/issues/3184

Has a few guards (I can remove if unneccessary) but the main functionality is simple - if DeepSpeed and ZeRO<3 then allows muon. Trains ok for me.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Muon optimizer compatibility with DeepSpeed by enabling support for ZeRO stages 1-2.
  * Enhanced validation error messages for DeepSpeed configuration incompatibilities.

* **Tests**
  * Added test coverage for Muon optimizer with DeepSpeed ZeRO stage 2.

### DIFF
--- a/src/axolotl/utils/schemas/validation.py
+++ b/src/axolotl/utils/schemas/validation.py
@@ -740,6 +740,31 @@ class RLValidationMixin:
 class OptimizationValidationMixin:
     """Validation methods related to optimization and performance."""
 
+    @staticmethod
+    def _get_deepspeed_zero_stage(deepspeed_cfg):
+        if not deepspeed_cfg:
+            return None
+
+        ds_config = None
+        if isinstance(deepspeed_cfg, (str, Path)):
+            try:
+                with open(deepspeed_cfg, "r", encoding="utf-8") as ds_fin:
+                    ds_config = json.load(ds_fin)
+            except (OSError, json.JSONDecodeError) as exc:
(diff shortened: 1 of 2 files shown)


winglian:
We probably want https://github.com/axolotl-ai-cloud/axolotl/pull/3261 since we'll need a newer version of deepspeed that we currently have in `main`

lhl:
> We probably want #3261 since we'll need a newer version of deepspeed that we currently have in `main`

Anything you need from my for this PR?

BTW, just as an FYI I actually got sidetracked the other and rewrote Muon in Axolotl and will submit that once the rest of the PRs I have get cleared out. Mentioned in Discord just now, but dvantage of new version:

- Full support for FSDP/DeepSpeed (including ZeRO-3) - yes, this was a huge pita lol
- Recovers throughput - as fast as AdamW (current implementation is like 10-15% slower?)
- Better numerics? Current path I think may be silently falling back to AdamW momentum or something else subtle. My Muon results differ and are better. I have unit and e2e tests and reproducible "real" runs (1epoch of large dataset on a llama-3.2 1b)
- Basic support for Muon Clip (qk_clip) - correct but slow (fast requires modifying attention kernels to pass more data out)

PR #3259, opened 2025-11-11
Moe aux loss free

lhl:
This is a new plugin that adds aux-loss-free balancing as a gating option to compatible MoE architectures without forking model code.

This is basically an plugin/monkeypatching version of DeepSeek's [Auxiliary-Loss-Free Load Balancing Strategy for Mixture-of-Experts](https://arxiv.org/pdf/2408.15664).

This code has been tested and the plugin itself should be pretty well documented.

Some more background about my ALFB approach: 
- https://chatgpt.com/share/69136736-9a1c-8012-b051-77f6cc157464

Compatibility:
- mixtral
- qwen3moe
- bailingv2
- llama4 (text layers)
- passhthrough for deepspeedv3

Btw I know @NanoCode012 had some comments on changing the telemetry/logging vars/configs?

Take a look at the README/config stuff, happy to make some updates to bring it inline.

## Release Notes

* **New Features**
  * Added Aux-Loss-Free MoE Router plugin enabling auxiliary-loss-free routing for supported Mixture-of-Experts models (Mixtral, Qwen3, Llama4, Bailing, Ring).

* **Documentation**
  * Added comprehensive documentation for configuring and using the Aux-Free MoE Router.

* **Configuration**
  * Added new MoE routing parameters: `moe_balance_type`, `moe_update_rate`, `moe_update_momentum`, `moe_bias_cap`, `moe_afb_warmup_steps`, `moe_bias_sync_group`, and `expert_parallel_size`.

* **Tests**
  * Added end-to-end and unit tests validating aux-free routing functionality and parity with traditional auxiliary-loss routing.

### DIFF
--- a/src/axolotl/integrations/aux_free_router/adapters.py
+++ b/src/axolotl/integrations/aux_free_router/adapters.py
@@ -189,8 +189,14 @@ class BailingAdapter(BaseMoEAdapter):
     family = "bailing_moe"
 
     def matches(self, model: nn.Module) -> bool:
-        model_type = getattr(getattr(model, "config", object()), "model_type", "")
-        return model_type in ("bailing_moe", "bailing_moe_v2")
+        cfg = getattr(model, "config", None)
+        if cfg is None:
+            return False
+        model_type = getattr(cfg, "model_type", "") or ""
+        if model_type in ("bailing_moe", "bailing_moe_v2", "ring_moe", "ring"):
+            return True
+        cfg_name = cfg.__class__.__name__.lower()
+        return "bailingmoev2" in cfg_name or "ring" in cfg_name
 
     def find_moe_layers(self, model: nn.Module) -> Iterable[nn.Module]:
         for m in model.modules():
--- a/src/axolotl/integrations/aux_free_router/adapters.py
+++ b/src/axolotl/integrations/aux_free_router/adapters.py
@@ -50,12 +50,7 @@ class BaseMoEAdapter:
             except Exception:  # pragma: no cover - non-critical
                 pass
 
-    def prepare(self, moe_layer: nn.Module, handle: LayerHandle, shim: AuxFreeShim) -> None:
-        """Attach per-layer buffers and mark as aux-free enabled.
-
-        Note: Forward rebind happens in concrete adapters once we implement full routing.
-        For now, we only attach buffers as placeholders to minimize disturbance.
-        """
+    def _register_aux_buffers(self, moe_layer: nn.Module, handle: LayerHandle) -> None:
         device = next(moe_layer.parameters(), torch.tensor(0)).device
         if not hasattr(moe_layer, "_afb_bias"):
             moe_layer.register_buffer("_afb_bias", torch.zeros(handle.num_experts, device=device))
@@ -65,6 +60,10 @@ class BaseMoEAdapter:
             moe_layer.register_buffer("_afb_ema", torch.zeros(handle.num_experts, device=device))
         moe_layer._afb_layer_idx = handle.layer_idx  # type: ignore[attr-defined]
         moe_layer._afb_top_k = handle.top_k  # type: ignore[attr-defined]
+
+    def prepare(self, moe_layer: nn.Module, handle: LayerHandle, shim: AuxFreeShim) -> None:
+        """Attach per-layer buffers and mark as aux-free enabled."""
+        self._register_aux_buffers(moe_layer, handle)
         self._patch_forward_with_aux_free(moe_layer)
 
     def _patch_forward_with_aux_free(self, moe_layer: nn.Module) -> None:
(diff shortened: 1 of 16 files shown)


src/axolotl/integrations/aux_free_router/adapters.py
```diff
@@ -0,0 +1,323 @@
...
+            )
+
+            expert_mask = F.one_hot(selected_experts, num_classes=self.num_experts).permute(2, 1, 0)
+            expert_hit = torch.greater(expert_mask.sum(dim=(-1, -2)), 0).nonzero()
+            for expert_idx in expert_hit:
+                expert_layer = self.experts[expert_idx]
+                idx, top_x = torch.where(expert_mask[expert_idx].squeeze(0))
+                current_state = flat_states[None, top_x].reshape(-1, hidden_dim)
+                current_hidden_states = expert_layer(current_state) * routing_weights[top_x, idx, None]
+                final_hidden_states.index_add_(0, top_x, current_hidden_states.to(flat_states.dtype))
```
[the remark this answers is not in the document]
lhl: mixtral fix updated

src/axolotl/utils/schemas/config.py
```diff
@@ -664,6 +664,44 @@ class AxolotlInputConfig(
 
     llama4_linearized_experts: bool | None = None
 
+    # MoE aux-loss-free (AFB) toggles
+    moe_balance_type: Literal["gshard", "noaux_tc"] | None = Field(
```
NanoCode012: Since this is integration, we can add these as integration config instead of setting in the global config.

Here is an example https://github.com/axolotl-ai-cloud/axolotl/blob/main/src/axolotl/integrations/kd/args.py

Then, these args will be loaded if a user specifies the plugin.

src/axolotl/utils/config/__init__.py
```diff
@@ -265,6 +265,7 @@ def validate_config(
     AxolotlInputConfig = AxolotlInputConfigBase
 
     if cfg.plugins:
+        prepare_plugins(cfg)
```
winglian, line 268: I think we want to remove this as `load_cfg` usually calls `prepare_plugins` before this function is called. I think in the tests below, we want to call `prepare_plugins()` before the `validate_config()` call. See `test_llm_compressor.py` as an example since that one is also it's own plugin. I think we have some mixed patterns in our tests on the ordering, but most of them don't really use plugins most of the time.

lhl: OK, so I double check on this. I've updated my tests so they mirror llm_compressor so they should follow the same pattern.

This guard was something that Codex added in a pass but I did a deep dive review and I think it's correct - the guard is correctly added because there are some paths (outside my plugin) that require it:

> Right now that extra prepare_plugins(cfg) inside validate_config is acting as a safety net for the parts of the tree that still call validate_config before preparing plugins (cut-cross-entropy, hooks, liger, diffusion, etc.). If we rip it out before those callers are fixed, their plugin args won’t register and they’ll start failing. So: yes, we added that guard in our PR, but no, we shouldn’t remove it yet—first we need those other tests updated (or their owners to confirm they’re okay with losing the fallback). Once the rest of the suite uses the prepare_plugins → validate_config order consistently, we can safely drop the guard without breaking anyone.
> 
> CLI paths (axolotl.cli.config.load_cfg) still call prepare_plugins(cfg) before validate_config(cfg), so end‑to‑end training from YAML is safe. The only reason that fallback inside src/axolotl/utils/config/__init__.py:267-269 still exists is to keep a few test modules—each of which builds configs by hand—functioning until they’re reordered.
> 
> Until those suites are updated to follow the prepare_plugins → validate_config order (like we just did for all aux-loss-free tests), removing the guard would drop their plugin registrations, causing schema validation to reject plugin-specific keys and their plugins never to run.

Here's the list I have:

  - Cut Cross Entropy plugin (axolotl.integrations.cut_cross_entropy.CutCrossEntropyPlugin): tests/e2e/integrations/
    test_cut_cross_entropy.py:55, :101, :129 call validate_config before prepare_plugins.
  - Liger plugin (axolotl.integrations.liger.LigerPlugin): tests/e2e/integrations/test_liger.py:56 and :102.
  - Log Hooks test plugin (tests.e2e.integrations.test_hooks.LogHooksPlugin): tests/e2e/integrations/test_hooks.py:155.
  - Diffusion plugin (axolotl.integrations.diffusion.DiffusionPlugin): both test cases in tests/e2e/test_diffusion.py:66 and :124 never call prepare_plugins at all.

NanoCode012: @lhl , I think we should probably fix those related cases instead by moving the order of calls? If you would like to keep this, I would say at `src/axolotl/integrations/base.py:register` to have a `if "plugin_name" in self.plugins: return`, so we don't waste trying to reload plugins.


A bit unrelated but seems like we got two separate definitions of `prepare_plugins`
- src/axolotl/cli/config.py
- src/axolotl/utils/config/__init__.py
and a possible redundant `plugin_set_cfg`

winglian:
@lhl I'm going to rebase this and add a fix to use the latest transformers as well as compose with our latest MoE kernels. Just for posterity before the rebase, the last original commit on here is https://github.com/shisa-ai/axolotl/commit/4dd305210c3f5758d1a4c5a239c4e9223781a1e5

Issue #3270, opened 2025-11-20
Bug Report: KD Trainer Evaluation Completely Broken


roycho96:
### Please check that this issue hasn't been reported before.

- [x] I searched previous [Bug Reports](https://github.com/axolotl-ai-cloud/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

- Evaluation should run without crashing
- Evaluation loss should be on the same scale as training loss (per-token normalized)

### Current behaviour

**Immediate crash**
```
TypeError: iteration over a 0-d tensor
```

### Steps to reproduce

1. Set up KD training with validation:
```yaml
datasets:
  - path: /path/to/kd_data.json
    type: axolotl.integrations.kd.chat_template
    field_messages: messages
    logprobs_field: target_logprobs

kd_trainer: true
val_set_size: 0.2  # Enable validation
eval_strategy: epoch
```

2. Run training:
```bash
accelerate launch -m axolotl.cli.train config.yml
```

3. **Crash during first evaluation step:**
```python
TypeError: iteration over a 0-d tensor
  File "/opt/venv/lib/python3.11/site-packages/transformers/trainer.py", line 4584, in prediction_step
    loss, outputs = self.compute_loss(model, inputs, return_outputs=True)
    ^^^^^^^^^^^^^
```

### Config yaml

```yaml
hf_use_auth_token: true
trust_remote_code: true
overrides_of_model_kwargs:
  cache_dir: /home/user/.cache/huggingface
dataset_prepared_path: /home/user/.cache/axolotl/prepared
wandb_mode: disabled
logging_steps: 1
save_safetensors: true
save_only_model: false
save_total_limit: 1
bf16: true
fp16: false
tf32: true
load_in_8bit: false
load_in_4bit: false
strict: false
flash_attention: true
gradient_checkpointing: true
gradient_checkpointing_kwargs:
  use_reentrant: false
lora_mlp_kernel: false
lora_qkv_kernel: false
lora_o_kernel: false
peft_use_dora: false
peft_use_rslora: false
val_set_size: 0.2
eval_strategy: epoch
save_strategy: epoch
saves_per_epoch: 1
base_model: /path/to/model
output_dir: ../../temp/rslt
model_type: llama
learning_rate: 5e-05
num_epochs: 2
micro_batch_size: 1
gradient_accumulation_steps: 1
special_tokens:
  pad_token: <|reserved_special_token_0|>
sequence_len: 4096
sample_packing: false
[...]

PR #3271, opened 2025-11-20
fix: Fix evaluation loss in KD trainer

roycho96:
# Description

Fixes issue https://github.com/axolotl-ai-cloud/axolotl/issues/3270
Fixes two critical bugs in the Knowledge Distillation (KD) trainer that completely broke evaluation:

1. **Evaluation crash**: `compute_loss` ignored `return_outputs` parameter, causing `TypeError: iteration over a 0-d tensor` when `prediction_step` tried to unpack the return value
2. **Loss normalization bug**: Even after fixing the crash, evaluation loss wasn't normalized by `num_items_in_batch`, making it ~1000x larger than training loss (total batch loss vs per-token loss)

**Changes:**
- Modified `compute_loss` to properly return `(loss, outputs)` tuple when `return_outputs=True`
- Added `prediction_step` override to calculate and pass `num_items_in_batch` during evaluation

## Motivation and Context

The KD trainer's evaluation was completely non-functional:
- Setting `val_set_size > 0` immediately crashed during the first evaluation step
- This made it impossible to monitor model performance during KD training
- Users had to disable validation entirely (`val_set_size: 0`) to run KD training

**Root cause:** The model's forward pass normalizes loss by `num_items_in_batch` to compute per-token loss:
```python
if num_items_in_batch is not None and num_items_in_batch > 0:
    loss = loss / num_items_in_batch
```

During training, this is automatically passed, but `prediction_step` didn't pass it, and `compute_loss` couldn't return outputs anyway.

### Test Environment
- **Hardware**: H100 GPU (80GB)
- **Framework**: DeepSpeed ZeRO Stage 2 with CPU offload
- **Models**: 
  - Teacher: Llama-3.2-3B-Instruct
  - Student: Llama-3.2-1B-Instruct
- **Precision**: BF16
- **Batch size**: micro_batch_size=1, gradient_accumulation_steps=1

### DeepSpeed Configuration
```json
{
    "zero_optimization": {
        "stage": 2,
        "offload_optimizer": {
            "device": "cpu",
            "pin_memory": true
        }
    },
    "bf16": {"enabled": true}
}
```

### Axolotl Configuration
- KD parameters: `kd_ce_alpha=0.7`, `kd_alpha=0.3`, `kd_temperature=0.6`
- Plugins: Liger kernels enabled
- Evaluation: `val_set_size=0.2`, `eval_strategy=epoch`
- Sequence length: 4096

### Test Data Preparation
Generated KD dataset with teacher model logits:
1. Loaded chat-formatted dataset with standard `messages` field
2. Used teacher model (Llama-3.2-3B) to generate top-k logprobs and token IDs for each position
3. Added `target_logprobs` and `target_token_ids` fields (v2 format: `List[List[float]]` and `List[List[int]]`)
4. No backward pass through teacher model (inference only)

### Test Results

**Before fix:**
```bash
# Evaluation immediately crashes
TypeError: iteration over a 0-d tensor
  File "trainer.py", line 4584, in prediction_step
    loss, outputs = self.compute_loss(model, inputs, return_outputs=True)
```

**After Bug 1 fix (compute_loss only):**
```
# Epoch 0 - Evaluation
{'eval_loss': 1326.0, 'eval_runtime': 14.15, 'epoch': 0}  # ~330x larger!

# Training
{'loss': 1.875, 'grad_norm': 248.98, 'learning_rate': 4.91e-05, 'epoch': 0.14}
{'loss': 2.0781, 'grad_norm': 104.96, 'learning_rate': 4.65e-05, 'epoch': 0.29}
{'loss': 5.9688, 'grad_norm': 82.22, 'learning_rate': 4.23e-05, 'epoch': 0.43}
{'loss': 5.7188, 'grad_norm': 68.43, 'learning_rate': 3.68e-05, 'epoch': 0.57}
{'loss': 4.3125, 'grad_norm': 63.78, 'learning_rate': 3.06e-05, 'epoch': 0.71}
{'loss': 5.25, 'grad_norm': 37.48, 'learning_rate': 2.39e-05, 'epoch': 0.86}
{'loss': 5.5625, 'grad_norm': 34.40, 'learning_rate': 1.73e-05, 'epoch': 1.0}
{'loss': 5.6562, 'grad_norm': 34.51, 'learning_rate': 1.73e-05, 'epoch': 1.0}

# Epoch 1 - Evaluation  
{'eval_loss': 2400.0, 'eval_runtime': 0.36, 'epoch': 1.0}  # ~430x larger!
```

**After complete fix (compute_loss + prediction_step):**
```
# Epoch 0 - Evaluation
{'eval_loss': 2.390625, 'eval_runtime': 11.85, 'epoch': 0}  # Comparable!

# Training
{'loss': 1.875, 'grad_norm': 249.09, 'learning_rate': 4.91e-05, 'epoch': 0.14}
{'loss': 2.0781, 'grad_norm': 104.96, 'learning_rate': 4.65e-05, 'epoch': 0.29}
{'loss': 5.9688, 'grad_norm': 82.69, 'learning_rate': 4.23e-05, 'epoch': 0.43}
{'loss': 5.7188, 'grad_norm': 69.87, 'learning_rate': 3.68e-05, 'epoch': 0.57}
{'loss': 4.375, 'grad_norm': 60.71, 'learning_rate': 3.06e-05, 'epoch': 0.71}
{'loss': 5.4062, 'grad_norm': 45.77, 'learning_rate': 2.39e-05, 'epoch': 0.86}
{'loss': 5.6562, 'grad_norm': 34.51, 'learning_rate': 1.73e-05, 'epoch': 1.0}

# Epoch 1 - Evaluation
{'eval_loss': 4.578125, 'eval_runtime': 0.51, 'epoch': 1.0}  # Comparable!
```

N/A

## Types of changes

- [x] Bug fix

## Social Handles (Optional)

GitHub: @roycho96 (feel free to tag me for follow-ups)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed loss normalization during evaluation to properly account for non-padding tokens, improving overall evaluation accuracy.
  * Improved robustness of loss extraction from model outputs to handle multiple output format types.

* **Documentation**
  * Updated documentation to reflect evaluation normalization improvements.


NanoCode012:
Thanks for the PR. I've let the tests run first while we take some time to confirm the original issue. Good catch on the num_items_in_batch, it's still haunting us till now :/

roycho96:
@NanoCode012 Thanks for the quick review! 

Also just pushed a third fix: `ChatTemplateStrategyWithKDv2` had a `super()` call order bug causing `KeyError: 'target_token_ids'` when using v2 format datasets. The parent's `transform_logprobs` was being called before `target_token_ids` was added to the sample. Fixed by calling grandparent directly and adding fields in the correct order.

roycho96:
I just pushed an additional fix.
The compute_loss method properly checks `hasattr(self.state, "total_tokens")` before accessing it, but the log method directly accesses it without checking, causing `AttributeError` when `include_tkps=True`.
This commit adds the same safety check to log() for consistency.

src/axolotl/integrations/kd/chat_template.py
```diff
@@ -284,12 +284,17 @@ def transform_logprobs(self, sample):
...
+        logprobs = prompt.pop(self.logprobs_field)
+        target_token_ids = prompt.pop("target_token_ids")
+        
+        tokenized_prompt = super(ChatTemplateStrategyWithKD, self)._tokenize_single_prompt(prompt)
+        
+        # Add KD fields before transform
+        tokenized_prompt[self.logprobs_field] = logprobs
+        tokenized_prompt["target_token_ids"] = target_token_ids
+        
+        tokenized_prompt = self.transform_logprobs(tokenized_prompt)
```
NanoCode012: Could you elaborate why you swapped to this change? The reason I updated to the current method was because `super()` would call `ChatTemplateStrategyWithKD` that handles logprob pop already.

NanoCode012 (commented):
Hey, sorry for getting back late. I updated the changes to be less breaking but have one question below.

src/axolotl/integrations/kd/chat_template.py
roycho96: The issue is that v2's `transform_logprobs()` requires `target_token_ids`:

```python
# v2's transfrom_logprobs
for token_pos_logprobs, pos_target_token_ids in zip(
    logprobs, sample["target_token_ids"], strict=False  # Needs this!
)
```
But when we call super()._tokenize_single_prompt():
Parent pops logprobs.
Parent calls self.transform_logprobs() and then, resolves to v2's version due to polymorphism.
Transform tries to access sample["target_token_ids"] which occurs KeyError!
Current code adds target_token_ids after super() returns which is too late.
The fix: Pop both fields upfront, skip parent's transform call by calling grandparent directly, then call v2's transform with both fields present.

NanoCode012: Is there some other way to simplify this? I'm not sure how clean it is to call the parent's parent. Would you be able to add a simple test to check for this behavior?

roycho96: @NanoCode012 I've refactored to use the Template Method pattern instead of the grandparent call.

I added a `_prepare_kd_fields()` hook in the parent class that gets called before transform. The base implementation does nothing, but v2 overrides it to add `target_token_ids` before the transform runs. This is much cleaner and follows the template method design pattern, making it easier to extend for future data formats.

I also found and fixed a bug in the refactored `compute_loss`. It was missing the model forward pass line, which caused a NameError during evaluation. Both training and evaluation work correctly now.

Added tests in `tests/integrations/test_kd_chat_template.py` to verify the hook adds fields correctly, handles missing fields, and that transform fails as expected without `target_token_ids`. All tests passing!

```
user:/project/work/llm-ops/SRC0001/axolotl$ python3 -m pytest tests/integrations/test_kd_chat_template.py -v
====================================================================================================== test session starts ======================================================================================================
platform linux -- Python 3.11.14, pytest-9.0.1, pluggy-1.6.0 -- /usr/bin/python3
cachedir: .pytest_cache
rootdir: /project/work/llm-ops/SRC0001/axolotl
configfile: pyproject.toml
plugins: anyio-4.11.0, langsmith-0.4.46
collected 3 items                                                                                                                                                                                                               

tests/integrations/test_kd_chat_template.py::TestChatTemplateStrategyWithKDv2::test_v2_prepare_kd_fields_adds_target_token_ids PASSED                                                                                     [ 33%]
tests/integrations/test_kd_chat_template.py::TestChatTemplateStrategyWithKDv2::test_v2_prepare_kd_fields_handles_missing_field PASSED                                                                                     [ 66%]
tests/integrations/test_kd_chat_template.py::TestChatTemplateStrategyWithKDv2::test_v2_transform_requires_target_token_ids PASSED                                                                                         [100%]

======================================================================================================= 3 passed in 1.20s =======================================================================================================
```

Does this approach look better?

src/axolotl/integrations/kd/chat_template.py
```diff
@@ -292,6 +301,14 @@ def _tokenize_single_prompt(self, prompt):
...
 
         return tokenized_prompt
+    
+    def _prepare_kd_fields(self, tokenized_prompt, original_prompt):
+        """
+        Add pre-tokenized target_token_ids for v2 format
+        """
+        if "target_token_ids" in original_prompt:
+            tokenized_prompt["target_token_ids"] = original_prompt.pop("target_token_ids")
+        return tokenized_prompt
```
NanoCode012, line 302:
```python
def _prepare_kd_fields(self, tokenized_prompt, original_prompt):
    target_token_ids = original_prompt.pop("target_token_ids", None)
    if target_token_ids is not None:
        tokenized_prompt["target_token_ids"] = target_token_ids
    return tokenized_prompt
```

Let's use pop None to be a bit cleaner.

src/axolotl/integrations/kd/chat_template.py
```diff
@@ -292,6 +301,14 @@ def _tokenize_single_prompt(self, prompt):
             tokenized_prompt["target_token_ids"] = target_token_ids
 
         return tokenized_prompt
```
NanoCode012: I think following the code flow now. We don't need this override in the V2 anymore. The hook handles this well!

tests/integrations/test_kd_chat_template.py
```diff
@@ -0,0 +1,80 @@
...
+        to be added by the hook.
+        """
+        sample = {
+            "input_ids": [1, 10, 20, 30, 2],
+            "labels": [1, 10, 20, 30, 2],
+            "logprobs": [[-0.1, -0.2], [-0.3, -0.4]],
+        }
+
+        with pytest.raises(KeyError, match="target_token_ids"):
+            v2_strategy.transform_logprobs(sample)
```
NanoCode012: Please check lint for this file

NanoCode012 (commented):
That indeed looks better. I just have some minor points left

roycho96:
@NanoCode012 
All fixed! Used pop(None), removed the v2 override, and fixed the lint issue.

NanoCode012:
> @NanoCode012 All fixed! Used pop(None), removed the v2 override, and fixed the lint issue.

Thanks, letting the test re-run

winglian (approved):
thanks! A couple of linting issues which I pushed up a commit for.

PR #3288, opened 2025-12-01
fix: perplexity metric

xzuyn:
# Description

`eval_causal_lm_metrics: ["perplexity"]` appears to be broken. I tried to fix it.

- The current code generates predictions, passes `references` to `Perplexity`, tokenizes them, then calculates the perplexity using a sliding window across them. This results in a *very* high value.
- This PR changes `Perplexity` to take in `eval_dataloader` instead, and calculate the perplexity across the batches without needing to predict or tokenize anything. If other options are used with `eval_causal_lm_metrics` it will predict new tokens as expected. That change brings the perplexity value to a much lower and reasonable one.

This version doesn't have the sliding window. It simply calculates the perplexity across the full batch, ignoring any `-100` labels (including throwing away any samples that are entirely masked).

image

image

## Summary by CodeRabbit

  * Optimized evaluation pipeline to skip unnecessary prediction generation when perplexity is the only requested metric.

* **Improvements**
  * Enhanced perplexity metric calculation with improved error handling and robustness for evaluation workflows.

### DIFF
--- a/src/axolotl/utils/callbacks/perplexity.py
+++ b/src/axolotl/utils/callbacks/perplexity.py
@@ -25,14 +25,12 @@ class Perplexity:
         return ["references"]
 
     def compute(
-        self,
-        model: PreTrainedModel,
-        eval_dataloader=None,
-        **kwargs,
+        self, model: PreTrainedModel, eval_dataloader=None, **_kwargs
     ) -> Dict[str, float]:
         model.eval()
 
-        scores = []
+        total_loss_sum = 0.0
+        total_token_count = 0
         for batch in tqdm(eval_dataloader, disable=not is_main_process()):
             if "position_ids" not in batch:
                 input_ids = batch["input_ids"].to(model.device)
@@ -52,16 +50,20 @@ class Perplexity:
                         labels=labels,
                     ).loss
 
-                scores.append(torch.exp(model_loss).item())
+                batch_token_count = (labels != -100).sum().item()
+                total_loss_sum += model_loss.item() * batch_token_count
+                total_token_count += batch_token_count
             else:
-                # do_causal_lm_eval + sample_packing already gives ValidationError. Extra protection.
+                # do_causal_lm_eval + sample_packing_packing already gives ValidationError. Extra protection.
                 LOG.debug(
                     "Packed evaluation samples not supported with perplexity metric"
                 )
                 return {"score": float("nan")}
 
-        if len(scores) == 0:
+        if total_token_count == 0:
             LOG.debug("No valid tokens found for perplexity metric")
             return {"score": float("nan")}
 
-        return {"score": sum(scores) / len(scores)}
+        return {
+            "score": float(torch.exp(torch.tensor(total_loss_sum / total_token_count)))
+        }
--- a/src/axolotl/utils/callbacks/perplexity.py
+++ b/src/axolotl/utils/callbacks/perplexity.py
@@ -34,41 +34,41 @@ class Perplexity:
         total_loss_sum = 0.0
         total_token_count = 0
         for batch in tqdm(eval_dataloader, disable=not is_main_process()):
-            if "position_ids" not in batch:
-                input_ids = batch["input_ids"].to(model.device)
-                attention_mask = (
-                    batch["attention_mask"].to(model.device)
-                    if "attention_mask" in batch
-                    else None
-                )
-                labels = batch["labels"].to(model.device)
+            input_ids = batch["input_ids"].to(model.device)
+            labels = batch["labels"].to(model.device)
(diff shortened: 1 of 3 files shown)


xzuyn:
Did not realize there were tests for it. PR is basically completely apart from the now failing `test_perplexity.py`

src/axolotl/utils/callbacks/perplexity.py
```diff
@@ -18,64 +18,52 @@ class Perplexity:
...
-                break
-
-        perplexity = torch.exp(torch.stack(losses).mean()).item()
+        if total_token_count == 0:
+            LOG.debug("No valid tokens found for perplexity metric")
+            return {"score": float("nan")}
 
         return {
-            "score": perplexity,
+            "score": float(torch.exp(torch.tensor(total_loss_sum / total_token_count)))
```
NanoCode012:
```suggestion
            "score": torch.exp(torch.tensor(total_loss_sum / total_token_count)).item()
```

We can probably simplify to use `math.exp` without tensor overhead if you want to optim more

src/axolotl/utils/callbacks/perplexity.py
```diff
@@ -18,64 +18,52 @@ class Perplexity:
...
-            with torch.no_grad():
-                outputs: CausalLMOutput = model(
-                    input_ids=input_ids_slice, labels=labels_slice
+        total_loss_sum = 0.0
+        total_token_count = 0
+        for batch in tqdm(eval_dataloader, disable=not is_main_process()):
+            if "position_ids" not in batch:
+                input_ids = batch["input_ids"].to(model.device)
+                labels = batch["labels"].to(model.device)
+
```
NanoCode012: I think you should also check if attention mask exists like

```
attention_mask = batch["attention_mask"].to(model.device) if "attention_mask" in batch else None
```

and pass to model forward below.

src/axolotl/utils/callbacks/perplexity.py
```diff
@@ -18,64 +18,52 @@ class Perplexity:
...
-        stride: int = 512,
-    ) -> None:
-        self.max_seq_len = max_seq_len
-        self.stride = stride
-        self.tokenizer = tokenizer
+    def __init__(self) -> None:
         self.name = "perplexity"
 
     def _feature_names(self) -> List[str]:
         return ["references"]
```
NanoCode012:
```suggestion
def _feature_names(self) -> List[str]:
return []
```

Unused (ignore my lack of spacing )

src/axolotl/utils/callbacks/perplexity.py
```diff
@@ -18,64 +18,52 @@ class Perplexity:
...
+
+                with torch.no_grad():
+                    model_loss = model(
+                        input_ids=input_ids,
+                        labels=labels,
+                    ).loss
+
+                batch_token_count = (labels != -100).sum().item()
+                total_loss_sum += model_loss.item() * batch_token_count
+                total_token_count += batch_token_count
```
NanoCode012: We should probably do an all-reduce if on multi-gpu, but it's not the scope of this PR. Did you run your tests on single or multi?

xzuyn: Only single gpu

xzuyn:
Would it be possible to instead just move this to where `eval/loss` is being calculated? It would make evaluation ~2x faster when only `eval/loss` and `eval/perplexity` is being computed, since loss only needs to be computed once instead of twice. Would also remove some randomness between the two metrics since they use the same loss.

NanoCode012:
> Would it be possible to instead just move this to where `eval/loss` is being calculated? It would make evaluation ~2x faster when only `eval/loss` and `eval/perplexity` is being computed, since loss only needs to be computed once instead of twice. Would also remove some randomness between the two metrics since they use the same loss.

I just asked gpt about this. It may be on to something 

```python
class PerplexityCallback(TrainerCallback):
    """
    A callback that calculates perplexity from eval_loss immediately
    after evaluation, without re-running model inference.
    """
    
    def on_log(self, args, state, control, logs=None, **kwargs):
        # We only care when 'eval_loss' is present in the logs
        if logs is not None and "eval_loss" in logs:
            try:
                # Perplexity = exp(CrossEntropyLoss)
                ppl = torch.exp(torch.tensor(logs["eval_loss"])).item()
                
                # Add it to the logs so it shows up in WandB/Tensorboard/Print
                logs["perplexity"] = ppl
                
                # Optional: print it nicely to console
                print(f"Computed Perplexity: {ppl:.4f}")
            except Exception:
                pass
```

We can possibly add it by default as it's just a torch.exp function that happens once every eval

PR #3293, opened 2025-12-02
compute loss only if training and update token metric naming

### LINKED ISSUE #3291: `train/total_tokens` increases from eval and resets when resuming from checkpoint

### Expected Behavior

- `train/total_tokens` should only increase from training steps, not evaluation steps.
- `train/total_tokens` should be resumed when using `resume_from_checkpoint`, not reset to 0.

### Current behaviour

- `train/total_tokens` increases during evaluation. You can see these as a jump on the step after eval. Meaning if you do evaluation every 10 steps, then step 11,21,31,41,51,etc will all show a jump. If you do evaluation every 50 steps, then step 51,101,151,201,etc will all show a jump. If you are evaluating often the final value will be very different from the true value.

These two runs use the same exact config and pre-processed dataset, but by step 50 the value is basically double the true value.
Image

- `train/total_tokens` resets to 0 when resuming.

### Steps to reproduce

Should happen with any config where evaluation is done, or the checkpoint is being resumed.

### Config yaml

```yaml

```

### Possible solution

_No response_

- [x] Linux
- [ ] macOS
- [ ] Windows

### Python Version

3.12.3

### axolotl branch-commit

main/b234532

### Acknowledgements

ved1beta:
fixes #3291 
image

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed token-per-second tracking to only occur when the model is actively training. Previously, tracking could run regardless of training state.

### DIFF
--- a/src/axolotl/core/trainers/base.py
+++ b/src/axolotl/core/trainers/base.py
@@ -648,9 +648,6 @@ class AxolotlTrainer(
             logs["tokens/train_per_sec_per_gpu"] = round(
                 self.state.last_tokens_per_second.item() / self.args.logging_steps, 2
             )
-            logs["tokens/total"] = int(self.state.tokens["total"].item())
-            logs["tokens/trainable"] = int(self.state.tokens["trainable"].item())
-
         del self._stored_metrics[train_eval]
 
         return super().log(logs, start_time)
--- a/src/axolotl/core/trainers/base.py
+++ b/src/axolotl/core/trainers/base.py
@@ -380,7 +380,6 @@ class AxolotlTrainer(
             # Store per-step trainable tokens for throughput calculation
             self.state.tokens["trainable_tokens"] = trainable_tokens.detach().cpu()
 
-
         if self.args.orpo_alpha:
             return self.orpo_compute_loss(
                 model,
@@ -650,6 +649,7 @@ class AxolotlTrainer(
                 self.state.last_tokens_per_second.item() / self.args.logging_steps, 2
             )
             logs["tokens/total"] = int(self.state.tokens["total"].item())
+            logs["tokens/trainable"] = int(self.state.tokens["trainable"].item())
 
         del self._stored_metrics[train_eval]
 
--- a/src/axolotl/core/trainers/base.py
+++ b/src/axolotl/core/trainers/base.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import json
 import os
 from collections import defaultdict
 from functools import partial, wraps
@@ -48,6 +49,8 @@ from axolotl.utils.logging import get_logger
 from axolotl.utils.samplers import MultipackBatchSampler, get_dataset_lengths
 
 LOG = get_logger(__name__)
+
+TOKENS_STATE_FILE = "tokens_state.json"
 
 REDUCTION_FNS = {
     "mean": torch.mean,
@@ -363,9 +366,9 @@ class AxolotlTrainer(
                 self.state.num_tokens = (inputs[inputs_key] != -100).sum().cpu()
 
             if hasattr(self.state, "total_tokens"):
-                self.state.total_tokens += num_tokens
+                self.state.total_tokens += num_tokens.cpu()
             else:
-                self.state.total_tokens = num_tokens
+                self.state.total_tokens = num_tokens.cpu()
 
         if self.args.orpo_alpha:
             return self.orpo_compute_loss(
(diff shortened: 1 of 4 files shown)


xzuyn:
Solves the eval increase part, but the reset on resume still happens.

Eval every 10 steps.
image

ved1beta:
ahh i thought it was just the recomputing here you goo , 
image

src/axolotl/utils/callbacks/tokens_per_second.py
```diff
@@ -10,22 +12,50 @@
...
+        """Restore total_tokens state when resuming from checkpoint."""
+        if not isinstance(self.resume_from_checkpoint, str):
+            return
+        tokens_state_path = os.path.join(self.resume_from_checkpoint, TOKENS_STATE_FILE)
+        if os.path.isfile(tokens_state_path):
+            with open(tokens_state_path, "r", encoding="utf-8") as f:
+                tokens_state = json.load(f)
+            state.total_tokens = torch.tensor(tokens_state.get("total_tokens", 0))
+            state.num_tokens = torch.tensor(tokens_state.get("num_tokens", 0))
+            LOG.info(f"Restored total_tokens: {state.total_tokens}")
```
NanoCode012: Is this how we should store total_tokens? Are we able to inject it into the trainer state file that gets created in checkpoints?

ved1beta: ummm TrainerState is a dataclass, it only serializes defined fields

xzuyn:
Resuming works too. `[2025-12-06 13:23:07,478] [INFO] [axolotl.utils.callbacks.tokens_per_second] Restored total_tokens: 544092`
image

NanoCode012 (changes requested):
As mentioned in chat, let's refactor to use `tokens/total` and `tokens/trainable`. I believe `self.num_tokens` is redundant?

ved1beta:
train eval working fine ,also checkpoints 
image

src/axolotl/core/trainers/base.py
```diff
@@ -348,24 +351,34 @@ def compute_loss(
...
+
+            # trainable tokens for throughput and total token slots for summaries
+            self.state.tokens["trainable"] = (
+                self.state.tokens["trainable"] + trainable_tokens.detach().cpu()
+            )
+            self.state.tokens["total"] = (
+                self.state.tokens["total"] + torch.as_tensor(total_tokens).cpu()
+            )
+            # Store per-step trainable tokens for throughput calculation
+            self.state.tokens["trainable_step"] = trainable_tokens.detach().cpu()
```
NanoCode012: Can be removed as is unused (not logged and too similar to others)

src/axolotl/core/trainers/base.py
```diff
@@ -625,13 +638,17 @@ def log(self, logs: dict[str, float], start_time: float | None = None) -> None:
...
+            and hasattr(self.state, "tokens")
+        ):
             # each rank will log its own tokens per second
             # for logging_steps > 1 we obtain a moving average of this metric
-            logs["tokens_per_second_per_gpu"] = round(
+            logs["tokens/trainable_per_second_per_gpu"] = round(
                 self.state.last_tokens_per_second.item() / self.args.logging_steps, 2
             )
-            logs["total_tokens"] = int(self.state.total_tokens.item())
+            logs["tokens/total"] = int(self.state.tokens["total"].item())
```
NanoCode012: I think we missed log `tokens/trainable`

src/axolotl/utils/callbacks/tokens_per_second.py
```diff
@@ -60,5 +95,15 @@ def on_log(
...
+        tokens = getattr(state, "tokens", None)
+        # Clear per-step tokens after logging
+        if tokens and "trainable_tokens" in tokens:
+            tokens["trainable_tokens"] = torch.zeros_like(tokens["trainable_tokens"])
+
+        if tokens and "total" in tokens:
+            logs["tokens/total"] = tokens["total"].item()
+
+        if tokens and "trainable" in tokens:
+            logs["tokens/trainable"] = tokens["trainable"].item()
```
NanoCode012, line 109: Is this duplicate log of base.py L651-652?

ved1beta: yess. redundant ?

NanoCode012 (approved):

winglian:
can we add a CI similar to TestResumeLlama that on resume the total tokens is correct?

ved1beta:
Okay!

xzuyn:
`tokens/total` and `tokens/trainable` aren't being sent to wandb. It is also being shown as a float instead of an int.

Issue #3301, opened 2025-12-04
Difference in starting Loss with and without Packing for Olmo3


ved1beta:
### Please check that this issue hasn't been reported before.

- [x] I searched previous [Bug Reports](https://github.com/axolotl-ai-cloud/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

difference should be less or same

### Current behaviour

<img width="1107" height="523" alt="Image" src="https://github.com/user-attachments/assets/64dd0152-6a2c-47da-bc22-e16f4105a0f2" />

red is sample packing false

### Steps to reproduce

run config https://discord.com/channels/1104757954588196865/1111279858136383509/1444067990080589995

### Config yaml

```yaml
# Weights and Biases logging config
wandb_project: omlo-3-issue
wandb_name: vedaya-self
wandb_entity:
wandb_watch:
wandb_log_model:

# Model checkpointing config
output_dir: ./Outputs/BS1GA8-4K-Packing-On
save_steps: 100
save_safetensors: true
save_total_limit: 10
save_only_model: false

# Model architecture config
base_model: allenai/Olmo-3-7B-Instruct
revision_of_model:
model_type: AutoModelForCausalLM
tokenizer_type: AutoTokenizer
chat_template_jinja:

# Mixed precision training config
bf16: true
fp16: false
tf32: false

# Model loading config
lora_on_cpu: false
low_cpu_mem_usage: true
strict: false
load_in_8bit: false
load_in_4bit: true
bnb_config_kwargs:
  bnb_4bit_quant_storage: uint8

# Sequence config
sequence_len: 4096
min_sample_len: 128
sample_packing: false
eval_sample_packing: false
batch_flattening:
pad_to_sequence_len: true
train_on_inputs: false
group_by_length: false

# LoRA adapter config
peft_init_lora_weights:
adapter: qlora
lora_r: 32
lora_alpha: 32
peft_use_rslora:
lora_dropout: 0
lora_target_linear: true
lora_target_modules:
lora_target_parameters:
lora_modules_to_save:
lora_mlp_kernel:
lora_qkv_kernel:
lora_o_kernel:

# Dataset config
dataset_num_proc:
datasets:
  - path: PJMixers-Dev/allenai_tulu-3-sft-olmo-2-mixture-0225-filtered-ShareGPT
    split: train[:10000]
    type: chat_template
    field_messages: conversations
    message_property_mappings:
      role: from
      content: value
test_datasets:
[...]

PR #3304, opened 2025-12-05
FIX loss difference with and without sample packing : ) 

ved1beta:
# Description
add pos_ids to features

#3301 

## How has this been tested?
image

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced batch processing to automatically generate missing positional information when not explicitly provided, improving reliability for diverse input formats.


xzuyn:
I'm not seeing any change with this PR. I understand that `train/loss` can differ a bit since it would see more tokens per step, but that shouldn't be the case for `eval/loss` and `eval/perplexity` at step 0 since with or without packing it sees the entire eval split.

image

NanoCode012:
Issue linked closed

PR #3317, opened 2025-12-11
feat: cheap ppl metric

xzuyn:
# Description

better way of doing #3288

doesn't mess with `eval_causal_lm_metrics`, so it uses `ppl` and `eval_ppl` to differentiate itself.

`{'eval_loss': 3.87862491607666, 'eval_runtime': 173.5787, 'eval_samples_per_second': 2.765, 'eval_steps_per_second': 0.346, 'eval_ppl': 48.35767347232679, 'memory/max_active (GiB)': 7.38, 'memory/max_allocated (GiB)': 7.38, 'memory/device_reserved (GiB)': 8.71, 'epoch': 0}`

## Summary by CodeRabbit

* **New Features**
  * Automatic perplexity metric computation added to training logs, derived from both standard and evaluation loss values during model training.

### DIFF
--- a/src/axolotl/core/trainers/base.py
+++ b/src/axolotl/core/trainers/base.py
@@ -10,6 +10,7 @@ from typing import Any, Callable, Literal, Optional
 import datasets
 import safetensors
 import torch
+import math
 from accelerate.state import AcceleratorState
 from datasets import Dataset
 from peft import PeftModel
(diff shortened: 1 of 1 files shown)


NanoCode012:
Any thoughts on having this for just eval? Training ppl could be noisy (open to suggestion). Additionally, should we add a simple `round(, 4)` to show less digits?

NanoCode012 (approved):
Good to go. If people find the train ppl noisy, we can adjust later

xzuyn:
> Any thoughts on having this for just eval? Training ppl could be noisy (open to suggestion).

It would be noisy yea, but it's still nice to have, and `train/loss` is already noisy where you want to add some sort of smoothing in W&B. Also without it you wouldn't be able to plot eval and train together in the same graph the way you can with loss.

> Additionally, should we add a simple round(, 4) to show less digits?

Oh I didn't even notice that the loss was being rounded. Yea, rounding makes sense.

PR #3321, opened 2025-12-18
feature: raise on long sequence drop

kallewoof:
# Description

It is sometimes not desired that sequences are silently dropped from the dataset, especially when the dataset has been carefully crafted and pre-fitted for the training context. This would then suggest that an error occurred somewhere in the process. This feature adds a third value for excess_length_strategy called 'raise', which will raise a ValueError if a sequence is encountered that is too long and would have normally been dropped/truncated.

## Motivation and Context

If the user has already pre-fitted the training data for the target context size, they would want to know if there were dropped sequences. Truncating and dropping are both inadequate in this use case.

## How has this been tested?

I tested this by providing a config with a too-big sequence length and set the strategy to 'raise'.

I have also added a new unit test to `test_data.py` which tests the new behavior.

## Types of changes

Adds new option to existing configuration parameter.

## Summary by CodeRabbit

* **New Features**
  * Added a "raise" option for handling inputs that exceed the maximum sequence length so the system can surface an error instead of truncating or dropping.

* **Improvements**
  * Made decision logic for excess-length handling consistent during dataset filtering so the chosen strategy is applied reliably.

* **Tests**
  * Added tests to verify excess-length behavior and updated public exports to support dataset/config workflows.

### DIFF
--- a/src/axolotl/utils/data/utils.py
+++ b/src/axolotl/utils/data/utils.py
@@ -188,7 +188,10 @@ def handle_long_seq_in_dataset(
         cfg: Dictionary mapping `axolotl` config keys to values.
 
     Returns:
-        Filtered dataset with long sequences removed.
+        Filtered dataset with long sequences handled according to the excess_length_strategy value:
+            'drop' (default)    excludes any sequence longer than sequence_len
+            'truncate'          truncates them down to sequence_len
+            'raise'             raises a ValueError if any sequence was found that was longer than sequence_len
     """
     if (
         hasattr(dataset, "column_names")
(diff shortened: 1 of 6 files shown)


kallewoof:
@coderabbitai review

kallewoof:
The Preview / preview check failed with
```
ERROR: Could not install packages due to an OSError: [Errno 28] No space left on device
```
Not sure what to do about that one.

kallewoof:
@coderabbitai generate docstrings

kallewoof:
@coderabbitai review

kallewoof:
I have no idea why the two tests are suddenly being canceled. There's no information in the logs. My last commit addresses the out of space issue in the "preview" test. I can move this into a separate pull request if desired.

winglian (approved):

PR #3388, opened 2026-02-03
fix: pass revision parameter to tokenizer and processor loaders

madScientist10:
## Summary

Fixes the `revision_of_model` config parameter to be properly passed to tokenizer and processor loaders, allowing users to load models from specific revisions/branches.

## Problem

When using `revision_of_model` in the config, the revision was not being passed to:
- `load_tokenizer()` - tokenizer loading
- `load_processor()` - processor loading for multimodal models
- `modify_tokenizer_files()` - when using `added_tokens_overrides`

This caused the tokenizer/processor to always load from the default branch even when a specific revision was specified.

## Changes

- **`src/axolotl/loaders/tokenizer.py`**:
  - Pass `revision` to `AutoTokenizer.from_pretrained()` when `cfg.revision_of_model` is set
  - Pass `revision` to `HFMistralTokenizer.from_pretrained()` for Mistral models
  - Add `revision` parameter to `modify_tokenizer_files()` function

- **`src/axolotl/loaders/processor.py`**:
  - Pass `revision` to `processor_cls.from_pretrained()` when loading processors

## Summary by CodeRabbit

* **New Features**
  * Added configurable model revision support for processor and tokenizer components. Users can now specify which model revision is loaded during system initialization. Defaults to "main" for backward compatibility, enabling flexible version management and improved control over model dependencies across the loading pipeline.

### DIFF
--- a/src/axolotl/loaders/processor.py
+++ b/src/axolotl/loaders/processor.py
@@ -19,6 +19,11 @@ def load_processor(cfg: DictDefault, tokenizer: PreTrainedTokenizerBase):
     if cfg.processor_type:
         processor_cls = getattr(transformers, cfg.processor_type)
 
+    # Build common kwargs for processor loading
+    processor_kwargs = {}
+    if cfg.revision_of_model:
+        processor_kwargs["revision"] = cfg.revision_of_model
+
     if cfg.tokenizer_use_mistral_common:
 
         def _patch_mistralcommontokenizer():
@@ -40,6 +45,7 @@ def load_processor(cfg: DictDefault, tokenizer: PreTrainedTokenizerBase):
         if processor_cls == VoxtralProcessor:
             return VoxtralProcessor.from_pretrained(
                 cfg.processor_config,
+                **processor_kwargs,
             )
 
         from axolotl.utils.mistral import Mistral3Processor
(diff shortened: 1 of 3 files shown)


src/axolotl/loaders/processor.py
```diff
@@ -38,8 +38,12 @@ def _patch_mistralcommontokenizer():
         from transformers import VoxtralProcessor
 
         if processor_cls == VoxtralProcessor:
+            kwargs = {}
+            if cfg.revision_of_model:
+                kwargs["revision"] = cfg.revision_of_model
             return VoxtralProcessor.from_pretrained(
                 cfg.processor_config,
+                **kwargs,
             )
```
NanoCode012: We can probably init kwargs earlier to refactor the below as well

src/axolotl/loaders/tokenizer.py
```diff
@@ -134,7 +135,8 @@ def _load_mistral_common_tokenizer(cfg: DictDefault):
         from axolotl.utils.mistral import HFMistralTokenizer
 
         # Load the HF-compatible wrapper around MistralTokenizer
-        tokenizer = HFMistralTokenizer.from_pretrained(cfg.tokenizer_config)
+        revision = cfg.revision_of_model or "main"
+        tokenizer = HFMistralTokenizer.from_pretrained(cfg.tokenizer_config, revision=revision)
```
NanoCode012: It may be better to pass in if not None.

src/axolotl/loaders/processor.py
madScientist10: Thanks, I made both updates. Please let me know if there are more changes required.

NanoCode012:
Can you add the test I made here https://github.com/NanoCode012/axolotl/commit/220118e5352313b63e2a012136a32af71540e678 

I can't push to your branch from local / create a PR for some reason

madScientist10:
Thanks for the commit and the test. I've incorporated both your formatting fix (e68344cc) and added the test file from your commit. all 6 tests pass locally

NanoCode012 (approved):

NanoCode012 (approved):

PR #3389, opened 2026-02-03
feat: add auto-chunking support for streaming pretraining datasets

madScientist10:
## Summary

Adds auto-truncation support for streaming pretraining datasets, ensuring no data is lost from long sequences.

Based on PR #3081 (feature/pretraining-auto-truncate) but:
1. Updated for the refactored codebase (pretraining.py → streaming.py, encode_pretraining → encode_streaming)
2. Extended to support `sample_packing: true` via a new `_chunk_long_sequences()` function

### `encode_streaming` (for `sample_packing: false`)
- Tokenizes without truncation to preserve all data
- Automatically splits long sequences into `max_tokens`-sized chunks
- Handles pad token fallback gracefully

### `_chunk_long_sequences` (NEW - for `sample_packing: true`)
- Chunks sequences longer than `max_seq_length` instead of dropping them
- Called in `encode_packed_streaming` before sample packing
- Preserves all data that would otherwise be lost by the "Dropping Long Sequences" filter

## Motivation

The original PR #3081 only addressed non-packed streaming. When `sample_packing: true`, long sequences were still being dropped silently. This PR ensures data preservation for both code paths.

## Test plan

- [x] Tested with `sample_packing: false` - sequences are chunked correctly
- [x] Tested with `sample_packing: true` - long sequences are chunked before packing instead of dropped
- [x] Verified no data loss with datasets containing very long samples

## Related

- Builds on #3081 (not merged)

## Summary by CodeRabbit

* **Improvements**
  * Enhanced streaming data processing with an improved sequence chunking strategy that uses fixed-size chunks instead of truncation-based approaches, providing significantly better handling of long input sequences during training data preparation workflows.
  * Implemented enhanced padding logic for more robust and stable processing of variable-length sequences in the data pipeline infrastructure.

### DIFF
--- a/src/axolotl/utils/data/streaming.py
+++ b/src/axolotl/utils/data/streaming.py
@@ -4,6 +4,7 @@ import functools
 from collections import defaultdict
 from typing import Callable, Dict, List, Optional
 
+import numpy as np
 import torch
 from datasets import Dataset
 from torch.utils.data import RandomSampler
@@ -253,13 +254,17 @@ def wrap_streaming_dataset(
 def _chunk_long_sequences(
     train_dataset: Dataset,
     max_seq_length: int,
-) -> Dataset:
+) -> Dict[str, list]:
     """
     Chunk sequences longer than max_seq_length into multiple smaller sequences.
 
     Instead of dropping long sequences (which loses data), this function splits
-    them into max_seq_length-sized chunks. This is especially useful for pretraining
-    datasets with very long samples (e.g., millions of tokens per example).
+    them into max_seq_length-sized chunks. Uses numpy for fast array splitting
+    to avoid the overhead of pure-Python list slicing on very long sequences
+    (100K+ tokens).
+
+    This is especially useful for pretraining datasets with very long samples
+    (e.g., millions of tokens per example).
 
     Note: This should only be used for pretraining. For SFT, long sequences should
     be dropped to maintain complete instruction-response pairs.
@@ -273,13 +278,13 @@ def _chunk_long_sequences(
         max_seq_length: Maximum sequence length for each chunk
 
     Returns:
-        Dataset with all sequences <= max_seq_length
+        Dict of lists with all sequences <= max_seq_length
     """
     columns = train_dataset.column_names
     has_labels = "labels" in columns
     has_attention_mask = "attention_mask" in columns
 
-    # Use batch column access for performance (avoids per-element __getitem__)
+    # Batch column access avoids per-element __getitem__ overhead
     all_input_ids: list = train_dataset["input_ids"]
     all_attention_mask: list = (
         train_dataset["attention_mask"] if has_attention_mask else []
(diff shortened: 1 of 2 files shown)


NanoCode012:
haven't read too much into changes, does this only apply to pretraining? For streaming sft, we want to ensure it drops.

madScientist10:
Good catch. For packed streaming, it would truncate for both pretraining and SFT. 
I added `is_pretraining` parameter to `encode_packed_streaming`, so that chunking is now conditional on `is_pretraining=bool(cfg.pretraining_dataset)` to prevent being applied on SFT.

For non-packed SFT, as per the existing note in streaming.py, it does not perform truncation since it follows a different pathway:
>  NOTE: This is not reachable for SFT datasets since we use the pre-existing
>  loading function for non-packed streaming datasets. Refer to
>  _prepare_streaming_datasets in sft.py for that code path.

NanoCode012:
> Good catch. For packed streaming, it would truncate for both pretraining and SFT. I added `is_pretraining` parameter to `encode_packed_streaming`, so that chunking is now conditional on `is_pretraining=bool(cfg.pretraining_dataset)` to prevent being applied on SFT.
> 
> For non-packed SFT, as per the existing note in streaming.py, it does not perform truncation since it follows a different pathway:
> 
> > NOTE: This is not reachable for SFT datasets since we use the pre-existing
> > loading function for non-packed streaming datasets. Refer to
> > _prepare_streaming_datasets in sft.py for that code path.

Thanks! I'll need to check back later on this.

winglian:
here are some issues Claude brought up:

Summary
The PR rewrites encode_streaming and adds a new _chunk_long_sequences helper to ensure long sequences in streaming pretraining datasets are chunked rather than truncated/dropped. The intent is good — preserving training data that would otherwise be silently lost.

Issues
1. encode_streaming rewrite removes EOS/PAD boundary tokens (Bug)

The original encode_streaming (main) explicitly appended EOS + PAD tokens between concatenated documents to mark document boundaries:

# OLD: main branch
input_ids[i] = torch.cat((input_ids[i], torch.tensor([tokenizer.eos_token_id, tokenizer.pad_token_id])), dim=0)

The new implementation completely removes this logic. When concatenate=True, it just torch.cats all tokenized inputs into one flat tensor and slices at max_tokens boundaries. This means:

Document boundaries are lost — the model sees a continuous stream of tokens with no separator between distinct documents.
This is a semantic regression for pretraining, where knowing where one document ends and another begins matters for loss masking and model learning.
This should either be restored or explicitly justified with reasoning about why it's acceptable.

2. encode_streaming rewrite drops the truncation=True, max_length=max_tokens - 2 guard

The old code reserved 2 tokens for EOS+PAD. The new code tokenizes fully without truncation. While this is intentional (to enable chunking), it means memory usage during tokenization is unbounded — a single extremely long text will be fully tokenized into a potentially enormous tensor before being chunked. For streaming pretraining with very long documents (e.g., books), this could OOM during the tokenizer() call itself.

3. _chunk_long_sequences iterates the dataset element-by-element (Performance)

streaming.py:219-223:

long_samples = sum(
    1 for i in range(total_samples)
    if len(train_dataset[i]["input_ids"]) > max_seq_length
)

Then it iterates again at line 238: for i in range(total_samples). This does two full passes over the dataset, accessing each element individually by index. For a Dataset backed by Arrow, per-element __getitem__ is much slower than batch/column access. Consider using train_dataset["input_ids"] to get the full column in one call, then operating on the list.

4. _chunk_long_sequences produces variable-length chunks but the downstream packer may not expect this

The last chunk of a split sequence can be shorter than max_seq_length (line 254: end = min(start + max_seq_length, seq_len)). This is fine for the packed path since process_pretraining_datasets_for_packing handles variable lengths, but it's worth noting that very short trailing chunks (e.g., a few tokens) add overhead to the packing bin without much training signal.

5. Existing test test_encode_pretraining is now broken

tests/test_data.py:45-59 asserts specific behavior of the old encode_streaming:

Line 45: self.assertEqual(len(result["input_ids"]), 3) — the number of chunks will differ with the new chunking logic.
Lines 53-59: Assert EOS token at position 5, PAD at position 6, BOS at position 7, etc. — the new code no longer inserts EOS/PAD between documents, so these assertions will all fail.
This test must be updated or the PR will break CI.

6. No new tests for _chunk_long_sequences

The new _chunk_long_sequences function has no test coverage. It should have unit tests covering:

No long sequences (early return path)
Sequences exactly at max_seq_length
Sequences requiring multiple chunks
Handling of the trailing short chunk
Datasets with and without labels / attention_mask
7. The concatenate=False path changed behavior silently

In the old code, concatenate=False returned individually truncated samples directly. In the new code, concatenate=False still runs the chunking loop (line 81-107), which means a single long document with concatenate=False now produces multiple output rows where it previously produced one (truncated) row. This is the intended feature, but callers expecting a 1:1 input-to-output mapping for concatenate=False may break.

Minor Nits
The PR title says "auto-truncation" but the implementation is "auto-chunking" (no data loss). The title should match the behavior.
inputs_ids (line 71) has a typo — should be input_ids for consistency with the dict key and standard naming.

madScientist10:
Thanks for the thorough review. I've addressed the necessary issues, and just leave a comment on the rest:

1. (EOS/PAD tokens) Fixed; restored the EOS + PAD appending logic before concatenation, matching the original behavior, so document boundaries are preserved.
2. (Unbounded tokenization memory) This is acknowledged, but it's inherent to the auto-chunking design (truncating first would defeat the purpose). The streaming_multipack_buffer_size controls how many examples are batched at once, limiting peak memory. I added a comment explaining this trade-off, but I'm open to ideas if we want to eliminate fully this risk.
3. (_chunk_long_sequences performance) Fixed; Replaced per-element __getitem__ with batch column access (train_dataset["input_ids"], etc.) and merged the counting pass into the main loop (now a single pass over the data).
4. (Variable-length trailing chunks) This is by design, since the downstream packer handles variable lengths, and dropping short trailing chunks would lose data. Added a comment noting this is intentional.
5. (Broken test) Fixed; The existing test_encode_pretraining passes as-is since the EOS/PAD logic was restored.
6. (Missing tests) Added 7 new unit tests for _chunk_long_sequences covering: no long sequences (early return), sequences exactly at max, multiple chunks, trailing short chunk, mixed short/long, without labels, and without attention_mask. All 10 tests pass.
7. (concatenate=False behavior change) This is the intended feature, yes. Added a note in the docstring explaining that individual samples exceeding max_tokens will be split into multiple chunks.

I also renamed PR title from "auto-truncation" to "auto-chunking", and "inputs_ids" to "input_ids".

madScientist10:
A couple of performance optimizations for `_chunk_long_sequences` and `encode_packed_streaming`:
I stress-tested the implementation on a streaming data corpus of many, very long docs (several million tokens each). The tokenization concern raised above was not an issue, but there were two other bottlenecks that are addressed in the latest commits:

`_chunk_long_sequences`: Pure-Python list slicing on very long sequences was slow. Now uses numpy (np.split) for fast array splitting. Also returns a plain Dict[str, list] instead of an HF Dataset, eliminating Dataset.from_dict() Arrow serialization overhead.
`encode_packed_streaming` (pretraining path): Bypasses process_pretraining_datasets_for_packing + get_dataset_lengths and builds sample dicts directly in a plain Python loop. This avoids HF Dataset .filter() / .map() / __getitem__ Arrow deserialization, which is significant with very long chunks. The same logic is replicated inline (drop short/long seqs, add position_ids, conditionally drop attention_mask). Also handles MultipackBatchSampler yielding list-of-indices for packed batches.

These changes reduce end-to-end data processing time from several minutes/hours (which would basically never even complete because they would crash due to NCCL timeouts) to seconds/minutes per streaming buffer on my tests. Tests are updated for the new return type.

Issue #3374, opened 2026-01-27
Transformers-v5 causes OOM on QLoRA config previously working on 4.57.6 (GLM-4.5-Air on 2x RTX Pro 6000)


Nero10578:
### Please check that this issue hasn't been reported before.

- [x] I searched previous [Bug Reports](https://github.com/axolotl-ai-cloud/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

Upgrading to transformers v5 in the axolotl transformers-v5 branch should have allowed configs that worked in transformers 4.57.6 to work as well.

### Current behaviour

Using the transformers-v5 branch of axolotl with transformers v5 causes a training config which previously worked with transformers 4.57.6 to OOM upon loading of model weights stage.

On model load the GPU usage spikes to 94.4GB and halfway on model load the GPU OOMs, seemingly as if the 4-bit quantization is not being applied to the model.

If I used the main axo branch with transformers 4.57.6 then the GPU VRAM usage is about 48GB on model load and then only creeps to 95.3GB per GPU as it starts to train. Allowing GLM-4.5-Air to be trained on 2x RTX Pro 6000 96GB GPUs with QLoRA.

### Steps to reproduce

I suppose this might affect a lot of MoE models that has the new fused moe implementation in transformers, but this should be the case on any glm4 moe models. 

Try running a QLoRA 4 bit train on transformers 4.57.6 and try again using transformers v5 and you will see the VRAM use explode as if the model isn't being quanted but as if its trained with full precision LoRA.

### Config yaml

```yaml
base_model: /home/arli/models/GLM-4.5-Air-Derestricted

plugins:
  - axolotl.integrations.cut_cross_entropy.CutCrossEntropyPlugin
cut_cross_entropy: true

load_in_8bit: false
load_in_4bit: true

datasets:
  - path: ./combined-dedup-nolasthuman.jsonl
    type: chat_template

    field_messages: conversations
    message_property_mappings:
      role: from
      content: value

dataset_prepared_path: ./last_run_prepared
val_set_size: 0
output_dir: ./outputs/lora-out-derestricted-tuned-try2

adapter: qlora
save_safetensors: true

sequence_len: 32768
sample_packing: true
pad_to_sequence_len: true

lora_r: 32
lora_alpha: 32
lora_dropout: 0.0
lora_target_linear: true
lora_target_modules:
  - gate_proj
  - down_proj
  - up_proj
  - q_proj
  - v_proj
[...]

PR #3395, opened 2026-02-09
FIX: monkey patch bitsandbytes oom on v5 

ved1beta:
usee [targate_params](https://github.com/huggingface/peft/blob/2cd96ed041620f74d239a0ce3f16207153c43413/src/peft/tuners/lora/config.py#L448)
to detect lora params 

#3418
#3374


src/axolotl/monkeypatch/moe_quant.py
```diff
@@ -0,0 +1,44 @@
...
+    target_params = MOE_EXPERT_PARAMS.get(model_config_type)
+    if not target_params:
+        return
+
+    count = 0
+    for module_name, module in model.named_modules():
+        for param_name in target_params:
+            if hasattr(module, param_name):
+                param = getattr(module, param_name)
+                if isinstance(param, torch.nn.Parameter) and param.ndim >= 2:
```
NanoCode012: Any specific reason for `param.ndim >= 2:`?

src/axolotl/loaders/model.py
```diff
@@ -173,6 +173,29 @@ def load(self) -> tuple[PreTrainedModel | PeftModelForCausalLM, PeftConfig | Non
...
+
+            bnb_config_params = inspect.signature(
+                BitsAndBytesConfig.__init__
+            ).parameters
+            if "target_parameters" not in bnb_config_params:
+                from axolotl.monkeypatch.moe_quant import (
+                    quantize_moe_expert_params,
+                )
+
+                quantize_moe_expert_params(self.model, self.cfg.model_config_type)
```
NanoCode012: Would regular bnb lora also have this issue?

src/axolotl/loaders/adapter.py
```diff
@@ -114,11 +115,22 @@ def load_lora(
     else:
         task_type = TaskType.CAUSAL_LM
 
+    # Exclude ParametrizationList modules created by MoE expert quantization.
+    # replace_parameter_4bit wraps quantized params in ParametrizationList child
+    # modules that PEFT doesn't support as LoRA targets.
+    exclude_modules = cfg.lora_exclude_modules or []
```
NanoCode012: I'm not sure this config exists

src/axolotl/common/architectures.py
```diff
@@ -19,3 +19,46 @@
     "lfm2_moe": "Lfm2MoeSparseMoeBlock",
     "afmoe": "AfmoeMoE",
 }
+
+# MoE architectures whose expert weights are 3D nn.Parameter tensors (not nn.Linear).
+# BnB 4-bit quantization skips these by default, causing OOM. This mapping provides
+# the parameter names needed for `target_parameters` in BitsAndBytesConfig or for
+# post-load quantization via bitsandbytes.nn.parametrize.
+# Verified against transformers 5.0.0 source.
+MOE_EXPERT_PARAMS = {
```
NanoCode012: I'm wondering whether there's a better way to handle this. This would require us to maintain this list. Could we use the dict above to detect whether it's a moe model?

The keys for most of the moe layers are the mostly the same either way. Would there be potential conflicts if we default to all of them?

src/axolotl/loaders/adapter.py
```diff
@@ -67,14 +67,63 @@ def find_all_linear_names(model):
     return list(lora_module_names)
 
 
+def find_moe_expert_param_names(model):
```
winglian:
```suggestion
def find_moe_expert_param_names(model: PreTrainedModel) -> list[str]:
```

src/axolotl/loaders/adapter.py
```diff
@@ -67,14 +67,63 @@ def find_all_linear_names(model):
...
+    ["mlp.experts.gate_up_proj", "mlp.experts.down_proj"]) suitable for
+    PEFT's LoraConfig target_parameters.
+    """
+    seen_suffixes = set()
+    for name, param in model.named_parameters():
+        if param.ndim >= 3:
+            parts = name.split(".")
+            # Find the layer index (first numeric segment) and extract the
+            # repeating suffix after it.
+            # e.g. "model.layers.0.mlp.experts.gate_up_proj" -> "mlp.experts.gate_up_proj"
```
NanoCode012, line 90: Maybe we should have some checks for the word "experts" or gate_up_proj / down_proj. They seem to be the common names used.

src/axolotl/loaders/model.py
```diff
@@ -173,6 +173,20 @@ def load(self) -> tuple[PreTrainedModel | PeftModelForCausalLM, PeftConfig | Non
...
+        self.model._moe_experts_quantized = False
+        self.model._moe_expert_param_names = []
+        if self.cfg.adapter in ("qlora", "lora") and (
+            self.cfg.load_in_4bit or self.cfg.load_in_8bit
+        ):
+            from axolotl.loaders.adapter import find_moe_expert_param_names
+            from axolotl.monkeypatch.moe_quant import quantize_moe_expert_params
+
+            self.model._moe_expert_param_names = find_moe_expert_param_names(self.model)
+            self.model._moe_experts_quantized = quantize_moe_expert_params(self.model)
```
NanoCode012, line 188: This is called for LoRa, despite the inner function calling `replace_parameter_4bit` specifically. Do we need to have a specific `replace_parameter_8bit` for lora ?

examples/glm4.7/glm4.7-flash-qlora.yaml
```diff
@@ -0,0 +1,65 @@
...
+pad_to_sequence_len: true
+
+lora_r: 32
+lora_alpha: 32
+lora_dropout: 0.0
+lora_target_linear: true
+lora_target_modules:
+  - gate_proj
+  - down_proj
+  - up_proj
```
NanoCode012:
```suggestion
```

examples/glm4.7/glm4.7-flash-qlora.yaml
```diff
@@ -0,0 +1,65 @@
...
+  - q_proj
+  - v_proj
+  - k_proj
+  - o_proj
+
```
NanoCode012, line 39: Maybe we should explicitly set the lora target parameters so it's clear that it's being trained on here.

It doesn't seem possible to Not target those layers as well.It seems to always be on

NanoCode012 (dismissed):

examples/glm4.7/glm4.7-flash-qlora.yaml
ved1beta: done !

src/axolotl/loaders/model.py
ved1beta: narrowed the guard to load_in_4bit only now

NanoCode012:
Mostly superseded by #3439 . We'll move the lora kernel specific changes to a new PR

PR #3401, opened 2026-02-10
Fix: excess_length_strategy truncation  method

rlronan:
Fix the sample truncation used in handle_long_seq_in_dataset invoked by setting excess_length_strategy to 'truncation' in the config. This is called during the SFT _load_raw_datasets function when sample_packing is False, skip_prepare_dataset is False, and streaming is false  at: 
https://github.com/axolotl-ai-cloud/axolotl/blob/06ac407b92168f79c512df33d42387b6af5aa7f9/src/axolotl/utils/data/sft.py#L346-L360

# Description

Removes the minimum sequence length filtering from the truncate_long_samples function, and converts the truncate_long_samples function called in handle_long_seq_in_dataset to use map. As flagged 4-5 times by the AI review in the [initial PR that introduced this feature](https://github.com/axolotl-ai-cloud/axolotl/pull/3068/changes#diff-dc284316c84f597e4e7126d804cf388e9df82a9ad29bb6f56a48e255d1cf5d54), filtering won't preserve the applied truncations, so they are never actually applied to the dataset. Instead map must be used. 

Additionally the PR adds test cases for handle_long_seq_in_dataset to validate that the initial code implementation does not correctly truncate, and that the new code from this PR does. 

## Motivation and Context

As above, this was flagged during the feature creation. I ran into this issue trying to train with long sequences, and truncate using excess_length_strategy='truncate', and setting increasingly small sequence_len values, but I keep running into OOM when the forward pass tried to allocate ~20gb for logits. I checked that manually truncating the samples to <~512 tokens did not produce an OOM, whereas setting excess_length_strategy='truncate', and sequence_len of 512 and smaller still caused OOM errors. 

## How has this been tested?

I tested this by writing test cases for handle_long_seq_in_dataset and running them directly with UV. 

I am testing in a local clone of the repo on mac osx. 

## AI Usage Disclaimer
Yes;
AI (Claude Code with Opus 4.6) was used to generate and assist with this PR. I asked Claude write the unit tests, to remove the low sample length filter from the truncate_long_samples function, and to reimplement it as a filter + map. I ran the tests, and reviewed the tests to validate they covered what I expected and reviewed the code changes to validate they were as expected

## Types of changes
[x] Bug Fix

## Summary by CodeRabbit

* **New Features**
  * Added automatic filtering of sequences that fall below the minimum required length from input datasets

* **Bug Fixes**
  * Improved sequence truncation handling with a more efficient two-stage processing approach: filtering sequences below minimum length before truncating those exceeding maximum length limits

* **Tests**
  * Comprehensive test coverage for all sequence processing strategies, edge cases, and configuration options

### DIFF
--- a/src/axolotl/utils/trainer.py
+++ b/src/axolotl/utils/trainer.py
@@ -205,6 +205,43 @@ def add_length(sample):
     return sample
 
 
+def check_max_seq_length(sample, sequence_len=2048, raise_on_long=False):
+    """
+    Check if samples exceed the maximum sequence length.
+
+    Works for both single-example (list[int]) or batched (list[list[int]]).
+
+    If raise_on_long is set, raises a ValueError if a sample exceeds sequence_len.
+    Otherwise returns True if sample is within limit, False if too long.
+    """
+    input_ids = sample["input_ids"]
+
+    # Edge case: if input_ids is empty, keep it (filtering short samples is separate)
+    if not input_ids:
+        return True
+
+    # Check if single example or batched by looking at the first element
+    if isinstance(input_ids[0], int):
+        # Single example (input_ids is a list of int)
+        length = len(input_ids)
+        if raise_on_long and length > sequence_len:
+            raise ValueError(
+                f"Sequence encountered with {length} tokens, which exceeds the maximum {sequence_len}."
+            )
+        return length <= sequence_len
+
+    # Batched (input_ids is a list of lists)
+    results = []
+    for seq in input_ids:
+        length = len(seq)
+        if raise_on_long and length > sequence_len:
+            raise ValueError(
+                f"Sequence encountered with {length} tokens, which exceeds the maximum {sequence_len}."
+            )
+        results.append(length <= sequence_len)
+    return results
+
+
 def drop_long_seq(sample, sequence_len=2048, min_sequence_len=2, raise_on_drop=False):
     """
     Drop samples whose sequence length is either too long (> sequence_len)
--- a/src/axolotl/utils/trainer.py
+++ b/src/axolotl/utils/trainer.py
@@ -205,47 +205,13 @@ def add_length(sample):
     return sample
 
 
-def check_max_seq_length(sample, sequence_len=2048, raise_on_long=False):
+def filter_sequences_by_length(
+    sample, sequence_len=2048, min_sequence_len=2, raise_on_drop=False
+):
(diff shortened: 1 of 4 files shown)


NanoCode012:
Thanks for looking into this. Will ping @ved1beta to give this a quick review

ved1beta (approved):
Looks good , if ci passes 🚀

NanoCode012 (approved):
Thanks for the PR. I took another look at it, did some refactoring, and added another test as well

src/axolotl/utils/trainer.py
```diff
@@ -205,6 +205,43 @@ def add_length(sample):
...
+        length = len(seq)
+        if raise_on_long and length > sequence_len:
+            raise ValueError(
+                f"Sequence encountered with {length} tokens, which exceeds the maximum {sequence_len}."
+            )
+        results.append(length <= sequence_len)
+    return results
+
+
 def drop_long_seq(sample, sequence_len=2048, min_sequence_len=2, raise_on_drop=False):
```
rlronan: Seems like there's another copy(copies?) of this function?

rlronan:
> Thanks for the PR. I took another look at it, did some refactoring, and added another test as well

Thanks for the refactor. Looks like the drop sequence function is redefined elsewhere. Maybe worth cleaning up or looking into all the sequence length handling stuff (in a separate pr?) 

Also what do you use for code formatting? The link in the contributions doc on codestyle doesn't go anywhere.

Lastly, do I need to request the CI be run on this PR?

src/axolotl/utils/trainer.py
NanoCode012: Thanks, fixed this

NanoCode012:
> > Thanks for the PR. I took another look at it, did some refactoring, and added another test as well
> 
> Thanks for the refactor. Looks like the drop sequence function is redefined elsewhere. Maybe worth cleaning up or looking into all the sequence length handling stuff (in a separate pr?)

Yup, will leave it as this.

> 
> Also what do you use for code formatting? The link in the contributions doc on codestyle doesn't go anywhere.

Weird, it should go here https://github.com/axolotl-ai-cloud/axolotl/blob/main/.github/CONTRIBUTING.md

Used to use black, but now we just use ruff

> Lastly, do I need to request the CI be run on this PR?

I started it!

PR #3409, opened 2026-02-13
Fix: load gemma3 text only via dynamic weights

NanoCode012:
# Description

Not tested yet!

This loads the Gemma3 multi-modal as a text only mode -> train -> offer script to run merge back to original model with vision.

- [ ] Figure out the flow for adapters. Train -> merge to base -> merge multi-modal?
- [ ] Check if adapter load with incomplete keys work (as it only targets language model layer) and how it interacts with dynamic weights

## AI Usage Disclaimer

Claude

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -193,3 +193,6 @@ out/
 
 # scm auto-versioning
 src/axolotl/_version.py
+
+# macOS
+.DS_Store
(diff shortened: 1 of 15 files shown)


NanoCode012:
Fixed in #3500

PR #3410, opened 2026-02-15
ScatterMoE LoRA support

winglian:
# Description

Uses local layer repo for ScatterMoE implementation instead of remote repo, making this easier to modify. 
ScatterMoE from https://arxiv.org/abs/2403.08245 and https://github.com/shawntan/scattermoe
Fuses LoRA computation with ScatterMoE
Uses some optimizations recommended in in [SonicMoE ](https://arxiv.org/abs/2512.14080) such as fused dX and fused gather

### DIFF
--- a/src/axolotl/core/trainers/base.py
+++ b/src/axolotl/core/trainers/base.py
@@ -719,13 +719,7 @@ class AxolotlTrainer(
         output_dir = output_dir if output_dir is not None else self.args.output_dir
         os.makedirs(output_dir, exist_ok=True)
         LOG.info(f"Saving model checkpoint to {output_dir}")
-        if state_dict is None:
-            state_dict = self.accelerator.get_state_dict(self.model)
-        if state_dict is not None:
-            state_dict = {
-                k: v.clone() if isinstance(v, torch.Tensor) else v
-                for k, v in state_dict.items()
-            }
+
         supported_classes = (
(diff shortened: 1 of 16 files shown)


src/axolotl/core/trainers/base.py
```diff
@@ -719,13 +719,7 @@ def _save(self, output_dir: Optional[str] = None, state_dict=None):
         output_dir = output_dir if output_dir is not None else self.args.output_dir
         os.makedirs(output_dir, exist_ok=True)
         LOG.info(f"Saving model checkpoint to {output_dir}")
-        if state_dict is None:
```
NanoCode012, line 722: This was added recently to solve some saving issue . Do the changes below solve it?

NanoCode012: Just did a quick pass on this, `.clone()` may be unintentionally placing tensors on GPU.

winglian: What do you mean? This was a cleanup from changes upstream.

NanoCode012: This was just added by ved a week ago to fix saving in context parallelism https://github.com/axolotl-ai-cloud/axolotl/commit/97a4f28511afd94e60d733a48a8dd193cea59868

PR #3419, opened 2026-02-17
feat: support dot-notation CLI args for nested config options

### LINKED ISSUE #2702: CLI Args should also handle nested options

### 🔖 Feature description

There is already an option to override yaml config values using cli arguments.
Looks like that: `axolotl train gsm8k.yaml --lr=0.0001`.
However it does not work with options that are defined by separate configs. My specific issue is with TRL config.
I'm running a multi-node GRPO training on SLURM cluster and i'm using one node for VLLM and all other nodes for actual training processes. There is no way to know which address would VLLM node have, so i need to be able to set it via cli args. Currently `axolotl train` does have a single --trl flag, but it is unusable. It can't even parse json dict.

### ✔️ Solution

Ideally it should be something like that:
`axolotl train grpo.yaml --trl.vllm-server-host="..." --trl.vllm-server-port="..."` which would override whatever values i have in my grpo.yaml.

### ❓ Alternatives

- no alternatives, apart from patching .yaml config during run

### 📝 Additional Context

The best way to do that is likely by adding additional if-clause into `add_options_from_config` (https://github.com/axolotl-ai-cloud/axolotl/blob/6aa41740df7623d1f8995a1efd3b668f4a57c5cf/src/axolotl/cli/utils.py#L123) which will handle cases when field of the config is another config.

### Acknowledgements

ManasVardhan:
## Summary

Adds support for overriding nested config fields (like TRL config) via CLI using dot-notation:

```bash
axolotl train grpo.yaml --trl.vllm-server-host=10.0.0.1 --trl.vllm-server-port=8000 --trl.beta=0.1
```

### `src/axolotl/cli/utils/args.py`
- `add_options_from_config`: When a field's type is a Pydantic `BaseModel` subclass, automatically generates dot-notation CLI options for all its sub-fields (e.g., `--trl.beta`, `--trl.vllm-server-host`). The kwargs use double-underscore as separator (`trl__beta`).
- Fixed `_strip_optional_type` to handle Python 3.10+ `X | None` union syntax (`types.UnionType`), which was preventing nested model detection.

### `src/axolotl/cli/config.py`
- `load_cfg`: Handles double-underscore kwargs by splitting on `__` and setting nested dict values on the config object.

### `tests/cli/test_nested_options.py`
- Tests for dot-notation option registration, boolean flags, mixed flat+nested options, and the nested kwargs application logic.

## Motivation

Fixes #2702 — Users running multi-node GRPO training on SLURM clusters need to dynamically set TRL config values (like `vllm_server_host`) via CLI args, since the VLLM node address isn't known until runtime.

## Summary by CodeRabbit

* **New Features**
  * Added support for nested CLI configuration options using dot-notation syntax (e.g., `--parent.child value`) to configure deeply nested settings more intuitively.
  * Enhanced boolean flag handling for nested options with `--flag`/`--no-flag` syntax.

* **Tests**
  * Added comprehensive tests validating nested configuration option handling and mixed flat/nested option usage.

### DIFF
--- a/src/axolotl/cli/config.py
+++ b/src/axolotl/cli/config.py
@@ -5,7 +5,7 @@ import os
 import tempfile
 from pathlib import Path
 from tempfile import NamedTemporaryFile
-from typing import Union
+from typing import Any, Union
 from urllib.parse import urlparse
 
 import requests
@@ -208,13 +208,36 @@ def load_cfg(
     # If there are any options passed in the cli, if it is something that seems valid
     # from the yaml, then overwrite the value
     cfg_keys = cfg.keys()
+
+    # Separate nested (dot-notation) kwargs from flat kwargs
+    nested_kwargs: dict[str, dict[str, Any]] = {}
+    flat_kwargs: dict[str, Any] = {}
     for key, value in kwargs.items():
+        if "__" in key:
+            parent, child = key.split("__", 1)
+            nested_kwargs.setdefault(parent, {})[child] = value
+        else:
+            flat_kwargs[key] = value
+
+    # Apply flat kwargs
+    for key, value in flat_kwargs.items():
         # If not strict, allow writing to cfg even if it's not in the yml already
         if key in cfg_keys or not cfg.strict:
             if isinstance(cfg[key], bool):
                 cfg[key] = bool(value)
             else:
                 cfg[key] = value
+
+    # Apply nested kwargs (e.g., trl__beta -> cfg.trl.beta)
+    for parent, children in nested_kwargs.items():
+        if parent not in cfg_keys and cfg.strict:
+            continue
+        if cfg[parent] is None:
+            cfg[parent] = {}
+        if not isinstance(cfg[parent], dict):
+            cfg[parent] = dict(cfg[parent]) if hasattr(cfg[parent], '__iter__') else {}
+        for child_key, child_value in children.items():
+            cfg[parent][child_key] = child_value
 
     try:
         device_props = torch.cuda.get_device_properties("cuda")
--- a/src/axolotl/cli/config.py
+++ b/src/axolotl/cli/config.py
@@ -235,7 +235,10 @@ def load_cfg(
         if cfg[parent] is None:
             cfg[parent] = {}
         if not isinstance(cfg[parent], dict):
-            cfg[parent] = dict(cfg[parent]) if hasattr(cfg[parent], '__iter__') else {}
+            LOG.warning(
+                "Overwriting non-dict value for '%s' with nested CLI overrides", parent
+            )
+            cfg[parent] = {}
         for child_key, child_value in children.items():
             cfg[parent][child_key] = child_value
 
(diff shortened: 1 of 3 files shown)


NanoCode012:
This is quite nice to have. Thanks for catching this. We'll get back to you when we get to review :)

src/axolotl/cli/config.py
```diff
@@ -208,14 +208,37 @@ def load_cfg(
...
+    # Apply nested kwargs (e.g., trl__beta -> cfg.trl.beta)
+    for parent, children in nested_kwargs.items():
+        if parent not in cfg_keys and cfg.strict:
+            continue
+        if cfg[parent] is None:
+            cfg[parent] = {}
+        if not isinstance(cfg[parent], dict):
+            cfg[parent] = dict(cfg[parent]) if hasattr(cfg[parent], '__iter__') else {}
+        for child_key, child_value in children.items():
+            cfg[parent][child_key] = child_value
```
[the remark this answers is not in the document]
ManasVardhan: Fixed in 2a95bf8 -- replaced the `dict(cfg[parent]) if hasattr(..., '__iter__')` fallback with a simple `cfg[parent] = {}` for all non-dict values, with a log warning. Also addressed the other review suggestions (docstring noting single-level nesting, type hints in args.py, and refactored the test to use extracted helper). Thanks for the catch!

tests/cli/test_nested_options.py
```diff
@@ -0,0 +1,171 @@
...
+                cfg[parent][child_key] = child_value
+
+        return cfg
+
+    def test_nested_kwargs_applied_to_cfg(self, tmp_path):
+        """Double-underscore kwargs should set nested config values."""
+        from axolotl.utils.dict import DictDefault
+
+        cfg = DictDefault({"trl": {"beta": 0.1}, "learning_rate": 0.01})
+        kwargs = {"trl__beta": 0.5, "trl__host": "192.168.1.1", "learning_rate": 0.02}
```
winglian: doesn't the CLI pass strings? If so args aren't likely to be cast to the appropriate types. so the `kwargs` would probably be closer to:
```python
kwargs = {"trl__beta": "0.5", "trl__host": "192.168.1.1", "learning_rate": "0.02"}
~``

Comment (ManasVardhan) [reply tests/cli/test_nested_options.py]: Good catch - this is handled! The PR includes a `_coerce_value()` function in `config.py` that does two things:

1. **If the config already has a typed value**, it casts to match (e.g., existing float 0.1 + CLI string "0.5" -> float 0.5)
2. **If no existing value**, it applies YAML-style inference ("true" -> bool, "42" -> int, "3.14" -> float, etc.)

So for your example:
```python
kwargs = {"trl__beta": "0.5", "trl__host": "192.168.1.1", "learning_rate": "0.02"}
```
- `trl.beta`: existing value is float 0.1, so "0.5" gets cast to float 0.5
- `trl.host`: no existing value, stays as string "192.168.1.1"
- `learning_rate`: existing value is float, so "0.02" gets cast to float 0.02

Tests for all of these cases are in `TestCoerceValue` and `TestLoadCfgNestedKwargs`.

winglian (approved):
Thanks for the feature improvement!

PR #3427, opened 2026-02-21
fix: Save de-duplicated dataset during pre-processing

### LINKED ISSUE #2719: Save De-duplicated Set During Pre-processing

### 🔖 Feature description

Currently the de-duplication is being run *after* the set has been saved during the pre-processing phase, and gets run again when starting the training phase (as in doing `-m axolotl.cli.preprocess` then doing `-m axolotl.cli.train`).

For small sets it might not matter much, but for larger sets (which would make more sense to be de-duplicating) this takes a *long* time. That means that if you had a large dataset that took like 15 minutes to de-duplicate you'd be spending 30 minutes on just de-duplicating, before you get to take a single training step, plus another 15 minutes any time you wanted to restart training for whatever reason, like OOM, hyper-parameter tuning, etc.

```
Example time to de-duplicate one of my configs: ~14 minutes.

Saving the dataset (23/23 shards): 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████| 162329/162329 [01:35<00:00, 1694.97 examples/s]
[2025-05-23 22:41:55,123] [INFO] [axolotl.utils.data.utils.deduplicate_and_log_datasets:134] [PID:122699] [RANK:0] Train dataset is None. Skipping deduplication.
[2025-05-23 22:41:55,123] [INFO] [axolotl.utils.data.utils.deduplicate_and_log_datasets:147] [PID:122699] [RANK:0] Eval dataset is None. Skipping deduplication.
[2025-05-23 22:41:55,124] [INFO] [axolotl.utils.data.utils.deduplicate_and_log_datasets:150] [PID:122699] [RANK:0] Starting deduplication for combined dataset. Original size: 162329
[2025-05-23 22:56:28,138] [INFO] [axolotl.utils.data.utils.deduplicate_and_log_datasets:154] [PID:122699] [RANK:0] Deduplication complete for combined dataset. New size: 159750
```

If this de-duplication step was instead done *only* during the pre-processing stage and then *saved*, you would only need to do it once and would also be able to push the de-duplicated set to huggingface.

### ✔️ Solution

Move the de-duplication to *before* saving in the pre-processing phase. Also prevent it from being run before the training phase.

### ❓ Alternatives

Not exactly an alternative, but an addition; making the de-duplication code multi-threaded. Currently it only seems to hit 1 or 2 threads. So that's probably where a lot of the slowness is coming from.

### 📝 Additional Context

_No response_

### Acknowledgements

ManasVardhan:
## Summary

Fixes #2719

Currently, de-duplication runs **after** the dataset is saved during preprocessing. This means the saved/cached dataset still contains duplicates, and deduplication has to run again on every subsequent load.

## Changes

Moved `deduplicate_and_log_datasets` to run **before** `save_preprocessed_dataset` in both:
- **SFT pipeline** (`sft.py` → `_load_raw_datasets`)
- **RL pipeline** (`rl.py` → `_load_split`)

Removed the now-redundant deduplication calls in `_handle_train_dataset_split` and `_handle_test_dataset_split` (SFT), since the dataset is already de-duplicated before it reaches those functions.

Note: `create_train_validation_split` in `shared.py` still has its own dedup call, which is now a harmless no-op (no duplicates left to remove) but provides a safety net.

## Tests

Added `tests/test_save_deduplicated.py` with tests verifying:
- Deduplication is called before save in SFT pipeline
- Deduplication is called before save in RL pipeline  
- Deduplication is not called when `dataset_exact_deduplication` is disabled

## Summary by CodeRabbit

* **Refactor**
  * Reorganized dataset deduplication to occur earlier in the preprocessing pipeline for improved data processing efficiency.

* **Tests**
  * Added unit tests to verify deduplication occurs at the correct stage in the data loading workflow.

### DIFF
--- a/src/axolotl/utils/data/sft.py
+++ b/src/axolotl/utils/data/sft.py
@@ -351,6 +351,10 @@ def _load_raw_datasets(
         if cfg.sample_packing:
             dataset, _ = process_datasets_for_packing(cfg, dataset, None)
 
+        # Deduplicate before saving so the saved dataset is already de-duplicated
+        if cfg.dataset_exact_deduplication:
+            dataset, _ = deduplicate_and_log_datasets(dataset=dataset)
+
         # Save the prepared dataset
         dataset_hash = generate_dataset_hash_from_config(
             cfg, datasets_configs, tokenizer.name_or_path
@@ -438,25 +442,16 @@ def _handle_train_dataset_split(
         )
         return train_dataset, eval_dataset
 
-    # No validation split - apply deduplication if needed and return as train dataset
-    if cfg.dataset_exact_deduplication:
-        train_dataset, _ = deduplicate_and_log_datasets(dataset=dataset)
-    else:
-        train_dataset = dataset
-
-    return train_dataset, None
+    # No validation split - deduplication already applied during preprocessing
+    return dataset, None
 
 
 def _handle_test_dataset_split(
     dataset: Dataset, cfg: DictDefault
 ) -> tuple[None, Dataset | None]:
     """Handle processing for test split."""
-    if cfg.dataset_exact_deduplication:
-        eval_dataset, _ = deduplicate_and_log_datasets(dataset=dataset)
-    else:
-        eval_dataset = dataset
-
-    return None, eval_dataset
+    # Deduplication already applied during preprocessing
+    return None, dataset
 
 
 def _apply_dataset_sharding(dataset: Dataset, cfg: DictDefault) -> Dataset:
--- a/src/axolotl/utils/data/sft.py
+++ b/src/axolotl/utils/data/sft.py
@@ -343,13 +343,6 @@ def _load_raw_datasets(
     # Merge datasets
     dataset = merge_datasets(datasets, cfg)
 
-    if cfg.skip_prepare_dataset and cfg.dataset_exact_deduplication:
-        LOG.warning(
-            "skip_prepare_dataset is enabled together with "
-            "dataset_exact_deduplication. Deduplication will be skipped "
-            "because dataset preparation is being skipped."
-        )
-
     if not cfg.skip_prepare_dataset and not streaming:
         if split == "test" and cfg.eval_sequence_len:
             dataset = handle_long_seq_in_dataset(dataset, cfg.eval_sequence_len, cfg)
--- a/src/axolotl/utils/data/sft.py
+++ b/src/axolotl/utils/data/sft.py
@@ -343,6 +343,13 @@ def _load_raw_datasets(
     # Merge datasets
     dataset = merge_datasets(datasets, cfg)
 
+    if cfg.skip_prepare_dataset and cfg.dataset_exact_deduplication:
+        LOG.warning(
+            "skip_prepare_dataset is enabled together with "
+            "dataset_exact_deduplication. Deduplication will be skipped "
+            "because dataset preparation is being skipped."
+        )
+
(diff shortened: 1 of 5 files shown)


ved1beta (approved):
good catch , nice tests

winglian (commented):
since the saved dataset contents now depends on deduplicaion, we should add that to the hash string in the `generate_dataset_hash_from_config` function

ved1beta:
The skip_prepare_dataset=True + dataset_exact_deduplication=True combination silently drops dedup. Since dedup runs as part of the  
  prepare pipeline, skipping the pipeline skips dedup too. Could you add a `ValueError` in the config validator (alongside              
  check_deduplication_with_streaming) that raises when both flags are set, with a message explaining why and pointing the user to     
  unset one of them? Something like:         
```                                                                                         
  dataset_exact_deduplication=True has no effect when skip_prepare_dataset=True.
  Deduplication runs as part of the prepare pipeline, which is skipped.                                                               
  Either set skip_prepare_dataset: false or disable dataset_exact_deduplication.
```

ManasVardhan:
Thanks for the feedback @winglian! I have added `dataset_exact_deduplication` to the hash string in `generate_dataset_hash_from_config` so that toggling dedup correctly invalidates the cached dataset.

Also addressed @ved1beta point about `skip_prepare_dataset=True` + `dataset_exact_deduplication=True` silently dropping dedup. Both SFT and RL paths now log a warning when that combination is detected.

ved1beta:
make sure ci is passing !

ManasVardhan:
Thanks for the reviews! Pushed fixes addressing all feedback:

**@winglian** - `dataset_exact_deduplication` was already added to the hash string in `generate_dataset_hash_from_config` (previous commit). Confirmed it is in the config string so toggling dedup correctly invalidates cached datasets.

**@ved1beta** - Replaced the runtime warning with a proper `ValueError` in the config validator (`check_deduplication_with_skip_prepare`). It now raises early when both `skip_prepare_dataset=True` and `dataset_exact_deduplication=True` are set, with a message explaining why and how to fix it.

**CI fixes:**
- Fixed the RL test (`test_dedup_called_before_save_rl`): was patching `axolotl.loaders.load_tokenizer` instead of `axolotl.utils.data.rl.load_tokenizer`, so the mock was not intercepting the imported reference and real code ran with a None config
- Fixed ruff lint (removed unused `call` and `pytest` imports) and formatting (line length)

src/axolotl/utils/data/sft.py
```diff
@@ -438,25 +442,16 @@ def _handle_train_dataset_split(
...
-        train_dataset, _ = deduplicate_and_log_datasets(dataset=dataset)
-    else:
-        train_dataset = dataset
-
-    return train_dataset, None
+    # No validation split - deduplication already applied during preprocessing
+    return dataset, None
 
 
 def _handle_test_dataset_split(
```
xzuyn: this always returns `None, dataset`. there is no use keeping this as a function

src/axolotl/utils/data/sft.py
xzuyn: this can just be replaced with `train_dataset, eval_dataset = None, dataset`

ManasVardhan:
Addressed @xzuyn's feedback: removed `_handle_test_dataset_split` and inlined `train_dataset, eval_dataset = None, dataset` directly. Thanks for the review!

winglian (approved):

PR #3433, opened 2026-02-25
Replace optimizer parameter with self.optimizer to match changes in Transformers Trainer

zinccat:
Match the update in https://github.com/huggingface/transformers/commit/1744f8f3fc19dfb812887afbe5651eb49995bed5 that 
https://github.com/huggingface/transformers/blame/b71de73468429eb02da18caa50e9b5200400a4ed/src/transformers/trainer.py#L1602
does not pass in optimizer arg anymore

# Description

Changes all `optimizer` to `self.optimizer` in `create_scheduler`

## Motivation and Context

Without the change, previous approach result in optimizer as None when creating scheduler, the change fixes the issue

## How has this been tested?

Yes, on a personal workload training model with cosine scheduler

## AI Usage Disclaimer

No

## Summary by CodeRabbit

* **Refactor**
  * Optimized learning rate scheduler initialization to use internal optimizer reference instead of parameter dependency, improving consistency in scheduler construction across all scheduler types. Public API and behavior remain unchanged.

### DIFF
--- a/src/axolotl/core/trainers/mixins/scheduler.py
+++ b/src/axolotl/core/trainers/mixins/scheduler.py
@@ -51,7 +51,7 @@ class SchedulerMixin(Trainer):
             plugin_manager = PluginManager.get_instance()
             lr_scheduler: LRScheduler | None = plugin_manager.create_lr_scheduler(
                 trainer=self,
-                optimizer=optimizer,
+                optimizer=self.optimizer,
                 num_training_steps=num_training_steps
             )
             if lr_scheduler is not None:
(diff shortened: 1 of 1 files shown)


zinccat:
the `optimizer` parameter is still kept to match the signature of the parent class

NanoCode012:
Thanks for bringing this to our attention. Seems like kallewoof also just made a PR on this too (linked above)

kallewoof:
This looks wrong to me. It should use `optimizer` if it is set, otherwise it should fall back to `self.optimizer`. See #3435.

NanoCode012:
I think we can close this PR in favor of the other as it closely follows upstream

PR #3437, opened 2026-02-26
fix: qwen3-next to use fla causal-conv1d to support packing

NanoCode012:
# Description

Since we ask users to uninstall triad's causal-conv1d, we've been fall backing to PyTorch's op, which doesn't handle cu seqlen. This PR fixes that as well as using a triton kernel for slightly more optimized performance.

Thanks to `morphism` for the report https://discord.com/channels/1104757954588196865/1104757955204743201/1476527723512987730

Context:

```
Hmm, though I notice that seq_idx is still None for the causal_conv1d part? Is that correct?

Ah, might be a tiny leak then?  Could maybe use fla.modules.convolution.causal_conv1d with cu_seqlens instead?
```

## Breaking change

Fails hard on packing + no FLA

## How has this been tested?

- [x] To manual run

## Release Notes

* **Documentation**
  * Simplified installation and setup instructions for clearer onboarding
  * Updated VRAM usage guidance with specific memory ranges for different configurations

* **Configuration Updates**
  * Enhanced training parameters and optimization settings
  * Added new configuration flags for improved performance and memory efficiency

  * Optimized computation handling for better resource utilization

### DIFF
--- a/examples/qwen3-next/README.md
+++ b/examples/qwen3-next/README.md
@@ -6,30 +6,13 @@ This guide shows how to fine-tune it with Axolotl with multi-turn conversations 
 
 ## Getting started
 
-1. Install Axolotl following the [installation guide](https://docs.axolotl.ai/docs/installation.html). You need to install from main as Qwen3-Next is only on nightly or use our latest [Docker images](https://docs.axolotl.ai/docs/docker.html).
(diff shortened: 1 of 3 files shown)


NanoCode012:
image

Depends on #3439

winglian (approved):

PR #3455, opened 2026-03-05
include number of params and rounded est of params so we can easily group in posthog

winglian:
# Description

Since we redact the model names when they aren't in the allowlist, we don't have a good understanding of the model sizes being trained. This PR calculates the number of parameters as well as handles some pre-rounding so we can quickly analyze and group.

## Release Notes

* **Chores**
  * Enhanced telemetry data collection to more accurately capture and report model properties and parameter estimates during training setup.


winglian:
Screenshot 2026-03-05 at 12 29 12 PM

shows up in posthog ^

PR #3460, opened 2026-03-05
Datasets dynamic work queue

Nero10578:
# Description

Update tokenizing to use a seperate work queue instead of relying on datasets.map

## Motivation and Context

The usual datasets.map method of tokenizing assigns fixed sets of examples per thread, which means that when threads finishes their set they just sit idle. The fix is to assign examples to tokenize to each thread as they finish working on them instead of assigning a fixed large set of examples to each thread.

## How has this been tested?

I merged it to the latest main code and it worked fine. It is isolated changes so it doesn't break anything. Checked with preprocess --debug and the tokenized output is the same.

Tested to go through tokenizing way faster than the current method. Especially apparent when there are very some long examples.

## AI Usage Disclaimer

Yes I used kilocode in vscode plugged into GLM 4.7 to help with this. 

## Social Handles (Optional)

@OwenArli in discord from arliai.com

## Summary by CodeRabbit

* **New Features**
  * Implemented parallel dataset tokenization leveraging multi-core CPU processing for improved performance across training pipelines.
  * Added support for simultaneous processing of multiple datasets with automatic load balancing across worker processes.
  * Enhanced dataset handling with comprehensive error recovery and per-dataset validation during the tokenization process.


src/axolotl/utils/data/sft.py
```diff
@@ -327,18 +327,108 @@ def _load_raw_datasets(
...
+                dataset = dataset.shuffle(seed=cfg.seed).shard(
+                    num_shards=dataset_config.shards, index=shards_idx
+                )
+
+            # Get the dataset strategy without processing
+            from axolotl.prompt_strategies import load
+
+            dataset_strategy = load(
+                dataset_config.type, tokenizer, cfg, dataset_config, processor=processor
+            )
```
NanoCode012, line 377: We should maybe have note for cases where this is skipped
```bash
# claude
- _is_dataset_already_tokenized check (already-tokenized datasets get re-processed)            
-  Custom dataset type handling (isinstance(dataset_config.type, DictDefault))                                                                                       
- cfg.skip_prepare_dataset check    # vlm                                                                                                                                        
- Bradley-Terry datasets                                                                                                                                                  
- Stepwise supervised datasets                                                                                                                                             - DATASET_HANDLERS dispatch (alpaca, oasst, etc.)
```

Is this right? Let me know if I misread this part.

src/axolotl/datasets_work_queue.py
```diff
@@ -0,0 +1,409 @@
...
+        self,
+        prompt_tokenizer: PromptTokenizingStrategy,
+        dataset: Dataset,
+        process_count: int | None = None,
+        keep_in_memory: bool | None = False,
+        **kwargs,
+    ):
+        self.prompt_tokenizer = prompt_tokenizer
+        self.process_count = process_count or mp.cpu_count()
+        self.keep_in_memory = keep_in_memory
```
NanoCode012, line 41: Not sure this is used anywhere

src/axolotl/utils/data/sft.py
```diff
@@ -327,18 +327,108 @@ def _load_raw_datasets(
...
+
+            if not dataset_strategy:
+                # Fallback to standard processing for this dataset
+                dataset_wrapper, dataset_prompter = _load_and_process_single_dataset(
```
NanoCode012, line 381: Just leaving this here. This could lead to cases where it's double `load` (here and in `load(` above) but not necessarily wrong. May be cleaner this way.

src/axolotl/utils/data/sft.py
```diff
@@ -327,18 +327,108 @@ def _load_raw_datasets(
...
+        strategies = []
+
+        for dataset_config in datasets_with_name_generator(datasets_configs):
+            # Load the raw dataset
+            dataset = load_dataset_with_config(
+                dataset_config, cfg.hf_use_auth_token, streaming=streaming
+            )
+
+            # Parse dataset type
+            d_base_type, d_prompt_style = _parse_dataset_type(dataset_config.type)
```
NanoCode012, line 351: I think this is unused

ved1beta (approved):
LGTM

PR #3466, opened 2026-03-06
install flash-linear-attention

winglian:
# Description

## Summary by CodeRabbit

* **Chores**
  * Added two new library dependencies to support enhanced functionality.
  * Updated installation configuration for ARM64 systems to exclude additional packages during setup.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -20,6 +20,9 @@ trl==0.29.0
 hf_xet==1.3.2
 kernels==0.12.2
 
+fla-core==0.4.1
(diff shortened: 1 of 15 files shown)


requirements.txt
```diff
@@ -20,6 +20,9 @@ trl==0.29.0
 hf_xet==1.3.2
 kernels==0.12.2
 
+fla-core==0.4.1
+flash-linear-attention==0.4.1
```
NanoCode012, line 24: Do we need the former?

PR #3469, opened 2026-03-06
fix: add guard for _initialize_missing_keys patch

NanoCode012:
# Description

Small followup for https://github.com/axolotl-ai-cloud/axolotl/pull/3464 adds a guard to prevent re-patch and a recursive `patch_fn(patch_fn(original_fn))` scenario

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where a system patch could be applied multiple times during repeated invocations, which could cause unintended side effects. The patch now applies only once, ensuring consistent and stable behavior across multiple operations.


winglian (approved):
thanks!

PR #3473, opened 2026-03-06
automatically enable tf32 if supported

winglian:
# Description

Most folks don't need full ieee fp32 precision, so enable automaticallly tf32 if the gpu supports it

## Summary by CodeRabbit

* **New Features**
  * Introduced CUDA TF32 capability detection with automatic configuration that intelligently enables or disables optimization based on hardware support.
  * TF32 now defaults to "auto" mode for seamless hardware-aware optimization.

* **Tests**
  * Updated test suite to validate TF32 functionality and configuration.

### DIFF
--- a/src/axolotl/utils/schemas/config.py
+++ b/src/axolotl/utils/schemas/config.py
@@ -1222,15 +1222,12 @@ class AxolotlConfigWCapabilities(AxolotlInputConfig):
 
     @model_validator(mode="after")
     def check_tf32(self):
-        if self.capabilities.tf32:
-            if self.tf32 is None or self.tf32 == "auto":
-                self.tf32 = True
-                LOG.info(
-                    "tf32 support detected, enabling tf32 automatically for this configuration."
-                )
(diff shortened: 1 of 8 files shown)


src/axolotl/utils/schemas/config.py
```diff
@@ -1218,6 +1220,19 @@ def check_bf16(self):
...
+        if self.capabilities.tf32:
+            if self.tf32 is None or self.tf32 == "auto":
+                self.tf32 = True
+                LOG.info(
+                    "tf32 support detected, enabling tf32 automatically for this configuration."
+                )
+        elif self.tf32 is None or self.tf32 == "auto":
+            self.tf32 = False
+            LOG.info("tf32 support not found, disabling tf32 for this configuration.")
+        return self
```
NanoCode012, line 1227: This condition feels weird to read at. Should it just be, if tf32 is auto, then cfg.tf32 = self.capabilities.tf32 ?

NanoCode012 (approved):

PR #3475, opened 2026-03-07
consolidate behavioud of routing in scattermoe kernels

winglian:
# Description

match behaviour of sonicmoe of softmax and sigmoid routing in scattermoe
capture the scattermoe autotuning best kernel metrics so we can prune them better later

## AI Usage Disclaimer
Claude

## Summary by CodeRabbit

* **Refactor**
  * Internal kernel code restructured with centralized helper functions for routing logic and expert computations, reducing code complexity while supporting multiple routing schemes and LoRA configurations.

* **Tests**
  * Comprehensive test coverage added for mixture-of-experts kernels, including multiple routing strategies (softmax and sigmoid variants), shared expert scenarios, and reference implementations for validation against actual performance.

### DIFF
--- a/src/axolotl/integrations/kernels/libs/scattermoe_lora/layers.py
+++ b/src/axolotl/integrations/kernels/libs/scattermoe_lora/layers.py
@@ -277,7 +277,10 @@ def _sigmoid_topk_route(
     e_score_correction_bias = getattr(base_gate, "e_score_correction_bias", None)
     if e_score_correction_bias is None:
         e_score_correction_bias = getattr(moe_block, "e_score_correction_bias", None)
-    scores_for_choice = router_probs + e_score_correction_bias
+    if e_score_correction_bias is not None:
+        scores_for_choice = router_probs + e_score_correction_bias
+    else:
+        scores_for_choice = router_probs
 
     # Group-based selection: pick top groups, mask the rest
     n_group = getattr(moe_block, "n_group", 1)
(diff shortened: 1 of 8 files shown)


src/axolotl/integrations/kernels/libs/scattermoe_lora/layers.py
```diff
@@ -220,6 +220,158 @@ def _unwrap_experts_lora(experts_module):
...
+    num_experts = base_gate.num_experts
+    routing_weights, selected_experts = torch.topk(routing_weights, top_k, dim=-1)
+
+    if getattr(base_gate, "norm_topk_prob", True):
+        routing_weights = routing_weights / routing_weights.sum(dim=-1, keepdim=True)
+
+    return routing_weights, selected_experts, top_k, num_experts
+
+
+def _sigmoid_topk_route(
```
NanoCode012, line 251: Not sure if we should duplicate code into this repo too? Should we not use a constant utils as done in the kernels folder?

tests/integrations/test_scattermoe_lora.py
```diff
@@ -321,3 +323,389 @@ def test_scatter_moe_gated_mlp_docstring_no_router_logits(self):
...
+
+
+# ============================================================================
+# 7. Routing strategy detection and sigmoid routing
+# ============================================================================
+
+
+def _make_softmax_gate(E=4, H=16, K=2):
+    """Create a mock softmax-style gate (Qwen/OLMoE)."""
+    return SimpleNamespace(
```
NanoCode012, line 335: I saw this class before as a suggestion. Does it really properly test what we are using in prod?

src/axolotl/integrations/kernels/libs/scattermoe_lora/layers.py
winglian: i think it's fine, this version supports passing in the lora weights for fused ops

Issue #3465, opened 2026-03-06
very high eval loss with gemma3 with sample packing on and eval packing off


sageof6path:
### Please check that this issue hasn't been reported before.

- [x] I searched previous [Bug Reports](https://github.com/axolotl-ai-cloud/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

Eval loss should be identical if eval sample packing is enabled vs disabled

### Current behaviour

Eval loss is very high when eval sample packing is disabled but when we enable eval sample packing it is coming low and the gap between them is very high.

### Steps to reproduce

You need to run the config provided below.

1) Run the config with sample packing enabled but eval packing is disabled. you will get high eval loss similar to the logs in https://github.com/sageof6path/axolotl/blob/main/examples/gemma3/high_eval_loss_logs.txt

2) Next, keep sample packing enabled but disable eval sample packing. you should get low eval loss. similar to logs in https://github.com/sageof6path/axolotl/blob/main/examples/gemma3/low_eval_loss.txt

### Config yaml

```yaml
base_model: google/gemma-3-1b-it

# gemma3 doesn't seem to play nice with ddp
ddp_find_unused_parameters: true
experimental_skip_move_to_device: true  # prevent OOM by NOT putting model to GPU before sharding

plugins:
- axolotl.integrations.cut_cross_entropy.CutCrossEntropyPlugin
- axolotl.integrations.liger.LigerPlugin
liger_rope: true
liger_rms_norm: true
liger_glu_activation: true
liger_layer_norm: true
liger_fused_linear_cross_entropy: true

chat_template: gemma3
eot_tokens:
- <end_of_turn>

dataset_prepared_path: last_run_prepared
output_dir: output

datasets:
  - path: philschmid/dolly-15k-oai-style
    type: chat_template
    split: train[:90%]
    chat_template: gemma3
    field_messages: messages

test_datasets:
  - path: philschmid/dolly-15k-oai-style
    type: chat_template
    split: train[90%:]
    chat_template: gemma3
    field_messages: messages

flash_attention: true

sequence_len: 8192
sample_packing: true
eval_sample_packing: false

use_kernels: true

micro_batch_size: 2
gradient_accumulation_steps: 1
num_epochs: 3
optimizer: adamw_torch_fused
learning_rate: 2e-4

lr_scheduler: cosine

bf16: true
tf32: true
logging_steps: 1

gradient_checkpointing: true
activation_offloading: true

[...]

PR #3478, opened 2026-03-08
fix: high eval loss w/ sample packing 

ved1beta:
# Description

process_datasets_for_packing() is called for ALL splits (both "train" and "test") whenever cfg.sample_packing is True
should only be done for train 

## Motivation and Context
#3465

## AI Usage Disclaimer
asked too claude to just investigate

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed sample packing behavior during test evaluation to respect configuration settings. When eval_sample_packing is disabled, sample packing is now properly skipped during test runs, reducing unnecessary data processing and improving evaluation efficiency while maintaining backward compatibility.

### DIFF
--- a/src/axolotl/utils/data/sft.py
+++ b/src/axolotl/utils/data/sft.py
@@ -348,7 +348,9 @@ def _load_raw_datasets(
             dataset = handle_long_seq_in_dataset(dataset, cfg.eval_sequence_len, cfg)
         else:
             dataset = handle_long_seq_in_dataset(dataset, cfg.sequence_len, cfg)
-        if cfg.sample_packing:
+        if cfg.sample_packing and not (
+            split == "test" and cfg.eval_sample_packing is False
+        ):
             dataset, _ = process_datasets_for_packing(cfg, dataset, None)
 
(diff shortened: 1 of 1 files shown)


src/axolotl/utils/data/sft.py
```diff
@@ -348,7 +348,9 @@ def _load_raw_datasets(
             dataset = handle_long_seq_in_dataset(dataset, cfg.eval_sequence_len, cfg)
         else:
             dataset = handle_long_seq_in_dataset(dataset, cfg.sequence_len, cfg)
-        if cfg.sample_packing:
+        if cfg.sample_packing and not (
+            split == "test" and cfg.eval_sample_packing is False
+        ):
```
NanoCode012: Should we (just thinking here) update this condition a bit better?

It may be easier to read if
```
if (split=='train' and cfg.sample_packing) or (split=='test' and cfg.eval_sample_packing)
```

this wouldn't require the double negs?

(feel free to clean up the formatting etc)

ved1beta: done !!

NanoCode012 (approved):

PR #3497, opened 2026-03-13
feat: merge adapter in fp32

NanoCode012:
# Description

A user mentioned that an adapter merged into base model lost knowledge of the SFT, while just being loaded ontop worked.

This enables `safe_merge` but waiting for clarification if this solves the OP's issue.

Limitation: merges in fp32, 2x more mem than before, so we should lock behind a config probably.


NanoCode012:
Could be wrong. Read def incorrectly

PR #3498, opened 2026-03-15
Fix double sequence partition during training with context-parallel

lorenzbaraldi:
# Description
This PR fixes an issue caused by double context partitioning when both Accelerate native Context Parallelism (CP) and the SequenceParallelContextManager are applied simultaneously.

## Motivation and Context

During training with context parallelism enabled, the token sequence was unintentionally split by a factor of 1 / cp_size².

This happened because:
	•	SequenceParallelContextManager already partitions the sequence by 1 / cp_size.
	•	At the same time, Accelerate applies additional context partitioning through maybe_context_parallel.

As a result, the sequence was partitioned twice, leading to an incorrect effective sequence length.

This patch prevents the double partitioning and ensures the sequence is split only once as intended.

## How has this been tested?

The fix was tested using a CP configuration with 8 GPUs.

Testing consisted of debugging the apply_sequence_parallelism function with and without the patch. Without the fix, the training loss was consistently higher than the evaluation loss, indicating incorrect training behavior. After applying the patch, the losses behaved as expected.

## AI Usage Disclaimer
Yes — Opus was used to assist with debugging.

## Types of changes

Bug fix

## Summary by CodeRabbit

* **Refactor**
  * Updated context parallel initialization handling in the accelerate integration for parallelism configuration.

### DIFF
--- a/src/axolotl/monkeypatch/accelerate/parallelism_config.py
+++ b/src/axolotl/monkeypatch/accelerate/parallelism_config.py
@@ -78,30 +78,21 @@ def patch_parallelism_config():
 
 
 def patch_prepare_cp():
-    import functools
+    import contextlib
 
-    import torch
     from accelerate import Accelerator
 
     def patched_prepare_cp(self, *args):
         if self.parallelism_config.cp_backend == "deepspeed":
             return args
 
-        from accelerate.big_modeling import _attach_context_parallel_hooks
-        from torch.distributed.tensor.experimental import context_parallel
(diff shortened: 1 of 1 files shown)


winglian (approved):

NanoCode012:
Thanks for PR! As a followup PR, I think we can also no-op the `_prepare_context_parallel_inputs` since we're bypassing transformers and remove unneeded bypass in `trainer_context_parallel`.

PR #3499, opened 2026-03-16
qwen docs + new config

ved1beta:
examples + docs 
image

## Release Notes

* **New Features**
  * Added new fine-tuning configuration examples for Qwen3.5 models (27B text-only and 9B multimodal)

* **Documentation**
  * Updated README with fine-tuning configurations, peak VRAM requirements, and new training commands
  * Expanded getting started section with full fine-tuning guidance and memory considerations

PR #3509, opened 2026-03-18
Update 7b-lora-vision.yaml

SicariusSicariiStuff:
9b, not 7b

# Description

## Summary by CodeRabbit

* **Chores**
  * Updated the base model configuration in the LoRA training example to use Qwen 3.5 9B instead of Qwen 3.5 7B.


NanoCode012:
Thanks, fixed in #3499

PR #3512, opened 2026-03-19
support offloading layers to CPU

winglian:
# Description

Similar to Zero-3 offloading, but seems to work better with 4bit. With qwen3.5 35B w qlora, uses 26GB reserved vs ~29GB reserved w/o offloading

## Summary by CodeRabbit

* **New Features**
  * Added layer offloading capability for GPU memory optimization during model training, enabling CPU offloading of model parameters with automatic prefetching during backward passes.
  * Added support for Qwen3.5 MoE models.
  * Improved device mapping configuration for quantized MoE expert training.

* **Configuration**
  * Added `layer_offloading` training argument to enable/disable layer offloading functionality.


NanoCode012:
Could we have more docs on this? I'm not sure how a user could use it and whether it's limited to single gpu or also fsdp/deepspeed/ddp.

I also saw some possibly unrelated changes for quantize moe experts and device_map.

PR #3516, opened 2026-03-19
Nemo gym integration

winglian:
## Release Notes

* **New Features**
  * Added NeMo Gym integration for reinforcement learning training with customizable reward environments and multi-turn support
  * Added OpenAI-compatible API endpoints (`/v1/models`, `/v1/chat/completions`) to vLLM server

* **Bug Fixes**
  * Fixed GRPO configuration handling to prevent invalid parameters in standard (non-async) training mode
  * Improved LoRA adapter synchronization with automatic fallback mechanism
  * Fixed entropy computation in fused decoding scenarios

* **Documentation**
  * Added comprehensive NeMo Gym integration guide with setup instructions and example configurations


NanoCode012 (commented):
Is this work an addition ontop of async GRPO?

I'm not so clear on the arch, is it a replacement for vllm generation?

Also, how does the uv packages interact with axolotl, if we were to use pip?

winglian:
> Is this work an addition ontop of async GRPO?

Yes, has some minor changes to work with Nemo Gym

> I'm not so clear on the arch, is it a replacement for vllm generation?

Not a replacement, it still uses vllm under the hood, but also has better multi-turn support

> Also, how does the uv packages interact with axolotl, if we were to use pip?

separate iirc. It mostly uses the venv with uv for the gym service

src/axolotl/core/trainers/grpo/async_trainer.py
```diff
@@ -2413,6 +2430,9 @@ def _get_per_token_logps_and_entropies(
                         logits, completion_ids, self.temperature
                     )
                     all_logps.append(logps)
+                    # Liger fused path doesn't compute entropy — append zeros
+                    if compute_entropy:
+                        all_entropies.append(torch.zeros_like(logps))
```
NanoCode012, line 2508: Should we: raise NotImplementedError here?

src/axolotl/integrations/nemo_gym/plugin.py
```diff
@@ -0,0 +1,492 @@
...
+                LOG.warning(
+                    "vLLM server does not expose /v1/load_lora_adapter. "
+                    "Set VLLM_ALLOW_RUNTIME_LORA_UPDATING=1 when starting vLLM, e.g.:\n"
+                    "  VLLM_ALLOW_RUNTIME_LORA_UPDATING=1 python -m vllm.entrypoints.openai.api_server "
+                    "--enable-lora --max-lora-rank 64 ..."
+                )
+        except Exception:
+            pass  # Server might not be up yet, sync will warn later
+
+    def _setup_lora_sync(self, trainer):
```
NanoCode012, line 394: Is this duplicate of `async_trainer.py._sync_lora_adapter`? Should we extract to shared util?

src/axolotl/integrations/nemo_gym/data_producer.py
```diff
@@ -0,0 +1,226 @@
...
+                prompt_text = (
+                    prompt[-1].get("content", "")
+                    if isinstance(prompt[-1], dict)
+                    else str(prompt[-1])
+                )
+            elif isinstance(prompt, str):
+                prompt_text = prompt
+
+            # Find the full dataset item, preserving agent_ref for routing
+            full_item = self._dataset_lookup.get(prompt_text, {})
```
NanoCode012, line 100: Is using prompt content safe way to hash? What if two same user prompt but diff assistant prompt? I'm also not sure if this includes multi-turn convos

src/axolotl/integrations/nemo_gym/data_producer.py
```diff
@@ -0,0 +1,226 @@
...
+                    dataset_items=expanded_items,
+                    agent_servers=self._agent_servers,
+                    timeout=self._request_timeout,
+                    max_completion_length=trainer.max_completion_length,
+                    temperature=trainer.temperature,
+                    top_p=getattr(trainer, "top_p", None) or 0.999,
+                )
+            )
+        finally:
+            loop.close()
```
NanoCode012, line 133: Is it expensive to create+re-create this event loop on each call? Should we persist and create during `init` and re-use?

src/axolotl/integrations/nemo_gym/rewards.py
```diff
@@ -0,0 +1,267 @@
...
+_verify_urls: dict[str, str] = {}
+
+
+def _get_verify_urls(head_port: int = 11000) -> dict[str, str]:
+    """Discover verify endpoints from the NeMo Gym head server.
+
+    Results are cached so that the HTTP round-trip only happens once per
+    process.
+    """
+    global _verify_urls
```
NanoCode012, line 74: Is storing this var state, in global the best way? Could there be chance of race condition?

Do we need to add a lock when editing?

Potential scenario by Claude 
```
When
  async_prefetch: true, the data producer runs in
   a background thread while the main training
  thread also calls reward functions. Two threads
   can both see _verify_urls as empty, both make
  the discovery HTTP call, and both write to the
  dict concurrently.
```

src/axolotl/core/trainers/grpo/async_trainer.py
winglian: this is correct, otherwise the liger grpo kernel wouldn't work. this happens upstream in trl too

src/axolotl/integrations/nemo_gym/data_producer.py
winglian: this is fast compared to the http call/rollouts.

src/axolotl/integrations/nemo_gym/rewards.py
winglian: should never race, but adding a thread lock

src/axolotl/integrations/nemo_gym/data_producer.py
winglian: NeMo Gym datasets have one user message per row; the server randomizes the task.

PR #3520, opened 2026-03-20
fix: make prepare_context_parallel_inputs no-op

NanoCode012:
# Description

Followup to #3498 

As we don't depend on transformers to call CP partition and use ring FA, we could bypass this call and remove a patch.

@lorenzbaraldi what do you think of this change?

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved context parallelism handling to prevent double input partitioning and enhance FlashAttention compatibility.

* **Tests**
  * Removed outdated context parallelism patch tests.


lorenzbaraldi:
Yes, this makes sense to me

PR #3537, opened 2026-03-23
provide early warning to users when dependencies are missing for a feature they may be using

kallewoof:
This attempts to at least warn users attempting to use packing when they haven't installed the flash-linear-attention package, to eliminate time waste.

# Description

Adds a simple warning log message when the import fails for `causal_conv1d`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for model configurations. The application now displays a warning message instead of crashing when required dependencies are missing, allowing users to identify and resolve configuration issues more easily.


kallewoof:
The rabbit is right that this might be noisy if people don't want the flash-linear-attention stuff (is there such a group of people?). Closing for now.

Issue #3303, opened 2025-12-05
Changing `output_dir` when using `added_tokens_overrides` changes dataset hash


xzuyn:
### Please check that this issue hasn't been reported before.

- [x] I searched previous [Bug Reports](https://github.com/axolotl-ai-cloud/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

Changing `output_dir` when using `added_token_overrides` should not change the dataset hash, as nothing about tokenization changes.

### Current behaviour

Using `added_token_overrides` causes the dataset hash to change when you change `output_dir`, meaning you either have to re-tokenize or rename the dataset cache to use the new hash.

---

For example, I process a dataset with this config;
```yaml
output_dir: ./run1
added_tokens_overrides:
  6: "<|unclean_text|>"
  7: "<|clean_text|>"
```
That results in a dataset hash of `abcd1234` (example). Then I start a training run with the config, and it looks for and loads `abcd1234`. Everything works as expected.

After that run I decide to do another run, so I edit the config;
```yaml
output_dir: ./run2
added_tokens_overrides:
  6: "<|unclean_text|>"
  7: "<|clean_text|>"
```
Then I start a training run with the config, resulting in "Unable to find prepared dataset" as it tries looking for hash `xyz555`.

---

Doing this without `added_tokens_overrides` lets you change `output_dir` however you want and the dataset hash remains the same, letting you train without needing to re-tokenize.

### Steps to reproduce

1. Pre-process any config with `added_tokens_overrides` and `output_dir`.
2. Note the dataset hash.
3. Edit `output_dir` in the config.
4. Pre-process the config.
5. Note the different hash.

### Config yaml

```yaml

```

### Which Operating Systems are you using?

- [x] Linux
- [ ] macOS
- [ ] Windows

### Python Version

3.12.3

### axolotl branch-commit

main/b234532

### Acknowledgements

- [x] My issue title is concise, descriptive, and in title casing.
- [x] I have searched the existing issues to make sure this bug has not been reported yet.
- [x] I am using the latest version of axolotl.
- [x] I have provided enough information for the maintainers to reproduce and diagnose the issue.

NanoCode012:
[...]

PR #3538, opened 2026-03-23
fix: DPO tool role KeyError (#3217), dataset hash output_dir (#3303), config validators

Edward-Zion-Saji:
## Summary

Three independent bug fixes with tests. Each is self-contained and can be reviewed/merged independently.

---

### Fix 1: DPO `tool` role raises `KeyError` (#3217)

**File:** `src/axolotl/prompt_strategies/dpo/chat_template.py`

The default `role_map_inv` in both `default()` and `argilla_chat()` was missing the `"tool"` role. Any DPO dataset whose `messages` list contained a `{"role": "tool", ...}` turn (tool-calling preference data) crashed with `KeyError: 'tool'`.

**Fix:** Add `"tool": ["tool"]` to the default mapping in both functions. Fully backwards-compatible — user-supplied `roles:` config still overrides the default.

---

### Fix 2: Dataset cache invalidated when `output_dir` changes with `added_tokens_overrides` (#3303)

**File:** `src/axolotl/utils/data/shared.py`

When `added_tokens_overrides` is set, Axolotl saves the modified tokenizer into `output_dir`, making `tokenizer.name_or_path` an absolute path like `/tmp/my_run/modified_tokenizer`. This path landed in the dataset fingerprint string, so renaming the output directory busted the cache and forced a full re-tokenization even though nothing had actually changed.

**Fix:** When `added_tokens_overrides` is present, derive the tokenizer fingerprint from `cfg.tokenizer_config` (the canonical HF model name) plus the sorted overrides dict — content that is stable across output directory renames. Without `added_tokens_overrides` the old behaviour is preserved unchanged.

---

### Fix 3: Three new Pydantic config validators

**File:** `src/axolotl/utils/schemas/config.py`

Added three `@model_validator(mode="before")` methods to `AxolotlConfigWCapabilities` that catch common misconfigurations at validation time rather than deep inside training:

| Validator | What it catches |
|---|---|
| `check_save_strategy_best_requires_metric` | `save_strategy: best` without `metric_for_best_model` — crashes inside HF Trainer otherwise |
| `check_streaming_with_val_set_size` | `streaming: true` + `val_set_size > 0` — documented as unsupported but previously not enforced |
| `check_lora_target_modules_regex` | Invalid Python regex patterns in `lora_target_modules` list — only discovered during tokenisation without this |

---

## Tests

| File | What is tested |
|---|---|
| `tests/prompt_strategies/test_dpo_chat_templates.py` | `TestDPOChatTemplateToolRole` — no KeyError on tool role; custom role mapping still respected |
| `tests/utils/data/test_hash.py` | Hash stability, regression for #3303, different overrides produce different hashes |
| `tests/utils/schemas/validation/test_config_validators.py` | Pass/fail paths for all three validators |

## Summary by CodeRabbit

* **New Features**
  * Added configuration validation to prevent invalid parameter combinations (e.g., requiring metrics for best-model checkpoints, preventing conflicting streaming settings).
  * Added support for tool messages in DPO chat template handling.

* **Bug Fixes**
  * Fixed dataset hash computation to remain stable when using token overrides, regardless of output directory changes.

* **Tests**
  * Added comprehensive validation tests for configuration parameters and DPO chat template functionality.

### DIFF
--- a/src/axolotl/utils/data/shared.py
+++ b/src/axolotl/utils/data/shared.py
@@ -516,11 +516,8 @@ def generate_dataset_hash_from_config(
     Returns:
         MD5 hash string representing the configuration.
     """
-    # When added_tokens_overrides is set the tokenizer is saved into output_dir,
-    # so tokenizer.name_or_path becomes an absolute path that includes output_dir.
-    # Changing output_dir would bust the dataset cache even though the tokenizer
-    # is identical. Use the canonical tokenizer config path plus the overrides
-    # content instead so the hash is stable across output_dir changes.
+    # When added_tokens_overrides is set, tokenizer.name_or_path contains output_dir.
+    # Use the canonical tokenizer config + overrides content so the hash is stable across output_dir changes.
     if cfg.get("added_tokens_overrides"):
         tokenizer_fingerprint = (
             f"{cfg.tokenizer_config}+overrides:"
--- a/src/axolotl/utils/data/shared.py
+++ b/src/axolotl/utils/data/shared.py
@@ -519,12 +519,8 @@ def generate_dataset_hash_from_config(
     # When added_tokens_overrides is set, tokenizer.name_or_path contains output_dir.
     # Use the canonical tokenizer config + overrides content so the hash is stable across output_dir changes.
     if cfg.get("added_tokens_overrides"):
-        tokenizer_fingerprint = (
-            f"{cfg.tokenizer_config}+overrides:"
-            + ",".join(
-                f"{k}={v}"
-                for k, v in sorted(cfg.added_tokens_overrides.items())
-            )
+        tokenizer_fingerprint = f"{cfg.tokenizer_config}+overrides:" + ",".join(
+            f"{k}={v}" for k, v in sorted(cfg.added_tokens_overrides.items())
         )
     else:
         tokenizer_fingerprint = tokenizer_name
--- a/src/axolotl/utils/data/shared.py
+++ b/src/axolotl/utils/data/shared.py
@@ -516,12 +516,28 @@ def generate_dataset_hash_from_config(
     Returns:
         MD5 hash string representing the configuration.
     """
+    # When added_tokens_overrides is set the tokenizer is saved into output_dir,
+    # so tokenizer.name_or_path becomes an absolute path that includes output_dir.
+    # Changing output_dir would bust the dataset cache even though the tokenizer
+    # is identical. Use the canonical tokenizer config path plus the overrides
+    # content instead so the hash is stable across output_dir changes.
+    if cfg.get("added_tokens_overrides"):
+        tokenizer_fingerprint = (
+            f"{cfg.tokenizer_config}+overrides:"
+            + ",".join(
+                f"{k}={v}"
+                for k, v in sorted(cfg.added_tokens_overrides.items())
+            )
+        )
+    else:
+        tokenizer_fingerprint = tokenizer_name
+
     config_str = (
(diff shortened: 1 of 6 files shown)


src/axolotl/utils/data/shared.py
```diff
@@ -516,12 +516,28 @@ def generate_dataset_hash_from_config(
     Returns:
         MD5 hash string representing the configuration.
     """
+    # When added_tokens_overrides is set the tokenizer is saved into output_dir,
+    # so tokenizer.name_or_path becomes an absolute path that includes output_dir.
+    # Changing output_dir would bust the dataset cache even though the tokenizer
+    # is identical. Use the canonical tokenizer config path plus the overrides
+    # content instead so the hash is stable across output_dir changes.
```
NanoCode012: Let's simplify these comment to one liner or two if possible.

tests/utils/data/test_hash.py
```diff
@@ -0,0 +1,123 @@
...
+        cfg = _base_cfg()
+        h1 = generate_dataset_hash_from_config(cfg, _datasets(), "NousResearch/Llama-3.2-1B")
+        h2 = generate_dataset_hash_from_config(cfg, _datasets(), "NousResearch/Llama-3.2-1B")
+        assert h1 == h2
+
+    def test_different_tokenizer_different_hash(self):
+        """A different tokenizer path produces a different hash."""
+        cfg = _base_cfg()
+        h1 = generate_dataset_hash_from_config(cfg, _datasets(), "NousResearch/Llama-3.2-1B")
+        h2 = generate_dataset_hash_from_config(cfg, _datasets(), "mistralai/Mistral-7B-v0.1")
```
NanoCode012: Maybe I missed it, but we don't pre-download this mistral model. Let's use another that we already re-use in another test to prevent downloading more model data in CI

NanoCode012 (commented):
Hey, thanks for this PR. The second one is a good catch, I couldn't figure it out why at the time.

For first issue, I'm a bit hesitant. Can you show me existing several HF DPO datasets using this role? Is it not `role: "tool_calling"` or `tool_call`? We don't want to create a new standard if possible

Edward-Zion-Saji:
Thanks for the review, @NanoCode012! Addressing all three points:

**Fix 1 — `"tool"` role justification**

`"tool"` is the [HuggingFace-standardised role name](https://huggingface.co/blog/unified-tool-use) for tool-response turns, adopted by `transformers` chat templates and the TRL trainers. Here are real HF datasets using it:

- **`trl-internal-testing/toolcall`** (official TRL test dataset, includes a `preference` DPO subset) — messages contain `{"role": "tool", ...}` turns directly: https://huggingface.co/datasets/trl-internal-testing/toolcall
- **`Nanbeige/ToolMind`** (369 k rows, conversations format) — each multi-turn trace uses `{"role": "tool", ...}` for tool responses: https://huggingface.co/datasets/Nanbeige/ToolMind
- **TRL DPO Trainer docs** explicitly document `"tool"` role messages in DPO preference data: https://huggingface.co/docs/trl/dpo_trainer

So `"tool"` is the established standard, not a new one — it mirrors the OpenAI chat format that HF adopted.

**Fix 2 — condensed comment in `shared.py`** ✅ (see latest commit)

**Fix 3 — replaced `mistralai/Mistral-7B-v0.1` with `HuggingFaceTB/SmolLM2-135M`** ✅ (see latest commit)

NanoCode012:
Thanks for clarifying the first point, letting CI run

NanoCode012:
@Edward-Zion-Saji CI failing from your new tests I think

PR #3545, opened 2026-03-25
feat: move to uv first

NanoCode012:
# Description

- [x] Generate lock file on GPU
- [x] Update docs

Also fixes some nightly, see issues https://github.com/axolotl-ai-cloud/axolotl/pull/3545#issuecomment-4153946726

## Release Notes

* **New Features**
  * Added `uv`-based installation workflow for simplified dependency management; new Docker image variants with `-uv` tag now available.

* **Documentation**
  * Updated installation guides to prioritize `uv` as the recommended package manager; added migration guide from pip to `uv`; revised Docker image documentation with `-uv` variant details.

* **Chores**
  * Upgraded minimum PyTorch requirement to 2.9.0; restructured project dependency declarations and build configuration.


NanoCode012:
2 CI failures on nightly build upstream which this PR attempts to fix:
- validate rope strictly fails loading `AutoTokenizer.from_pretrained("microsoft/Phi-3-medium-128k-instruct")` https://github.com/huggingface/transformers/pull/41250
- loralinear signature change which impacts our lora efficient merge test https://github.com/huggingface/peft/pull/2960

pyproject.toml
```diff
@@ -67,5 +214,44 @@ markers = [
...
+        { package = "axolotl" },
+        { extra = "llmcompressor" },
+    ],
+]
+
 [tool.uv.extra-build-dependencies]
 axolotl = ["huggingface_hub"]
+mamba-ssm = ["torch"]
+causal-conv1d = ["torch"]
+flash-attn = ["torch"]
```
BrownianNotion: Is it worth setting `match_runtime = true` to ensure packages like flash-attn/deepspeed use the same version of torch https://docs.astral.sh/uv/concepts/projects/config/#augmenting-build-dependencies?

This seems to be the preferred approach over using `--no-build-isolation`

NanoCode012: Thank you for the suggestion, I will look into this

NanoCode012: @BrownianNotion , I read through this, is this applicable only for `uv sync`? I'm deleting the lockfile and usually recommending folks to do `uv pip install..` which may not be impacted by this?

It's tricky with lock file as we would need to maintain it updated per requirements change and it doesn't really work with N torch x cuda combinations. If anyone has better recommendation on how to handle this, we're all ears.

Edit: I'll add this anyways but I think it could be a no-op with no build isolation. I'll see if we still need that flag later.

BrownianNotion: Thanks for looking into it! I think `match_runtime` is indeed for `uv sync` only. 

Fair enough, I don't know of a nice way to support N torch x cuda combinations with the .lock file, you might have to enumerate all the combinations and their indices for example.

pyproject.toml
```diff
@@ -67,5 +214,44 @@ markers = [
...
+    ],
+    [
+        { package = "axolotl" },
+        { extra = "fbgemm-gpu" },
+    ],
+    [
+        { package = "axolotl" },
+        { extra = "llmcompressor" },
+    ],
+]
```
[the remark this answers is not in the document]
NanoCode012: @BrownianNotion , are you familiar whether this is an issue?

winglian (approved):

PR #3552, opened 2026-03-26
support flattening/packing for GRPO

winglian:
Screenshot 2026-03-25 at 9 53 04 PM
Pink is baseline w/o packing. This is about a 10% speedup over padded sequences for this particular dataset, and this is over the async prefetch case too. 

## Summary by CodeRabbit

* **New Features**
  * Configurable worker extension classes for vLLM serving
  * Support for token ID lists in addition to string prompts during vLLM serving
  * New batch flattening optimization option for GRPO training to improve performance

* **Bug Fixes**
  * Improved error handling for vLLM batch update endpoint with fallback to individual requests
  * Added error checking in vLLM worker outputs before response flattening

* **Refactor**
  * Lazy initialization of prompt data loaders in GRPO training for better resource management
  * Updated scoring computation with padding-free optimization option


src/axolotl/core/trainers/grpo/async_trainer.py
```diff
@@ -2428,6 +2510,114 @@ def _prepare_inputs_legacy_async(self, generation_batch):
...
+                start = cu_seqlens[i].item() + plen - 1
+                start = max(0, start)
+                actual = min(n_compl, total_tokens - 1 - start)
+                if actual > 0:
+                    all_logps[chunk_start + i, :actual] = flat_logps[
+                        start : start + actual
+                    ]
+
+            del logits, flat_logits, flat_logps, flat_ids
+            torch.cuda.empty_cache()
```
NanoCode012, line 2618: Do we need to empty each run?

src/axolotl/core/trainers/grpo/async_trainer.py
```diff
@@ -2428,6 +2510,114 @@ def _prepare_inputs_legacy_async(self, generation_batch):
...
+
+            with autocast_ctx:
+                logits = model(
+                    input_ids=flat_ids,
+                    position_ids=flat_pos,
+                    use_cache=False,
+                    cu_seq_lens_q=cu_seqlens,
+                    cu_seq_lens_k=cu_seqlens,
+                    max_length_q=seq_lens.max().item(),
+                    max_length_k=seq_lens.max().item(),
```
NanoCode012, line 2578: This part, I'm unfamiliar. Does all model API support passing these params?

NanoCode012 (commented):
Is the 10% improvement directly from this change and not from variance?

src/axolotl/core/trainers/grpo/async_trainer.py
```diff
@@ -1619,15 +1648,25 @@ def _compute_deferred_scores(self, rollout: dict) -> dict:
...
+                    old_per_token_logps = self._get_per_token_logps_flattened(
+                        self.model,
+                        prompt_completion_ids,
+                        attention_mask,
+                        logits_to_keep,
+                        batch_size=logprob_batch_size,
+                        prompt_mask=prompt_mask,
+                    )
+                else:
+                    old_per_token_logps, _ = self._get_per_token_logps_and_entropies(
```
NanoCode012, line 1662: Looks like Liger isn't compat against packing too? Liger uses this old path

src/axolotl/core/trainers/grpo/async_trainer.py
winglian: yeah, helps with memory pressure

src/axolotl/core/trainers/grpo/async_trainer.py
winglian: yeah, if you look at the transformers function, `prepare_fa_kwargs_from_position_ids()`, it basically returns these, so should apply to all models

src/axolotl/core/trainers/grpo/async_trainer.py
winglian: yeah, we handle liger already I think and don't use it for the fwd+bwd, but we can use it for the scoring path (fwd only)

PR #3508, opened 2026-03-18
super nemo support

ved1beta:
# super nemo support 

https://huggingface.co/nvidia/NVIDIA-Nemotron-3-Super-120B-A12B-BF16

no sample packing T_T'

`examples/nemotron-h/nano-30b-a3b-qlora.yaml`
max_vram :69Gbs 
```
', 'grad_norm': '0.2281', 'learning_rate': '1.452e-05', 'ppl': '2.731', 'memory/max_active (GiB)': '68.22', 'memory/max_allocated (GiB)': '68.22', 'memory/device_reserved (GiB)': '77.14', 'tokens/train_per_sec_per_gpu': '189', 'tokens/total': 2623744, 'tokens/trainable': 704234, 'epoch': '0.02908'}
{'loss': '0.9717', 'grad_norm': '0.2569', 'learning_rate': '1.454e-05', 'ppl': '2.642', 'memory/max_active (GiB)': '69.89', 'memory/max_allocated (GiB)': '69.89', 'memory/device_reserved (GiB)': '77.14', 'tokens/train_per_sec_per_gpu': '145.2', 'tokens/total': 2628352, 'tokens/trainable': 705280, 'epoch': '0.02913'}
{'loss': '0.6611', 'grad_norm': '0.2496', 'learning_rate': '1.457e-05', 'ppl': '1.937', 'memory/max_active (GiB)': '67.66', 'memory/max_allocated (GiB)': '67.66', 'memory/device_reserved (GiB)': '77.14', 'tokens/train_per_sec_per_gpu': '48.39', 'tokens/total': 2633856, 'tokens/trainable': 706148, 'epoch': '0.02917'}
{'loss': '0.8849', 'grad_norm': '0.4', 'learning_rate': '1.459e-05', 'ppl': '2.423', 'memory/max_active (GiB)': '67.95', 'memory/max_allocated (GiB)': '67.95', 'memory/device_reserved (GiB)': '77.14', 'tokens/train_per_sec_per_gpu': '42.57', 'tokens/total': 2637184, 'tokens/trainable': 707015, 'epoch': '0.02922'}
[...]

PR #3554, opened 2026-03-27
qwen3.5  configs

ved1beta:
change lora_target_modules:
   - gate_up_proj
   - down_proj

## Release Notes

* **Documentation**
  * Updated configuration examples to clarify targeting options for shared and routed expert modules in mixture-of-experts models.
  * Added guidance on using alternative configuration parameters for different expert types.

### DIFF
--- a/examples/qwen3.5/122b-a10b-moe-qlora-fsdp.yaml
+++ b/examples/qwen3.5/122b-a10b-moe-qlora-fsdp.yaml
@@ -31,10 +31,11 @@ lora_target_modules:
   - k_proj
   - v_proj
   - o_proj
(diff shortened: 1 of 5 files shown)


NanoCode012:
Could you just test one of these configs to ensure step by step that the below are correct?
1. leaving them all commented: low trainable params
2. target experts: more trainable params
3. target experts + shared experts: even more params

ved1beta:
tested on nemo  super branch #3508  
image

NanoCode012 (commented):
Could you also update L60 in the README on `Shared Experts` to point to this new case?

NanoCode012 (approved):

PR #3555, opened 2026-03-27
Add precompute_ref_log_probs to config schema

### LINKED ISSUE #3546: DPO `precompute_ref_log_probs` missing from config

### Expected Behavior

For DPO, TRL supports the option `precompute_ref_log_probs` to precompute the reference model's log probs and save memory during training. Axolotl should also support this flag by plumbing it from the config to the trainer.

### Current behaviour

In Axolotl `precompute_ref_log_probs` isn't present in the master config `src/axolotl/utils/schemas/config.py::AxolotlInputConfig`, even though the logic to handle the arg exists in the rl builder https://github.com/axolotl-ai-cloud/axolotl/blob/main/src/axolotl/core/builders/rl.py#L245-L248. This means that even if that flag is set in the `.yml` file, it is dropped on the final step of `src/axolotl/utils/config/__init__.py::validate_config` since `AxolotlInputConfig` doesn't contain the field.

### Steps to reproduce

Specify `precompute_ref_log_probs: true` in any DPO .yml config.

### Possible solution

From my understanding it should be a simple fix - just add the field to the config. The existing logic should correctly plumb it through to the DPOTrainer. But let me know if there are potential side effects/other things to consider. Happy to try a PR.

- [x] Linux
- [ ] macOS
- [ ] Windows

### Python Version

3.12

### axolotl branch-commit

main/b55706b9

### Acknowledgements

joaquinhuigomez:
Fixes #3546

The RL builder reads `cfg.precompute_ref_log_probs` to configure DPO training, but the field is missing from `AxolotlInputConfig`. Since the schema uses Pydantic validation, setting this field in a config file has no effect — it gets silently dropped.

Added `precompute_ref_log_probs: bool | None = None` alongside the existing DPO fields.

## Summary by CodeRabbit

* **New Features**
  * Added a new optional configuration parameter for precomputing reference log probabilities.

### DIFF
--- a/src/axolotl/core/builders/rl.py
+++ b/src/axolotl/core/builders/rl.py
@@ -242,10 +242,6 @@ class HFRLTrainerBuilder(TrainerBuilderBase):
             and self.cfg.rl not in (RLType.GRPO, RLType.ORPO, RLType.EBFT)
         ):
             trainer_kwargs["peft_config"] = self.peft_config
-        if self.cfg.precompute_ref_log_probs is not None:
-            trainer_kwargs["precompute_ref_log_probs"] = (
-                self.cfg.precompute_ref_log_probs
-            )
 
         trainer_cls, trainer_cls_args = self._get_trainer_cls(trainer_kwargs)
 
--- a/src/axolotl/core/trainers/dpo/__init__.py
+++ b/src/axolotl/core/trainers/dpo/__init__.py
@@ -34,4 +34,8 @@ class DPOStrategy:
             training_args_kwargs["dpo_norm_loss"] = cfg.dpo_norm_loss
         if cfg.dpo_use_liger_kernel is not None:
             training_args_kwargs["use_liger_kernel"] = cfg.dpo_use_liger_kernel
+        if cfg.precompute_ref_log_probs is not None:
+            training_args_kwargs["precompute_ref_log_probs"] = (
+                cfg.precompute_ref_log_probs
+            )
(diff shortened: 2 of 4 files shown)


joaquinhuigomez:
Just checking in — this is a small schema addition (one field) to fix #3546. Happy to make any changes if needed.

winglian (approved):

joaquinhuigomez:
Thanks for the approval @winglian! Let me know if anything else is needed to get this merged.

joaquinhuigomez:
All checks are green — anything else needed to merge?

BrownianNotion:
@joaquinhuigomez thanks for the PR, just checking did you test this works and the config change was enough?

BrownianNotion:
Was thinking it'd be good to add a test in `tests/core/test_builders.py`. Specifically, add `precompute_ref_log_probs: True` here https://github.com/axolotl-ai-cloud/axolotl/blob/main/tests/core/test_builders.py#L99, then assert it's true here https://github.com/axolotl-ai-cloud/axolotl/blob/main/tests/core/test_builders.py#L300. Thoughts @winglian?

NanoCode012 (approved):
Thanks for catching this. I just updated it a bit to add description

joaquinhuigomez:
Added the test. Also moved precompute_ref_log_probs to DPOStrategy.set_training_args_kwargs to match the pattern of use_weighting and other DPO config fields.

PR #3556, opened 2026-03-28
train only on last assistant message

VarunGumma:
# Description
A flag to enable loss computation and training only on the last assistant turn of a conversation. 

## Motivation and Context

- In vanilla multi-turn training, the loss over all the assistant tokens is calculated. But in case of some special cases (maybe in-context examples in a chat format), the only loss over the last assistant turn needs to be calculated, and the previous turns just demonstrate the pattern, and we want to take them as fixed inputs and not compute the loss. 

## How has this been tested?

- A basic test has been added to `tests/prompt_strategies/test_chat_templates.py`

- No special environment, no additional code changes.

## AI Usage Disclaimer

- Yes, GitHub CoPilot assisted. I asked it to make the patch and manually tested it. 

NA

## Types of changes

- Added a `train_on_last_assistant_only` flag, and it goes along with `train_on_inputs: false`
- Once this flag is detected, we just loop through the messages, and set the weight of the assistant turns to 0 expect the last one. ([diff](https://github.com/axolotl-ai-cloud/axolotl/compare/main...VarunGumma:axolotl:main#diff-84365377db47760a349b310d6fb8e831fde45479ceaccb0beca877b2f2831872))

## Social Handles (Optional)
- GitHub: https://github.com/VarunGumma
- X: https://x.com/VarunGumma23

## Release Notes

* **New Features**
  * Added `train_on_last_assistant_only` configuration parameter to enable selective training on only the final assistant message in multi-turn conversations, while disabling training on prior assistant messages.

* **Tests**
  * Added test coverage validating the new `train_on_last_assistant_only` feature behavior.

### DIFF
--- a/src/axolotl/core/datasets/transforms/chat_builder.py
+++ b/src/axolotl/core/datasets/transforms/chat_builder.py
@@ -8,6 +8,7 @@ from typing import Any, Mapping
 
 def chat_message_transform_builder(
     train_on_inputs=False,
+    train_on_last_assistant_only=False,
     conversations_field: str = "messages",
     message_field_role: str | list[str] | None = None,  # commonly "role"
     message_field_content: str | list[str] | None = None,  # commonly "content"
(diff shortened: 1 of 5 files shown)


VarunGumma:
@winglian do take a look at this!

NanoCode012:
@VarunGumma , we do support this. Please see this https://docs.axolotl.ai/docs/dataset-formats/conversation.html#training-on-last-message

It's kind of legacy, but I don't think we removed it yet. If this solves your case, could you please close the PR?

PR #3558, opened 2026-03-30
feat: add moe kernel support for non-glu

NanoCode012:
# Description

This is required for NemotronH and can be useful for any future models.

Sonicmoe has internal handling for non-glu, we just needed to patch our end to pass it properly.
Scattermoe required patching to handle this.

## Summary by CodeRabbit

* **New Features**
  * Added support for the nemotron_h model type with optimized Mixture of Experts routing
  * Extended expert layer handling to support both GLU and non-GLU expert architectures
  * Enhanced MoE integration with optional latent projection layers for improved computation paths

### DIFF
--- a/src/axolotl/integrations/kernels/constants.py
+++ b/src/axolotl/integrations/kernels/constants.py
@@ -36,6 +36,8 @@ SPARSE_MOE_BLOCK = {
     "glm4v_moe": "Glm4vMoeTextMoE",
     # sigmoid -> topk routing (no group selection)
     "minimax_m2": "MiniMaxM2SparseMoeBlock",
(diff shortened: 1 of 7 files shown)


ved1beta:
tried running on #3508 with scatter working fine :saluting_face:

winglian:
Did we run any correctness tests/checks for the scattermoe changes?

ved1beta:
the training runs  look similar i thnk that should be enough 
image

winglian:
we have some tests in `tests/integrations/test_scattermoe_lora*.py` and tests/e2e/integrations/test_scattermoe_lora_kernels.py`  that we should include the new functionality

PR #3559, opened 2026-03-30
Add troubleshooting note for GLM4 GGUF MTP mismatch

mariozupan:
Adds a troubleshooting note for a common issue when converting GLM-4 / GLM4-MoE models to GGUF.

If `num_nextn_predict_layers > 0` is present in config but MTP weights are not included (e.g. after PEFT/QLoRA), llama.cpp fails to load with missing tensor errors.

Solution: set `num_nextn_predict_layers = 0` before conversion.

This helps users avoid a non-obvious failure during llama.cpp loading.

## Summary by CodeRabbit

* **Documentation**
  * Added troubleshooting guidance for GGUF model-loading failures reporting missing tensor errors. Provides configuration solutions to resolve tensor alignment mismatches in GLM-4 and GLM4-MoE models.


NanoCode012 (approved):
Thanks

Issue #3565, opened 2026-03-31
Support DPO `loss_type` and `loss_weights`.


BrownianNotion:
### ⚠️ Please check that this feature request hasn't been suggested before.

- [x] I searched previous [Ideas in Discussions](https://github.com/axolotl-ai-cloud/axolotl/discussions/categories/ideas) didn't find any similar feature requests.
- [x] I searched previous [Issues](https://github.com/axolotl-ai-cloud/axolotl/labels/enhancement) didn't find any similar feature requests.

### 🔖 Feature description

As of TRL v0.29.0, the `loss_type` parameter now only takes a list of strings. https://huggingface.co/docs/trl/v0.29.1/en/dpo_trainer#multi-loss-combinations 

As a result, previously supported loss functions are now unavailable as Axolotl does not expose either of these two parameters. For example:

1. RPO, previously configured in both TRL and axolotl by `rpo_alpha` has now been deprecated and is instead configured by passing `loss_type=["sigmoid", "sft"]` and `loss_weights=[1.0, alpha]`. https://github.com/huggingface/trl/blob/main/docs/source/paper_index.md#iterative-reasoning-preference-optimization
2. Length-normalised DPO loss, used in the OLMo/TULU models. 

See https://github.com/axolotl-ai-cloud/axolotl/pull/3560 for more context.

### ✔️ Solution

1. Expose and support `loss_type` and `loss_weights` as config parameters (perhaps with names `dpo_loss_type` and `dpo_loss_weights` so they don't conflict with the string type of the experimental preference trainers), plumb them through to the `DPOTrainer` via `DPOStrategy`. 
Not every `loss_type` needs to be supported, but the previously supported loss functions (eg. RPO/length-normalised DPO) should made available.
2. Consider deprecating `IPO` as a separate `RLType` as it is just DPO with a separate loss function.

Happy to work on this

### 📝 Additional Context

Was difficult to say whether this is an enhancement or bug, as part of the change ensures we retain previous functionality with the new TRL update.

### Acknowledgements

- [x] My issue title is concise, descriptive, and in title casing.
- [x] I have searched the existing issues to make sure this feature has not been requested yet.
- [x] I have provided enough information for the maintainers to understand and evaluate this request.

BrownianNotion:
Draft PR here, let me know if any high-level design changes or if we want to remove IPO and use the DPO loss type instead: https://github.com/axolotl-ai-cloud/axolotl/pull/3566

[...]

Issue #3548, opened 2026-03-25
DPO `dpo_norm_loss` no longer works on `trl==0.29.0`


BrownianNotion:
### Please check that this issue hasn't been reported before.

- [x] I searched previous [Bug Reports](https://github.com/axolotl-ai-cloud/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

Length-normalised DPO normalises the chosen/rejected log probs by their sequence lengths, and is used in the OLMo model family for preference training. The objective function is given below 
<img width="645" height="73" alt="Image" src="https://github.com/user-attachments/assets/370fe92b-1e11-4a3c-bcfd-803efb8eb49b" />
**Source:** https://arxiv.org/pdf/2411.15124

### Current behaviour

This loss type no longer works in Axolotl with `trl==0.29.0`. My understanding is that this is due to the refactoring of the loss computation in `0.29.0` where:
1. [`concatenated_forward`](https://github.com/huggingface/trl/blob/v0.28-release/trl/trainer/dpo_trainer.py#L1517), which used to perform both the concatenation of chosen/rejected responses and the loss computation was removed. 
2. The concatenation is now performed in the data collator https://github.com/huggingface/trl/blob/main/trl/trainer/dpo_trainer.py#L175
3. The loss computation is performed separately in `_compute_loss` (or the liger version if using liger). https://github.com/huggingface/trl/blob/main/trl/trainer/dpo_trainer.py#L1147

In Axolotl, the `concatenated_forward` still exists here https://github.com/axolotl-ai-cloud/axolotl/blob/main/src/axolotl/core/trainers/dpo/trainer.py#L98 but is not called anywhere (and would raise an AttributeError if it was). Instead, the loss computation goes through TRL, which does not support length-normalised DPO. https://github.com/huggingface/trl/issues/2964

### Steps to reproduce

Add `dpo_norm_loss: true` to any DPO config.

### Possible solution

My suggestion:
1. Remove deprecated `concatenated_forward` from `AxolotlDPOTrainer`
2. Let TRL handle length-normalisation (I will hopefully implement a PR soon for this)
3. If TRL implements length-normalisation as a separate loss_type, remove the `dpo_norm_loss` config flag and support the new loss type.

### Which Operating Systems are you using?

- [x] Linux
- [ ] macOS
- [ ] Windows

### Python Version

3.12

### axolotl branch-commit

main/b55706b9

### Acknowledgements

[...]

PR #3566, opened 2026-03-31
DPO support loss types

BrownianNotion:
# Description
Support multiple loss types and weights as part of TRL >= v0.29.0 updates. #3565 

**Key changes:**
1. Adds two new config parameters `dpo_loss_type` and `dpo_loss_weights` which are passed to Axolotl's DPO Trainer via `DPOStrategy`. Note that `dpo_loss_type` expects a list of strings.
2. Partially deprecates `RLType.IPO/rl: ipo` in favour of passing `rl: dpo, dpo_loss_type: ["ipo"]`. Warns users of upcoming deprecation.

## Motivation and Context

This PR exposes the full list of loss functions available in TRL and the ability to combine multiple dpo losses with custom weightings. 

This will also restore losses previously available in axolotl such as RPO (and hopefully [length-normalised DPO](https://github.com/huggingface/trl/pull/5406) once TRL supports this), but are now broken on TRL >= 0.29 due to refactors/parameter deprecation [see #3548  #3560]. This PR aims to update Axolotl to work with the newer TRL version and restore this functionality. 

This PR also begins cleaning up `RLType.IPO`, which is no longer needed as a separate RL type.

## How has this been tested?
Unit tests.

## AI Usage Disclaimer
Claude for light review, updates/fixes written by me.

## Types of changes
Config additions, new losses for DPO, partial deprecation of `RLType.IPO`.

## Summary by CodeRabbit

* **New Features**
  * Added configurable DPO loss types and weights via `dpo_loss_type` and `dpo_loss_weights` parameters for flexible training customization.

* **Deprecation**
  * The `rl: ipo` configuration is now deprecated; use `rl: dpo` with `dpo_loss_type: ["ipo"]` instead.

* **Documentation**
  * Updated training method guidance to reflect the new DPO configuration approach and IPO deprecation notice.

### DIFF
--- a/src/axolotl/core/trainers/dpo/__init__.py
+++ b/src/axolotl/core/trainers/dpo/__init__.py
@@ -20,14 +20,15 @@ class DPOStrategy:
     @classmethod
     def set_training_args_kwargs(cls, cfg):
         training_args_kwargs = {}
-        if cfg.rl is RLType.IPO:
-            training_args_kwargs["loss_type"] = ["ipo"]
-        else:
+        if cfg.rl is RLType.DPO:
             if cfg.dpo_loss_type is not None:
                 training_args_kwargs["loss_type"] = cfg.dpo_loss_type
 
             if cfg.dpo_loss_weights is not None:
                 training_args_kwargs["loss_weights"] = cfg.dpo_loss_weights
+
+        if cfg.rl is RLType.IPO:
+            training_args_kwargs["loss_type"] = ["ipo"]
 
         # Label smoothing is not compatible with IPO
         if cfg.rl is RLType.DPO and cfg.dpo_label_smoothing:
--- a/src/axolotl/core/trainers/dpo/__init__.py
+++ b/src/axolotl/core/trainers/dpo/__init__.py
@@ -22,6 +22,13 @@ class DPOStrategy:
         training_args_kwargs = {}
         if cfg.rl is RLType.IPO:
             training_args_kwargs["loss_type"] = ["ipo"]
+        else:
+            if cfg.dpo_loss_type is not None:
+                training_args_kwargs["loss_type"] = cfg.dpo_loss_type
+
+            if cfg.dpo_loss_weights is not None:
+                training_args_kwargs["loss_weights"] = cfg.dpo_loss_weights
+
         # Label smoothing is not compatible with IPO
         if cfg.rl is RLType.DPO and cfg.dpo_label_smoothing:
             training_args_kwargs["label_smoothing"] = cfg.dpo_label_smoothing
--- a/src/axolotl/utils/schemas/validation.py
+++ b/src/axolotl/utils/schemas/validation.py
@@ -776,16 +776,20 @@ class RLValidationMixin:
             )
 
         if rl == "dpo":
+            if dpo_loss_weights is not None and dpo_loss_type is None:
+                raise ValueError(
+                    "`dpo_loss_weights` requires `dpo_loss_type` to be set"
+                )
             if (
-                dpo_loss_type
-                and dpo_loss_weights
+                dpo_loss_type is not None
+                and dpo_loss_weights is not None
                 and len(dpo_loss_type) != len(dpo_loss_weights)
             ):
                 raise ValueError(
-                    f"`dpo_loss_type` and `dpo_dpo_loss_weights` must be the same length, "
+                    f"`dpo_loss_type` and `dpo_loss_weights` must be the same length, "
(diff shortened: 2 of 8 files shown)


BrownianNotion:
@NanoCode012 when you have some time, would love to get your thoughts on the current state of the changes :) In particular, are we happy to support all loss_types and loss_weights that TRL does? these should reasonably well-tested upstream, so hopefully there isn't too much risk in supporting them.

src/axolotl/core/trainers/dpo/__init__.py
```diff
@@ -22,6 +22,13 @@ def set_training_args_kwargs(cls, cfg):
         training_args_kwargs = {}
         if cfg.rl is RLType.IPO:
             training_args_kwargs["loss_type"] = ["ipo"]
+        else:
```
BrownianNotion: Probably better to have this as if `cfg.rl` is `RLType.DPO` rather than in the else, but will change it pending on what we decide to do with IPO

NanoCode012:
@BrownianNotion , what other types do you have in mind? or do you mean a passthrough like you've done?

BrownianNotion:
> @BrownianNotion , what other types do you have in mind? or do you mean a passthrough like you've done?

I mean passing through the `loss_type` from config to Axolotl's DPOTrainer. Based on my understanding, this wasn't previously exposed as a config parameter, or at least, it wasn't plumbed through (correct me if I'm wrong). So all loss types except `sigmoid` (default) and `ipo` (`RLType.IPO`) weren't available. https://huggingface.co/docs/trl/dpo_trainer#loss-types

src/axolotl/utils/schemas/validation.py
```diff
@@ -770,6 +770,20 @@ def check_gdpo(cls, data):
             )
         return data
 
+    @model_validator(mode="before")
+    @classmethod
+    def check_dpo(cls, data):
+        if data.get("rl") == "dpo":
```
NanoCode012: Can we add validation, if `dpo_loss_type` or the other is set, and not `rl: dpo`, we should also raise ValueError given it's dpo specific.

src/axolotl/utils/schemas/validation.py
```diff
@@ -770,6 +770,20 @@ def check_gdpo(cls, data):
...
+            loss_types = data.get("dpo_loss_type")
+            loss_weights = data.get("dpo_loss_weights")
+
+            if loss_types and loss_weights and len(loss_types) != len(loss_weights):
```
NanoCode012: Re: loss type validation mentioned by the other user on the Issue, does trl not validate that config upstream and error appropriately?

I'm a bit hesitant on hardcoding any constant/lists for loss types.

NanoCode012 (commented):
Sorry this took a while, just a bit of cleanup would do. I think this config naming is ok with me since it follows existing convention.

NanoCode012:
>  Remove IPO RLType and use DPO loss_type instead.

We can recommend people to use the new method instead on the ipo docs, while backwards compat re-routing to passing loss types if rl: ipo is passed.  I think this is a good transition.

src/axolotl/utils/schemas/validation.py
BrownianNotion: Your call - I figured we would want to error early rather than have the user wait for the runtime trl error. But I understand that this would have added maintenance costs given the hard coding (there isn't an enum we can just import from trl if I remember correctly)

src/axolotl/utils/schemas/validation.py
```diff
@@ -770,6 +771,42 @@ def check_gdpo(cls, data):
...
 
+    @model_validator(mode="before")
+    @classmethod
+    def check_dpo(cls, data):
+        dpo_loss_type = data.get("dpo_loss_type")
+        dpo_loss_weights = data.get("dpo_loss_weights")
+        rl = data.get("rl")
+
+        if rl == "ipo":
+            warnings.warn(
```
NanoCode012: Please use our logging.warn

NanoCode012 (commented):
Only minor nit left

src/axolotl/utils/schemas/validation.py
BrownianNotion: done :)

BrownianNotion: thanks for fixing this up, realised it wasn't done properly in hindsight. Will take more care next time

NanoCode012 (approved):

winglian:
rebased and fixed conflicts. Should be good to merge once tests pass

PR #3575, opened 2026-04-02
Skip redundant evaluation when resuming from checkpoint

### LINKED ISSUE #3305: Resuming triggers evaluation if `eval_steps` and `save_steps` line up

### Expected Behavior

If you are resuming from `checkpoint-20`, it should immediately start training step 21.

### Current behaviour

When resuming `checkpoint-20` it evaluates step 20 again, then starts training step 21.

### Steps to reproduce

1. Use a config which has `eval_steps` and `save_steps` set to values that overlap. For example `eval_steps: 10` + `save_steps: 10`.
2. Train until it saves a checkpoint. For example `checkpoint-20`.
3. Stop training.
4. Resume `checkpoint-20` using `resume_from_checkpoint`.
5. Observe that evaluation runs for step 20 again, then training for step 21 occurs.

### Config yaml

```yaml
save_steps: 10
eval_steps: 10
```

### Possible solution

I'm not sure about fixing it, but viewing in wandb this *looks* better, since it makes the lines from the original run connect with the resumed run.

*But* it shouldn't be done this way. It should probably just read *all* the info from the trainer_state and send that for the resumed step.

- [x] Linux
- [ ] macOS
- [ ] Windows

### Python Version

3.12.3

### axolotl branch-commit

main/b234532

### Acknowledgements

joaquinhuigomez:
Fixes #3305

When `eval_steps` and `save_steps` overlap, resuming from a checkpoint re-triggers evaluation for the checkpoint step because HuggingFace Trainer fires `_maybe_log_save_evaluate` before the first new training step. This adds a `SkipEvalOnResumeCallback` that records the resumed `global_step` at `on_train_begin` and suppresses `should_evaluate` for any step at or before that point. The callback is a no-op for fresh runs (resume step is 0).

## Release Notes

* **Bug Fixes**
  * Fixed unnecessary evaluation execution when resuming training from a checkpoint at evaluation step boundaries.

* **Tests**
  * Added test suite to validate evaluation behavior during checkpoint resume scenarios.

### DIFF
--- a/src/axolotl/core/builders/base.py
+++ b/src/axolotl/core/builders/base.py
@@ -119,7 +119,8 @@ class TrainerBuilderBase(abc.ABC):
             plugin_manager.add_callbacks_pre_trainer(cfg=self.cfg, model=self.model)
         )
 
-        callbacks.append(SkipEvalOnResumeCallback())
+        if self.cfg.resume_from_checkpoint:
+            callbacks.append(SkipEvalOnResumeCallback())
 
         if self.cfg.gc_steps:
             callbacks.append(GCCallback(gc_steps=self.cfg.gc_steps))
--- a/src/axolotl/core/builders/base.py
+++ b/src/axolotl/core/builders/base.py
@@ -41,6 +41,7 @@ from axolotl.utils.callbacks import (
     GCCallback,
     SaveAxolotlConfigtoWandBCallback,
     SaveModelOnFirstStepCallback,
+    SkipEvalOnResumeCallback,
 )
 from axolotl.utils.callbacks.profiler import PytorchProfilerCallback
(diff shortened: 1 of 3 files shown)


src/axolotl/core/builders/base.py
```diff
@@ -118,6 +119,8 @@ def get_callbacks(self) -> list[TrainerCallback]:
             plugin_manager.add_callbacks_pre_trainer(cfg=self.cfg, model=self.model)
         )
 
+        callbacks.append(SkipEvalOnResumeCallback())
```
winglian: Might be good to conditionally add this like `if self.cfg.resume_from_checkpoint:`

winglian (approved):

PR #3580, opened 2026-04-04
Fix DO_NOT_TRACK not being correctly handled

maximegmd:
# Description

DO_NOT_TRACK was not properly handled and was completely skipped.

## Motivation and Context

DO_NOT_TRACK should be accepted.

## Summary by CodeRabbit

* **Refactor**
  * Simplified telemetry opt-out environment variable handling. Telemetry is now disabled only when `AXOLOTL_DO_NOT_TRACK` or `DO_NOT_TRACK` is explicitly set to "1" or "true".


winglian (approved):
thanks for catching this @maximegmd. I went ahead and added some easy to read unit tests for this.

PR #3553, opened 2026-03-27
MX QAT patch

ved1beta:
# Description
use MXLinearConfig for mxfp4 PTQ, MXFakeQuantizeConfig only for QAT

## Motivation and Context
#3595
`AssertionError: Passing a generic Callable to `quantize_` is no longer recommended...`

## How has this been tested?
tested manually after repro 

## Release Notes

* **Refactor**
  * Updated MXFP4 quantization configuration framework to improve system stability and compatibility.
  * Enhanced error handling for quantization configuration registration to support additional error conditions.

* **Tests**
  * Updated MXFP4 quantization configuration tests to ensure validation accuracy with the new implementation.

Issue #3539, opened 2026-03-23
MXFP4 Post Training Quantization Causes Torchao AssertionError


jjbuschhoff:
### Please check that this issue hasn't been reported before.

- [x] I searched previous [Bug Reports](https://github.com/axolotl-ai-cloud/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

The trained model should be quantized to mxfp4.

### Current behaviour

The quantization fails with:

```
[2026-03-23 12:02:45,444] [INFO] [axolotl.cli.quantize.do_quantize:80] [PID:1081475] Quantizing model with configuration: 
	weight_dtype: TorchAOQuantDType.mxfp4
	activation_dtype: TorchAOQuantDType.mxfp4
	group_size: 32
	quantize_embedding: None
Traceback (most recent call last):
  File "/projects/training/.venv/bin/axolotl", line 10, in <module>
    sys.exit(main())
             ^^^^^^
  File "/projects/training/.venv/bin/lib/python3.12/site-packages/axolotl/cli/main.py", line 347, in main
    cli()
  File "/projects/training/.venv/bin/lib/python3.12/site-packages/click/core.py", line 1485, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/projects/training/.venv/bin/lib/python3.12/site-packages/click/core.py", line 1406, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/projects/training/.venv/bin/lib/python3.12/site-packages/click/core.py", line 1873, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/projects/training/.venv/bin/lib/python3.12/site-packages/click/core.py", line 1269, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/projects/training/.venv/bin/lib/python3.12/site-packages/click/core.py", line 824, in invoke
    return callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/projects/training/.venv/bin/lib/python3.12/site-packages/axolotl/cli/utils/args.py", line 48, in wrapper
    return func(*args, **filtered_kwargs)
[...]

PR #3581, opened 2026-04-04
fix un supported tensor for pTq

ved1beta:
#3539 

PR #3553  switched to MXDynamicActivationMXWeightConfig for PTQ added _attach_torchao_quantizer() for saving. 

However, _attach_torchao_quantizer sets up TorchAoHfQuantizer, which calls flatten_tensor_state_dict() during save — and that function doesn't support MXTensor, giving `ValueError: Unsupported tensor type: <class 'torchao.prototype.mx_formats.mx_tensor.MXTensor'>.`

tested with scripts present in issue 

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error handling when saving quantized models with automatic fallback recovery.

* **Improvements**
  * Improved MX format quantization configuration and serialization support.
  * Updated quantization-aware training configuration handling.

### DIFF
--- a/src/axolotl/cli/quantize.py
+++ b/src/axolotl/cli/quantize.py
@@ -93,10 +93,19 @@ def do_quantize(
         weight_dtype, activation_dtype, group_size
     )
 
-    LOG.info(f"Saving quantized model to: {str(Path(output_dir) / 'quantized')}.")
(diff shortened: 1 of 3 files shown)


NanoCode012 (approved):
Thanks you!

PR #3584, opened 2026-04-08
fix ddp/fsdp w gemma4

winglian:
## Release Notes

* **New Features**
  * Added configuration option to freeze multimodal modules during text-only training of multimodal models.
  * Enhanced Gemma4 model evaluation to properly handle multimodal token type identifiers.

* **Bug Fixes**
  * Fixed model type detection for Gemma4 in distributed training scenarios (DDP/FSDP).
  * Improved distributed training configuration auto-detection for Gemma4 models.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -38,6 +38,8 @@ Agent-specific references:
 - [docs/agents/grpo.md](docs/agents/grpo.md) — GRPO online RL with reward functions
 - [docs/agents/reward_modelling.md](docs/agents/reward_modelling.md) — outcome and process reward models
 - [docs/agents/pretraining.md](docs/agents/pretraining.md) — continual pretraining
+- [docs/agents/model_architectures.md](docs/agents/model_architectures.md) — model-specific quirks (Gemma4, Qwen3.5 MoE, etc.)
+- [docs/agents/new_model_support.md](docs/agents/new_model_support.md) — debugging and adding support for new model architectures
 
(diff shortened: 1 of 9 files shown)


NanoCode012:
Verified by two users here:
- https://discord.com/channels/1104757954588196865/1111279858136383509/1491552422219481098
- https://discord.com/channels/1104757954588196865/1111279858136383509/1491573756194193500

The former had an act offloading error that was solved by turning the config off:
```
RuntimeError: Expected to mark a variable ready only once. This error is caused by one of the following reasons: 1) Use of a module parameter outside the `forward` function. Please make sure model parameters are not shared across multiple concurrent forward-backward passes. or try to use _set_static_graph() as a workaround if this module graph does not change during training loop.2) Reused parameters in multiple reentrant backward passes. For example, if you use multiple `checkpoint` functions to wrap the same part of your model, it would result in the same set of parameters been used by different reentrant backward passes multiple times, and hence marking a variable ready multiple times. DDP does not support such use cases in default. You can try to use _set_static_graph() as a workaround if your module graph does not change over iterations.
[rank3]: Parameter at index 814 with name base_model.model.model.language_model.layers.59._checkpoint_wrapped_module.mlp.gate_proj.lora_A.default.weight has been marked as ready twice. This means that multiple autograd engine  hooks have fired for this particular parameter during this iteration.
```

TheLocalDrummer:
Confirmed working well for me

PR #3601, opened 2026-04-13
dpo collation/padding

winglian:
# Description

When using DPO/ORPO w Qwen 3.5, the flash linear attention autotunes when there are new sequence lengths. by padding to buckets, we can minimize the excessive memory for autotuning without having to pad everything out to max length.

## Release Notes

* **New Features**
  * Added configurable `pad_to_multiple_of` option to customize batch padding behavior during training
  * Enhanced support for DPO, ORPO, IPO, and SIMPO training modes with improved data collation

* **Improvements**
  * Optimized tensor operations for improved performance during training
  * Extended trainer configuration schema for better control over padding strategies

### DIFF
--- a/src/axolotl/core/builders/rl.py
+++ b/src/axolotl/core/builders/rl.py
@@ -228,8 +228,45 @@ class HFRLTrainerBuilder(TrainerBuilderBase):
 
         return training_args, trainer_kwargs
 
+    def build_collator(self, **kwargs):
+        """Build a data collator for preference-tuning trainers.
+
+        Returns None for RL types that provide their own collator (e.g. GRPO,
+        KTO), letting the trainer construct its default. For DPO/IPO/ORPO/SIMPO
+        returns an ``AxolotlDPODataCollatorWithPadding`` when
+        ``pad_to_multiple_of`` is set, otherwise None (so the trainer
+        falls back to the TRL default).
+        """
+        if self.cfg.rl not in (
+            RLType.DPO,
+            RLType.IPO,
+            RLType.ORPO,
+            RLType.SIMPO,
+        ):
+            return None
+
+        pad_to_multiple_of = getattr(self.cfg, "pad_to_multiple_of", None)
+        if not pad_to_multiple_of:
+            return None
+
+        from axolotl.utils.collators.dpo import AxolotlDPODataCollatorWithPadding
+
+        LOG.info(
+            f"Using AxolotlDPODataCollatorWithPadding with pad_to_multiple_of="
+            f"{pad_to_multiple_of}"
+        )
+        return AxolotlDPODataCollatorWithPadding(
+            pad_token_id=self.tokenizer.pad_token_id,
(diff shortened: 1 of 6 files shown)


src/axolotl/core/builders/rl.py
```diff
@@ -228,9 +228,46 @@ def _build_training_arguments(self, total_num_steps):
...
+            return None
+
+        from axolotl.utils.collators.dpo import AxolotlDPODataCollatorWithPadding
+
+        LOG.info(
+            f"Using AxolotlDPODataCollatorWithPadding with pad_to_multiple_of="
+            f"{pad_to_multiple_of}"
+        )
+        return AxolotlDPODataCollatorWithPadding(
+            pad_token_id=self.tokenizer.pad_token_id,
```
NanoCode012, line 260: Would this raise any issue given they migrated to self.processor/self.processing_class?

src/axolotl/utils/collators/dpo.py
```diff
@@ -0,0 +1,121 @@
...
+
+    pad_to_multiple_of: int | None = None
+
+    def __call__(self, features: list[dict[str, Any]]) -> dict[str, Any]:
+        pad_to_mult = self.pad_to_multiple_of
+
+        padded_batch: dict[str, Any] = {}
+        for k in features[0].keys():
+            if k.endswith(
+                ("_input_ids", "_attention_mask", "_labels", "_pixel_values")
```
NanoCode012, line 43: Is there a list we can find for keys to pad? Do we need to also include other multi-modal keys? mm_token_ids etc or is this enough?

src/axolotl/utils/collators/dpo.py
```diff
@@ -0,0 +1,121 @@
...
+            ):
+                if self.is_encoder_decoder:
+                    to_pad = [torch.LongTensor(ex[k]) for ex in features]
```
NanoCode012: Should check dtype

src/axolotl/utils/schemas/config.py
```diff
@@ -663,6 +663,12 @@ class AxolotlInputConfig(
             "description": "Pad inputs so each step uses constant sized buffers. This will reduce memory fragmentation and may prevent OOMs, by re-using memory more efficiently. Defaults to True if `sample_packing` enabled"
         },
     )
+    pad_to_multiple_of: int | None = Field(
```
NanoCode012, line 666: Potential conflict: KD has the same config name. Should be safe for now.

NanoCode012: Additionally, we used to just base this off sequence_len, would we need to refactor those handling? It could break any existing tokenized dataset though.

src/axolotl/core/builders/rl.py
winglian: I think this is a follow on fix as the rl builder only seems to support self.tokenizer atm.

PR #3611, opened 2026-04-19
fix: [gemma4] fix VRAM leak in hybrid FA2+SDPA (hybrid attentiuon) path under activation check…

### LINKED ISSUE #3610: VRAM leak in Gemma-4 hybrid attention path under activation checkpointing

### Expected Behavior

VRAM usage should remain relatively flat across training steps when using gemma4_hybrid_attn_impl: true (FA2 for sliding-window layers + SDPA for global layers), matching the flat VRAM profile of the all-flex-attention path but at the ~22× higher throughput the  hybrid path delivers.

### Current behaviour

With gemma4_hybrid_attn_impl: true, VRAM climbs ~0.47 GiB/step from a 42 GiB baseline and OOMs around step 73 (~94 GiB peak).   

The climb is independent of actual sequence length and image size — it happens regardless of input shape variability.
                                                                                                                            
With flex_attention: true (all 60 layers flex), VRAM is flat at 64 GiB but training runs at ~30 s/step vs ~1.3 s/step on hybrid — not a viable workaround.
                                                                                                                                  
Introduced by PR #3598 (commit b8358aa5, "[gemma4] use mixed Flash Attention and SDPA and add fused RMSNorm+RoPE Triton kernels"), which added the hybrid flag, the unconditional patch_gemma4_fused_attn() monkeypatch, and the fused RMSNorm+RoPE kernels together. 

### Steps to reproduce

  1. Install axolotl at HEAD 323da791 (post PR #3598).                                                                            
  2. Launch training on Gemma-4 31B multimodal with the following config elements:
    - gemma4_hybrid_attn_impl: true                                                                                               
    - adapter: lora with lora_r: 256 on the language layers; vision tower + embed_vision trained full                             
    - sample_packing: false                                                                                                       
    - pad_to_sequence_len: false                                                                                                  
    - sequence_len: 5248                                                                                                          
    - FSDP2 with reshard_after_forward: true                      
    - gradient_checkpointing: true                                                                                                
  3. Hardware: 2× RTX PRO 6000 Blackwell 96 GB (also reproducible on any FSDP2 + activation-checkpointing setup per the analysis).
  4. Monitor nvidia-smi or torch.cuda.memory_allocated() per step — observe linear ~0.47 GiB/step climb from the ~42 GiB baseline until OOM around step 73.    

### Config yaml

```yaml
base_model: /home/trainer/model_shims/gemma-4-31B-it-max-1120
tokenizer_type: AutoTokenizer
processor_type: AutoProcessor

seed: 42
data_seed: 42

load_in_8bit: false
load_in_4bit: false
strict: false

adapter: lora
lora_r: 256
lora_alpha: 256          # rsLoRA -> alpha == r is optimal
lora_dropout: 0.05
peft_use_dora: false     # disabled for fast iteration; flip back on for final runs
peft_use_rslora: true    # rank-stabilized LoRA - critical at high rank
peft_autocast_adapter_dtype: false

lora_target_modules: 'model\.language_model\.layers\.[\d]+\.(_checkpoint_wrapped_module\.)?(mlp|self_attn)\.(up|down|gate|q|k|v|o)_proj'

# Optimized LoRA kernels (Liger / KernelsPlugin LoRA_QK for attention_k_eq_v)
#lora_mlp_kernel: true
#lora_qkv_kernel: true
#lora_o_kernel: true

# lora_modules_to_save:
#   - embed_tokens


multimodal: true
skip_prepare_dataset: true
remove_unused_columns: false
sample_packing: false
freeze_mm_modules: false

# Gemma-4 ships its own chat template in axolotl
chat_template: gemma4

default_system_message:
train_on_inputs: false
group_by_length: false

datasets:
  - path: /home/out/train.normalized.jsonl
    type: chat_template

# Held-out validation set, overwritten in place by RUN_FILTER_DATASET.sh
test_datasets:
  - path: /home/out/val.normalized.jsonl
    type: chat_template
    split: train          # local jsonl - axolotl expects this split name

sequence_len: 5248
pad_to_sequence_len: false
eval_sample_packing: false

output_dir: /home/output
num_epochs: 3
micro_batch_size: 1
learning_rate: 0.0001         # default LR for the LoRA A/B adapters on LM layers

embedding_lr: 0.00003         # 3e-5 — conservative, prevents subtle gen quality drift

lr_groups:
  - name: vision_tower
    modules:
      - vision_tower          # catches model.vision_tower.*
    lr: 0.00003               # 3e-5 — adapts for OOD Rashi script without nuking general visual priors
  - name: projector
    modules:
      - embed_vision          # catches model.embed_vision.embedding_projection.*
    lr: 0.00007               # 7e-5 — bridge layer benefits from faster adaptation
lr_scheduler: cosine
warmup_ratio: 0.03
weight_decay: 0.0
label_smoothing_factor: 0.0

max_grad_norm: 1.0
# 8-bit AdamW (torchao) — required to fit full vision FT + r=512 DoRA + embed
# trainable. FSDP2-compatible, ~negligible quality impact for LoRA.
optimizer: adamw_torch_8bit
adam_beta1: 0.9
adam_beta2: 0.999
adam_epsilon: 0.00000001

save_strategy: steps
save_steps: 200
eval_strategy: steps
eval_steps: 200
logging_steps: 10
# gc_steps: 1   # deprecated — replaced by the two fields below (PR #3604)
torch_empty_cache_steps: 10   # HF-native torch.cuda.empty_cache() every 10 steps
gc_collect_steps: 10          # Python gc.collect() every 10 steps
save_total_limit: 12
save_safetensors: true

load_best_model_at_end: true
metric_for_best_model: eval_loss
greater_is_better: false

bf16: true
tf32: true

fp8: false
fp8_enable_fsdp_float8_all_gather: false

gradient_checkpointing: false

dataloader_num_workers: 8
dataloader_pin_memory: true
dataloader_persistent_workers: true
dataloader_prefetch_factor: 8

# =============================================================================
# Attention - use Flex Attention instead of Gemma4 hybrid Flash/SDPA.
# This avoids Flash Attention limits for Gemma4 global_head_dim=512.
# =============================================================================
gemma4_hybrid_attn_impl: true
#flex_attention: true
#flash_attention: true
attention_dropout: 0.0
xformers_attention: false
sdp_attention: false
use_reentrant: false

# =============================================================================
# Plugins - CCE + KernelsPlugin + Liger
# =============================================================================
plugins:
  - axolotl.integrations.cut_cross_entropy.CutCrossEntropyPlugin
  - axolotl.integrations.kernels.KernelsPlugin
  - axolotl.integrations.liger.LigerPlugin

#use_kernels: true       # enables LoRA_QK fused kernel for attention_k_eq_v=true

# Liger fused ops
liger_layer_norm: true
liger_rope: true
liger_rms_norm: true
liger_glu_activation: true
liger_rms_norm_gated: true
liger_fused_linear_cross_entropy: false   # CCE handles the LM-head loss instead

torch_compile: true
compile: true
use_fast: true

# Early stopping
early_stopping_patience: 25

# Gemma-4 is in-tree in transformers >=5.5 — trust_remote_code must be OFF
# because the LoRA fused kernels (lora_mlp_kernel / lora_qkv_kernel /
# lora_o_kernel) are incompatible with custom-code paths.
trust_remote_code: false

# Validation
val_set_size: 0           # disabled - using test_datasets (real held-out val) instead
debug: false
debug_data_preprocessing: false
eval_batch_size: 1

# =============================================================================
# FSDP2 Configuration (2x RTX PRO 6000 Blackwell, 96GB each)
# =============================================================================
fsdp_version: 2
fsdp_config:
  offload_params: false           # 192GB total; keep optimizer on GPU
  # NOTE: cpu_ram_efficient_loading must be false when using FP8 (torchao) +
  # FSDP2 — accelerate explicitly rejects that combo. Each rank will load the
  # full bf16 model into CPU RAM (~62 GB); with 377 GB system RAM that's fine.
  cpu_ram_efficient_loading: false
  auto_wrap_policy: TRANSFORMER_BASED_WRAP
  # Gemma-4 decoder + vision encoder layer classes
  transformer_layer_cls_to_wrap: Gemma4TextDecoderLayer,Gemma4VisionEncoderLayer
  state_dict_type: SHARDED_STATE_DICT
  reshard_after_forward: true
  activation_checkpointing: true
  sync_module_states: true

local_rank:

# Dataset processing
dataset_prepared_path:
push_dataset_to_hub:
hf_use_auth_token:
preprocess_in_chat_template: true
shuffle_merged_datasets: true
dataset_num_proc: 8       # renamed from dataset_processes (deprecated)

# Loss
ignore_index: -100
roles_to_train:
  - assistant
train_on_eos: turn

# Grad accumulation
gradient_accumulation_kwargs:
  sync_each_batch: false

save_model_complete: true
save_best_only: false
```

### Possible solution

**Root cause.** 

HF's Gemma4TextModel.forward allocates one shared_kv_states = {} dict per forward and passes it as a kwarg to every  decoder_layer(...) call. Producer layers (store_full_length_kv=True) write K/V tensors into this dict; consumer layers (is_kv_shared_layer=True) read from it.                                                                                         
                                                                  
  GradientCheckpointingLayer.__call__ forms partial(super().__call__, **kwargs) when checkpointing is active. That partial — which holds a reference to the shared dict — is captured by the checkpoint implementation (axolotl's CPU_Offloaded_Gradient_Checkpointer via ctx.forward_function = forward_function at                        
  src/axolotl/monkeypatch/gradient_checkpointing/offload_cpu.py:51, or torch's stock torch.utils.checkpoint.checkpoint). Because  the dict is a mutable Python container (not an autograd-managed tensor), the K/V tensors it holds are pinned for the full  duration of backward. Combined with delayed ref-cycle cleanup in torch's caching allocator under FSDP2 + activation checkpointing, the residual bleeds across training steps.

  The all-flex path does not leak because it compiles all 60 layers identically into a single graph context where saved tensors are managed by the compiled graph rather than Python refcounting.
                                                                                                                                  
  Violated invariant: anything crossing an activation-checkpoint boundary must be a tensor (refcounted by autograd) or plain Python data — never a mutable container holding tensor references.

  **Possible Solution**                                                                                                               
                                                                  
  Route shared_kv_states through a thread-local side channel instead of the decoder-layer kwargs, so the checkpoint partial never 
  references the dict. All changes in src/axolotl/monkeypatch/models/gemma4/fused_attn.py:
                                                                                                                                  
  1. Add a threading.local() store with _get_shared_kv_states() / _set_shared_kv_states() helpers.
  2. Add _patch_decoder_layer_call() — monkeypatches Gemma4TextDecoderLayer.__call__ to pop shared_kv_states from kwargs and stash it in TLS before delegating to radientCheckpointingLayer.__call__. The partial formed downstream no longer references the dict.
  3. Modify fused_forward to prefer the TLS store, falling back to the kwarg for callers that bypass the patched __call__ (e.g.,  direct attention invocation during eval).                                                                                       
  4. Wire _patch_decoder_layer_call() into patch_gemma4_fused_attn(). Idempotent via a _axolotl_shared_kv_tls_patched sentinel on the class.                                                                                                                      
                                                                  
  TLS is overwritten on each new step's first decoder-layer call, so the previous step's dict is released promptly. No changes to  the hybrid dispatch, FSDP wrap policy, or any config behavior. 
                                                                                                                                  
  **Hypotheses considered and rejected during diagnosis:**                                                                            
  - H1 — per-layer shallow-copied config breaks FSDP/dynamo equivalence: contributes to path divergence but is not the leak  source.                                                                                                                         
  - H2 — silent cfg.flash_attention = True activates downstream per-step allocations: checked all cfg.flash_attention consumers;
  none allocate per-step state on the reported config (sample_packing: false, SM120 skips FA4 auto-patch).                        
  - H4 — FA2 workspace retention under mixed dispatch: FA2 workspace is stateless per-call; no accumulation.                      
   
  **Validation performed:**                                                                                                           
  - ast.parse clean.                                              
  - Patch installs: both Gemma4TextDecoderLayer.__call__ and Gemma4TextAttention.forward are replaced.                            
  - TLS get/set round-trips correctly.                                                                
  - Double-invocation of patch_gemma4_fused_attn() does not double-wrap.                                                          
  - End-to-end VRAM-flatness across steps to be confirmed by reporter.  

  | step | was (pre-patch) | is (post-patch) |                                                                 
  |-----:|---------------:|---------------:|                                                                   
  | eval | 42.5 GiB | 42.6 GiB |                                                                               
  | 9    | 57.2 GiB | 55.3 GiB |                                                                               
  | 19   | 64.8 GiB | 56.3 GiB |                                                                               
  | 29   | 62.0 GiB | 48.7 GiB |                                                                               
  | 39   | 72.5 GiB | 55.0 GiB |
  | 49   | 74.7 GiB | 52.1 GiB |                                                                               
  | 59   | 79.2 GiB | 49.0 GiB |                                  
  | 69   | 85.6 GiB | 52.9 GiB |                                                                               
  | 73   | **OOM (~94 GiB)** | 57.1 GiB (step 79) |                                                            
  | 89   | — | 52.7 GiB |                                                                                      
  | 99   | — | 57.5 GiB |                                                                                      
  | 109  | — | 48.5 GiB |                                                                                      
  | 119  | — | 53.4 GiB |                                                                                      
  
  **Pre-patch:** monotonic climb ~0.47 GiB/step → OOM at step 73.                                              
  **Post-patch:** oscillates in a 48–58 GiB band with no trend across 120+ steps. Peak at step 99 (57.5 GiB)
  matches early-run peak at step 19 (56.3 GiB). 

- [x] Linux
- [ ] macOS
- [ ] Windows

### Python Version

3.12

### axolotl branch-commit

main/323da791

### Acknowledgements

thad0ctor:
# [gemma4] fix VRAM leak in hybrid FA2+SDPA path under activation checkpointing

Route `shared_kv_states` through a thread-local side channel instead of the decoder-layer kwargs so the checkpoint partial never references the dict.

# Description

HF's `Gemma4TextModel.forward` passes `shared_kv_states` (a mutable dict used for cross-layer K/V sharing) as a kwarg to every `decoder_layer(...)` call. `GradientCheckpointingLayer.__call__` then forms `partial(super().__call__, **kwargs)`, and whichever checkpoint implementation runs — axolotl's `CPU_Offloaded_Gradient_Checkpointer` (via `ctx.forward_function = forward_function` at `src/axolotl/monkeypatch/gradient_checkpointing/offload_cpu.py:51`) or torch's stock `torch.utils.checkpoint.checkpoint` — captures that partial. The partial holds a reference to the dict, which holds the K/V tensors produced by `store_full_length_kv` layers. Those tensors stay pinned for the full duration of backward, and delayed ref-cycle cleanup in torch's caching allocator under FSDP2 + activation checkpointing bleeds the residual across training steps.

**Violated invariant:** anything crossing an activation-checkpoint boundary must be a tensor (refcounted by autograd) or plain Python data — never a mutable container holding tensor references.

**All changes live in `src/axolotl/monkeypatch/models/gemma4/fused_attn.py` (+58 / −2):**

- `threading.local()` store with `_get_shared_kv_states()` / `_set_shared_kv_states()` helpers.
- `_patch_decoder_layer_call()` — monkeypatches `Gemma4TextDecoderLayer.__call__` to pop `shared_kv_states` from kwargs and stash it in TLS before delegating to `GradientCheckpointingLayer.__call__`. The partial formed downstream no longer references the dict.
- `fused_forward` reads TLS first and falls back to the kwarg for callers that bypass the patched `__call__` (e.g. direct attention invocation during eval hooks).
- Wired into `patch_gemma4_fused_attn()`; idempotent via an `_axolotl_shared_kv_tls_patched` sentinel.

TLS is overwritten on each new step's first decoder-layer call, so the previous step's dict is released promptly. No changes to the hybrid dispatch, FSDP wrap policy, config schema, or any public behavior. The patch is unconditional — it works correctly for hybrid, flex, and eager paths, because the TLS rerouting is semantically equivalent to the kwarg path (it is purely a capture-avoidance change).

## Motivation and Context

Fixes #3610.

Introduced by PR #3598 (commit `b8358aa5`, _"[gemma4] use mixed Flash Attention and SDPA and add fused RMSNorm+RoPE Triton kernels"_), which added the hybrid flag, the unconditional `patch_gemma4_fused_attn()` monkeypatch, and the fused RMSNorm+RoPE kernels together.

**Observed symptom (pre-patch):**

| Path                                   | VRAM behavior                                           | Throughput (steady-state) |
| -------------------------------------- | ------------------------------------------------------- | ------------------------- |
| `gemma4_hybrid_attn_impl: true`        | Climbs ~0.47 GiB/step from 42 GiB baseline; OOMs at step 73 (~94 GiB peak) | ~13–14 s/step             |
| `flex_attention: true` (all 60 layers) | Flat at 64 GiB                                          | ~30 s/step (~2× slower)  |

The climb is independent of actual sequence length or image size — confirmed by the reporter, which rules out shape variability, `torch.compile` recompiles, and head_dim=512 SDPA peak allocations (which would scale with T²). The flex path is flat but ~2× slower and not a practical workaround for a multi-day training run.

**Hypotheses considered and rejected during diagnosis:**

- **Per-layer shallow-copied config breaks FSDP/dynamo equivalence** — contributes to path divergence but is not the leak source.
- **Silent `cfg.flash_attention = True` activates downstream per-step allocations** — audited all `cfg.flash_attention` consumers; none allocate per-step state under the reported config (`sample_packing: false`, SM120 skips the FA4 auto-patch).
- **FA2 workspace retention under mixed dispatch** — FA2 workspace is stateless per-call; no accumulation.

The violated-invariant analysis also cleanly explains why the all-flex path doesn't leak: flex compiles all 60 layers identically into a single graph context where saved tensors are managed by the compiled graph rather than by Python refcounting, so the dict's refs don't pin anything.

## How has this been tested?

**Static / structural validation (performed):**

- `ast.parse` on the modified module — clean.
- Patch installation smoke test — both `Gemma4TextDecoderLayer.__call__` and `Gemma4TextAttention.forward` are replaced after `patch_gemma4_fused_attn()`.
- Sentinel (`_axolotl_shared_kv_tls_patched`) prevents double-wrapping on repeated invocations.
- TLS `_get_shared_kv_states()` / `_set_shared_kv_states()` round-trip correctly.
- Byte-identical to the editable install being used on the reporter's hardware.
- Ran `pre-commit run --files src/axolotl/monkeypatch/models/gemma4/fused_attn.py` on the modified file:       
   
  | Hook                     | Result     |                                                                    
  | ------------------------ | ---------- |                       
  | ruff (legacy alias)      | ✅ Passed  |                                                                    
  | ruff format              | ✅ Passed  |                                                                    
  | mypy                     | ✅ Passed  |
  | bandit                   | ✅ Passed  |                                                                    
  | fix end of files         | ✅ Passed  |                       
  | trim trailing whitespace | ✅ Passed  |          

**End-to-end VRAM-flatness validation (confirmed on reporter's hardware):**

Reran the failing config — Gemma-4 31B multimodal, LoRA r=256 on language layers, vision tower + `embed_vision` trained full, `sample_packing: false`, `pad_to_sequence_len: false`, `sequence_len: 5248`, FSDP2 with `reshard_after_forward: true`, `activation_checkpointing: true` under `fsdp_config` (HF Trainer-level `gradient_checkpointing: false`), on 2× RTX PRO 6000 Blackwell 96 GB. Ran 200+ steps past the previous OOM point (step 73) including a full eval + checkpoint-save cycle at step 200. VRAM is flat; throughput matches pre-patch hybrid.

| step               | was (pre-patch)    | is (post-patch)   |
| ------------------:| ------------------:| -----------------:|
| eval (pre-train)   | 42.5 GiB           | 42.6 GiB          |
| 9                  | 57.2 GiB           | 55.3 GiB          |
| 19                 | 64.8 GiB           | 56.3 GiB          |
| 29                 | 62.0 GiB           | 48.7 GiB          |
| 39                 | 72.5 GiB           | 55.0 GiB          |
| 49                 | 74.7 GiB           | 52.1 GiB          |
| 59                 | 79.2 GiB           | 49.0 GiB          |
| 69                 | 85.6 GiB           | 52.9 GiB          |
| 73                 | **OOM (~94 GiB)**  | —                 |
| 79                 | —                  | 57.1 GiB          |
| 89                 | —                  | 52.7 GiB          |
| 99                 | —                  | 57.5 GiB          |
| 109                | —                  | 48.5 GiB          |
| 119                | —                  | 53.4 GiB          |
| 129                | —                  | 52.3 GiB          |
| 139                | —                  | 58.5 GiB          |
| 149                | —                  | 49.9 GiB          |
| 159                | —                  | 57.9 GiB          |
| 169                | —                  | 53.7 GiB          |
| 179                | —                  | 54.3 GiB          |
| 189                | —                  | 66.2 GiB          |
| 199                | —                  | 49.3 GiB          |
| eval @ 200         | —                  | 44.6 GiB          |
| save @ 200         | —                  | ✅ checkpoint saved |

Post-patch VRAM oscillates in a ~48–58 GiB band with one excursion to 66.2 GiB at step 189; no upward trend. Step time is ~13–14 s/it (steady-state), identical to pre-patch hybrid. Eval loss dropped 5.43 → 0.69 over the first 200 steps; the first `save_strategy: steps` checkpoint wrote cleanly under `SHARDED_STATE_DICT` FSDP2.

**Testing environment of record:**

- `transformers` 5.5.4
- Hardware: 2× RTX PRO 6000 Blackwell 96 GB (SM120); FSDP2; activation checkpointing enabled.
- Model: Gemma-4 31B (shimmed to enforce max image tokens in training)
- `torch` 2.11.0+cu130
- main/323da791 served as basis for fork with these changes 

**What reviewers may want to sanity-check:**

- The `Gemma4TextDecoderLayer.__call__` intercept sits _above_ `GradientCheckpointingLayer.__call__`, so the partial formed downstream does not re-capture the popped kwarg.
- The kwarg fallback in `fused_forward` preserves behavior for any caller that invokes attention without going through the patched decoder-layer `__call__` (e.g. custom eval hooks that reach into `self_attn` directly).
- The unconditional application to the flex and eager paths is intentional and benign: the TLS rerouting is a capture-avoidance change, not a behavior change.

## AI Usage Disclaimer

Claude  Code (Opus 4.7) was used in the following scopes:

- **Root-cause investigation** — reading the axolotl + HF transformers source for the Gemma-4 hybrid attention path, `CPU_Offloaded_Gradient_Checkpointer`, `GradientCheckpointingLayer`, and the `shared_kv_states` dataflow through the decoder loop; validating four hypotheses against the code; producing the "violated invariant" analysis.
- **Patch design** — proposing the thread-local side channel as the minimal diff shape that preserves the hybrid dispatch and FSDP wrap policy.
- **Patch implementation** — writing the TLS helpers, the `_patch_decoder_layer_call()` intercept, and the `fused_forward` fallback; wiring into `patch_gemma4_fused_attn()`; adding the idempotency sentinel.
- **Static validation** — the AST parse, patch-installation smoke test, TLS round-trip, and idempotency verification.

Supplied by the human author: the training config, observed VRAM numbers (0.47 GiB/step, step-73 OOM, 42 GiB baseline, 94 GiB peak, 13–14 vs 30 s/step steady-state throughput), hardware details, confirmation that the climb is independent of sequence length / image size, and the end-to-end 200+ step VRAM-flatness rerun on the reporter's 2× RTX PRO 6000 Blackwell hardware including a full eval + checkpoint-save cycle.

N/A — the change is a structural fix with no user-visible surface.

## Types of changes

- [x] Performance / memory optimization (eliminates per-step VRAM accumulation on the hybrid path)
- [ ] Documentation update

## Social Handles (Optional)

_(fill in if desired)_

## Summary by CodeRabbit

* **Refactor**
  * Improved internal memory management and state handling for Gemma4 model attention processing, enhancing overall efficiency and system reliability.

### DIFF
--- a/src/axolotl/monkeypatch/models/gemma4/fused_attn.py
+++ b/src/axolotl/monkeypatch/models/gemma4/fused_attn.py
@@ -10,7 +10,6 @@ Usage:
 """
 
 import logging
-import threading
 from typing import Callable
 
 import torch
@@ -18,7 +17,7 @@ import torch
 logger = logging.getLogger(__name__)
 
 
-# Thread-local side channel for the shared-KV dict. We route the dict through
+# Module-level side channel for the shared-KV dict. We route the dict through
 # this instead of a kwarg on the decoder-layer forward so that when HF's
 # GradientCheckpointingLayer forms `partial(super().__call__, **kwargs)` and
 # axolotl's CPU_Offloaded_Gradient_Checkpointer captures that partial in
@@ -27,15 +26,28 @@ logger = logging.getLogger(__name__)
 # pinned across the full backward pass (and, via Python ref-cycle delays in
 # torch's caching allocator, bleed across training steps), causing VRAM to
 # climb ~0.47 GiB/step under the hybrid FA2+SDPA path.
-_GEMMA4_SHARED_KV_TLS = threading.local()
+#
+# NOTE: originally `threading.local()`, but PyTorch's C++ autograd engine
+# (`_engine_run_backward`) spawns per-device worker threads to dispatch
+# backward. When HF-Trainer gradient_checkpointing (`torch.utils.checkpoint`,
+# non-reentrant / saved-tensor-hooks) fires `unpack_hook` -> `recompute_fn`
+# during backward, it runs on the autograd worker thread -- whose
+# `threading.local()` is empty, so the dict set on the main thread during
+# forward is invisible and `_get_shared_kv_states()` returns None, crashing
+# the consumer-layer lookup (`shared_kv_states[self.kv_shared_layer_index]`)
+# with `'NoneType' object is not subscriptable`. A plain module-level dict is
+# shared across all threads and works for both paths. The container is
+# overwritten each forward, so the previous step's dict is released promptly
+# -- same lifecycle guarantee the TLS variant gave.
+_GEMMA4_SHARED_KV_STORE: dict = {"store": None}
 
 
 def _set_shared_kv_states(store):
-    _GEMMA4_SHARED_KV_TLS.store = store
+    _GEMMA4_SHARED_KV_STORE["store"] = store
 
 
 def _get_shared_kv_states():
-    return getattr(_GEMMA4_SHARED_KV_TLS, "store", None)
+    return _GEMMA4_SHARED_KV_STORE["store"]
 
 
 def _make_fused_forward(original_forward):
@@ -170,7 +182,7 @@ def _patch_decoder_layer_call():
     """
     from transformers.models.gemma4.modeling_gemma4 import Gemma4TextDecoderLayer
 
-    if getattr(Gemma4TextDecoderLayer, "_axolotl_shared_kv_tls_patched", False):
+    if getattr(Gemma4TextDecoderLayer, "_axolotl_shared_kv_patched", False):
         return
 
     original_call = Gemma4TextDecoderLayer.__call__
@@ -185,7 +197,7 @@ def _patch_decoder_layer_call():
         return original_call(self, *args, **kwargs)
 
     Gemma4TextDecoderLayer.__call__ = patched_call
-    Gemma4TextDecoderLayer._axolotl_shared_kv_tls_patched = True
+    Gemma4TextDecoderLayer._axolotl_shared_kv_patched = True
 
 
 def patch_gemma4_fused_attn():
--- a/src/axolotl/monkeypatch/models/gemma4/fused_attn.py
+++ b/src/axolotl/monkeypatch/models/gemma4/fused_attn.py
@@ -10,11 +10,32 @@ Usage:
 """
 
 import logging
+import threading
 from typing import Callable
 
 import torch
 
 logger = logging.getLogger(__name__)
+
+
+# Thread-local side channel for the shared-KV dict. We route the dict through
+# this instead of a kwarg on the decoder-layer forward so that when HF's
+# GradientCheckpointingLayer forms `partial(super().__call__, **kwargs)` and
+# axolotl's CPU_Offloaded_Gradient_Checkpointer captures that partial in
+# `ctx.forward_function`, the ctx does not hold a reference to the dict --
+# otherwise the K/V tensors stored in the dict for the producer layers stay
+# pinned across the full backward pass (and, via Python ref-cycle delays in
+# torch's caching allocator, bleed across training steps), causing VRAM to
+# climb ~0.47 GiB/step under the hybrid FA2+SDPA path.
+_GEMMA4_SHARED_KV_TLS = threading.local()
+
+
+def _set_shared_kv_states(store):
+    _GEMMA4_SHARED_KV_TLS.store = store
+
+
+def _get_shared_kv_states():
+    return getattr(_GEMMA4_SHARED_KV_TLS, "store", None)
 
 
 def _make_fused_forward(original_forward):
@@ -30,7 +51,7 @@ def _make_fused_forward(original_forward):
         hidden_states: torch.Tensor,
         position_embeddings: torch.Tensor,
         attention_mask: torch.Tensor | None,
-        shared_kv_states: dict[int, tuple[torch.Tensor, torch.Tensor]],
+        shared_kv_states: dict[int, tuple[torch.Tensor, torch.Tensor]] | None = None,
         past_key_values=None,
         **kwargs,
     ) -> tuple[torch.Tensor, torch.Tensor | None]:
@@ -38,6 +59,12 @@ def _make_fused_forward(original_forward):
         from transformers.models.gemma4.modeling_gemma4 import (
             eager_attention_forward,
         )
+
+        # Prefer the thread-local store (populated by the patched decoder-layer
+        # __call__) so the dict is not captured by the checkpoint partial.
+        tls_store = _get_shared_kv_states()
+        if tls_store is not None:
+            shared_kv_states = tls_store
 
         input_shape = hidden_states.shape[:-1]
         hidden_shape = (*input_shape, -1, self.head_dim)
@@ -133,14 +160,43 @@ def _make_fused_forward(original_forward):
     return fused_forward
 
 
+def _patch_decoder_layer_call():
+    """Strip `shared_kv_states` from the decoder-layer kwargs and route it via
+    thread-local storage instead. This breaks the capture chain
+    `ctx.forward_function -> partial -> kwargs -> shared_kv_states dict` inside
+    the CPU-offload activation checkpointer, so the dict (and the K/V tensors
+    it holds for the producer layers) is not pinned for the duration of the
+    backward pass.
+    """
+    from transformers.models.gemma4.modeling_gemma4 import Gemma4TextDecoderLayer
+
+    if getattr(Gemma4TextDecoderLayer, "_axolotl_shared_kv_tls_patched", False):
+        return
+
+    original_call = Gemma4TextDecoderLayer.__call__
+
+    def patched_call(self, *args, **kwargs):
+        shared_kv = kwargs.pop("shared_kv_states", None)
+        if shared_kv is not None:
+            _set_shared_kv_states(shared_kv)
+        return original_call(self, *args, **kwargs)
+
+    Gemma4TextDecoderLayer.__call__ = patched_call
+    Gemma4TextDecoderLayer._axolotl_shared_kv_tls_patched = True
+
+
 def patch_gemma4_fused_attn():
     """
-    Monkeypatch Gemma4TextAttention.forward to use fused RMSNorm+RoPE kernels.
+    Monkeypatch Gemma4TextAttention.forward to use fused RMSNorm+RoPE kernels,
+    and route `shared_kv_states` via thread-local storage to avoid a VRAM leak
+    under activation checkpointing.
     """
     from transformers.models.gemma4.modeling_gemma4 import Gemma4TextAttention
 
     original_forward = Gemma4TextAttention.forward
     Gemma4TextAttention.forward = _make_fused_forward(original_forward)
+
+    _patch_decoder_layer_call()
 
     logger.info(
         "Patched Gemma4TextAttention.forward with fused RMSNorm+RoPE Triton kernels"
--- a/src/axolotl/monkeypatch/models/gemma4/fused_attn.py
+++ b/src/axolotl/monkeypatch/models/gemma4/fused_attn.py
@@ -6,7 +6,8 @@ kernels, eliminating intermediate tensor allocations from rotate_half / apply_ro
 
 Usage:
     from axolotl.monkeypatch.models.gemma4.fused_attn import patch_gemma4_fused_attn
-    patch_gemma4_fused_attn()
+    # Pass install_shared_kv_workaround=True when activation checkpointing is enabled.
+    patch_gemma4_fused_attn(install_shared_kv_workaround=True)
 """
 
 import logging
@@ -16,29 +17,8 @@ import torch
 
 logger = logging.getLogger(__name__)
 
-
-# Module-level side channel for the shared-KV dict. We route the dict through
-# this instead of a kwarg on the decoder-layer forward so that when HF's
-# GradientCheckpointingLayer forms `partial(super().__call__, **kwargs)` and
-# axolotl's CPU_Offloaded_Gradient_Checkpointer captures that partial in
-# `ctx.forward_function`, the ctx does not hold a reference to the dict --
-# otherwise the K/V tensors stored in the dict for the producer layers stay
-# pinned across the full backward pass (and, via Python ref-cycle delays in
-# torch's caching allocator, bleed across training steps), causing VRAM to
-# climb ~0.47 GiB/step under the hybrid FA2+SDPA path.
-#
-# NOTE: originally `threading.local()`, but PyTorch's C++ autograd engine
-# (`_engine_run_backward`) spawns per-device worker threads to dispatch
-# backward. When HF-Trainer gradient_checkpointing (`torch.utils.checkpoint`,
-# non-reentrant / saved-tensor-hooks) fires `unpack_hook` -> `recompute_fn`
-# during backward, it runs on the autograd worker thread -- whose
-# `threading.local()` is empty, so the dict set on the main thread during
-# forward is invisible and `_get_shared_kv_states()` returns None, crashing
-# the consumer-layer lookup (`shared_kv_states[self.kv_shared_layer_index]`)
-# with `'NoneType' object is not subscriptable`. A plain module-level dict is
-# shared across all threads and works for both paths. The container is
-# overwritten each forward, so the previous step's dict is released promptly
-# -- same lifecycle guarantee the TLS variant gave.
+# Module-level dict used as a side channel for shared KV states avoiding kwarg and TLS
+# to prevent memory leak on gradient checkpoint enabled training (PR #3611)
 _GEMMA4_SHARED_KV_STORE: dict = {"store": None}
 
 
@@ -72,11 +52,9 @@ def _make_fused_forward(original_forward):
             eager_attention_forward,
         )
 
-        # Prefer the thread-local store (populated by the patched decoder-layer
-        # __call__) so the dict is not captured by the checkpoint partial.
-        tls_store = _get_shared_kv_states()
-        if tls_store is not None:
-            shared_kv_states = tls_store
+        store = _get_shared_kv_states()
+        if store is not None:
+            shared_kv_states = store
 
         input_shape = hidden_states.shape[:-1]
         hidden_shape = (*input_shape, -1, self.head_dim)
@@ -173,12 +151,8 @@ def _make_fused_forward(original_forward):
 
 
 def _patch_decoder_layer_call():
-    """Strip `shared_kv_states` from the decoder-layer kwargs and route it via
-    thread-local storage instead. This breaks the capture chain
-    `ctx.forward_function -> partial -> kwargs -> shared_kv_states dict` inside
-    the CPU-offload activation checkpointer, so the dict (and the K/V tensors
-    it holds for the producer layers) is not pinned for the duration of the
-    backward pass.
+    """Strip `shared_kv_states` from decoder-layer kwargs and route via the
+    module-level side channel so the checkpoint partial cannot pin it (PR #3611).
     """
     from transformers.models.gemma4.modeling_gemma4 import Gemma4TextDecoderLayer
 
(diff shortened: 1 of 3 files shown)


src/axolotl/monkeypatch/models/gemma4/fused_attn.py
```diff
@@ -133,15 +160,44 @@ def fused_forward(
...
+    if getattr(Gemma4TextDecoderLayer, "_axolotl_shared_kv_tls_patched", False):
+        return
+
+    original_call = Gemma4TextDecoderLayer.__call__
+
+    def patched_call(self, *args, **kwargs):
+        shared_kv = kwargs.pop("shared_kv_states", None)
+        if shared_kv is not None:
+            _set_shared_kv_states(shared_kv)
+        return original_call(self, *args, **kwargs)
```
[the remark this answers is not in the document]
thad0ctor: Edge case addressed via commit [251021e](https://github.com/axolotl-ai-cloud/axolotl/pull/3611/commits/251021e1f347bef0d8628ba63d7889deda5fa47f)

NanoCode012 (commented):
Ok, I think this makes sense. I'm a bit unclear whether having the store on module scope is the best overall workaround (possible clash with any async runs tho likely rare).

I think we can also make that big chunk of comments more concise. If necessary, just point it to this PR with the explanation here.

Edit: Could we also add unit test (no need e2e) where appropriate? Maybe one to check that the shared kv is popped, the moe issue you met, ?

thad0ctor:
> Ok, I think this makes sense. I'm a bit unclear whether having the store on module scope is the best overall workaround (possible clash with any async runs tho likely rare).
> 
> I think we can also make that big chunk of comments more concise. If necessary, just point it to this PR with the explanation here.
> 
> Edit: Could we also add unit test (no need e2e) where appropriate? Maybe one to check that the shared kv is popped, the moe issue you met, ?

Good point on the scope, let me think of if there needs to be a gating check to prevent any regressions across other functionality - I'm leaning towards this as other options would likely be too invasive elsewhere and add latency.  

I'll address the comments and unit tests for sure.

winglian (approved):

winglian:
thanks! lmk when this is ready to review again

thad0ctor:
> thanks! lmk when this is ready to review again

I just pushed the changes with a gate (for checkpointed training), unit tests and shorter docstrings

Please let me know if you need me to update the PR summary itself as some of it is now OBE

Edit: I missed a test for the MoE 'NoneType' error I observed and corrected in e3669b2c , commit [1bdc84c](https://github.com/axolotl-ai-cloud/axolotl/pull/3611/commits/1bdc84c1b4ff95121ec6296ebbdb2c0f7b066953) adds a 5th test per @NanoCode012 's comments, should be good for review now.  Thanks!

src/axolotl/monkeypatch/models/gemma4/fused_attn.py
```diff
@@ -16,6 +17,18 @@
 
 logger = logging.getLogger(__name__)
```
NanoCode012: Not due to your change, but we should use the axolotl logger as shown in other files.

thad0ctor: Do you need me to update this?

winglian: I'll fix this

winglian (approved):

PR #3612, opened 2026-04-19
feat: add processor_kwargs YAML field forwarded to from_pretrained

thad0ctor:
## Description

Adds a new `processor_kwargs: dict[str, Any] | None` field on `ModelInputConfig`. Its contents are forwarded as `**kwargs` to the processor class's `from_pretrained()` call inside `load_processor`, giving users a YAML hook for overriding fields that live in a model's on-disk `processor_config.json` without having to maintain a local directory that shadows the real weights just to swap one file or update the file directly.

**Scope:** applies to any model whose load path goes through `load_processor` (i.e. `cfg.processor_type` is set). HF's `ProcessorMixin.from_pretrained` propagates top-level kwargs to sub-processors (image / video / feature-extractor / tokenizer), so the same field works across image, video, and audio processors with no model-specific wiring.

**Changes:**
- `src/axolotl/utils/schemas/model.py`: new `processor_kwargs` field on `ModelInputConfig`, placed next to `processor_type` and mirroring the existing `model_quantization_config` / `model_quantization_config_kwargs` pairing. Docstring notes (a) the name overlap with transformers' own call-time `processor_kwargs` argument and (b) that `revision` / `trust_remote_code` should stay on the top-level keys to avoid inconsistent precedence across loader branches.
- `src/axolotl/loaders/processor.py`: forwards `cfg.processor_kwargs` into the shared kwargs dict before the Voxtral and main `from_pretrained()` calls. The `Mistral3Processor` path does not call `from_pretrained` and is intentionally untouched.
- `tests/test_revision_parameter.py`: two new tests covering the forward path and the no-op default.

**Example (Gemma-4):**

```yaml
processor_kwargs:
  image_seq_length: 1120
  max_soft_tokens: 1120
```

Default behavior is preserved when the field is unset (`None`): the guard in `load_processor` skips the update and nothing new reaches `from_pretrained`.

## Motivation and Context

There is currently no YAML hook for per-processor-field overrides. Users who want to change defaults like Gemma-4's `image_seq_length` / `max_soft_tokens` or Qwen-VL's `min_pixels` / `max_pixels` have to maintain a local model directory that symlinks every real file except `processor_config.json`, which they override on disk or edit the model files directly.

## How has this been tested?

1. **Unit tests** (mocked `AutoProcessor`, added in `tests/test_revision_parameter.py`):
   - `test_load_processor_forwards_processor_kwargs` asserts `image_seq_length` and `max_soft_tokens` reach `AutoProcessor.from_pretrained` call kwargs.
   - `test_load_processor_omits_processor_kwargs_when_unset` asserts no-op default.
   - All 8 tests in the file pass (6 pre-existing + 2 new).

2. **Schema round-trip** against the real Gemma-4 YAML config: `AxolotlInputConfig(**yaml.safe_load(...))` with `processor_kwargs` populated preserves the value intact through Pydantic validation.

3. **End-to-end image tokenization** against `google/gemma-4-31B-it`: same 896×896 test image fed through two processors loaded via `load_processor`.

   | metric | baseline | with override `image_seq_length=1120, max_soft_tokens=1120` | ratio |
   |---|---:|---:|---:|
   | `image_seq_length` attr | 280 | 1120 | 4× |
   | `pixel_values` patches | 2520 | 10080 | 4× |
   | `image_position_ids` entries | 2520 | 10080 | 4× |

   The patch-count scaling exactly matches the configured token budget, confirming the override reaches the live image processor's encode path.

**Environment:** Python 3.12, torch 2.9 / transformers 5.5.4, tested on CPU (no GPU-specific paths touched).

## AI Usage Disclaimer

Yes — Claude (Anthropic) was used to assist with tracing the config-loading plumbing, drafting the schema field and forwarder, and authoring the tests.

N/A.

## Types of changes

- [ ] Documentation update
- [ ] Refactor / code cleanup

## Social Handles (Optional)

N/A.

## Release Notes

* **New Features**
  * Added a new configuration option to customize processor initialization, allowing you to override default processor settings at load time.
  * Processor loading now properly forwards and applies custom configuration values for enhanced control over processor behavior.

### DIFF
--- a/src/axolotl/utils/schemas/model.py
+++ b/src/axolotl/utils/schemas/model.py
@@ -67,21 +67,7 @@ class ModelInputConfig(BaseModel):
     processor_kwargs: dict[str, Any] | None = Field(
         default=None,
         json_schema_extra={
-            "description": (
-                "Extra kwargs forwarded to the processor's from_pretrained(), overriding "
-                "fields in the processor's on-disk config (e.g. image_seq_length, "
-                "max_soft_tokens for image processors; min_pixels, max_pixels for "
-                "Qwen-VL). Useful when you want to tune per-processor-field values "
-                "without editing processor_config.json on disk.\n\n"
-                "Note: despite the name, this is distinct from transformers' own "
-                "`processor_kwargs` dict argument to ProcessorMixin.__call__ (which "
-                "controls per-sub-processor call-time kwargs). This field only "
-                "affects from_pretrained load-time kwargs.\n\n"
-                "Do not set `revision` or `trust_remote_code` here -- use the "
-                "top-level `revision_of_model` / `trust_remote_code` keys instead. "
-                "Setting either inside `processor_kwargs` will be rejected by the "
-                "schema validator with a ValueError at config parse time."
-            )
+            "description": "kwargs forwarded to the processor's from_pretrained(), overriding processor config (e.g. image_seq_length, min_pixels, etc.)."
         },
     )
     tokenizer_save_jinja_files: bool | None = Field(
--- a/src/axolotl/utils/schemas/model.py
+++ b/src/axolotl/utils/schemas/model.py
@@ -64,6 +64,26 @@ class ModelInputConfig(BaseModel):
     processor_type: str | None = Field(
         default=None, json_schema_extra={"description": "transformers processor class"}
     )
+    processor_kwargs: dict[str, Any] | None = Field(
+        default=None,
+        json_schema_extra={
+            "description": (
+                "Extra kwargs forwarded to the processor's from_pretrained(), overriding "
+                "fields in the processor's on-disk config (e.g. image_seq_length, "
+                "max_soft_tokens for image processors; min_pixels, max_pixels for "
+                "Qwen-VL). Useful when you want to tune per-processor-field values "
+                "without editing processor_config.json on disk.\n\n"
+                "Note: despite the name, this is distinct from transformers' own "
+                "`processor_kwargs` dict argument to ProcessorMixin.__call__ (which "
+                "controls per-sub-processor call-time kwargs). This field only "
+                "affects from_pretrained load-time kwargs.\n\n"
+                "Do not set `revision` or `trust_remote_code` here -- use the "
+                "top-level `revision_of_model` / `trust_remote_code` keys instead. "
+                "Setting them inside `processor_kwargs` has inconsistent precedence "
+                "across axolotl's loader branches."
+            )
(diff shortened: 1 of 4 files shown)


src/axolotl/utils/schemas/model.py
```diff
@@ -64,6 +64,26 @@ class ModelInputConfig(BaseModel):
     processor_type: str | None = Field(
         default=None, json_schema_extra={"description": "transformers processor class"}
     )
+    processor_kwargs: dict[str, Any] | None = Field(
+        default=None,
+        json_schema_extra={
+            "description": (
+                "Extra kwargs forwarded to the processor's from_pretrained(), overriding "
```
NanoCode012, line 71: I don't think we need such long descriptions :)

NanoCode012 (commented):
I think this looks good overall, will let CI run

src/axolotl/utils/schemas/model.py
```diff
@@ -107,6 +127,22 @@ def hint_trust_remote_code(cls, trust_remote_code):
...
+        reserved = {"revision", "trust_remote_code"}
+        conflicts = reserved.intersection(processor_kwargs)
+        if conflicts:
+            raise ValueError(
+                "Do not set reserved keys "
+                f"{sorted(conflicts)} inside `processor_kwargs`; "
+                "use the top-level `revision_of_model` / `trust_remote_code` "
+                "config keys instead."
+            )
+        return processor_kwargs
```
NanoCode012, line 144: Should also add a check that this is not compatible with `cfg.tokenizer_use_mistral_common`

src/axolotl/utils/schemas/model.py
thad0ctor: Thank you, haha noted :) I kept the (very) verbose comments in there to give you a bit more context as part of the review, I shortened this in commit - [96d8536](https://github.com/axolotl-ai-cloud/axolotl/pull/3612/commits/96d8536fe6f9418084ff9ad8201b7f70bd9543d5)

src/axolotl/utils/schemas/model.py
thad0ctor: Added via [96d8536](https://github.com/axolotl-ai-cloud/axolotl/pull/3612/commits/96d8536fe6f9418084ff9ad8201b7f70bd9543d5)

winglian (approved):
thanks!

thad0ctor:
> thanks!

you're welcome, keep up the good work!

PR #3613, opened 2026-04-20
support for vllm 0.19.1

winglian:
## Summary by CodeRabbit

* **Chores**
  * Updated a core dependency to include patch-level improvements and stability enhancements.


winglian:
shouldn't be needed now that we have it >= using uv

PR #3629, opened 2026-04-25
feat: add multimodal continued pre-training (raw image+text)

thad0ctor:
# Description

Adds a streaming-first multimodal CPT path: raw `(text, images[])` rows are tokenized once with a placeholder-count guardrail, batched through a hardened collator, and fed to a VLM with image-family tokens masked out of labels. Gated by `type: multimodal_pretrain` (or `multimodal: true`) on a `pretraining_dataset` entry; works end-to-end for train and eval, including multi-entry eval and mixed image/text batches.

## Features

- **Streaming MM CPT encoder** (`encode_streaming_multimodal`): counts placeholders by token id (not substring), enforces `placeholders == len(images)` per row, and rejects rows that exceed `sequence_len` instead of silently truncating mid-placeholder.
- **MM CPT collator** (`MultiModalPretrainDataCollator`): image loader with remote-URL rejection (SSRF) and an `image_base_dir` sandbox (relative, absolute, and symlink paths all realpath-checked), per-row image cap (`max_images_per_row`, default 32), processor-call retry that pinpoints — and with `skip_bad_images` drops — the offending row, and label-side masking of every image-family token id.
- **Mixed image/text batches**: text-only rows in a batch take a tokenizer-only fallback (no `pixel_values`); rows with images go through the processor as usual.
- **Eval support**: `test_datasets` accepts MM entries via a dedicated `MultiModalEvalDataset` model so per-entry `text_column` / `image_column` / `image_base_dir` / `image_token` survive validation. Multi-entry MM eval streams are concatenated.
- **`dispatch_batches: true` support**: non-main ranks get a placeholder dataset that mirrors the configured text + image columns.
- **Config validation gates**: `processor_type` required, `sample_packing: false` enforced, `chat_template` rejected, single `pretraining_dataset` entry required, MM eval entries must share `image_base_dir` / `image_token`, mixed MM/non-MM eval rejected, incompatible processor classes (Mllama, Pixtral, InternVL) rejected at startup. `remove_unused_columns` is auto-set to `false` with an INFO log.
- **Docs**: new section in `docs/multimodal.qmd` covering the YAML shape, placeholder-token table, eval contract, and supported/rejected model families.

## YAML example

```yaml
base_model: HuggingFaceTB/SmolVLM-500M-Instruct
processor_type: AutoProcessor

pretraining_dataset:
  - path: /path/to/shards/*.jsonl
    ds_type: json
    type: multimodal_pretrain
    text_column: text
    image_column: images
    image_base_dir: /path/to/images

streaming: true
sequence_len: 2048
sample_packing: false
```

## Motivation and Context

Existing axolotl multimodal paths assume conversational SFT (chat templates with image placeholders inside a structured turn format). Raw image+text continued pre-training — feeding a VLM unstructured documents that interleave text and images — wasn't supported. This PR adds that path without disturbing the existing chat-template SFT flow: it's a separate `type: multimodal_pretrain` opt-in that reuses the streaming pretraining infrastructure and the existing processor/tokenizer wiring.

## How has this been tested?

100 unit tests across five suites:

- `tests/prompt_strategies/test_multimodal_pretrain.py` — encoder/strategy: placeholder count guardrail, sequence-length enforcement, falsy-image rejection, processor compatibility gate.
- `tests/test_multimodal_streaming.py` — streaming encoder + collator: end-to-end batch construction, label masking, mixed-batch and all-text-batch handling, security gates (path traversal incl. absolute/symlink escapes, remote URLs, per-row image cap, processor-failure skip, error sanitization), and Qwen vision-marker wrapping.
- `tests/utils/schemas/validation/test_multimodal_cpt.py` — config validation: processor/packing/template gates, single pretraining-entry rule, MM eval schema preservation, eval homogeneity (`image_base_dir`/`image_token`/`skip_bad_images`/`allow_remote_images`), eval `path` + explicit `split` requirements, `remove_unused_columns` auto-set INFO log.
- `tests/test_mm_cpt_collator_routing.py` — `build_collator` routing (MM CPT -> MM collator, text pretraining untouched) and a real streaming `.map` end-to-end run (jsonl + PNGs -> processor -> collated batch).
- `tests/utils/data/test_mm_cpt_eval.py` — eval data path: placeholder schema for `dispatch_batches`, multi-entry eval merge, mixed MM/non-MM rejection, eval-aware collator config source.

Tested with `HuggingFaceTB/SmolVLM-500M-Instruct`, Gemma 4, Gemma 3, Qwen 2.5VL and Qwen 3.5, as the reference processor. Lint (ruff/ruff-format) and mypy clean against the patches added in this branch.

**Real world testing:**

In addition to the unit suite, the path was validated against two end-to-end LoRA-CPT runs on a OCR corpus  (raw <image>\ntext rows, streamed via type: multimodal_pretrain):             
                                                                                
  - Gemma-4-31B-it, 2× GPU FSDP — LoRA r=256/α=256 (LM proj layers) with vision_tower, embed_vision, and embed_tokens in modules_to_save. 4,400 steps of a 49,000-step plan; eval_loss 9.42 → 1.53 (ppl 12,330 → 4.64), train loss0.41 at step 4,400, ~78 GB device-reserved at peak. Eight checkpoints written at 200-step cadence.                                                       
  - Gemma-4-E2B-it, single RTX 5090 — same LoRA shape, smaller base. 10,600 steps of a 98,000-step plan; eval_loss 9.82 → 0.99 (ppl 18,350 → 2.70), train loss 0.33 at step 10,600, ~18 GB peak.
                                                                                
  Both runs trained to multi-thousand-step stability without divergence or NaN, with periodic eval against a separate test_datasets entry and image-family label masking applied throughout. Confirms the streaming encoder + collator + eval path work end-to-end in real training under FSDP and single-GPU configs.

## AI Usage Disclaimer

Yes — Claude Code (Opus 4.7) was used for design discussion, code review, refactoring suggestions, and test generation. All code was reviewed and validated before commit.  Codex was used for additional vetting, testing and review.  Coderabbit reviews were addressed on personal fork for every commit.  

## Types of changes

- [x] Documentation update

## Summary by CodeRabbit

* **New Features**
  * Multimodal continued pretraining: train/evaluate on (image, text) pairs with processor-aware tokenization, streaming encoding, and a collator that handles mixed image/text batches, processor compatibility checks, hardened image-loading, and masks image tokens in loss.

* **Documentation**
  * Comprehensive multimodal guide: data format, placeholder/token rules, example configs, eval behavior, and loss‑masking semantics.

* **Configuration & Validation**
  * Schema fields and prevalidation gates for multimodal datasets; rejects incompatible settings (sample packing, chat templates), enforces one-entry multimodal train, and validates eval/train modality consistency.

* **Tests**
  * Extensive end-to-end and unit tests covering tokenization, streaming, collation, validation, image-loading security, fixtures, and error cases.


src/axolotl/utils/collators/mm_pretrain.py
```diff
@@ -0,0 +1,306 @@
...
+                    f"(individual rows all succeed; see __cause__ for details)"
+                )
+            else:
+                location = f"batch of {len(texts)} rows (retry inconclusive)"
+            raise RuntimeError(
+                f"MultiModalPretrainDataCollator: processor call failed on "
+                f"{location} ({type(exc).__name__}: {exc}). Common causes: "
+                f"placeholder token absent from the row's text, image count "
+                f"mismatch, or an unsupported processor class."
+            ) from exc
```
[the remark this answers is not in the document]
thad0ctor: This potential limitation will be documented in the documentation in my next commit.  This approach was tested as operable against Gemma 3/4, Qwen 3.5, SmolVLM/SmolVLM2 and Qwen 2.5VL.  

Will consider the following approach for a future PR / Commit pending repo owner feedback:

  2. Per-subset homogeneous fallback. When any(len(im)==0) and not 
  all(len(im)==0), split the batch: run imaged rows through self.processor, run 
  text-only rows through self.tokenizer, pad to common seq_len, insert zero
  pixel_values for text-only positions, restore original row order. ~30-40 lines
   + a regression test. Bulletproof but adds a code path with its own edge cases
   (pixel_values shape varies per processor — Qwen2-VL uses image_grid_thw,
  others use pixel_values of varying ndim).

The goal was min viable without adding additional code complexity to facilitate review. 

Note: testing of Gemma 3 as part of this comment yielded an autodetection error that will be patched in next commit

src/axolotl/utils/data/sft.py
```diff
@@ -160,35 +165,73 @@ def _prepare_streaming_dataset(
...
+                else concatenate_datasets(eval_streams)
+            )
+        else:
+            _, eval_dataset, _ = _load_and_prepare_datasets(
+                tokenizer,
+                cfg,
+                split="test",
+                processor=processor,
+                streaming=False,
+            )
```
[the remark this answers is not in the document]
thad0ctor: Valid, next commit will  honor eval seq length config and add a regression testing as well

src/axolotl/utils/data/sft.py
```diff
@@ -160,35 +165,73 @@ def _prepare_streaming_dataset(
...
+            "split": entry.get("split", "train"),
+            "data_files": entry.get("data_files"),
+            "type": entry.get("type", "pretrain"),
+            "text_column": entry.get("text_column", "text"),
+            "multimodal": entry.get("multimodal"),
+            "image_column": entry.get("image_column", "images"),
+            "image_base_dir": entry.get("image_base_dir"),
+            "image_token": entry.get("image_token"),
+        }
+    )
```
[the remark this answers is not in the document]
thad0ctor: Valid.  Summary of fix in pending changes:

  Summary of fix:                                                               
                                                                                
  - src/axolotl/utils/data/sft.py:                                              
    - _pretraining_config_from_entry: now reads trust_remote_code (default      
  False) from the entry.                                                        
    - _extract_pretraining_config: defaults trust_remote_code: False for the    
  bare-string-path case.                                                        
    - _load_streaming_dataset: passes                                           
  trust_remote_code=pretraining_config.get("trust_remote_code", False) to       
  load_dataset(...).                                                            
  - Tests: test_pretraining_config_from_entry_preserves_trust_remote_code covers
   both the explicit-True and unset (default-False) cases.                      
                                                                             
  A user who sets trust_remote_code: true on a multimodal_pretrain entry (in    
  either pretraining_dataset or test_datasets) now actually gets it threaded    
  through to load_dataset. Default behavior unchanged for everyone else (False).

src/axolotl/utils/data/sft.py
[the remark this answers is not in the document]
thad0ctor: Pending changes: an is_eval=False flag threaded through _load_streaming_dataset
   → wrap_streaming_dataset. When is_eval=True and cfg.eval_sequence_len is set,
   the encoder uses that as max_tokens. The encoder no longer "always reads     
  cfg.sequence_len directly" — it picks cfg.eval_sequence_len for eval.      

  CodeRabbit's suggestion: stash eval_config["sequence_len"] =                  
  cfg.eval_sequence_len on the per-entry config dict, and have the encoder read
  its cap from the per-entry config rather than from cfg.                       
                                                                             
  Functionally equivalent

src/axolotl/utils/schemas/datasets.py
```diff
@@ -238,6 +238,88 @@ class PretrainingDataset(BaseModel):
...
+    image_base_dir: str | None = Field(
+        default=None,
+        json_schema_extra={"description": "Base directory for relative image paths."},
+    )
+    image_token: str | None = Field(
+        default=None,
+        json_schema_extra={
+            "description": "Override the image placeholder token (autodetected from processor if unset)."
+        },
+    )
```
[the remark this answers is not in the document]
thad0ctor: Valid.

Addressed as follows in pending commit:
                                
  - src/axolotl/utils/schemas/datasets.py — ds_type: str | None = Field(...)    
  added to both PretrainingDataset and MultiModalEvalDataset so the loader hint 
  survives validation.                                                          
  - src/axolotl/utils/data/sft.py — _pretraining_config_from_entry and
  _extract_pretraining_config now thread ds_type through.                       
  _load_streaming_dataset branches on ds_type: when set, calls
  load_dataset(ds_type, data_files=path or data_files, ...) (so path:           
  /data/*.jsonl + ds_type: json works as documented). Default behavior unchanged
   when ds_type is unset.
  - Tests — test_pretraining_config_from_entry_preserves_ds_type covers schema
  preservation; test_load_streaming_dataset_routes_ds_type_to_loader asserts    
  load_dataset is called with the right loader name and data_files glob.
                                                                                
  The documented YAML example (path: /path/to/shards/*.jsonl + ds_type: json)   
  now actually works end-to-end.

src/axolotl/utils/data/sft.py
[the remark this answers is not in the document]
thad0ctor: the fallback to cfg.sequence_len when eval_sequence_len is omitted is the documented schema contract, not a regression .  Next commit adds encoder-side test coverage for that case alongside the collator-side one. Removing the fallback would silently break every config that hasn't set eval_sequence_len.

src/axolotl/utils/schemas/validation.py
```diff
@@ -1340,6 +1340,93 @@ def check_streaming_w_multiple_datasets(cls, data):
...
+                    raise ValueError(
+                        f"Multimodal CPT eval requires `{key}` to be either "
+                        f"unset on all `test_datasets` entries or identical "
+                        f"across them. The eval collator resolves "
+                        f"`image_base_dir` and `image_token` once from the "
+                        f"first entry, so heterogeneous values would silently "
+                        f"miscollate later entries. Got: {sorted(map(str, values))}."
+                    )
+
+        return data
```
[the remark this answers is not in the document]
thad0ctor: Valid.

Next commit addresses as follows:

  - src/axolotl/utils/schemas/validation.py — check_multimodal_cpt now does     
  three modality cross-checks at config-load time, regardless of training     
  modality:                                                                  
    a. Mixed MM + non-MM entries inside test_datasets → reject ("All eval     
  entries must share modality").                                                
    b. All-MM test_datasets with non-MM training → reject.                   
    c. Non-MM test_datasets with MM training → reject (this was the runtime     
  check in sft.py; now caught at config parse).                                 
  - Tests — three new tests in test_multimodal_cpt.py covering each rejection
  case.

tests/utils/schemas/validation/test_multimodal_cpt.py
```diff
@@ -0,0 +1,270 @@
...
+        assert validated.remove_unused_columns is False
+        matches = [
+            r
+            for r in caplog.records
+            if r.levelno == logging.INFO and "Auto-set" in r.getMessage()
+        ]
+        assert matches, "expected an INFO record about auto-setting remove_unused_columns"
+        msg = matches[0].getMessage()
+        assert "remove_unused_columns" in msg
+        assert "previous value: None" in msg
```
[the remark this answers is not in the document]
thad0ctor: Will be resolved in next commit

thad0ctor:
@coderabbitai review

tests/test_multimodal_streaming.py
```diff
@@ -0,0 +1,584 @@
...
+        "images": [[str(two_tiny_images[0])] * 3],
+    }
+    out = encode_streaming_multimodal(
+        examples,
+        tokenizer=smolvlm_processor.tokenizer,
+        max_tokens=4096,
+        image_token=spec.image_token,
+        image_token_id=spec.image_token_id,
+    )
+    assert sum(1 for t in out["input_ids"][0] if t == spec.image_token_id) == 3
```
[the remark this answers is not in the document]
thad0ctor: - Filler bumped from × 20 → × 400 (~120 tokens → ~2000+ tokens of filler between placeholders).                                                     
  - Added a len(ids) > 2000 sanity assert so the test self-documents that it's  exercising a long input                                                               
  - Comment rewritten to state the actual property: the last placeholder must remain countable when it's hundreds of tokens deep. Now any regression that adds e.g. tokenizer(text, truncation=True, max_length=128) will drop the third placeholder and trip the placeholder-mismatch error in the encoder before reaching the count assertion.

thad0ctor:
@coderabbitai review

thad0ctor:
@coderabbitai review ,  your last suggestion introduced a regression in gemma4, resolved in last commit

tests/test_multimodal_streaming.py
```diff
@@ -0,0 +1,732 @@
...
+
+
+def test_build_image_token_spec_gemma3_swaps_to_boi_token():
+    """Gemma-3: `image_token` is the post-expansion soft token; placeholder is `boi_token`."""
+    tok = _StubTokenizer(
+        {"<image_soft_token>": 262144, "<start_of_image>": 255999}
+    )
+    proc = _StubProcessor(
+        tok, image_token="<image_soft_token>", boi_token="<start_of_image>"
+    )
```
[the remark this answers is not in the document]
thad0ctor: Collapsed in follow on commit

tests/utils/schemas/validation/test_multimodal_cpt.py
```diff
@@ -0,0 +1,334 @@
...
+        monkeypatch.setattr(
+            logging.getLogger("axolotl.utils.schemas.validation"), "propagate", True
+        )
+        cfg = _mm_cpt_cfg(min_base_cfg, remove_unused_columns=False)
+        with caplog.at_level(logging.INFO, logger="axolotl.utils.schemas.validation"):
+            validate_config(cfg)
+        assert not any(
+            "Auto-set" in r.getMessage() and "remove_unused_columns" in r.getMessage()
+            for r in caplog.records
+        )
```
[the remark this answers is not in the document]
thad0ctor: Will push a fix that mirrors the positive test (patches logging.getLogger("axolotl")) plus a comment explaining why the leaf wouldn't work. Verified mechanically: emitting an "Auto-set" record under that logger inside the caplog.at_level(...) block now lands in caplog.records and trips the assertion as intended.

thad0ctor:
@coderabbitai review

src/axolotl/utils/schemas/datasets.py
```diff
@@ -236,8 +236,102 @@ class PretrainingDataset(BaseModel):
...
+    multimodal: bool | None = Field(
+        default=None,
+        json_schema_extra={
+            "description": "Opt in to multimodal eval. Auto-enabled when type='multimodal_pretrain'."
+        },
+    )
+    text_column: str | None = Field(
+        default="text",
+        json_schema_extra={"description": "Column holding the row's text."},
+    )
```
ved1beta: MultiModalEvalDataset has text_column but PretrainingDataset does not , intentional ? 
both ultimately read the same dict so it should work,  but the schemas don't match.

src/axolotl/prompt_strategies/multimodal_pretrain.py
```diff
@@ -0,0 +1,331 @@
...
+
+class MultimodalPretrainTokenizationStrategy(PretrainTokenizationStrategy):
+    def __init__(
+        self,
+        *args: Any,
+        image_token: str,
+        image_token_id: int,
+        image_column: str = "images",
+        image_base_dir: str | None = None,
+        image_token_spec: ImageTokenSpec | None = None,
```
ved1beta: image_token_spec is not used any where , image_base_dir to

src/axolotl/utils/collators/mm_pretrain.py
```diff
@@ -0,0 +1,347 @@
...
+            )
+        self._image_family_token_ids = set(self.image_token_spec.image_family_token_ids)
+        if self.image_base_dir is not None:
+            self._base_dir_real = os.path.realpath(self.image_base_dir)
+
+    def _resolve_image_path(self, p: str) -> str:
+        if not isinstance(p, str):
+            raise ValueError(f"Image path must be str, got {type(p).__name__}.")
+        if "\x00" in p:
+            raise ValueError("Image path contains embedded NUL byte.")
```
ved1beta: none of this exists in mm_chat.py.

src/axolotl/utils/schemas/datasets.py
thad0ctor: Both schemas already have text_column, the diff was likely cut off so the train-side declaration on line ~235 wasn't visible. I'll align the field metadata (description + data_files type) for symmetry though

ved1beta (changes requested):

ved1beta:
No blockers, mostly bloat-trim.

src/axolotl/prompt_strategies/multimodal_pretrain.py
thad0ctor: good catch, I was playing around with a few things locally and this got orphaned after I settled on the code.  will clean up in next commit

src/axolotl/utils/collators/mm_pretrain.py
thad0ctor: I'll strip out the checks in a commit, these came from code rabbit review and are largely not an issue practices good data management

thad0ctor:
> No blockers, mostly bloat-trim.

I just committed based on your feedback with the cleanup and stripping out the associated tests

thad0ctor:
@ved1beta sorry forgot to tag you on my last response

ved1beta:
lint please

thad0ctor:
> lint please

@ved1beta just committed lint fixes

thad0ctor:
@ved1beta assuming test failure is unrelated to PR since it previously passed and only failed after syncing with main

src/axolotl/utils/schemas/datasets.py
```diff
@@ -232,11 +232,108 @@ class PretrainingDataset(BaseModel):
...
+    image_base_dir: str | None = Field(
+        default=None,
+        json_schema_extra={"description": "Base directory for relative image paths."},
+    )
+    image_token: str | None = Field(
+        default=None,
+        json_schema_extra={
+            "description": "Override the image placeholder token (autodetected from processor if unset)."
+        },
+    )
```
ved1beta, line 291: most of the fields here exist in `PretrainingDataset`
```
text_column, ds_type, skip, trust_remote_code, multimodal, image_column, image_base_dir, image_token,
```

docs/multimodal.qmd
```diff
@@ -364,6 +364,137 @@ Here is an example of a multi-modal dataset:
...
+(`<image>`, `<\|image_pad\|>`, `<\|vision_start\|>`, `<\|vision_end\|>`,
+`<start_of_image>`, `<end_of_image>`, `<image_soft_token>`, `<\|image\|>`,
+etc.) to `-100` in the labels tensor.
+
+Supported processors expand image placeholders into vision-specific token ids.
+If those ids contribute to loss, the model is trained to predict patch or
+marker tokens rather than only the target text. On architectures like
+Qwen-VL and SmolVLM, leaving them unmasked substantially increases loss and can
+destabilize training.
+
```
ved1beta, line 527: why is there no docs for non streaming path ? what happens in non streaming path  ? 
will probably need to 
```
datasets:
  type: multimodal_pretrain
```
have you tested this path ?

ved1beta (changes requested):

docs/multimodal.qmd
ved1beta: i feel the non streaming path is broken , consider removing it completely. remove the strategy class + load()  and its tests ` Keep build_image_token_spec, check_processor_compatibility, and ImageTokenSpec`

thad0ctor: > why is there no docs for non streaming path ? what happens in non streaming path ? will probably need to
> 
> ```
> datasets:
>   type: multimodal_pretrain
> ```
> 
> have you tested this path ?

The non-streaming is a known limitation of this PR.  Only the streaming path is enabled to reduce the amount of code this touches, I will remove the non-streaming path from this PR to make sure it doesn't trigger and make clear in docs.

src/axolotl/utils/schemas/datasets.py
thad0ctor: good catch, I'll push a commit to change this to inherit and strip associated tests

docs/multimodal.qmd
thad0ctor: Please note, I do have a follow-on PR staged that addresses the resumption speed associated with paused streaming training in MM streaming path @ved1beta

thad0ctor:
@ved1beta Pushed the carve-out (389e0f2) which strips the non-streaming strategy class + load() and associated unit tests.

Addressed MultiModalEvalDataset into PretrainingDataset via inheritance (only the type default and the _require_mm_markers validator differ now) and added a docs callout that only the streaming pretraining_dataset route is currently wired. 

Impacted tests passed (60/60) and ran multi-step test on a local Qwen3-VL completes end-to-end with the streaming MM CPT path

ved1beta:
mostly good , config using type: multimodal_pretrain under datasets: will now fail at strategy load time with a raw AttributeError. make it a small ValueError("multimodal_pretrain is only supported via pretraining_dataset with streaming: true — see docs/multimodal.qmd")

ved1beta (approved):
looks good  after the val error , requesting nano for quick look if i missed anything LGTM

thad0ctor:
> looks good after the val error , requesting nano for quick look if i missed anything LGTM

thank you, just added the valueerror per your request 

I do have a follow up PR planned to address the non-streaming path and improve resumption speed on streaming path @NanoCode012

thad0ctor:
@ved1beta @NanoCode012 I just wanted to ping you guys on this, if you want I can close and re-commit with the expansions here on my fork - https://github.com/thad0ctor/axolotl/pull/31

ved1beta:
i feel we can do this one by one. ,  anything is fine

thad0ctor:
@NanoCode012 providing additional benchmark information as requested in https://discord.com/channels/1104757954588196865/1104758010959634503/1522134068655952035

Training Setup:
- ~300K text image pairs across ~1 full epoch
- Model Qwen3 VL (8B) 
- Text image pairs are full transcription of real and synthetic data for a font-type found in historical manuscripts with very poor public dataset coverage (trained via MM-CPT)

Method: identical eval for both models: a held-out set of 150 real manuscript samples (121 full pages, 28 line crops, 1 word), greedy decoding, full-resolution image input. Both models run as Q8_0 GGUF in llama.cpp with the same F16 vision projector. Base = `unsloth/Qwen3-VL-8B-Instruct-GGUF` (Q8_0) — the untuned model the fine-tune started from. Metrics: CER = character error rate (lower is better), chrF (higher is better), exact = whole-sample exact-match rate.

Benchmarks:

| Model | CER ↓ | chrF ↑ | Exact ↑ | Containment ↑ |
|---|---|---|---|---|
| Base Qwen3-VL-8B-Instruct (Q8_0) | 0.655 | 0.358 | 0% | 0% |
| Fine-tuned (CPT, Q8_0) | **0.173** | **0.845** | **14%** | **15%** |

That is a **74% relative reduction in character error rate** (0.655 → 0.173) and chrF more than doubling (0.358 → 0.845). The base model never produces a fully correct transcription (0% exact / 0% containment) on this script; the fine-tune does.

Per-bucket CER (lower is better):

| Bucket | Base | Fine-tuned |
|---|---|---|
| Line | 0.988 | 0.066 |
| Page | 0.581 | 0.198 |
| Word | 0.391 | 0.043 |

The base model is effectively unusable on line crops (CER ~0.99 — near-total error) and only rough on full pages; after training, line/word transcription is near-clean and full-page CER drops ~3x. Results were stable across late checkpoints (steps 12000–14800 land within noise of each other).

A note on method: I first tried SFT, but found it was more sensitive and tended to perturb/regress the base model's existing output behavior while teaching the new script. Continued pre-training (CPT) was able to add the OCR / image-recognition ability **without** regressing the model's general output behavior

winglian:
Majors

  1. The claimed input hardening mostly doesn't exist (mm_pretrain.py). Of the five advertised guards, two are real (relative path traversal, remote URLs). There is no NUL-byte check, no multi-frame guard, no pixel cap, and no per-row image cap — and absolute paths bypass the image_base_dir sandbox entirely (reproduced: ../../x rejected, /abs/path/x loaded verbatim). One hostile row with many placeholders yields multi-GB collator output — an OOM-DoS with all guardrails "on". Either implement the guards or
  rewrite the claims; the absolute-path hole is the one I'd insist on fixing.
  2. sequence_len is advisory for image rows. Rejection happens pre-expansion only; a single 32×32 image expands a 6-word row to 1,134 tokens against max_length=512 with just a warning. Nothing bounds or drops post-expansion rows — VRAM spikes/position overflow surface mid-run.
  3. skip_bad_images doesn't cover processor-stage failures — a row whose image loads but crashes preprocessing (tiny/extreme-aspect images on Qwen, common in web corpora) aborts the whole run even with skip enabled. Reproduced. For the streaming-noisy-corpus use case this feature targets, this is the difference between surviving and dying at step N.
  4. Qwen docs row is factually wrong (independently verified by two agents): the processor does not auto-wrap <|image_pad|> with <|vision_start|>/<|vision_end|> — users following the docs train Qwen without vision markers, silently degrading mrope image handling. Docs should require explicit markers (or the encoder should wrap). Ironically the collator masks vision-marker ids that will never appear.
  5. ds_type routing regression (sft.py:281-293): existing streaming SFT configs that set ds_type alongside a Hub path + data_files previously ignored ds_type and loaded from the Hub; they now resolve against the local filesystem and crash at startup. Gate the branch or call out the break.
  6. The two seams most likely to break are untested: build_collator routing in causal.py has zero coverage (delete the interception and the whole suite stays green — every MM run would silently get the text collator), and nothing runs the real streaming .map pipeline end-to-end. A ~5s test (3-row jsonl + 2 tiny PNGs → real processor → one batch) would close both.

  Minors (selected)

  - trust_remote_code forwarding is dead in pinned datasets==4.8.x — True now produces an alarming error log on every run for configs that predate this PR. Drop it.
  - MM eval entries inherit split: "train" (silent eval-on-train-split) and path is optional (validates, then dies with an unhelpful traceback).
  - Eval homogeneity enforced for image_base_dir/image_token but not skip_bad_images/allow_remote_images, which silently take entry[0]'s policy.
  - The "is MM CPT" predicate is duplicated 4× with one .strip() drift — consolidate before they diverge further.
  - Structural-token masking draws from get_added_vocab() including non-special tokens — an added plain-text token containing "image"/"patch" gets silently -100'd.
  - Every row is tokenized twice and the encoder ships dead input_ids/labels columns (~3× stream payload); per-step PIL decode runs in the collator with no num_workers=0 warning on the CPT path.
  - Dead second MM intercept in build_collator's else-branch (unreachable, contradicts the module docstring); retry-offender indices shift after skipped rows; docs nits (Gemma-4 is 280 not 256; "59 tests" is stale — it's 75+2).

thad0ctor:
Nice, I reproduced all of it. Pushed fixes (`f5889d3c0`..`3c6399998`):

- **abs paths**: `image_base_dir` is a real sandbox now — absolute paths and symlink escapes are rejected, not just `../` traversal
- **ds_type**: only routes local paths/globs; Hub path + `ds_type` keeps the old path-first load, so existing configs don't break
- **skip_bad_images**: now covers processor-stage failures — the retry already isolated the offender, so it drops it and re-runs the batch (repro'd your extreme-aspect Qwen case). Offender indices report original row positions too
- **Qwen**: confirmed no autowrap — the encoder now wraps bare `<|image_pad|>` with `<|vision_start|>`/`<|vision_end|>` itself, so mrope gets its markers and the collator's marker masking actually does something. Docs fixed (Gemma-4 = 280 as well)
- **tests**: `build_collator` routing coverage + a real streaming `.map` e2e (jsonl + PNGs → processor → batch); deleted the unreachable second intercept
- **OOM row**: added `max_images_per_row` (default 32)
- minors: dropped `trust_remote_code`, MM eval entries now require `path` + explicit `split`, homogeneity covers `skip_bad_images`/`allow_remote_images`, structural masking limited to special=True tokens, predicate consolidated into one helper, num_workers warning wired into the CPT path, and the encoder no longer ships the dead token columns (~3x payload)

On the hardening claims: the PR description was stale, not the code lying — those guards were deliberately stripped in `f413bee57` as speculative for a CLI trainer, and the description never caught up. It's updated to match reality now. FWIW NUL paths still fail safely (`realpath` raises) and multi-frame is inert on this path (`load_image` decodes frame 0 only).

Post-expansion `sequence_len` stays warn-only on purpose — truncating after expansion desyncs `input_ids`/`pixel_values`. The per-row cap bounds the worst case now; happy to add an opt-in drop-on-overflow if you'd rather have that.

thad0ctor:
@winglian @NanoCode012 just wanted to follow up on this one before it gets buried, thanks!

Issue #3655, opened 2026-05-14
"KeyError: 0" when trying to fine-tune on text-only conversational dataset during pre-processing step


mags0ft:
### Please check that this issue hasn't been reported before.

- [x] I searched previous [Bug Reports](https://github.com/axolotl-ai-cloud/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

Using the Axolotl Docker image `axolotlai/axolotl-uv:main-latest`, training a model with the attached config is expected to work; however, there seems to be an error with the dataset tokenization / pre-processing. For testing purposes, I made [this conversational dataset](https://huggingface.co/datasets/mags0ft/gsm8k-chatml) on Hugging Face to train on.

What I thought would happen is that the pre-processing completes without any problems and the training process starts. Unfortunately, this does not happen.

### Current behaviour

Axolotl promptly crashes before starting the training at the dataset pre-processing state. It's throwing a KeyError. I don't quite understand where it's coming from, but I believe it is dataset-related...

This is the most relevant part of the error message in short:

```
  File "/workspace/axolotl/src/axolotl/prompt_strategies/chat_template.py", line 418, in tokenize_prompt
    tokenized_prompt = self._tokenize_single_prompt(
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/axolotl/src/axolotl/prompt_strategies/chat_template.py", line 520, in _tokenize_single_prompt
    turn_start_idx, turn_end_idx = self.find_turn(
                                   ^^^^^^^^^^^^^^^
  File "/workspace/axolotl/src/axolotl/prompt_strategies/chat_template.py", line 725, in find_turn
    if dummy_ids[i] != full_ids[i]:
       ~~~~~~~~~^^^
KeyError: 0
```

<details>
<summary>
Here is the full log.
</summary>

```
[2026-05-14 19:15:18,925] [WARNING] [torchao] Skipping import of cpp extensions due to incompatible torch version. Please upgrade to torch >= 2.11.0 (found 2.9.1+cu128).

     #@@ #@@      @@# @@#
    @@  @@          @@  @@           =@@#                               @@                 #@    =@@#.
    @@    #@@@@@@@@@    @@           #@#@=                              @@                 #@     .=@@
      #@@@@@@@@@@@@@@@@@            =@# @#     ##=     ##    =####=+    @@      =#####+  =#@@###.   @@
[...]

PR #3657, opened 2026-05-15
fix AssertionError: Original QKV code not found

ved1beta:
# Description
when training Gemma 4 with LoRA. The Gemma 4 fused RMSNorm+RoPE monkeypatch replaces Gemma4TextAttention.forward wholesale before the LoRA QKV/O source-rewrite patch runs, so the rewrite can no longer find its expected pattern and asserts.

## Motivation and Context

#3655

## How has this been tested?

config runs without  AssertionError: Original QKV code not found

## AI Usage Disclaimer

claude dignose 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved LoRA compatibility with optimized/fused attention paths to avoid incorrect patching.
  * Gemma4 models now bypass incompatible forward-rewrite logic and use the LoRA output hook when available, preventing misapplied patches and ensuring safer fallback behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/axolotl-ai-cloud/axolotl/pull/3657?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/axolotl/monkeypatch/lora_kernels.py
+++ b/src/axolotl/monkeypatch/lora_kernels.py
@@ -76,85 +76,11 @@ QKV_PATCHES = [
     value_states = value_states.view(hidden_shape).transpose(1, 2)
 """.lstrip("\n"),
     ),
-    # Gemma4: norm between proj and transpose, RoPE between norm and transpose,
-    # conditional KV sharing (is_kv_shared_layer), v_proj may be None (attention_k_eq_v).
-    # We only fuse the projection calls; norms, RoPE, and KV sharing stay as-is.
-    (
-        """
-    query_states = self.q_proj(hidden_states).view(hidden_shape)
-    query_states = self.q_norm(query_states)
-    query_states = apply_rotary_pos_emb(query_states, cos, sin, unsqueeze_dim=2)
-    query_states = query_states.transpose(1, 2)
-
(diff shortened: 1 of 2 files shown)


winglian:
@coderabbitai review

src/axolotl/monkeypatch/lora_kernels.py
```diff
@@ -76,85 +76,11 @@
...
-    else:
-        key_states = key_states.view(hidden_shape)
-        value_states = value_states.view(hidden_shape) if self.v_proj is not None else key_states
-""".lstrip("\n"),
-    ),
+    # NOTE: Gemma4 intentionally has no QKV_PATCHES entry. It always runs
+    # through the fused attention monkeypatch (patch_gemma4_fused_attn),
+    # whose hand-written forward already calls self.apply_qkv/self.apply_o.
+    # patch_self_attn_lora skips Gemma4 by class, so a source-rewrite pattern
+    # here would be permanently dead. See that skip for the rationale.
```
winglian: I don't think this is correct, it's only when we setup that patch, right?

ved1beta: ok agreed it was misleading  it's not specific to Gemma4 ,ig it holds because patch_manager applies patch_gemma4_fused_attn unconditionally for gemma4 before patch_self_attn_lora runs, corrected the comment 🫡

NanoCode012 (approved):
Thanks

PR #3685, opened 2026-05-26
Feature/mcp integration

pollardsg:
# Description


winglian:
way too many artifacts introduced, like exe files, venv, etc. +21K lines isn't reviewable

PR #3691, opened 2026-05-30
feat: automatic LoRA rank recommendation based on dataset size

jonyszone:
When lora_r is not set and adapter is lora/qlora, axolotl now auto-selects a sensible rank after the training dataset is loaded:

  < 1 000 samples  → rank  8
  < 10 000 samples → rank 16
  < 100 000 samples → rank 32
  ≥ 100 000 samples → rank 64

The heuristic is exposed as axolotl.utils.lora.recommend_lora_r(n) so users can call it programmatically. An explicit lora_r in the config always takes precedence.

Closes #<issue>

# Description

## Summary by CodeRabbit

* **New Features**
  * Automatic LoRA rank recommendation: System now automatically selects an appropriate LoRA rank based on dataset size when LoRA/QLoRA adapters are enabled without explicit rank configuration.

* **Tests**
  * Added comprehensive test coverage validating rank recommendation thresholds, boundary conditions, and dataset integration behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/axolotl-ai-cloud/axolotl/pull/3691?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/axolotl/common/datasets.py
+++ b/src/axolotl/common/datasets.py
@@ -13,6 +13,7 @@ from axolotl.telemetry.errors import send_errors
 from axolotl.utils.data import prepare_datasets, prepare_preference_datasets
 from axolotl.utils.dict import DictDefault
 from axolotl.utils.logging import get_logger
+from axolotl.utils.lora import recommend_lora_r
 from axolotl.utils.schemas.enums import RLType
 from axolotl.utils.tokenization import check_dataset_labels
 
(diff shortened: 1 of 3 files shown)


winglian:
Thanks for the PR. Are these ranks empirically validated against the dataset sizes? From a maintainer perspective, it might be easier simply to default it to something like 64 when unset.

PR #3692, opened 2026-05-30
feat(compile): expose torch_compile_options for Inductor flag tuning

thad0ctor:
# Description

Adds `torch_compile_options: dict[str, bool|int|float|str] | None` to config, an opt-in allowlist of 10 `torch._inductor.config` knobs that HF Trainer's `TorchDynamoPlugin` (transformers 5.8.1) never forwards. Default `None` → behavior identical to today. The allowlist lives in `enums.py` (`INDUCTOR_COMPILE_OPTIONS_ALLOWLIST`) as the single source of truth for the validators and the runtime apply.

```yaml
torch_compile: true
torch_compile_mode: max-autotune
torch_compile_options:
  max_autotune_gemm: true   # −2.6% on sm_120 (≥32 GB); regresses on sm_86 — arch-specific
```

- **Validated at `axolotl preprocess`:** a `field_validator` rejects non-allowlisted keys; a `model_validator` rejects options without `torch_compile`; a second `model_validator` warns on `triton.cudagraphs` + `sample_packing` (cudagraphs need static shapes, packed sequences are dynamic).
- **Runtime apply** (`TrainerBuilderBase._apply_torch_compile_options`) `setattr`s each validated key on `torch._inductor.config` before compile; dotted keys (`triton.cudagraphs`) work natively.
- Adds an example YAML and updates the `docs/optimizations.qmd` sm_86 note. Purely additive on the existing `torch_compile*` fields.

## Motivation and Context

HF Trainer doesn't forward kwargs to `torch._inductor.config`, so flags like `max_autotune_gemm` previously required monkey-patching before `axolotl train`. These flags have no single correct default  their payoff is architecture ands training-specific (see below) so better suited as opt-in.

## How has this been tested?

**Unit tests** (torch 2.11.0+cu130 / transformers 5.8.1):
- Schema (`TestTorchCompileValidation`, 9): allowlist round-trip, rejects disallowed key, requires `torch_compile`, and the new `triton.cudagraphs`+`sample_packing` warning.
- Runtime (`TestApplyTorchCompileOptions`, 7): flags actually flip `torch._inductor.config`, all 10 allowlist keys verified present (rename sentinel, incl. dotted key), dotted/multi/empty-dict apply, wiring.

**Benchmarks** — Qwen3-8B + LoRA r=16 + bf16 + sdpa + seq 2048 + grad_ckpt + alpaca, 2 seeds, fresh process per (cond, seed), tail-100-median ms/step. Conds: **a** eager · **b** `fused_attn_kernel` (baseline) · **c** b + `torch_compile` · then single inductor flags on c.

`torch_compile` (c vs b) is a large win that scales with arch, and cuts peak memory 19,992 → 19,614 MB on every card:

| c vs b | vanilla 3090 (sm_86) | 3090 Ti (sm_86) | 5090 (sm_120) |
|---|---|---|---|
| compile win | −14.25% | −16.55% | −19.06% |

Of the 10 flags, only `max_autotune_gemm` has a real, arch-dependent effect (Δ vs plain compile); the rest are washes or torch-2.11-default no-ops on both arches:

| flag | sm_86 (3090 Ti) | sm_120 (5090) |
|---|---|---|
| `max_autotune_gemm` | +13% regression | **−2.57% gain** |
| `coordinate_descent_tuning` | wash | wash |
| `triton.cudagraphs` | wash (+mem pool) | wash (+mem pool) |

`max_autotune_gemm` seems to be  best run alone (at least on sm_120) a 6-combo `f+X` probe on sm_120 found nothing beats it (`coordinate_descent_tuning` and `triton.cudagraphs` each erode it). It hits the same 101 KB shared-mem limit on both arches (consumer Blackwell ≈ Ampere here), but the autotuner nets better tiles on sm_120, and its +2.2 GB peak needs ≥32 GB.

Full sm_120 single-flag sweep (RTX 5090), showing what does and doesn't help:

| cond | config | mean ms | peak MB | Δ vs b | Δ vs c |
|---|---|---|---|---|---|
| a | eager | 928.9 | 19,992 | −0.78% | — |
| b | fused (baseline) | 936.2 | 19,992 | 0.00% | — |
| c | + `torch_compile` | 757.8 | 19,614 | −19.06% | 0.00% |
| d | + `coordinate_descent_tuning` | 763.0 | 19,614 | −18.50% | +0.69% wash |
| e | + `shape_padding`+`epilogue_fusion` | 756.0 | 19,614 | −19.24% | −0.23% wash |
| **f** | **+ `max_autotune_gemm`** | **738.3** | 21,836 | **−21.14%** | **−2.57%** |
| g | + `triton.cudagraphs` | 757.7 | 20,207 | −19.06% | −0.00% wash |
| l | + all 10 keys | 752.6 | 21,828 | −19.60% | −0.67% |

## AI Usage Disclaimer

Yes — Opus 4.8 used throughout

## Types of changes

- [x] New feature (non-breaking) — `torch_compile_options` config field; opt-in Inductor flag plumbing

## Summary by CodeRabbit

* **New Features**
  * Configurable compile-time options for model compilation exposed in training configs
  * New Qwen3-8B LoRA example demonstrating fused attention + compiled training optimizations

* **Documentation**
  * Expanded optimization guide with performance deltas, cold-start notes, and architecture-specific tuning guidance

* **Tests**
  * Added test coverage and a warning-capture helper to validate compile-option validation and related warnings (e.g., CUDA-graphs vs. sample-packing)


winglian:
@coderabbit review

docs/optimizations.qmd
```diff
@@ -82,7 +82,9 @@ fused_attn_kernel: true
...
+- **Baseline perf:** stacking `torch_compile: true` is a steady-state win that scales with arch — ~−14% to −16.5% on sm_86 (vanilla RTX 3090 / 3090 Ti), ~−19% on sm_120 (Blackwell/RTX 5090), measured on Qwen3-8B + LoRA r=16 + torch 2.11. Also trims peak memory ~2%.
+- **Compile cold-start:** adds 5–15 min on the first step.
+- **Tuning (`torch_compile_options`):** an opt-in allowlist of `torch._inductor.config` knobs — `coordinate_descent_tuning`, `coordinate_descent_check_all_directions`, `shape_padding`, `epilogue_fusion`, `max_autotune_gemm`, `fx_graph_cache`, `assume_aligned_inputs`, `comprehensive_padding`, `decompose_mem_bound_mm`, `triton.cudagraphs`. Payoff is arch-specific: on **sm_120 + ≥32 GB**, `max_autotune_gemm: true` adds ~−2.6% over plain compile (the only flag measured to help, and best run alone); the same flag **regresses ~+13% on sm_86**. `coordinate_descent_tuning` and the rest are within seed noise on every card tested. `triton.cudagraphs` needs static shapes — incompatible with `sample_packing`.
```
NanoCode012: Maybe simplify the wording here? Hard to understand from a glance

src/axolotl/core/builders/base.py
```diff
@@ -543,6 +543,17 @@ def _configure_torch_compile(self, training_args_kwargs: dict):
                 )
             if self.cfg.torch_compile_mode:
                 training_args_kwargs["torch_compile_mode"] = self.cfg.torch_compile_mode
+            if self.cfg.torch_compile_options:
```
NanoCode012, line 561: Does transformers upstream expose this? What's the standard naming for this? First time we're using `_options` I think.

NanoCode012 (commented):
I think this PR mostly ready, just some standardization + rebase needed.

docs/optimizations.qmd
NanoCode012: Similar for the included yaml config I think

thad0ctor: I consolidate qmd  yml in follow-on commit

thad0ctor:
> I think this PR mostly ready, just some standardization + rebase needed.

All set, I cleared the merge conflicts

src/axolotl/core/builders/base.py
thad0ctor: Checked both the pinned transformers (5.12.x) and current upstream HEAD: `TrainingArguments` only has `torch_compile` / `torch_compile_backend` / `torch_compile_mode` and no options field. 

Trainer builds `TorchDynamoPlugin(backend=..., mode=...)` with only those two args (trainer.py `_create_accelerator_args`), so there's no plumbing for inductor flags even though accelerate's `TorchDynamoPlugin` *does* have a field for this named `options`, matching PyTorch's own `torch.compile(..., options=...)`, whose docs use these same flags (`epilogue_fusion`, `max_autotune`, `shape_padding`, `triton.cudagraphs`) as the examples. 

HF never populates the plugin's `options` and there's no env route (`ACCELERATE_DYNAMO_OPTIONS` doesn't exist), setting `torch._inductor.config` directly is the logical path.

winglian:
1. Confusing hard-error when torch_compile: auto resolves to Fals: on a host where auto-resolution disables compile (old torch / no dynamo), a config with torch_compile: auto + torch_compile_options fails validation with "Set torch_compile: true (or torch_compile: auto) to use these flags" — but the user already has auto; it's the cause of the failure. The mode="before" validator check_torch_compile_auto (config.py:1969) overwrites the value before the new after-validator runs, so it can't tell "user set false" from "auto resolved false". Suggest either warning-and-ignoring in the auto-resolved case (inside check_torch_compile_auto, which still sees the original value), or making the message auto-aware.
  2. New example uses the deprecated flash_attention: true. Validating it emits "deprecated and will be removed in a future release" — a brand-new example should use attn_implementation: flash_attention_2.

thad0ctor:
> 1. Confusing hard-error when torch_compile: auto resolves to Fals: on a host where auto-resolution disables compile (old torch / no dynamo), a config with torch_compile: auto + torch_compile_options fails validation with "Set torch_compile: true (or torch_compile: auto) to use these flags" — but the user already has auto; it's the cause of the failure. The mode="before" validator check_torch_compile_auto (config.py:1969) overwrites the value before the new after-validator runs, so it can't tell "user set false" from "auto resolved false". Suggest either warning-and-ignoring in the auto-resolved case (inside check_torch_compile_auto, which still sees the original value), or making the message auto-aware.
> 2. New example uses the deprecated flash_attention: true. Validating it emits "deprecated and will be removed in a future release" — a brand-new example should use attn_implementation: flash_attention_2.

Thanks! just pushed changes

winglian (approved):

PR #3717, opened 2026-06-08
FIX loss diff for streaming n non streaming 

ved1beta:
# Description
pretrain (streaming) drops every document longer than sequence_len via `filter_sequences_by_length`

completion (non-streaming) splits long docs into sequence_len chunks and keeps all tokens.

## fix 
pretrain strategy chunk long documents instead of dropping them.

## Motivation and Context
#3341

## How has this been tested?
unit test enough 

## AI Usage Disclaimer

dignose claude 

## Release Notes

* **Bug Fixes**
  * Improved tokenization stride calculation for pretraining to ensure proper overlap behavior with longer sequences
  * Fixed sequence length parameter configuration to use specified value directly instead of scaled multiplier

* **Tests**
  * Added test coverage for long document handling and edge cases in pretraining tokenization

### DIFF
--- a/src/axolotl/prompt_strategies/pretrain.py
+++ b/src/axolotl/prompt_strategies/pretrain.py
@@ -30,13 +30,16 @@ class PretrainTokenizationStrategy(PromptTokenizingStrategy):
     def _tokenize(
         self, prompt: str, add_eos_token: bool = True, strip_bos_token: bool = False
     ) -> BatchEncoding:
+        # keep the overlap below the window size so small sequence_len values don't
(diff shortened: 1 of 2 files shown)


winglian (approved):

NanoCode012:
Do you have some charts to show this after fix?

Yes, this is a known limitation, we documented it on our docs too https://docs.axolotl.ai/docs/dataset-formats/#non-streaming-smaller-datasets 

However, it's good to close this gap.

ved1beta:
https://wandb.ai/vedaya-self/axolotl-pr3717/workspace?nw=nwuservedaya

PR #3725, opened 2026-06-11
fix(chat_templates): extract reasoning_content before reassigning content in qwen3_5 template

### LINKED ISSUE #3724: Qwen3.5 chat template drops reasoning for inline <think> assistant turns (two set statements swapped)
### Summary

The bundled `qwen3_5.jinja` template parses inline `<think>...</think>` reasoning incorrectly for assistant messages. Two `{%- set %}` statements are in the wrong order relative to the official Qwen template: `content` is mutated before `reasoning_content` is extracted from it. As a result the reasoning trace is silently dropped and the final answer is duplicated into the `<think>` block.

### Location

Assistant branch, the `else` path that handles inline reasoning (used when `message.reasoning_content` is not provided as a separate field):

https://github.com/axolotl-ai-cloud/axolotl/blob/22bcb9a773cb9de49bdcabca5be61c09b420b736/src/axolotl/utils/chat_templates/templates/qwen3_5.jinja#L94-L98

```jinja
{%- else %}
{%- if '</think>' in content %}
{%- set content = content.split('</think>')[-1].lstrip('\n') %}
{%- set reasoning_content = content.split('</think>')[0].rstrip('\n').split('<think>')[-1].lstrip('\n') %}
{%- endif %}
{%- endif %}
```

The official Qwen3.5-2B template does the two assignments in the opposite order:

https://huggingface.co/Qwen/Qwen3.5-2B/blob/main/chat_template.jinja#L94-L97

```jinja
{%- if '</think>' in content %}
{%- set reasoning_content = content.split('</think>')[0].rstrip('\n').split('<think>')[-1].lstrip('\n') %}
{%- set content = content.split('</think>')[-1].lstrip('\n') %}
{%- endif %}
```

### Root cause

Both assignments read `content`, so they are order-dependent:

1. `content = content.split('</think>')[-1]...` reassigns `content` to everything *after* `</think>` (the answer).
2. `reasoning_content = content.split('</think>')[0]...split('<think>')[-1]...` then reads the **already-mutated** `content`. That string no longer contains `</think>` or `<think>`, so both splits are no-ops and `reasoning_content` collapses to the answer.

The reasoning must be extracted before `content` is overwritten. The Axolotl copy has lines 96 and 97 in the reverse of the upstream order.

### Impact

For an assistant turn like `<think>Let me compute 2+2.</think>The answer is 4.`

current output:
```
<|im_start|>assistant
<think>
The answer is 4.
</think>

The answer is 4.
```

expected output:
```
<|im_start|>assistant
<think>
Let me compute 2+2.
</think>

The answer is 4.
```

The reasoning trace is dropped entirely and the answer is duplicated inside the `<think>` block. This only affects datasets where assistant reasoning is embedded inline in `content` (the `else` branch); turns passing a separate `reasoning_content` field are unaffected. For SFT on reasoning data this silently corrupts the training targets — the model is trained to put its answer inside `<think>` and then repeat it.

### Proposed fix

Swap lines 96–97 so `reasoning_content` is computed before `content` is reassigned, matching upstream.

### Secondary note (optional)

The guard also differs from upstream: Axolotl uses `message.reasoning_content is defined and message.reasoning_content is not none`, while the official template uses `message.reasoning_content is string`. Not the cause of this bug, but worth aligning while touching this block.

Anai-Guo:
## Summary

Fixes #3724.

The bundled `qwen3_5.jinja` template parses inline `<think>...</think>` reasoning incorrectly for assistant messages. In the `else` branch that handles inline reasoning, the two `{%- set %}` statements were in the wrong order relative to the official Qwen3.5 template: `content` was mutated **before** `reasoning_content` was extracted from it.

Both assignments read `content`, so they are order-dependent:

```jinja
{%- set content = content.split('</think>')[-1].lstrip('\n') %}
{%- set reasoning_content = content.split('</think>')[0].rstrip('\n').split('<think>')[-1].lstrip('\n') %}
```

After the first line, `content` no longer contains `</think>` (or `<think>`), so the second line's `content.split('</think>')[0].split('<think>')[-1]` just returns the already-truncated answer. The result: the reasoning trace is silently dropped and the answer is duplicated into the `<think>` block of the rendered prompt.

## Fix

Swap the two statements so `reasoning_content` is extracted from the original `content` **before** `content` is reassigned, matching the [official Qwen3.5 template](https://huggingface.co/Qwen/Qwen3.5-2B/blob/main/chat_template.jinja#L94-L97):

```jinja
{%- set reasoning_content = content.split('</think>')[0].rstrip('\n').split('<think>')[-1].lstrip('\n') %}
{%- set content = content.split('</think>')[-1].lstrip('\n') %}
```

## Test plan

- [ ] Render an assistant turn whose `content` contains an inline `<think>...</think>` block (and no separate `message.reasoning_content` field) and confirm the reasoning text lands in the `<think>` block and the answer follows after `</think>`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed parsing and display of reasoning content in model responses to ensure accurate extraction.

### DIFF
--- a/src/axolotl/utils/chat_templates/templates/qwen3_5.jinja
+++ b/src/axolotl/utils/chat_templates/templates/qwen3_5.jinja
@@ -93,8 +93,8 @@
             {%- set reasoning_content = message.reasoning_content %}
         {%- else %}
             {%- if '</think>' in content %}
-                {%- set content = content.split('</think>')[-1].lstrip('\n') %}
                 {%- set reasoning_content = content.split('</think>')[0].rstrip('\n').split('<think>')[-1].lstrip('\n') %}
+                {%- set content = content.split('</think>')[-1].lstrip('\n') %}
             {%- endif %}
         {%- endif %}
         {%- if loop.index0 > ns.last_query_index %}


ved1beta (approved):
good catch

NanoCode012:
Ooh, this was the same comment I had on your PR @ved1beta . I think you can choose whichever to merge first

PR #3731, opened 2026-06-12
feat(grpo): add RLOO and REINFORCE++ advantage estimators

### LINKED ISSUE #3676: Add RLOO and REINFORCE++ Advantage Estimators as GRPO Extensions

### 🔖 Feature description

 RLOO (REINFORCE Leave-One-Out) and REINFORCE++ are closely related online-RL algorithms that share the same rollout-and-reward structure but use different baselines/advantage estimation

### ✔️ Solution

-Add an advantage-estimator selector either by extending loss_type 
- Implement the differing advantage computation in the GRPO strategy/trainer

### ❓ Alternatives

NA

### 📝 Additional Context

NA

### Acknowledgements

Ayushhgit:
# Description

Closes #3676.

Adds a `trl.advantage_estimator` config option that selects the baseline used when computing GRPO advantages. RLOO and REINFORCE++ share GRPO's rollout-and-reward structure and differ only in baseline/advantage estimation, so they are implemented as a selector inside the GRPO strategy/trainers (the approach suggested in the issue) rather than as separate trainers:

- `group_mean` (default) — standard GRPO group-mean baseline; **zero behavior change** for existing configs.
- `rloo` — leave-one-out baseline over the group ([Ahmadian et al., 2024](https://huggingface.co/papers/2402.14740)).
- `reinforce_plus_plus` — global-batch baseline ([Hu, 2025](https://huggingface.co/papers/2501.03262)); pairs with `scale_rewards: batch` for the paper's normalization.

The estimator only controls the baseline. Scaling stays governed by `scale_rewards`, which the schema now accepts as TRL's string values (`group`/`batch`/`none`) in addition to the previous bool — needed to even express REINFORCE++'s batch normalization from a YAML config.

```yaml
rl: grpo
trl:
  advantage_estimator: rloo
  scale_rewards: none
```

## Implementation notes

- **`core/trainers/grpo/advantages.py` (new)** — a single `compute_advantages(rewards, num_generations, estimator, scale_rewards)` helper shared by all three GRPO trainer variants, so the estimator math lives in one place.
- **`AxolotlGRPOTrainer`** — TRL inlines the advantage computation inside its ~600-line `_generate_and_score_completions`, so instead of copying that method, the trainer stashes the gathered rewards in a `_calculate_rewards` override and recomputes advantages after `super()` returns (only when a non-default estimator is configured). The logged advantages are replaced too, so `_logs` stays consistent with what is actually trained on.
- **Sequence-parallel trainer** — its existing inline advantage block is replaced with the helper. This also fixes a latent bug: `if self.args.scale_rewards:` was always truthy because TRL's `GRPOConfig.__post_init__` normalizes `scale_rewards` to the strings `"group"`/`"none"`, so `scale_rewards: false` still divided by the group std.
- **Async trainer** — both scoring paths (full-batch and streaming) now use the helper; the streaming path additionally gains `"batch"` scaling support it previously lacked.
- **Validation** — config-parse-time errors for `advantage_estimator` with non-GRPO trainers or with `multi_objective_aggregation: normalize_then_sum`, plus warnings for non-canonical scaling pairings (RLOO is canonically unscaled; REINFORCE++ normalizes with batch statistics).

## Motivation and Context

Requested in #3676. RLOO/REINFORCE++ are commonly used GRPO alternatives (both available in TRL/verl) and need only a different baseline given axolotl's existing GRPO rollout machinery.

## How has this been tested?

New `tests/core/test_grpo_advantages.py` (19 tests, CPU-only, no model downloads):

- Estimator math vs hand-computed formulas for all three estimators × group/batch/none scaling, including the identity `A_rloo = A_grpo · G/(G−1)` for unscaled rewards and zero-mean checks.
- The `AxolotlGRPOTrainer` override: advantages and logged advantages both replaced for `rloo`, exact pass-through for `group_mean`.
- Config plumbing (`GRPOStrategy.set_training_args_kwargs`) and the new validator (rejects non-GRPO and `normalize_then_sum`, accepts valid configs).

All 19 pass locally; `ruff check`/`ruff format` clean.

## Types of changes

- New feature (non-breaking change which adds functionality)
- Bug fix (the SP-trainer `scale_rewards: false` truthiness fix noted above)

## Release Notes

* **New Features**
  * Added three advantage estimator options for GRPO training: `group_mean`, `rloo`, and `reinforce_plus_plus`.
  * Extended reward scaling configuration to support group-level, batch-level, or disabled scaling modes.

* **Documentation**
  * Updated GRPO configuration reference with new advantage estimator options and usage examples.

* **Tests**
  * Added comprehensive test coverage for advantage computation and estimator configuration validation.

### DIFF
--- a/docs/grpo.qmd
+++ b/docs/grpo.qmd
@@ -534,8 +534,31 @@ All GRPO-specific options live under the `trl:` key in your config. Standard tra
 | `epsilon` | float | `null` | PPO clipping lower bound |
 | `epsilon_high` | float | `null` | PPO clipping upper bound |
 | `loss_type` | str | `null` | Loss formulation: `grpo`, `bnpo`, or `dr_grpo` |
-| `scale_rewards` | bool | `true` | Normalize rewards by standard deviation |
+| `scale_rewards` | bool or str | `true` | Reward scaling: `group` (or `true`), `batch`, or `none` (or `false`) |
+| `advantage_estimator` | str | `group_mean` | Baseline for advantages: `group_mean` (GRPO), `rloo`, or `reinforce_plus_plus` |
 | `mask_truncated_completions` | bool | `false` | Exclude truncated completions from loss |
+
+#### Advantage estimators
+
+GRPO, RLOO, and REINFORCE++ share the same rollout-and-reward structure and differ
+only in how the baseline is subtracted from rewards. `advantage_estimator` selects
+the baseline while `scale_rewards` independently controls normalization:
+
+- `group_mean` (default): each completion is baselined against the mean reward of
+  its prompt group — standard GRPO.
+- `rloo`: each completion is baselined against the mean reward of the *other*
+  completions in its group ([RLOO](https://huggingface.co/papers/2402.14740)).
+  Canonically used with `scale_rewards: none`.
+- `reinforce_plus_plus`: completions are baselined against the mean reward of the
+  whole batch ([REINFORCE++](https://huggingface.co/papers/2501.03262)).
+  Pair with `scale_rewards: batch` for the normalization used in the paper.
+
+```yaml
+rl: grpo
+trl:
+  advantage_estimator: rloo
+  scale_rewards: none
+```
 
 ### Reward functions
 
--- a/src/axolotl/core/trainers/grpo/__init__.py
+++ b/src/axolotl/core/trainers/grpo/__init__.py
@@ -118,6 +118,9 @@ class GRPOStrategy:
         if trl.scale_rewards is not None:
             grpo_args_kwargs["scale_rewards"] = trl.scale_rewards
 
+        if trl.advantage_estimator is not None:
+            grpo_args_kwargs["advantage_estimator"] = trl.advantage_estimator
+
         if trl.loss_type is not None:
             grpo_args_kwargs["loss_type"] = trl.loss_type
         if trl.mask_truncated_completions is not None:
--- a/src/axolotl/core/trainers/grpo/advantages.py
+++ b/src/axolotl/core/trainers/grpo/advantages.py
@@ -1,0 +1,109 @@
+"""Advantage estimators for GRPO-style RL trainers.
+
+Implements the selectable baseline/advantage computations requested in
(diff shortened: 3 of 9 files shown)


Ayushhgit:
Addressed review feedback in eda3df05: explicit `group_mean` now goes through the `rl: grpo` check (still allowed with `normalize_then_sum` since it is a no-op there), the REINFORCE++ warning fires for any non-`batch` scaling, the SP trainer uses `self.scale_rewards`, and the advantages module docstring is trimmed. Added tests for the new validation behavior (21 passing).

NanoCode012:
Hello, we do have an existing PR on this #3693 . Could you share how your PR differs?

Ayushhgit:
Thanks @NanoCode012 — you're right that #3693 is the earlier PR for this feature and has priority; I missed it since it wasn't linked on the issue. Happy to close this in its favor.

Before that, two differences that might be worth folding into #3693 (your call):

1. **REINFORCE++ semantics.** In #3693, `reinforce_plus_plus` computes a per-group mean/std baseline, which is mathematically the same as the default GRPO advantage with scaling. The REINFORCE++ formulation (Hu, 2025) uses a **global-batch** baseline/normalization rather than a per-group one. This PR implements that global-batch baseline, and keeps the estimator orthogonal to `scale_rewards` (widened to accept `group`/`batch`/`none`) so the batch normalization is actually expressible from config. If the intent in #3693 is the per-group variant, it may be worth renaming to avoid confusion with the paper.

2. **Latent `scale_rewards` bug.** The sequence-parallel trainer's `if self.args.scale_rewards:` is always truthy because TRL normalizes `scale_rewards` to the strings `group`/`none`, so `scale_rewards: false` still divides by the group std. This PR routes that path through the shared helper to fix it; #3693 leaves it as-is (moot for rloo/reinforce there, but still affects the default GRPO path).

Neither is a knock on #3693 — just flagging in case they're useful. I'm glad to send the REINFORCE++ baseline + the scale_rewards fix as a small follow-up PR against main once #3693 lands, if that's helpful. Otherwise feel free to close this one.

PR #3732, opened 2026-06-12
perf(lora): route GatedDeltaNet projections through fused LoRA kernels

thad0ctor:
## Description
Routes Qwen3.5 GatedDeltaNet (linear-attention) LoRA projections (`in_proj_qkv/z/b/a`, `out_proj`) through axolotl's existing fused LoRA kernel instead of the plain peft module forward. peft's `_cast_input_dtype` round-trips the full `[B, S, hidden]` activation bf16→fp32→bf16 on every forward, gradient-checkpoint recompute, and backward (numerics-free copy traffic that scales with the activation, not the LoRA rank). The fused path keeps the activation in its compute dtype and casts only the rank-r adapters.

Added as default behavior when opt-in via the existing `lora_qkv_kernel` / `lora_o_kernel` flags are enabled. When they're off (or `sample_packing` is off, or a projection isn't LoRA-wrapped) the routing falls back to the exact peft call.

## Motivation and Context
At 9B there are 24 GatedDeltaNet layers × 5 projections on this path; eliminating the round-trip removes hundreds of MB of copy traffic per step. The mechanism is model-agnostic (any LoRA module invoked via a plain peft forward pays the same tax).

## Performance
All on RTX 5090. **Micro-benchmark** — isolated GatedDeltaNet projection set at 9B dims (hidden 4096, fwd+bwd, CUDA-event timed), fused vs peft, per GDN layer:

| seq | peft | fused | speedup | peak mem saved |
|-----|------|-------|---------|----------------|
| 4096  | 12.10 ms | 10.99 ms | 1.10× | 0.03 GiB |
| 8192  | 24.19 ms | 20.72 ms | 1.17× | 0.06 GiB |
| 16384 | 50.27 ms | 42.49 ms | 1.18× | 0.13 GiB |
| 32768 | 99.33 ms | 83.53 ms | 1.19× | 0.25 GiB |

Speedup grows with sequence length — the signature of activation-scaling copy traffic.

**Full-model** — Qwen3.5-9B LoRA (seq 16384, sample packing, gradient checkpointing, CCE), steady-state (warmup excluded):

| | throughput | ms/step |
|--|-----------|---------|
| was (main, peft GDN) | 1252 tok/s | 13086 |
| **is (fused GDN)** | **1316 tok/s** | **12453** |
| **Δ** | **+5.1%** | **−633 ms (−4.8%)** |

Qwen3.5-2B (seq 8192, 100 steps, RTX 3090), tail-50 steady-state:

| | throughput | ms/step |
|--|-----------|---------|
| was (main, peft GDN) | 1773 tok/s | 4619 |
| **is (fused GDN)** | **1877 tok/s** | **4364** |
| **Δ** | **+5.9%** | **−256 ms (−5.5%)** |

Both runs show a clean, consistent gain (9B per-window throughput fully separated; 2B tail averages +5.9%/+6.5% across two windows). The relative gain holds at small scale. One-time ~20 s triton compile on the first fused step amortizes over a real run.

## How has this been tested?
RTX 5090 + 2×RTX 3090 (torch 2.11.0, transformers 5.9.0, trl 1.5.1, CCE). "was-vs-is" = `main` vs this branch, identical config/seed; step-1 loss is the cleanest comparison (LoRA Δ = 0 at init).

**Numerics (tensor-level, `apply_lora_linear` vs peft)**

| Check | Result |
|-------|--------|
| forward + input-grad | **bit-identical** (rel-L2 = 0) |
| adapter-grad (`grad_B`) | cos 0.99999 (sub-ulp float noise) |
| DoRA branch (`use_dora`) | out cos 0.99999, grad cos 0.99999 |

**End-to-end loss parity (was-vs-is)**

| Config | step-1 loss | max Δ over run |
|--------|-------------|----------------|
| Qwen3.5-0.8B (bf16 LoRA) | bit-identical | 1.01e-2 |
| Qwen3.5-2B | bit-identical | 1.2e-2 |
| Qwen3.5-9B / seq 16384 | bit-identical | (perf run) |
| qlora 4-bit (`W_quant` path) | bit-identical | 1.3e-2 |
| FSDP2, Qwen3.5-2B, 2×3090 | bit-identical | 2.15e-2 |
| **kernels OFF — branch vs main (fallback)** | **bit-identical** | **2.05e-2** |
| gemma3 (non-GDN, fix dormant) | bit-identical | 1.8e-2 |

A same-code run-to-run baseline (0.8B, is-vs-is) gave max Δ 8.7e-3, so every Δ above sits at the nondeterminism floor.

**Coverage / safety**
- **Other GDN — qwen3_next** (`in_proj_qkvz`/`in_proj_ba`): unit test asserts it is **not** patched.
- **Qwen3.5-MoE**: shares the same routed forward (`_make_qwen3_5_gated_delta_forward`).
- **Fallback**: `lora_qkv_kernel`/`lora_o_kernel` off → exact peft path, bit-identical to `main` (above), and no `apply_<proj>` attached.
- **FSDP2**: real Qwen3.5 GDN on 2×3090 trains cleanly (DTensor unshard of GDN adapters OK); existing FSDP2 fused-kernel suite **4 passed**.
- **Regression suites**: lora-kernel patching **22 passed** (Llama / Qwen3-MoE / gemma).
- **7 new committed tests** (`test_lora_kernel_patching.py`): exclusion logic + numeric parity (incl. DoRA).

## AI Usage Disclaimer
Opus 4.8 used throughout, coderabbit feedback addressed on my fork as well

## Types of changes
- [x] Performance improvement (non-breaking, numerics-neutral)
- [x] New tests
- [ ] Breaking change

## Summary by CodeRabbit

* **New Features**
  * Extended LoRA support to linear attention modules with optimized computation paths.

* **Tests**
  * Added comprehensive test coverage for LoRA patching with linear attention modules, validating forward pass accuracy and gradient computations.


src/axolotl/monkeypatch/models/qwen3_5/modeling.py
```diff
@@ -141,13 +149,13 @@ def patched_forward(
...
+        mixed_qkv = _la_proj_fwd(self, "in_proj_qkv", hidden_states)  # [B, T, D]
 
-        z = self.in_proj_z(hidden_states)
+        z = _la_proj_fwd(self, "in_proj_z", hidden_states)
         z = z.reshape(batch_size, seq_len, -1, self.head_v_dim)
 
-        b = self.in_proj_b(hidden_states)
-        a = self.in_proj_a(hidden_states)
+        b = _la_proj_fwd(self, "in_proj_b", hidden_states)
+        a = _la_proj_fwd(self, "in_proj_a", hidden_states)
```
winglian: while lora adapters on these are probably unnecessary since they already project down and are already low rank, I'll keep this here for now.

thad0ctor:
@winglian 

great input, thank you

src/axolotl/monkeypatch/lora_kernels.py
```diff
@@ -350,6 +352,42 @@ def find_self_attn_in_layer(
...
+    apply_linear.__name__ = f"apply_{proj_name}"
+    return apply_linear
+
+
+def _make_apply_lora_gdn_in_proj(proj_names: tuple[str, ...]):
+    def apply_in_proj(self, X: torch.Tensor) -> dict:
+        return apply_lora_gdn_in_proj(self, X, proj_names)
+
+    return apply_in_proj
+
```
ved1beta, line 389: GDN projections gated on lora_qkv_kernel OR lora_o_kernel, unlike self-attn which gates independently. 
we could do something like  in_proj => lora_qkv_kernel and out_proj =>lora_o_kernel

src/axolotl/monkeypatch/lora_kernels.py
```diff
@@ -557,6 +625,12 @@ def apply_lora_kernel_patches(
                         "Cannot patch some MLP layers - requires LoRA adapters"
                     )
 
+    if linear_attn_patched_layers:
+        LOG.info(
+            f"Patched linear-attention LoRA projections {linear_attn_patched_projs} "
+            f"with fused kernels on {linear_attn_patched_layers} layer(s)"
+        )
```
ved1beta, line 647: accumulate across layers instead of reassigning, so the INFO log reflects every patched projection, not just the last layer's.

src/axolotl/kernels/lora.py
```diff
@@ -1791,6 +1791,201 @@ def apply_lora_o(self, X: torch.Tensor) -> torch.Tensor:
...
+    single dropout mask be shared (and saved once for backward) across the group
+    instead of one per projection, and accumulates the input gradient in a single
+    buffer instead of N separate ``dY @ W`` reductions. Each projection is
+    independent: one may carry a LoRA adapter (and DoRA) while another is
+    base-only (A/B None).
+
+    Tensors are passed flat, ``_FUSED_PROJ_NUM_TENSORS`` per projection as
+    ``(W, b, A, B, lora_bias, magnitude)``; scales/quant-states ride alongside as
+    non-tensor tuples. Supports bias, dropout, and DoRA.
+    """
```
ved1beta, line 1814: document that the shared dropout mask diverges from peft's per-projection dropout (not bit-identical when lora_dropout > 0); parity tests only cover dropout=0. also make this sort

src/axolotl/monkeypatch/models/qwen3_5/modeling.py
```diff
@@ -110,6 +110,25 @@ def _patched_decoder_forward(
...
+
+
+def _la_in_proj_fwd(module, x):
+    """Fused shared-input GDN in-projections when patched, else plain per-proj peft."""
+    fused = getattr(module, "apply_in_proj_fused", None)
+    if fused is not None:
+        return fused(x)
+    return {name: getattr(module, name)(x) for name in _GDN_IN_PROJ_NAMES}
+
+
```
ved1beta, line 128: patched_forward hardcodes all four in_proj[...] keys, so the is not None filter on in_projs is dead defensiveness; drop it and build the group from LINEAR_ATTN_IN_PROJS (and dedupe
  _GDN_IN_PROJ_NAMES against it).

ved1beta (changes requested):

src/axolotl/monkeypatch/models/qwen3_5/modeling.py
thad0ctor: dropped the filter and build the group straight from LINEAR_ATTN_IN_PROJS. Deduped _GDN_IN_PROJ_NAMES away; modeling.py now imports the single LINEAR_ATTN_IN_PROJS (verified no import cycle)

src/axolotl/kernels/lora.py
thad0ctor: Documented the shared-mask divergence in the docstring (not bit-identical to peft at lora_dropout>0) and trimmed the docstrings. 

Also added a backward-under-dropout parity test to capture this in the future

src/axolotl/monkeypatch/lora_kernels.py
thad0ctor: All set,  switched linear_attn_patched_projs to a set accumulated across layers and log sorted(...), so it reflects every patched projection, not just the last layer's

src/axolotl/monkeypatch/lora_kernels.py
thad0ctor: split it to mirror self-attn: in_proj now gates on lora_qkv_kernel and out_proj on lora_o_kernel independently. Added a parametrized test covering both single-flag cases.

ved1beta (approved):
lgtm

thad0ctor:
Resolved merge conflicts [08f013b](https://github.com/axolotl-ai-cloud/axolotl/pull/3732/commits/08f013b8e6b646ea875105ba410a924033515bc1)****

winglian (approved):

PR #3733, opened 2026-06-13
feat(offload): hidden_states activation offloading + fix legacy/full-param paths

winglian:
## Summary

Adds `activation_offloading: hidden_states` — an ALST-style activation offloader — and fixes the existing TRL-based offload paths.

`hidden_states` is gradient checkpointing that offloads **only the per-layer input** (`hidden_states`) to CPU and recomputes the intermediates, overlapping the transfer with compute on a side stream (async d2h with bounded in-flight + backward h2d prefetch). Unlike the existing TRL offloader (which offloads *every* saved tensor and is PCIe-saturated for full-parameter training), this moves one tensor/layer, so PCIe stays within budget and at long sequence the copy hides behind the recompute.

It replaces torch's reentrant `CheckpointFunction`, so it is framework-agnostic (no DeepSpeed dependency) and DTensor-aware (sequence/context parallel).

## Why

For long-context **full-parameter** training, offloading *everything* (existing `activation_offloading: true`) is slower and uses *more* GPU memory than plain gradient checkpointing (the offload backlog exceeds PCIe bandwidth). Offloading only the checkpoint input is the right granularity.

## Results (Qwen3-8B full-param, single GPU, bf16)

| seq | GC-only | hidden_states | vs GC |
|---|---|---|---|
| 16k | 63.3 GB / 2325 tok/s | 58.9 GB / 2243 | 0.96× speed, 1.08× less mem |
| 32k | 73.1 GB / 1954 | 64.3 GB / 2000 | 1.02× speed, 1.14× less mem |
| 64k | 92.9 GB / 1605 | 75.7 GB / 1565 | 0.98× speed, 1.23× less mem |
| 128k | **OOM** | fits | reaches context where GC OOMs |

Speed is tied with GC; the memory advantage widens with sequence length.

## Also fixes the existing TRL offload paths

- **Decouple offload from recompute.** `_apply_activation_checkpointing` is now adapter-aware: LoRA offloads *instead of* recomputing (pure offload is leaner and faster — recompute pins the offloaded tensors and balloons memory); full finetune keeps recompute and offloads only the checkpoint boundaries.
- **Wire `legacy`/`disk`.** These string modes were dead (gated on `is True`, and the trainer arg type `bool | None` couldn't hold the string). `legacy` now selects the synchronous (non-streamed) TRL offloader.

## Validation

- Bit-exact gradients vs plain checkpointing (forward output, input grad, weight grads all `0.0` in fp32).
- **FSDP2, 2×GPU, full-parameter:** runs clean under ZeRO-3 (`reshard_after_forward: true`) and ZeRO-2 (`false`) — reentrant checkpoint + offload + FSDP2 compose. (ZeRO-1 is not an FSDP2 sharding mode.)
- Validation forces `use_reentrant=True` for `hidden_states` and warns when used with LoRA/QLoRA.
- Tests: new `test_hidden_states_offload_full_param`, `test_offload_mode_wiring`, `test_offload_is_adapter_aware` + existing offloading/leak tests (11 pass).

## Usage

```yaml
gradient_checkpointing: true
activation_offloading: hidden_states   # full-parameter long-context
```

## Summary by CodeRabbit

* **New Features**
  * Added a new activation offloading mode that keeps checkpointed hidden states on CPU with streamed transfer support.
  * Expanded activation offloading options to support additional modes, including legacy and disk-based behavior.

* **Bug Fixes**
  * Improved checkpointing behavior so the correct offloading and recomputation path is selected automatically.
  * Added safeguards for compatible checkpointing settings when hidden-states offloading is enabled.

### DIFF
--- a/src/axolotl/core/trainers/mixins/activation_checkpointing.py
+++ b/src/axolotl/core/trainers/mixins/activation_checkpointing.py
@@ -30,13 +30,18 @@ class ActivationOffloadingMixin(Trainer):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         if self.args.activation_offloading:
+            # "legacy" uses the previous synchronous implementation (no CUDA
+            # streams), which keeps far fewer activations resident on-GPU than
+            # the stream-overlapped path (True/"disk"); the streams path stashes
+            # several activations to overlap copies, inflating peak reserved.
+            use_streams = self.args.activation_offloading != "legacy"
             if isinstance(self.model, PeftModel):
                 self.activation_offload_context = get_lora_act_offloading_ctx_manager(
-                    self.model, use_streams=True
+                    self.model, use_streams=use_streams
                 )
             else:
                 self.activation_offload_context = get_act_offloading_ctx_manager(
-                    self.model, use_streams=True
+                    self.model, use_streams=use_streams
                 )
         else:
             self.activation_offload_context = contextlib.nullcontext()
--- a/src/axolotl/core/training_args_base.py
+++ b/src/axolotl/core/training_args_base.py
@@ -235,9 +235,12 @@ class AxolotlTrainingMixins:
         },
     )
 
-    activation_offloading: bool | None = field(
+    activation_offloading: bool | str | None = field(
         default=None,
-        metadata={"help": "Use activation offloading with CUDA streams for training."},
+        metadata={
+            "help": "Activation offloading mode: True (stream-overlapped), "
+            "'legacy' (synchronous), 'disk', or False."
+        },
     )
 
     layer_offloading: bool | None = field(
--- a/src/axolotl/core/training_args_base.py
+++ b/src/axolotl/core/training_args_base.py
@@ -3,7 +3,7 @@ Base Axolotl Training Mixins shared across various trainer configs
 """
 
 from dataclasses import dataclass, field
-from typing import Optional
+from typing import Literal, Optional
 
 from PIL.Image import Resampling
 
@@ -235,7 +235,9 @@ class AxolotlTrainingMixins:
         },
     )
 
-    activation_offloading: bool | str | None = field(
+    # "hidden_states" never reaches the trainer (it's handled via the checkpoint
+    # monkeypatch in the model loader), so the trainer only sees these values.
+    activation_offloading: Literal["legacy", "disk"] | bool | None = field(
         default=None,
         metadata={
             "help": "Activation offloading mode: True (stream-overlapped), "
--- a/src/axolotl/monkeypatch/activation_offload_checkpoint.py
+++ b/src/axolotl/monkeypatch/activation_offload_checkpoint.py
@@ -1,0 +1,255 @@
+"""Activation-checkpoint input ("hidden_states") offloading, ALST-style.
+
+Gradient checkpointing already discards a layer's intermediate activations and
+recomputes them in backward; it still keeps the *checkpoint input* (the layer's
+hidden_states) resident. For long sequences those per-layer inputs
+(num_layers x [seq, hidden]) dominate GPU memory. This offloads only that one
+tensor per checkpoint to CPU and brings it back for the backward recompute —
+minimal PCIe traffic (one tensor/layer, not every activation).
+
+The copies are overlapped with compute on a side stream (forward async d2h with
+bounded in-flight + backward h2d prefetch), so at long seq the transfer hides
+behind the recompute.
+
+It replaces torch's reentrant ``CheckpointFunction`` (use_reentrant=True), so it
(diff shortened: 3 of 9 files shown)


src/axolotl/core/trainers/mixins/activation_checkpointing.py
```diff
@@ -30,13 +30,18 @@ class ActivationOffloadingMixin(Trainer):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         if self.args.activation_offloading:
```
NanoCode012, line 32: Could we update the https://docs.axolotl.ai/docs/gradient_checkpointing.html docs as well for this new function and when to choose each?

src/axolotl/utils/schemas/config.py
```diff
@@ -604,11 +604,18 @@ class AxolotlInputConfig(
             "description": "Additional kwargs to pass to the trainer for gradient checkpointing"
         },
     )
-    activation_offloading: Literal["legacy", "disk"] | bool | None = Field(
-        default=False,
-        json_schema_extra={
-            "description": "Whether to offload activations. Available options are: true, false, 'legacy', 'disk'."
-        },
+    activation_offloading: Literal["legacy", "disk", "hidden_states"] | bool | None = (
```
NanoCode012, line 607: I think in patch manager, we check for "offload_disk" whereas the pydantic model uses `disk`

src/axolotl/monkeypatch/activation_offload_checkpoint.py
```diff
@@ -0,0 +1,255 @@
...
+            gpu_t, ev = self.prefetched.pop(tid)
+        else:
+            gpu_t, ev = self._bring_back(tid)
+        nxt = (
+            tid - 1
+        )  # next input needed (reverse order) — prefetch to overlap recompute
+        if nxt >= 0 and nxt in self.cpu and nxt not in self.prefetched:
+            self.prefetched[nxt] = self._bring_back(nxt)
+        torch.cuda.current_stream().wait_event(ev)
+        _, _, requires_grad = self.cpu[tid]
```
NanoCode012, line 112: This one's Claude review, follows upstream trl to prevent a use-after-free issue
```
gpu_t.record_stream(torch.cuda.current_stream())
```

src/axolotl/core/training_args_base.py
```diff
@@ -235,9 +235,12 @@ class AxolotlTrainingMixins:
         },
     )
 
-    activation_offloading: bool | None = field(
+    activation_offloading: bool | str | None = field(
```
NanoCode012: Should we set the same exact literal as in our validator to prevent drift?

tests/e2e/test_activation_offloading.py
```diff
@@ -81,6 +81,172 @@ def test_activation_offloading(
...
+            # validator forces reentrant for hidden_states
+            assert cfg.gradient_checkpointing_kwargs["use_reentrant"] is True
+            normalize_config(cfg)
+            dataset_meta = load_datasets(cfg=cfg)
+            train(cfg=cfg, dataset_meta=dataset_meta)
+            assert ckpt.CheckpointFunction is HiddenStatesOffloadCheckpoint
+            check_model_output_exists(temp_dir, cfg)
+        finally:
+            unpatch_hidden_states_offload()
+
```
NanoCode012, line 249: Should we add a tiny model bit correctness check between regular and this offloading ckpt?

src/axolotl/monkeypatch/activation_offload_checkpoint.py
```diff
@@ -0,0 +1,255 @@
...
+        self.cpu = {}  # id -> (cpu_tensor, device, requires_grad)
+        self.prefetched = {}  # id -> (gpu_tensor, event)
+
+    # ---- forward ----
+    def offload(self, t):
+        s0 = torch.cuda.current_stream()
+        self.s1.wait_stream(s0)  # input must be produced before we copy it
+        with torch.cuda.stream(self.s1):
+            cpu_t = t.detach().to("cpu", non_blocking=True)
+        ev = self.s1.record_event()
```
NanoCode012, line 73: Claude's suggestion, potential perf improvement:

```
torch.empty(..., pin_memory=True) + .copy_(t, non_blocking=True)
```

upstream TRL creates an empty pinned buffer and copies to it unlike in this case `.to("cpu", non_blocking=True)` where it targets a pageable memory and is sync

winglian: pinned memory usage is way worse and seems to provide no improvement on any axis.

src/axolotl/core/training_args_base.py
winglian: this is what we send to HF's TrainingArgs, so it differs from our config.

PR #3735, opened 2026-06-15
fix(diffusion): exclude padding from bidirectional attention mask

Ayushhgit:
# Description

Fixes a correctness bug in the diffusion LM integration: in non-packed (padded) batches, real tokens attended to padding positions in the bidirectional attention mask, corrupting their representations and the training loss.

`create_bidirectional_attention_mask` short-circuited to an all-to-all mask whenever `sample_packing` was `False`:

```python
if attention_mask is None or not sample_packing:
    return torch.ones(batch_size, 1, seq_len, seq_len, dtype=torch.bool, device=device)
```

The returned 4D mask is passed to the model as `attention_mask`, which **replaces** the model's own padding handling. So with `sample_packing: false` (the default) and a padded batch, padding tokens were fully attended to by every real token in both directions. For a bidirectional (diffusion) model this contaminates the hidden states of all real tokens and therefore the loss/gradients.

## Why training was affected but generation wasn't

The two call sites disagreed:

- **Training** (`DiffusionTrainer._compute_diffusion_loss`) passes `sample_packing=self.axolotl_cfg.sample_packing` — `False` by default, so the buggy branch fired on every padded batch.
- **Generation** (`generation.py`) already worked around it by passing `sample_packing=(attention_mask is not None)`, i.e. it forced the masked path whenever a mask existed.

So generation effectively had the correct behavior; training did not.

## Fix

Honor the `attention_mask` whenever it is provided, regardless of `sample_packing`. The existing rule

```python
(mask_i == mask_j) & (mask_i > 0)
```

already produces the correct mask for **both** cases:

- binary padding mask (`[1,1,1,0]`) → valid↔valid only, padding excluded;
- packed segment ids (`[1,1,2,2]`) → same-sample only, cross-sample blocked.

Only the early-return guard needed to change (`not sample_packing` removed). No behavior change for the `attention_mask is None` path (still full all-to-all) or for the existing packed path.

## How has this been tested?

`tests/integrations/test_diffusion.py`:

- New `test_bidirectional_attention_mask_excludes_padding` — non-packed batch with a binary padding mask asserts no attention to/from the padding position.
- Existing `test_bidirectional_attention_mask_no_packing` (no mask → all-ones) and `..._with_packing` (segment ids) still pass.

Full file: 13 passed locally. `ruff check`/`ruff format` clean.

## Types of changes

- Bug fix (non-breaking change which fixes an issue)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed attention mask behavior to consistently exclude padding tokens during diffusion model training. The system now reliably prevents padding tokens from interfering with attention calculations, improving training consistency.

* **Tests**
  * Added test coverage verifying proper exclusion of padding tokens in attention mask operations.

### DIFF
--- a/src/axolotl/integrations/diffusion/utils.py
+++ b/src/axolotl/integrations/diffusion/utils.py
@@ -129,13 +129,17 @@ def create_bidirectional_attention_mask(
 ) -> torch.Tensor:
     """
     Create bidirectional attention mask to override default causal masking.
-    Handles sample-packed sequences where different samples are identified
-    by different attention mask values.
+
+    When an ``attention_mask`` is provided it is always honored, whether it is a
+    binary padding mask (non-packed batches) or per-sample segment ids (packed
+    batches): a query may attend to a key only if they share the same non-zero
+    value. Passing the 4D mask to the model replaces its own padding handling,
+    so padding must be excluded here or real tokens would attend to padding.
 
     Args:
         input_ids: Input token ids [batch_size, seq_len]
         attention_mask: Attention mask [batch_size, seq_len]
-        sample_packing: Whether sample packing is enabled
+        sample_packing: Unused; segment-vs-padding masks are handled uniformly.
 
     Returns:
         bidirectional_mask: 4D attention mask [batch_size, 1, seq_len, seq_len]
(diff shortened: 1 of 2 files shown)


NanoCode012:
I've been reading this to understand it better, still unfamiliar fully with the diffusion side of things.

Does this mean, the issue doesn't occur with packing on? You mentioned it only occurs with full padded sequences, which shouldn't happen? When does it even happen?

winglian (approved):

PR #3738, opened 2026-06-16
fix(offload): make activation_offloading: disk actually offload to disk

winglian:
## Summary

`activation_offloading: disk` was effectively dead — it silently behaved like `true` (TRL CPU offloader) and the `Disco` disk checkpointer never ran. This PR makes `disk` actually offload activations to disk, and fixes a memory regression in the disk path that made it look strictly worse than plain gradient checkpointing.

## What was broken

`disk` never reached the Disco disk checkpointer because of three independent issues:

1. **Stale string check** — `patch_manager` gated the Disco patch on `activation_offloading == "offload_disk"`, but after validation the value is `"disk"`, so the branch never matched.
2. **Builder disabled gradient checkpointing** — any truthy `activation_offloading` (including `"disk"`) was routed into the TRL offloader with `gradient_checkpointing=False`, so HF never enabled the checkpointing that drives the Disco patch.
3. **Model loader double-wrapped** — for full-finetune it also applied the TRL recompute wrap (`ac_wrap_hf_model`), colliding with Disco.

Net effect: `disk` did stream-overlapped CPU offload (identical to `true`), while the docs claimed disk offload.

## Changes

- **Wiring**: `disk` now keeps HF gradient checkpointing enabled, installs the Disco patch, and skips the TRL offloader / recompute wrap.
- **Disco prefetch fix**: defaulted `prefetch_to_gpu=False`. The previous default did the host→device copy on a background thread (`torch.load(map_location="cuda")`), allocating off the main stream and fragmenting the caching allocator — reserved memory climbed ~5 GB above plain GC while *allocated* was actually ~1.3 GB below it. The background thread now loads to CPU and the existing `load_tensor` moves it to GPU on the main thread (no change to the save/recompute/cleanup ordering).
- **Disco prefetch-worker bug**: fixed a missing `continue` that caused `task_done() called too many times` (and a latent `KeyError`); only surfaced once disk actually ran.
- Schema description + mixin comment corrected; `disk` is no longer mislabeled "TRL offloader".
- **Tests**: added `test_disk_offload_wiring` (asserts the Disco patch is installed, TRL context is off, training runs and writes output) and removed the now-incorrect `("disk", True)` case from the TRL wiring test.

## Benchmarks

Qwen2.5-1.5B, full-parameter, micro-batch 1, sample packing, 8 steps (median of 5 post-warmup), single RTX PRO 6000. Profiled allocated vs reserved to isolate the fragmentation.

**Disco prefetch fix — multi-step alloc vs reserved @ 16k (the regression):**

| variant | ms/step | reserved GB | allocated GB | loss |
|---|--:|--:|--:|--:|
| none (plain GC) | 1680 | 48.82 | 42.61 | 12.63787 |
| disk `prefetch_to_gpu=True` (before) | 2276 | 50.25 | 41.53 | 12.63757 |
| disk `prefetch_to_gpu=False` (after) | 2237 | **48.88** | **41.34** | 12.63750 |

Reserved drops to ≈none, allocated stays below none, marginally faster, loss unchanged.

**Real `train()` path, chunked cross-entropy (logits removed to expose activation memory):**

| seq | none (resv / ms) | disk after fix (resv / ms) | disk before fix (resv) |
|--:|--:|--:|--:|
| 16384 | 25.64 / 1170 | 25.66 / 1755 | 27.38 |
| 32768 | 40.76 / 2362 | 39.59 / 3572 | (higher) |

After the fix `disk` is no longer worse on memory: reserved ≈ none (below it at 32k) and allocated ~1.3 GB under none.

**Tradeoff note:** `disk` remains ~1.3–1.5x slower than no-offload. That is inherent disk I/O bandwidth (CPU↔disk), the expected cost of the mode — it exists to fit longer context when CPU RAM is exhausted, not to be fast. On this 1.5B model the per-layer-input saving is small (~1.3 GB); the benefit grows with deeper models / longer context where those inputs dominate.

## Testing

`tests/e2e/test_activation_offloading.py` full suite passes (10 passed), including the new `test_disk_offload_wiring`; `true`/`legacy`/`lora`/`qlora`/`hidden_states` paths unaffected. No thread errors from Disco after the prefetch-worker fix.

## Not in scope

The `legacy` patch_manager branch (`CPU_Offloaded` wrapper) is similarly vestigial now that `legacy` = TRL-synchronous; left untouched to keep this PR scoped to `disk`. A CUDA-stream rewrite of Disco's GPU↔CPU copies could overlap that leg for more speed, but the disk-I/O floor remains and it carries correctness risk against the current hardened implementation — deferred.

## Release Notes

* **Bug Fixes**
  * Improved handling of disk activation offloading to prevent unintended duplicate optimization paths.
  * Fixed early exit logic in background prefetch worker to prevent incorrect queue accounting.

* **Documentation**
  * Expanded configuration documentation for activation offloading options, clarifying available modes and their behaviors.

* **Tests**
  * Added dedicated test coverage for disk offloading mode to ensure proper patch installation and behavior isolation.

### DIFF
--- a/src/axolotl/monkeypatch/gradient_checkpointing/offload_disk.py
+++ b/src/axolotl/monkeypatch/gradient_checkpointing/offload_disk.py
@@ -49,13 +49,18 @@ class DiskOffloadManager:
     def __init__(
         self,
         prefetch_size: int = 3,
-        prefetch_to_gpu: bool = True,
+        prefetch_to_gpu: bool = False,
         save_workers: int = 4,
     ):
         """
         Args:
             prefetch_size: Maximum number of tensors to prefetch in the background.
             prefetch_to_gpu: Whether to prefetch tensors directly to GPU memory.
+                Default False: prefetching to GPU does the host->device copy on a
+                background thread (torch.load map_location="cuda"), allocating off
+                the main stream and fragmenting the caching allocator (inflates
+                reserved memory by several GB). With False the background thread
+                loads to CPU and load_tensor moves it to GPU on the main thread.
             save_workers: Maximum number of concurrent save operations.
         """
         self.temp_dir = tempfile.mkdtemp(prefix="disco_")
@@ -435,7 +440,7 @@ class Disco(torch.autograd.Function):
     _manager = None
 
     @staticmethod
-    def get_instance(prefetch_size=1, prefetch_to_gpu=True, save_workers=4):
+    def get_instance(prefetch_size=1, prefetch_to_gpu=False, save_workers=4):
         """Get or create the offload manager"""
         if Disco._manager is None:
             Disco._manager = DiskOffloadManager(
@@ -453,7 +458,7 @@ class Disco(torch.autograd.Function):
         hidden_states,
         *args,
         prefetch_size=1,
-        prefetch_to_gpu=True,
+        prefetch_to_gpu=False,
         save_workers=4,
     ):
         """Forward pass that offloads activations to disk asynchronously"""
--- a/src/axolotl/monkeypatch/gradient_checkpointing/offload_disk.py
+++ b/src/axolotl/monkeypatch/gradient_checkpointing/offload_disk.py
@@ -167,6 +167,9 @@ class DiskOffloadManager:
                         or self.file_status[file_path] == "deleted"
                     ):
                         self.prefetch_queue.task_done()
+                        # without this, we fall through and either double-count
+                        # task_done() or KeyError on the status checks below
+                        continue
                     if file_path in self.prefetch_cache:
                         self.prefetch_queue.task_done()
                         continue
--- a/tests/e2e/test_activation_offloading.py
+++ b/tests/e2e/test_activation_offloading.py
@@ -84,15 +84,16 @@ class TestActivationOffloading:
 
     @pytest.mark.parametrize(
         "offload_mode,expect_streams",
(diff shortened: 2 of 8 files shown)


src/axolotl/monkeypatch/gradient_checkpointing/offload_disk.py
```diff
@@ -49,13 +49,18 @@ class DiskOffloadManager:
     def __init__(
         self,
         prefetch_size: int = 3,
-        prefetch_to_gpu: bool = True,
+        prefetch_to_gpu: bool = False,
```
NanoCode012, line 52: I don't think we provide a way to configure this right, so it's internal only?

tests/e2e/test_activation_offloading.py
```diff
@@ -144,6 +145,71 @@ def capture_step(self, *args, **kwargs):
...
+                "save_first_step": False,
+            }
+        )
+        try:
+            cfg = validate_config(cfg)
+            normalize_config(cfg)
+            dataset_meta = load_datasets(cfg=cfg)
+            train(cfg=cfg, dataset_meta=dataset_meta)
+        finally:
+            transformers.modeling_utils.checkpoint = original_ckpt
```
NanoCode012, line 203: I'm wondering if there's a better way to test without having to boot up training, like hook into trainer to detect the ckpt set?

PR #3607, opened 2026-04-16
Feat: add support for datasets with `str` saved `messages` field

brightwind26:
This PR supports datasets that saves `messages` field in `str` format. This happens for agentic datasets with large traces such as https://huggingface.co/datasets/allenai/Sera-4.6-Lite-T2

To reproduce, simply run `axolotl preprocess config.yaml --debug` with the config below:

```yaml
base_model: Qwen/Qwen3.5-0.8B
# Automatically upload checkpoint and final model to HF
# hub_model_id: username/custom_model_name

datasets:
  - path: allenai/Sera-4.6-Lite-T2
    type: chat_template
    field_messages: messages
    split: "train[:1%]"
    train_on_eot: turn
    roles_to_train: ["assistant"]
    message_property_mappings:
      role: role
      content: content
    message_field_training: "train"
    roles:
      assistant:
        - gpt
        - model
        - assistant
      user:
        - human
        - tool
        - user

val_set_size: 0.1
output_dir: ./outputs/lora-out

adapter: lora
lora_model_dir:

sequence_len: 32768
sample_packing: true
eval_sample_packing: true


lora_r: 16
lora_alpha: 32
lora_dropout: 0.05
lora_target_modules:
  - gate_proj
  - down_proj
  - up_proj
  - q_proj
  - v_proj
  - k_proj
  - o_proj

wandb_project:
wandb_entity:
wandb_watch:
wandb_name:
wandb_log_model:

gradient_accumulation_steps: 2
micro_batch_size: 2
num_epochs: 1

optimizer: adamw_8bit
lr_scheduler: cosine
learning_rate: 0.0002

bf16: auto
tf32: false

gradient_checkpointing: true
resume_from_checkpoint:
logging_steps: 1
flash_attention: true

loss_watchdog_threshold: 5.0
loss_watchdog_patience: 3

warmup_ratio: 0.1
evals_per_epoch: 4
saves_per_epoch: 1
weight_decay: 0.0
special_tokens:
  pad_token: "<|end_of_text|>"

# save_first_step: true  # uncomment this to validate checkpoint saving works with your config
```

With the fixes proposed in this PR, datasets such as Sera are properly loaded and tokenized

## Summary by CodeRabbit

* **New Features**
  * Expanded input format support to accept JSON-encoded strings for tool and message specifications.
  * Enhanced batch detection to recognize both string and list-formatted prompt values.

PR #3739, opened 2026-06-17
fix: correct NameError in _get_messages when JSON decodes to non-list

JSap0914:
# Description

When `field_messages` contains a JSON string that decodes to a non-list value (e.g. a JSON object / dict), `ChatTemplateStrategy._get_messages` raises an `UnboundLocalError` instead of the intended `AssertionError` with a helpful message.

The assertion on line 1069 reads:

```python
assert isinstance(messages, list), (
    f"... got {type(message)}"   # ← 'message' is undefined here
)
```

`message` is only bound inside the `for i, message in enumerate(messages):` loop that follows. Before the loop runs, referencing `type(message)` raises `UnboundLocalError: cannot access local variable 'message' where it is not associated with a value` (Python 3.12+).

The fix is one character: `type(message)` → `type(messages)`.

## Motivation and Context

Introduced in #3607 (support for str-encoded `messages` field). A dataset where the messages JSON field is accidentally serialised as a dict rather than a list of dicts will hit this error path and receive an unhelpful crash instead of a clear validation message.

## How has this been tested?

New isolated test file `tests/prompt_strategies/test_get_messages_str_format.py` (no GPU, no network, no real tokenizer):

```
pytest tests/prompt_strategies/test_get_messages_str_format.py -v --noconftest
```

Before fix → `UnboundLocalError` on `test_non_list_json_raises_assertion_error_with_type`  
After fix  → 5/5 passed

## AI Usage Disclaimer

> GitHub Copilot / Claude assisted with drafting the test and commit message.

## Types of changes

## Release Notes

* **Bug Fixes**
  * Corrected assertion error message to accurately display the type of the provided messages variable when JSON decoding fails or results in invalid data structures.

* **Tests**
  * Added comprehensive test coverage for JSON-encoded message handling, including scenarios for valid messages, type validation, dictionary element validation, malformed JSON, and missing messages.

### DIFF
--- a/src/axolotl/prompt_strategies/chat_template.py
+++ b/src/axolotl/prompt_strategies/chat_template.py
@@ -1066,7 +1066,7 @@ class ChatTemplateStrategy(PromptTokenizingStrategy):
                 LOG.error(f"Error parsing messages as JSON. Error: {e}")
                 raise
             assert isinstance(messages, list), (
-                f"For SFT datasets that are stored in `str` format, the turns must be saved in a list of dictionaries, got {type(message)}"
+                f"For SFT datasets that are stored in `str` format, the turns must be saved in a list of dictionaries, got {type(messages)}"
             )
 
             # Extra check here to make sure decoded json is a list of dicts.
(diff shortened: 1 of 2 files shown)


NanoCode012 (approved):
Thanks for the catch. I updated the test file a bit to be more targeted.

PR #3631, opened 2026-04-29
fix: pin torchvision per matrix entry to prevent ABI drift

ved1beta:
fix the torchvision version fail 

## Summary by CodeRabbit

* **Chores**
  * Updated Docker build and CI/CD pipeline configurations to use explicit torchvision version pinning across different environments.
  * Introduced torchvision version specifications for various PyTorch configurations to ensure consistent dependency resolution during builds and testing.

PR #3740, opened 2026-06-17
fix: remove non-uv images and general docker improvements

NanoCode012:
# Description

- This PR cleans up and deprecates the old non-uv paths.
- This reduces our CI concerns and also simplifies our build pipeline.
- Remove dead workflow docker next / nightly that weren't called

There were also some intermittent issue that occurred, and this would also solve it. Closes #3631 

Adds a lot of misc general fixes to our workflow.

## Summary by CodeRabbit

* **New Features**
  * Docker images now use UV-based builds with a relocatable virtual environment instead of Miniconda + pip.

* **Documentation**
  * Updated Docker documentation to reflect the new image configurations and setup approach.

* **Refactor**
  * Consolidated Docker image variants; non-UV and UV tags now reference the same underlying images.
  * Simplified base image configurations and CI/CD workflows.


NanoCode012:
The nightly fails correctly! We just merged the gemma4 fix upstream, and this test detected the bug doesn't exist anymore for nightly

ved1beta (approved):

docker/Dockerfile-base-nightly
winglian, line 1: are we sure we were supposed to delete this one?

NanoCode012: No workflow was built against this image, it was orphaned.

Issue #3443, opened 2026-03-01
torch.distributed.DistBackendError after rank 0 is loading weights.


savitha-suresh:
### Please check that this issue hasn't been reported before.

- [x] I searched previous [Bug Reports](https://github.com/axolotl-ai-cloud/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

I am training Qwen3 235B model on 4A100 and 860 gb RAM. Rank0 is loading weights that takes more than 30 mins, other gpus need to weight for more than 30 minutes. 
ddp_timeout: 7200  should increase the timeout

### Current behaviour

I am training Qwen3 235B model on 4A100 and 860 gb RAM. Rank0 is loading weights that takes more than 30 mins, other gpus need to weight for more than 30 minutes.  But the time taken to load the weights is greater than 30 mins and because of this I get a crash .
```torch.distributed.DistBackendError: [3] is setting up NCCL communicator and retrieving ncclUniqueId from [0] via c1
0d key-value store by key '0', but store->get('0') got error: wait timeout after 1800000ms, keys: /default_pg/0//cuda//0    
[rank3]: Exception raised from doWait at /pytorch/torch/csrc/distributed/c10d/TCPStore.cpp:597 (most recent call first):    
```
Why does setting ddp_timeout: 7200 is not having any affect? How to increase the timeout?

### Steps to reproduce

ddp_timeout to 7200 should increase the nccl timeout. But its not happening. 

### Config yaml

```yaml
base_model: /home/savitha/model/qwen3-235b-thinking-2507
trust_remote_code: true

load_in_4bit: true
adapter: qlora
bnb_4bit_quant_type: nf4
bnb_4bit_compute_dtype: bfloat16

lora_r: 16
lora_alpha: 32
lora_dropout: 0.05

lora_target_modules:
  - q_proj
  - k_proj
  - v_proj
  - o_proj
chat_template: qwen3

ddp_timeout: 7200          

fsdp_version: 2
fsdp:
  - full_shard
  - auto_wrap
fsdp_config:
  activation_checkpointing: true
  use_orig_params: true
  offload_params: false                              
  sync_module_states: true
  cpu_ram_efficient_loading: true                    
  auto_wrap_policy: TRANSFORMER_BASED_WRAP
  transformer_layer_cls_to_wrap: Qwen3MoeDecoderLayer 
  state_dict_type: SHARDED_STATE_DICT
  reshard_after_forward: true                         
[...]

PR #3741, opened 2026-06-17
FIX: ddp_timeout  for weight loading

ved1beta:
# Description
 NOW when cfg.ddp_timeout is set, pass it to the early Accelerator() via InitProcessGroupKwargs(timeout=...), so the process group created before weight loading honors it.
## Motivation and Context
#3443
## How has this been tested?

timeout just works 

## AI Usage Disclaimer

claude dignosed

## Summary by CodeRabbit

* **New Features**
  * Added support for configurable DDP process-group initialization timeout, allowing users to customize the timeout duration for distributed training on Ray workers.


NanoCode012 (commented):
This seems to be an isolated ray train fix despite the original issue being regular fsdp2?

We have this env called AXOLOTL_NCC_TIMEOUT set in `def init_distributed_state():`, this is Claude's suggestion to apply it across diff env (fsdp, ray, deepspeed):
```py
  # in prepare_optim_env, before the first distributed call
  if cfg.ddp_timeout:
      os.environ.setdefault("AXOLOTL_NCCL_TIMEOUT", str(cfg.ddp_timeout))
```

ved1beta:
asked claude to have a look , fix looks sound

src/axolotl/cli/train.py
```diff
@@ -138,8 +139,16 @@ def ray_train_func(kwargs: dict):
     if cfg.deepspeed and hasattr(cfg.deepspeed, "to_dict"):
         cfg.deepspeed = cfg.deepspeed.to_dict()
 
-    # initialize accelerator before model instantiation
-    Accelerator(gradient_accumulation_steps=cfg.gradient_accumulation_steps)
+    # this call creates the process group, so ddp_timeout must be applied here
```
NanoCode012, line 195: I'm not sure you need the changes in this file now

ved1beta: dk how missed it , done !

NanoCode012: I think the model got confused. It changed the wrong file :)

PR #3730, opened 2026-06-12
fix: KTO user_defined dataset transform crashes on every documented config

Ayushhgit:
# Description

Fixes #2757.

The `user_defined.default` KTO dataset strategy (`src/axolotl/prompt_strategies/kto/user_defined.py`) crashed in every configuration, including the exact config documented in `docs/rlhf.qmd`:

1. **`completion_format` provided** (the documented usage): the fallback branch assigned the default to a misnamed `chosen_format` variable, so the variable was never defined when a `completion_format` was given, and the transform raised `NameError: cannot access free variable 'chosen_format'`.
2. **`completion_format` omitted**: the generated default placeholder (`{completion}`) did not match the `.format()` keyword used (`chosen=`), raising `KeyError: 'completion'`.
3. **Custom `field_prompt`**: prompt formatting read the hardcoded `sample["prompt"]` instead of `sample[field_prompt]`, raising `KeyError` (the DPO equivalent already does this correctly).

The function only accidentally worked with `field_completion: chosen` and default `field_prompt`.

This PR also addresses the error-reporting chain that made #2757 hard to diagnose: `prompt_strategies/base.py` swallowed the underlying exception and returned `None`, which then crashed downstream in `utils/data/rl.py` with the unhelpful `TypeError: None is not a callable object` shown in the issue. Now:

- `base.load` logs the actual exception (with traceback) in its warning.
- `_load_split` raises a clear `ValueError` naming the dataset type when the transform fails to load, instead of crashing later in `inspect.signature(None)`.

## Changes

- `src/axolotl/prompt_strategies/kto/user_defined.py`: fix `completion_format` handling and use `sample[field_prompt]`; placeholders follow the documented canonical names (`{prompt}`, `{system}`, `{completion}`).
- `src/axolotl/prompt_strategies/base.py`: include the underlying exception in the strategy-load warning.
- `src/axolotl/utils/data/rl.py`: raise an actionable error when a dataset transform fails to load.
- `tests/prompt_strategies/test_kto_user_defined.py`: new unit tests covering the documented config from `docs/rlhf.qmd`, defaults, custom field names, system-in-prompt formatting, and the loader path.

## Verification

All three failure modes reproduce on `main`:

```
NameError: cannot access free variable 'chosen_format' where it is not associated with a value in enclosing scope
KeyError: 'completion'
KeyError: 'prompt'
```

With this PR, the new tests pass:

```
tests/prompt_strategies/test_kto_user_defined.py ......  [100%]
6 passed
```

[...]

PR #3742, opened 2026-06-19
fix(dpo): user_defined strategy KeyError on custom field names

### LINKED ISSUE #2645: Axolotl complains the mapper for user_defined data type of dpo training has an issue

### Expected Behavior

user_defined data type for dpo works as documented

### Current behaviour

https://github.com/axolotl-ai-cloud/axolotl/blob/0d71b0aa5fd4a35a011a574a9233eecb4f2793bb/src/axolotl/prompt_strategies/dpo/user_defined.py#L6

This type does not work.

https://github.com/axolotl-ai-cloud/axolotl/blob/0d71b0aa5fd4a35a011a574a9233eecb4f2793bb/src/axolotl/utils/data/rl.py#L62

The exception has occurred here.

### Steps to reproduce

Use user_defined data type for dpo

### Config yaml

```yaml

```

### Possible solution

Using another data type

- [ ] Linux
- [ ] macOS
- [ ] Windows

### Python Version

3.10

### axolotl branch-commit

latest

### Acknowledgements

vineethsaivs:
# Description

Fixes #2645.

The `user_defined` DPO dataset strategy (`src/axolotl/prompt_strategies/dpo/user_defined.py`) raised `KeyError` and did not work as documented in two related ways.

**1. Custom field names (the reported bug).** When `prompt_format` / `chosen_format` / `rejected_format` are not provided, the defaults were built from the *configured field names*:

```python
chosen_format = "{" + field_chosen + "}"   # e.g. "{my_chosen}"
```

but `transform_fn` formats them with the *canonical* keyword:

```python
sample["chosen"] = chosen_format.format(chosen=sample[field_chosen])
# "{my_chosen}".format(chosen=...) -> KeyError: 'my_chosen'
```

So a config such as:

```yaml
datasets:
  - path: ...
    type:
      field_prompt: question
      field_chosen: my_chosen
      field_rejected: my_rejected
```

raised `KeyError: 'question'` (surfacing in `utils/data/rl.py`, as the issue notes). The strategy only worked when every field kept its default name.

**2. `{system}` with a missing/empty system value.** With `prompt_format: "{system} {prompt}"`, any row whose system field is missing, empty, or `None` (common in mixed datasets where only some rows carry a system prompt) fell through to the `else` branch and raised `KeyError: 'system'` on the unfilled placeholder.

## Motivation and Context

This is the DPO counterpart of the KTO `user_defined` bug fixed in #3730: the same placeholder/keyword mismatch existed in both files, and #3730 left the DPO version live.

- The default format strings now use the canonical placeholders (`{prompt}`, `{chosen}`, `{rejected}`) that match the keywords passed to `.format()`, and the system check uses `"{system}" in prompt_format`. Field names are still read via `sample[field_*]`, so custom columns are honored; only the placeholder names are canonicalized.
- When `prompt_format` contains `{system}`, a system value is always supplied via `sample.get(field_system) or ""`, so missing / empty / `None` system values render as an empty string instead of crashing (and never the literal text `"None"`).

(The merged KTO strategy has the same guarded `{system}` pattern and the same latent edge case; happy to send a follow-up to align it.)

## How has this been tested?

Added `tests/prompt_strategies/test_dpo_user_defined.py` (mirrors `test_kto_user_defined.py`), covering:

- custom `field_prompt` / `field_chosen` / `field_rejected` (the issue repro: raises `KeyError` on `main`, passes here),
- defaults with no explicit formats,
- explicit `prompt_format` / `chosen_format` / `rejected_format`,
- `{system}` formatting when present, when the field is missing, and when it is `None`,
- a non-dict `type` raising a clear `ValueError`,
- the loader path returning a callable.

The custom-field-name and missing/None-system tests fail before this change (`KeyError`) and pass after; the others pass before and after.

## Types of changes

### DIFF
--- a/src/axolotl/prompt_strategies/dpo/user_defined.py
+++ b/src/axolotl/prompt_strategies/dpo/user_defined.py
@@ -15,22 +15,19 @@ def default(cfg, dataset_idx=0, **kwargs):
     field_rejected = ds_cfg.get("field_rejected", "rejected")
     prompt_format = ds_cfg.get("prompt_format")
     if not prompt_format:
-        prompt_format = "{" + field_prompt + "}"
+        prompt_format = "{prompt}"
     chosen_format = ds_cfg.get("chosen_format")
     if not chosen_format:
-        chosen_format = "{" + field_chosen + "}"
+        chosen_format = "{chosen}"
     rejected_format = ds_cfg.get("rejected_format")
     if not rejected_format:
-        rejected_format = "{" + field_rejected + "}"
+        rejected_format = "{rejected}"
 
     def transform_fn(sample):
-        if (
-            "{" + field_system + "}" in prompt_format
-            and field_system in sample
-            and sample[field_system]
-        ):
+        if "{system}" in prompt_format:
             sample["prompt"] = prompt_format.format(
-                system=sample[field_system], prompt=sample[field_prompt]
+                system=sample.get(field_system) or "",
+                prompt=sample[field_prompt],
             )
         else:
             sample["prompt"] = prompt_format.format(prompt=sample[field_prompt])
--- a/tests/prompt_strategies/test_dpo_user_defined.py
+++ b/tests/prompt_strategies/test_dpo_user_defined.py
@@ -39,8 +39,7 @@ class TestDPOUserDefined:
         assert sample["rejected"] == "bad"
 
     def test_defaults_without_formats(self):
-        """Transform falls back to canonical {prompt}/{chosen}/{rejected} formats
-        when no explicit formats are configured."""
+        """Falls back to canonical {prompt}/{chosen}/{rejected} when no formats given."""
         cfg = make_cfg({})
         transform_fn = default(cfg)
         sample = transform_fn({"prompt": "hello", "chosen": "good", "rejected": "bad"})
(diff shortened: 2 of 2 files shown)


vineethsaivs:
Good catch, addressed in `bee606b`. The `else` branch could indeed raise `KeyError: 'system'` when `prompt_format` contains `{system}` but a row's system field is missing or empty.

Now, when `{system}` is in `prompt_format`, a system value is always supplied via `sample.get(field_system) or ""`, so missing, empty, and `None` system values all render as an empty string. I used `or ""` rather than `.get(field_system, "")` on purpose: datasets often store an absent optional system as `None`, and `.get(field_system, "")` would render that as the literal text `"None"`. Added `test_system_in_prompt_format_missing_system_field` and `test_system_in_prompt_format_none_system_field`.

Note: the merged KTO `user_defined` strategy (#3730) has the same guarded `{system}` pattern and the same latent edge case. Happy to send a follow-up to align it if you'd like.

NanoCode012 (approved):
Thanks for catching these

vineethsaivs:
Thanks for the review and the test tidy-up, @NanoCode012. The parametrized empty-system case and the custom `field_system` test read nicely. Glad to help.

PR #3743, opened 2026-06-19
perf(scattermoe): NVFP4+LoRA experts via dequant→bf16 grouped GEMM

Entrpi:
## Summary

The fused NVFP4 expert kernel re-dequantizes the 4-bit weights **inside** the grouped-GEMM K-loop (and again in the backward), which starves the tensor cores: the dequant, not the matmul, dominates. For LoRA fine-tuning — where the expert weights are **frozen** — we can dequantize each expert to bf16 **once** and run the existing fast bf16 LoRA grouped GEMM instead.

This routes NVFP4 + LoRA through **dequant→bf16 (one pass) + the bf16 LoRA fwd/bwd kernels**, behind a flag. It is **10.5× faster fwd+bwd at the expert layer on B200** (sm_100) and **2.5× on RTX PRO 6000** (sm_120), bit-exact dequant, all LoRA/dX gradients validated, and FSDP2-safe.

## The change (2 files + 1 test)

**`scattermoe_lora/experts.py`**
- `_prepare_weights_and_lora`: new first branch for `is_nv and gup_lora and down_lora`, gated by `_NVFP4_LORA_DEQUANT_BF16` (env `AXOLOTL_NVFP4_LORA_FUSED=1` reverts). Dequantizes the frozen NVFP4 experts to bf16 `[E, in, out]` and returns a **plain `Tensor`, not an `MXWeights`**. The fused-vs-bf16 dispatch in `ScatterMoELoRA.forward` keys only on `isinstance(expert_weights, MXWeights)` — returning bf16 flips `is_mx=False` and selects the plain bf16 fwd/bwd kernels. A `.recipe` (rebuilds the bf16 weight) keeps the transient **recomputed in backward**, not pinned across depth.
- **Fused bf16 backward**: `is_mx=True` force-enables the fused dX/gather backward, but the new `is_mx=False` path was defaulting it off (the slower non-fused base + Python LoRA dX). Threaded a `use_fused_bwd` flag so the bf16 path takes the fused `scatter2scatter_lora_dX`, matching the MXFP4 path. Default on; env `AXOLOTL_NVFP4_LORA_NONFUSED_BWD=1` reverts.

**`scattermoe_lora/selective_dequant_kernel.py`**
- `dequant_nvfp4_full_triton` + `_nvfp4_dequant_kernel`: a one-pass NVFP4→bf16 Triton kernel (int64 offsets, pow2-`C` constexpr shifts, E2M1 LUT, in-kernel E4M3 block-16 scale decode). Folds the optional `per_tensor_scale` **in-kernel in fp32** (scalar / per-expert `[E]` / `[E,1,1]`), single-rounding.
- Why a custom kernel: torchao's eager `NVFP4Tensor.dequantize()` materializes ~145 GB of unfused fp32 intermediates (~164 ms at E=256) — slower than the fused kernel; and `torch.compile` of the subclass `.dequantize()` *miscompiles* (fast but numerically wrong). The custom kernel is ~7 ms and bit-exact.

**`tests/.../scattermoe_lora/test_nvfp4_dequant.py`** (new, GPU-gated): asserts the dequant is **bit-exact** (`torch.equal`) with the validated linear fold for `per_tensor_scale` = None / scalar / per-expert, in fp32 and bf16.

## Validation

| | result | evidence |
|---|---|---|
| Forward FAST-vs-fused | rel 2.9e-3 (bf16 floor) | — |
| fwd+bwd grads (dX, dA/dB) | match fused ref (≤1.0e-2) | — |
| Dequant incl. `per_tensor_scale` | **bit-exact** vs validated fold (None/scalar/per-expert) | new test, 6/6 |
| Expert layer fwd+bwd | **10.5× B200 / 13.2× B200 E128 / 2.5× RTX** | — |
| Full reduced-model step (B200) | **6.38×** (experts = 93% of per-layer time) | — |
| **FSDP2** (2×RTX, single node) | sharded grads match unsharded; dX bit-exact | — |

## Design decisions

- **dequant→bf16, not fp8/fp4.** FP4×FP4-native wins (3.3×) only on the bandwidth-bound RTX PRO 6000 *and* needs FP4 activations (a quality risk). bf16 is the universal, safe win. See the follow-up below for fp8-read.
- **NVFP4 only.** MXFP4 stays on the fused path (different block-32 ue8m0 scale decode). Future work.
- **Memory:** the bf16 transient is ~2× the packed-FP4 bytes (+~5.6 GB/layer at E=128, +~11 GB at E=256), recomputed-in-backward via `.recipe` so it isn't pinned across depth. The flag lets users trade speed for memory.

## Honest caveats

- **FSDP2 validated** on 2×RTX PRO 6000 (single node); not yet exercised multi-node (cross-node NCCL) — the sharding logic is identical.
- The baseline is the as-shipped (autotuned) fused kernel.
- **Dense routing assumed** (LoRA fine-tuning at non-trivial batch×seq hits all experts): the path dequantizes all E experts. A sparse-routing selective dequant would help tiny-batch/eval — deferred.

## Follow-up opportunity (not in this PR): fp8-read grouped GEMM

The bf16 path reads the full bf16 weight each pass. A custom **fp8-read** grouped kernel — read the fp8-packed weight (~half the bytes), upcast to bf16 **in-register**, bf16 MMA — halves the weight bandwidth. Measured (forward GEMM microbench): **1.79× on RTX PRO 6000** (bandwidth-bound) but only **~1.1–1.15× on B200/B300** (not bandwidth-bound at HBM3e). Capturing it needs a custom fp8-read kernel for **both** forward and the dX backward (and both weight orientations) plus end-to-end validation. Deferred because the datacenter payoff is small and the win is workstation-skewed; the bf16 path here is the load-bearing, universally-validated win.

### DIFF
--- a/src/axolotl/integrations/kernels/libs/scattermoe_lora/selective_dequant_kernel.py
+++ b/src/axolotl/integrations/kernels/libs/scattermoe_lora/selective_dequant_kernel.py
@@ -177,3 +177,110 @@ def selective_dequant_nf4_triton(
     )
 
     return out.reshape(num_active, *expert_shape)
+
+
+# --- NVFP4 (E2M1 + E4M3 block-16) fast dequant -----------------------------------------
+# torchao's eager ``NVFP4Tensor.dequantize()`` materializes large unfused fp32 intermediates
+# (~145 GB / ~164 ms at E=256) and torch.compile miscompiles its subclass dispatch, so the
+# NVFP4+LoRA fast path needs a dedicated kernel. One memory-bound pass: unpack the E2M1
+# nibble, gather the codebook, multiply the linear E4M3 block-16 scale. Validated bit-exact
+# vs torchao at E up to 256. (per_tensor_scale is folded outside the kernel.)
+#
+# OCP E2M1 codebook (low nibble first); ``scale`` is the un-swizzled linear E4M3 block scale.
+_NVFP4_E2M1_LUT = [0.0, 0.5, 1.0, 1.5, 2.0, 3.0, 4.0, 6.0,
+                   -0.0, -0.5, -1.0, -1.5, -2.0, -3.0, -4.0, -6.0]
+
+
+@triton.jit
+def _nvfp4_dequant_kernel(
+    QDATA,  # [R, C/2] uint8 packed E2M1 (2 nibbles/byte, low=col 2j, high=col 2j+1)
+    SCALE,  # [R, C/16] E4M3 linear block scale
+    LUT,    # [16] fp32 E2M1 codebook
+    OUT,    # [R*C] output, dtype = OUT.dtype.element_ty
+    PTS,    # fp32 per_tensor_scale: [1] (scalar) or [E] (per-expert); dummy ptr if unused
+    total,
+    C: tl.constexpr,    # last (contracted) dim; power-of-2 so //C,%C,>>1,>>4 are shifts
+    C2: tl.constexpr,   # C // 2
+    C16: tl.constexpr,  # C // 16
+    ROWS_PER_E: tl.constexpr,  # rows per expert (R // E); for per-expert pts indexing
+    PTS_MODE: tl.constexpr,    # 0 none, 1 scalar, 2 per-expert
+    BLOCK: tl.constexpr,
+):
+    # flat 1D over output elements; int64 offsets (r*C2 reaches ~4e9, int32 overflows).
+    g = tl.program_id(0).to(tl.int64) * BLOCK + tl.arange(0, BLOCK).to(tl.int64)
+    m = g < total
+    r = g // C
+    c = (g % C).to(tl.int32)
+    q = tl.load(QDATA + r * C2 + (c >> 1), mask=m, other=0)
+    nib = tl.where((c & 1) == 1, (q >> 4) & 0xF, q & 0xF).to(tl.int32)
+    val = tl.load(LUT + nib)
+    sc = tl.load(SCALE + r * C16 + (c >> 4), mask=m, other=0.0).to(tl.float32)
+    # Fold per_tensor_scale onto the block scale in fp32, then a single bf16 round at the
+    # store. Folding onto `sc` (not the product) reproduces the validated path's exact
+    # `block_scale * per_tensor` order (mx_weights.py) -> bit-identical, single-rounding.
+    # (A post-store bf16 multiply double-rounds, ~2x the error.)
+    if PTS_MODE == 1:
+        sc = sc * tl.load(PTS).to(tl.float32)
+    elif PTS_MODE == 2:
+        e = r // ROWS_PER_E
+        sc = sc * tl.load(PTS + e, mask=m, other=0.0).to(tl.float32)
+    tl.store(OUT + g, (val * sc).to(OUT.dtype.element_ty), mask=m)
+
+
+def dequant_nvfp4_full_triton(nv_param, dtype: torch.dtype = torch.bfloat16) -> torch.Tensor:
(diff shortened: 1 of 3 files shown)


src/axolotl/integrations/kernels/libs/scattermoe_lora/selective_dequant_kernel.py
```diff
@@ -177,3 +177,110 @@ def selective_dequant_nf4_triton(
...
+    E4M3-block-16 layout the axolotl loader produces (see ``selective_nvfp4_weights_fwd``),
+    folding the optional ``per_tensor_scale`` in-kernel in fp32 (single rounding, torchao
+    parity). Accepts a scalar pts (the loader's shape) or a per-expert ``[E]`` / ``[E,1,1]``
+    (the FSDP carry shape).
+    """
+    qd, sc = nv_param.qdata, nv_param.scale
+    lead = qd.shape[:-1]
+    C2 = qd.shape[-1]
+    C = C2 * 2
+    assert C & (C - 1) == 0, f"NVFP4 fast dequant needs power-of-2 last dim, got {C}"
```
ved1beta, line 243: relax to assert C % 16 == 0 (correct, slightly slower div/mod on non-pow2) — or keep the pow2 fast path but fall back to the fused path when C isn't pow2 instead of asserting. Either way, add a non-pow2 K
  case to test_nvfp4_dequant.py.

ved1beta (changes requested):

ved1beta:
also would love to see benchmarks for these

thad0ctor:
I think I put my comment on the wrong PR, it may make more sense here:

I've been doing a fair bit of NVFP4 training + attention work on Blackwell and tested your kernel locally against a couple of variants I've discovered through my own NVFP4 flows and wanted to share results in case they're useful, happy to fold them in as a follow-on PR if you don't want to tweak your approach.

**Recommended Dequant Method** A tweaked single-pass kernel — each lane consumes one packed byte and emits both nibbles via a contiguous 2-wide store, arithmetic E2M1 decode instead of the codebook gather, `BLOCK=8192` — runs ~4–5% faster in interleaved A/B median (up to ~10–13% on smaller, launch-bound shapes in single-run timing). It's a modest win because the kernel is already memory-bound at ~90% of bandwidth (~1.3 TB/s of a ~1.46 TB/s ceiling here) — your version is close to optimal. This method stays pure Triton, no new build dependency.

I attached the patch file and standalone replacement `.py` file for reference.

## Performance (interleaved A/B median)

| Shape (E × N × K) | this PR | tweaked | speedup |
|---|---|---|---|
| 128 × 2048 × 2048 (large, MoE-weight-like) | 1.053 ms | 1.014 ms | **1.04×** |
| 64 × 768 × 2048 | 0.217 ms | 0.199 ms | **1.09×** |
| 8 × 2048 × 2048 (small, launch-bound) | 0.067 ms | 0.056 ms | **1.20×** |

The large-tensor case is the representative one for full-expert weight dequant (memory-bound, ~4%); the smaller shapes gain more as the reduced index math and single vectorized store cut launch overhead.

**Setup:** RTX PRO 6000 Blackwell (sm120a), CUDA 13.3 / Triton 3.7 / torchao 0.18. All variants validated **bit-identical** to `NVFP4Tensor.dequantize()` and to your kernel across all three `per_tensor_scale` modes (none / scalar / per-expert).

[optionA_selective_dequant_kernel.py](https://github.com/user-attachments/files/29170416/optionA_selective_dequant_kernel.py)
[optionA_dequant_triton_tweak.patch](https://github.com/user-attachments/files/29170417/optionA_dequant_triton_tweak.patch)

PR #3744, opened 2026-06-19
perf(scattermoe): NVFP4+LoRA fp8-read split path (sm_120 memory win)

Entrpi:
> **Stacked on #3743** (the NVFP4→bf16 fast path). Until that merges, this PR's diff also shows its commit; the fp8-read commit is the top one. Rebase/retarget onto the branch once #3743 lands.

## Summary

#3743 dequantizes the frozen NVFP4 experts to **bf16** once and runs the bf16 LoRA grouped GEMM
(10.5× B200 / 2.5× RTX vs the fused FP4 kernel). This follow-up adds an **fp8-read** variant: keep
the dequantized weight in **fp8 (e4m3)** instead of bf16, read it in the grouped GEMM, and upcast to
bf16 **in-register** (bf16 MMA, bf16 activations — W8-read / A16, no fp8 activations). The fp8
transient weight is **half** the bf16 one, so it **halves the per-layer weight materialization**.

On the bandwidth-bound workstation expert GEMM this is **~1.36× lower peak memory** (≈6.3 GB/layer
saved at E=256) and **~1.07× faster** fwd+bwd — i.e. **a memory optimization that is roughly
speed-neutral**, behind a flag, default off. **sm_120 (workstation Blackwell) only.**

## The key finding: fp8-read must use the SPLIT path, not the fused LoRA kernel

A standalone grouped GEMM reads fp8 weights ~1.7–1.8× faster than bf16 (weight-bandwidth-bound,
the fp8→bf16 upcast is a near-free HW op). But routing fp8 through the **fused** LoRA kernel
(`scatter2scatter_lora`) is **0.73× — a regression**: the fused kernel's two dots + extra LoRA
A-tile load per K-iteration serialize the in-register fp8→bf16 conversion and expose its latency.

Isolating it: the **base** `scatter2scatter` (no LoRA, same scatter routing) reads fp8 at **1.70×** —
so it is the LoRA *fusion*, not the scatter structure, that breaks fp8. fp8-read therefore routes
through the **split / non-fused** path (base fp8 grouped GEMM + separate bf16 LoRA passes; non-fused
base-fp8 dX + bf16 LoRA-dX) — the opposite of the bf16 path, which prefers the fused kernel. The
isolated split-fp8 forward is ~1.27× over fused-bf16; on the full fwd+bwd this dilutes to ~1.07×
because the dA/dB grouped-Gram, the LoRA passes, and the activation work are all dtype-independent.

## The change (5 files)

- **`kernels/ops.py`** — base `_compute_expert_block`: `tl.dot(x, w.to(x.dtype), ...)` (in-register
  weight upcast; no-op for bf16/fp16, the half-bandwidth upcast for an fp8 weight). Makes the base
  grouped GEMM fp8-read-capable.
- **`kernels/lora_ops.py`** — `scatter2scatter_lora`: route an fp8 (`float8_e4m3fn`) weight to the
  existing `_scatter2scatter_lora_split` (base fp8 + bf16 LoRA) regardless of expert count.
- **`parallel_linear_lora.py`** — `ScatterMoELoRA.forward`: skip the host bf16 upcast for an fp8
  weight (it would discard the fp8-read bandwidth/memory win); the kernels upcast in-register. The
  non-fused dX path (`use_fused_dX=False`) already reads the base weight via `scatter2scatter`, now
  fp8-capable.
- **`experts.py`** — `_prepare_weights_and_lora`: new fp8-read branch (before the bf16 branch, which
  is left untouched) gated by `_NVFP4_LORA_FP8_READ` (env `AXOLOTL_NVFP4_LORA_FP8_READ=1`, default
  off). Dequant NVFP4→fp8, return the fp8 tensor with `use_fused_bwd=False` (split/non-fused) and a
  `.recipe` that rebuilds fp8 in backward.
- **`selective_dequant_kernel.py`** — `dequant_nvfp4_fp8_triton`: one-pass NVFP4→fp8 (reuses the b5
  NVFP4 dequant kernel with an fp8 store). `scale_mode="none"` (no per-expert scale — a pow2 scale
  does **not** improve output quality, b19, because the subnormal weights are small and don't move
  the output).

## Validation (RTX PRO 6000, sm_120)

| | result | evidence |
|---|---|---|
| Forward / grads fp8 vs bf16-read | 3.2e-2 / 3–9e-2 (fp8 weight-error floor) | b21, E=256 |
| Expert-layer fwd+bwd, E=256 | **1.07× speed, 1.36× peak memory** (44.9 vs 48.1 ms; 17.5 vs 23.8 GB) | b21 |
| Expert-layer fwd+bwd, E=128 | 1.01× speed, 1.34× memory | b21 |
| Base scatter2scatter fp8 vs bf16 | 1.70× (the isolated win) | b20d |
| Fused LoRA kernel fp8 (rejected) | 0.73× — why split is used | b20/b20b/b20c |
| Weight-quant error (fp8 vs bf16-read) | +0.0315 relL2 = +2.5% on the NVFP4 error | b19 |
| **FSDP2** (2×RTX, single node) | **PASS — sharded grads bit-exact** (dX + LoRA grads 0.00e+00) | b22 |

## Design decisions

- **Memory, not speed, is the value.** The full fwd+bwd is ~speed-neutral (1.07×); the durable win
  is halving the transient bf16 weight (≈6.3 GB/layer at E=256) on the 96 GB workstation.
- **sm_120-only, enforced.** On datacenter (B200/B300) the expert GEMM is compute/MMA-bound, not
  bandwidth-bound, so fp8 ≈ bf16; and the split structure is slower than fused for bf16 — so fp8-read
  would regress there. An arch gate (`_fp8_read_arch_ok`) makes the flag a **no-op off sm_120**: the
  path auto-falls-back to bf16-read, so the flag is safe on both targets (datacenter not re-measured;
  gated by arch, with a GPU-free mock unit test of the gate).
- **W8-read / A16.** Weights fp8, activations + MMA bf16 — no fp8/fp4 activation quality risk.
- **No per-expert scale** (b19). **NVFP4 only** (MXFP4 keeps the fused path). Default **off**.

## Honest caveats

- The speed win is marginal (1.07× expert layer; ~1.04–1.06× full step after non-expert dilution).
  Ship it for the **memory** headroom, not speed.
- +2.5% on the (already accepted) NVFP4 weight error — the frozen base is perturbed slightly more;
  LoRA adapts around it. Synthetic-weight worst case; real-weight impact likely smaller.
- Datacenter is gated off by arch, not re-measured in this PR.

### DIFF
--- a/src/axolotl/integrations/kernels/libs/scattermoe_lora/experts.py
+++ b/src/axolotl/integrations/kernels/libs/scattermoe_lora/experts.py
@@ -23,9 +23,20 @@ _NVFP4_LORA_DEQUANT_BF16 = os.environ.get("AXOLOTL_NVFP4_LORA_FUSED", "0") != "1
 # the non-fused backward (fallback + A/B measurement). dA/dB are unaffected (the fused-gather
 # threshold leaves the grouped-Gram path, which is already optimal at high expert counts).
 _NVFP4_LORA_FUSED_BWD = os.environ.get("AXOLOTL_NVFP4_LORA_NONFUSED_BWD", "0") != "1"
+# fp8-READ variant of the NVFP4+LoRA fast path (workstation Blackwell / sm_120). Materialize the
+# frozen NVFP4 weight as fp8 (e4m3) instead of bf16 and read it in the grouped GEMM, upcasting to
+# bf16 in-register (bf16 MMA, bf16 activations) — half the weight bytes on the bandwidth-bound
+# expert GEMM (~1.3x fwd+bwd on sm_120). Routed through the split/non-fused kernels (the fused
+# LoRA kernel serializes the fp8 upcast and loses). Default OFF: the datacenter gain is small and
+# the split overhead regresses there, and it adds ~2.5% to the (already accepted) NVFP4 weight
+# error. AXOLOTL_NVFP4_LORA_FP8_READ=1 enables (NVFP4 + LoRA only). See selective_dequant_kernel.
+_NVFP4_LORA_FP8_READ = os.environ.get("AXOLOTL_NVFP4_LORA_FP8_READ", "0") == "1"
 from .parallel_experts import flatten_sort_count, parallel_linear
 from .parallel_linear_lora import get_lora_params_from_wrapper, parallel_linear_lora
-from .selective_dequant_kernel import dequant_nvfp4_full_triton
+from .selective_dequant_kernel import (
+    dequant_nvfp4_fp8_triton,
+    dequant_nvfp4_full_triton,
+)
 from .selective_dequant import (
     get_active_experts,
     is_mxfp4_param,
@@ -103,6 +114,22 @@ def _get_base_param(param):
     except ImportError:
         pass
     return param
+
+
+def _fp8_read_arch_ok(device=None) -> bool:
+    """Whether the fp8-read path should activate on this GPU.
+
+    fp8-read wins only in the weight-bandwidth-bound regime: workstation Blackwell (sm_120,
+    GDDR7). On datacenter Blackwell (sm_100 / sm_103, HBM) the expert GEMM is compute/MMA-bound,
+    fp8 ≈ bf16, and the split structure is slower than the fused kernel -> a regression. So gate
+    fp8-read to the validated sm_120; everywhere else falls back to the universal bf16-read path.
+    """
+    if not torch.cuda.is_available():
+        return False
+    try:
+        return torch.cuda.get_device_capability(device) == (12, 0)
+    except Exception:
+        return False
 
 
 def _parallel_linear_maybe_lora(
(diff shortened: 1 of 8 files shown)


thad0ctor:
I've been doing a fair bit of NVFP4 training + attention work on Blackwell and tested your kernel locally against a couple of variants I've discovered through my own NVFP4 flows and wanted to share results in case they're useful, happy to fold them in as a follow-on PR if you don't want to tweak your approach.

**Recommended Dequant Method** A tweaked single-pass kernel — each lane consumes one packed byte and emits both nibbles via a contiguous 2-wide store, arithmetic E2M1 decode instead of the codebook gather, `BLOCK=8192` — runs ~4–5% faster in interleaved A/B median (up to ~10–13% on smaller, launch-bound shapes in single-run timing). It's a modest win because the kernel is already memory-bound at ~90% of bandwidth (~1.3 TB/s of a ~1.46 TB/s ceiling here) — your version is close to optimal. This method stays pure Triton, no new build dependency.

I attached the patch file and standalone replacement `.py` file for reference.

## Performance (interleaved A/B median)

| Shape (E × N × K) | this PR | tweaked | speedup |
|---|---|---|---|
| 128 × 2048 × 2048 (large, MoE-weight-like) | 1.053 ms | 1.014 ms | **1.04×** |
| 64 × 768 × 2048 | 0.217 ms | 0.199 ms | **1.09×** |
| 8 × 2048 × 2048 (small, launch-bound) | 0.067 ms | 0.056 ms | **1.20×** |

The large-tensor case is the representative one for full-expert weight dequant (memory-bound, ~4%); the smaller shapes gain more as the reduced index math and single vectorized store cut launch overhead.

**Setup:** RTX PRO 6000 Blackwell (sm120a), CUDA 13.3 / Triton 3.7 / torchao 0.18. All variants validated **bit-identical** to `NVFP4Tensor.dequantize()` and to your kernel across all three `per_tensor_scale` modes (none / scalar / per-expert).

[optionA_selective_dequant_kernel.py](https://github.com/user-attachments/files/29170416/optionA_selective_dequant_kernel.py)
[optionA_dequant_triton_tweak.patch](https://github.com/user-attachments/files/29170417/optionA_dequant_triton_tweak.patch)

winglian:
this PR was cherry picked into the other branch already and included co-authored-by trailers. Thanks!

PR #3746, opened 2026-06-21
perf(nvfp4): native Blackwell quantization

thad0ctor:
# Description

Adds an arch-gated, NVIDIA-native NVFP4 quantizer for Blackwell that replaces torchao's software FP4 pack with the hardware `cvt.rn.satfinite.e2m1x2.f32` instruction in a fused Triton kernel. It works by monkeypatching `NVFP4Tensor.to_nvfp4` — one interception covering both producers (PTQ 2D swizzled/per-tensor weights and MoE 3D non-swizzled experts) — and is **byte-identical** to torchao (verified, including exact rounding ties).

It **engages automatically on Blackwell**: `quantize_model()` (PTQ, behind `axolotl quantize`) and `convert_qat_model()` (QAT convert) now call a gated `_maybe_install_native_nvfp4()` that installs the fast path only when `weight_dtype == nvfp4` and Blackwell is present. The per-call gate additionally requires CUDA ≥ 12.8, Triton, the tensor's own device to be Blackwell (multi-GPU safe), a power-of-2 K, and contiguous bf16/fp32 — anything else (incl. non-Blackwell, D192, non-contiguous) transparently delegates to torchao, so default behavior is unchanged everywhere else.

## Motivation and Context

PyTorch exposes no hardware FP4 conversion, so torchao's `to_nvfp4` packs FP4 with a software bit-manipulation cast (`f32_to_f4_unpacked`) — making NVFP4 weight quantization an order of magnitude slower than the hardware allows on Blackwell. Blackwell's `e2m1x2` cvt does the FP32→FP4 pack in hardware. This recovers that, byte-for-byte; the only subtlety was matching torchao's e4m3 block-scale floor `E4M3_EPS = finfo(e4m3).tiny = 2^-6`, after which the hardware RNE rounding matches torchao's software cast exactly, including ties.

### Performance — RTX PRO 6000 Blackwell (sm120), `[131072, 2048]` bf16, interleaved median

| Path | Was (torchao) | Is (native) | Speedup |
|---|---|---|---|
| MoE / non-swizzled (single-level, fully-fused kernel) | 32.4 ms | 2.0 ms | **16.1×** |
| PTQ / swizzled (two-level: fused pack + eager scale math) | 32.6 ms | 7.96 ms | **4.1×** |

Both produce byte-identical output to torchao. The two-level/PTQ path fuses only the pack (its scale math + `to_blocked` swizzle stay eager); both quant paths are one-time weight-cast costs, not training-hot.

## Update — reshape-free kernel + bf16-direct single-level (commit 7e81a96)

Follow-up fixing a latent correctness bug and adding a single-level speedup:

- **Correctness:** both kernels reshaped the loaded `[BR,K]` tile into `[BR,NG,16]`. `tl.reshape` miscompiles at exactly **1024 bytes/row** (K=256 fp32 / K=512 bf16): `amax` reduces the wrong 16-element group, silently corrupting ~50% of `qdata`. Width-specific, so it slipped past the original suite (covers K=512 bf16, not K=256 fp32). Rewrote `_quant_per_block_kernel` and `_pack_only_kernel` to load the even (low-nibble) and odd (high-nibble) block columns as two separate **strided tiles** straight into the cvt — no `tl.reshape`/`tl.split` on the loaded tile. Correct at every K×dtype, byte-identical to the old path where it was correct.
- **Speed:** the single-level path fed bf16 via a `.float()` pre-cast (a full fp32 `[R,K]` transient). The kernel widens to fp32 in-register (exact), so bf16 now goes straight in. Dropping the transient gives **2.4–4.2× end-to-end** on bf16 weights, and **1.26–2.4×** over the arithmetic-encode kernel in #3747.

| Path (bf16, RTX 5090, interleaved median) | `.float()` (was) | bf16-direct (is) | Speedup |
|---|---|---|---|
| 16384×4096 | 0.494 ms | 0.119 ms | **4.15×** |
| 131072×2048 | 1.908 ms | 0.455 ms | **4.19×** |
| 8192×8192 | 0.609 ms | 0.226 ms | **2.70×** |

Re-validated **byte-exact vs torchao 0.18** (single-level + two-level + swizzled, fp32 + bf16, K 256→8192, all 0 diffs — including the previously-corrupted K=256/512), plus a 0-broken edge sweep (masked rows, tiny K, single row). `ruff` 0.15.8 clean.

## How has this been tested?

A 21-case pytest gate asserts **0 byte mismatches** vs torchao `to_nvfp4` (qdata, scale, per_tensor_scale) across: single-level non-swizzled (MoE, 2D + 3D); two-level swizzled scalar `per_tensor_scale` (PTQ); two-level per-expert `[E,1,1]` (swizzled + non-swizzled) — each in fp32 + bf16. Plus fallback parity (non-pow2 K=192, forced-non-Blackwell) and monkeypatch install/uninstall hygiene. **All 21 pass.** `ruff check`/`format` clean. Env: RTX PRO 6000 Blackwell (sm120), CUDA 13.3, Triton 3.7, torchao 0.18. QAT training-time fake-quant is intentionally left on torchao — only `to_nvfp4` weight casting is accelerated.

## AI Usage Disclaimer

Opus 4.8 throughout

## Types of changes

- [x] New feature (non-breaking; auto-engages only on Blackwell + nvfp4, no-op otherwise)
- [x] Performance improvement
- [ ] Breaking change
- [ ] Documentation

## Summary by CodeRabbit

* **New Features**
  * Added optimized FP4 quantization support for NVIDIA Blackwell GPUs with automatic fallback to standard implementation when hardware requirements aren't met.

* **Tests**
  * Added integration tests to validate FP4 quantization correctness and fallback behavior.
  * Added performance benchmarks for FP4 quantization.


thad0ctor:
Pushed `7e81a96`: rewrote both quant kernels **reshape-free** — load the even/odd block columns as two separate strided tiles straight into the cvt instead of reshaping the `[BR,K]` load into `[BR,NG,16]`. That `tl.reshape` miscompiles at exactly **1024 bytes/row** (K=256 fp32 / K=512 bf16), reducing the wrong 16-element group and corrupting ~50% of `qdata` at those widths (slipped past the suite, which covers K=512 bf16 but not K=256 fp32). Single-level now feeds bf16 straight in (kernel widens to fp32 in-register, exact) instead of pre-casting via `.float()` → **2.4–4.2×** on bf16 weights. Re-validated **byte-exact vs torchao 0.18** across single/two-level + swizzled, fp32+bf16, K 256→8192; `ruff` 0.15.8 clean.

NanoCode012:
Quick note: this was built against torchao0.18 while we're still in 0.17

thad0ctor:
@NanoCode012  Fair point, I built this off the wrong venv so that's an oversight on my part.

I will defer to you guys, the quantization method works the same way (not torchao dependent) but the tests' numerical validation is impacted by the torchao bump albeit not from a CI standpoint as this doesn't run on Blackwell

The version changes is the reference oracle the parity test diffs against which surfaces two known 0.17→0.18 reference changes:

1. bf16 single-level (~0.1% of codes): 0.18 changed its reference from true division to reciprocal-multiply 
(x*(1/s)) to match the MSLK Triton kernel (the actual GEMM quantization path). My kernel already targets that
production path, so it's byte-exact on 0.18; on 0.17 it differs by ≤1 ULP on the bf16 grid-boundary cases. fp32 is exact on both.

2. 3D per-expert (MoE): 0.17's reference nvfp4_quantize crashes on [E,N,K] input (view(orig_shape) RuntimeError); 0.18 fixed this. So the 4 per-expert tests can't be validated against 0.17 at all 

Nothing in the PR fails functionally on 0.17 (the parity tests are Blackwell-gated, so CI skips them);  these are reference-side differences, not kernel bugs. 

So if you guys are interested the options are: 1) rewrite this to 0.17 knowing it will require an update at 0.18 and have a numerical divergence 2) apply version aware test paths (more code that will eventually become OBE) 3) defer this until it is bumped to 0.18

thad0ctor:
Also, given how much NVFP4 work is in flight right now, I'd be happy to use this PR to harmonize the NVFP4 quant paths before they diverge further:

  - the MoE-LoRA code now on `main` (#3747)
  - the expert-merge draft (#3753), which inlines its own `_requant_proj_weight` / `_dequant_fused`
  - native FP4 training (#3721), which has `_to_nvfp4_chunked`, `_abs_amax`, `_BLOCK_SIZE`, an eligibility guard (`shape[-1] %
  BLOCK_SIZE == 0`), and `_mslk_dequant` (if this gets picked up)

A shared module exposing:

  - `quantize_to_nvfp4(w, per_tensor_scale=None)` — thin wrapper over the path this PR accelerates, so no need to keep calling `to_nvfp4(block_size=16, ...)`
  - `dequantize_nvfp4(qdata, scale, per_tensor_scale)` — both #3721's `_mslk_dequant` and #3753's per-expert dequant reimplement  it today, so they could share one tested path
  - `BLOCK_SIZE` + the divisibility/eligibility guard in one place (the merge requant in #3753 currently has none; #3747 added  one only on its dequant side)

  This gets most valuable if/when native FP4 training (#3721) is upstreamed, at which point the training producer, this quantizer, and the merge path are the same primitive applied at three points in the lifecycle (train → save → merge). Sharing one tested round-trip implementation means a fix here (e.g. the reshape-free 1024 B/row correctness fix) propagates everywhere instead of needing to be re-discovered per fork.

Happy to expand this PR and route #3721 + #3753 through it, or start fresh.  Although it would need to be dependent on a torchao bump and source install (it's a pretty light build) pending full release of 0.18 - not entirely out of bounds for NVFP4 users who are used to delving into experimental builds.

Issue #2396, opened 2025-03-07
EXTREMELY SLOW (unusable) towards end of tokenization of dataset with long multi turn conversations


Nero10578:
### Please check that this issue hasn't been reported before.

- [x] I searched previous [Bug Reports](https://github.com/axolotl-ai-cloud/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

Expected tokenization to work extremely fast as in the old commits in from this commit and older 339f3c67e2d6855340b5958274ea539517829baa

### Current behaviour

Ever since a change to how the chat_templates works back in this commit 10cfecf02e8829de749708c2588dc76be3a156d2 the tokenization has become literally unusable for my dataset that contains long multi turn conversation. The speed seems to become exponentially slower towards the end of the tokenization since it uses less and less CPU cores as it finishes most of the examples.

Since that commit I have just been replacing the chat_template.py with the older one and it worked fine, but now since this commit b194e17c2847348bb102d24e86775dfb77c9c5fe this hacky fix stopped working as the new chat_template.py is much different. So now newer axolotl version are unusable to me.

![Image](https://github.com/user-attachments/assets/faeec121-7aa7-41cd-8494-068217a94c4f)

I can help the performance to be better by changing num_proc in datasets.py to a huge number but it still eventually slows to a crawl at the end.

### Steps to reproduce

Run preprocess on using the latest axolotl commit on a long multi turn conversation dataset.

### Config yaml

```yaml
base_model: /home/owen/models/Mistral-Small-24B-Instruct-2501
tokenizer_type: AutoTokenizer

train_on_inputs: false
group_by_length: false
load_in_8bit: false
load_in_4bit: true
strict: false

sequence_len: 8192
bf16: auto
tf32: false
flash_attention: true

shuffle_merged_datasets: true

# lora_on_cpu: true
# gpu_memory_limit: 20GiB

#Data
datasets:
  - path: /home/owen/datasets/RPMax-v1.4-formatted/conversations-escaped-formatted_split_1.json
    type: chat_template
    field_messages: conversations
    message_field_role: from
    message_field_content: value


warmup_steps: 20
dataset_prepared_path: ./lora_last_run_prepared

# Iterations
num_epochs: 1
saves_per_epoch: 8
saves_total_limit: 8

# Evaluation
val_set_size: 0.0025
eval_max_new_tokens: 128
eval_sample_packing: false
evals_per_epoch: 8

# LoRA
[...]

PR #3750, opened 2026-06-23
fix token inf time

ved1beta:
# Description
casuses inf time for tokenization for conservation dataset , 

## Motivation and Context
#2396 

### Speed (`_tokenize_single_prompt`, Qwen2.5, ~120 words/turn)

| turns | tokens | before | after | speedup |
|------:|-------:|-------:|------:|--------:|
| 4     | 2.4k   | 7.4 ms    | 5.9 ms  | 1.3× |
| 16    | 10k    | 63.7 ms   | 15.1 ms | 4.2× |
| 40    | 28k    | 380.8 ms  | 46.5 ms | 8.2× |
| 80    | 57k    | 1577.6 ms | 96.4 ms | **16.4×** |

Scaling per doubling: before ~4.1× (quadratic), after ~2.1× (linear). Gains grow
with conversation length — short chats were never the bottleneck.

## Summary by CodeRabbit

* **Refactor**
  * Optimized chat template token span computation for improved performance during prompt processing.

### DIFF
--- a/src/axolotl/prompt_strategies/chat_template.py
+++ b/src/axolotl/prompt_strategies/chat_template.py
@@ -170,7 +170,6 @@ class ChatTemplatePrompter(Prompter):
     def build_prompt_text(
         self,
         conversation: list[dict],
-        add_generation_prompt=False,
         tools=None,
     ) -> str | None:
         """Render a conversation to text without tokenizing.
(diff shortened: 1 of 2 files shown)


NanoCode012:
Could you explain more about this issue? When would `turn` be null? What are the other changes? I thought the issue is dataset specific but did they provide a sample data (been a while since I followed that issue)

ved1beta:
find_turn re-renders + re-tokenizes the whole conversation prefix once per turn O(N²)  (never finishes slowdown) . new _fast_content_spans renders + tokenizes once, then uses char→token offset mappings to locate each turn's content span in a single linear pass (O(N)).

NanoCode012 (approved):

PR #3751, opened 2026-06-23
fix: avoid IndexError in metharme/reflection _tokenize on empty fields

vineethsaivs:
# Description

`MetharmePromptTokenizingStrategy._tokenize` and `ReflectionPromptTokenizingStrategy._tokenize` index `result["input_ids"][-1]` without the empty-result guard that the base `PromptTokenizingStrategy._tokenize` already has, so a field that tokenizes to nothing raises `IndexError` and aborts the whole preprocessing run.

The Metharme override even logs `"Tokenizer result is empty. You may want to audit your dataset"` and then immediately indexes the empty list on the very next line; the Reflection override has no guard at all. This change returns an empty `BatchEncoding` in that case, exactly as the base class does.

## Motivation and Context

A row with an empty field crashes preprocessing when the tokenizer does not prepend a BOS token (gpt2 / GPT-NeoX / Pythia families), because `tokenizer("")` then returns `input_ids == []`. For Metharme this is reachable through `parse_instruction_fields` returning `(prompt["prompt"], "", prompt["generation"])`, where an empty `generation` is tokenized on its own. The base class was hardened for the empty case; these two overrides drifted out of sync. The warning text in the Metharme override shows such rows were meant to be tolerated, not fatal.

No linked issue; found by reading the code.

## How has this been tested?

Added `tests/test_empty_tokenization.py` with a regression test for each strategy: with a gpt2 tokenizer (which does not prepend BOS, so `_tokenize("")` yields an empty result), the strategy must return an empty `BatchEncoding` rather than raise. Both tests fail on `main` (`IndexError: list index out of range`) and pass with this change. The existing fixtures use huggyllama, which prepends BOS and so masks the bug, so the new test uses gpt2 to exercise the empty path.

## Types of changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed crashes in prompt tokenization strategies when processing empty or invalid input data, improving system stability.

* **Tests**
  * Added regression tests to prevent recurrence of empty tokenization edge cases.

### DIFF
--- a/src/axolotl/prompt_strategies/metharme.py
+++ b/src/axolotl/prompt_strategies/metharme.py
@@ -37,8 +37,7 @@ class MetharmePromptTokenizingStrategy(InstructionPromptTokenizingStrategy):
         )
         if len(result["input_ids"]) == 0:
             LOG.warning("Tokenizer result is empty. You may want to audit your dataset")
-            # Bail out like the base class instead of indexing into an empty result below;
-            # an empty field with a tokenizer that does not prepend BOS yields no tokens.
+            # An empty field (no-BOS tokenizer) yields no tokens; bail out like the base class.
             return BatchEncoding(data={"input_ids": [], "attention_mask": []})
         # If there's already an EOS token there, subtract from the number added
         if result["input_ids"][-1] == self.tokenizer.eos_token_id:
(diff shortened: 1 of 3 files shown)


vineethsaivs:
Good call, condensed the comment to a single WHY-focused line per the repo's comment style.

NanoCode012 (approved):
Waiting on CI

PR #3747, opened 2026-06-22
feat(kernels): fast + low-memory MoE-LoRA for gemma4 (NVFP4/Marlin + bnb) and DeepSeek-V4 on sm120/sm90/sm100

winglian:
## Summary

Makes **MoE LoRA / QLoRA fine-tuning fast and memory-efficient** for large sparse-MoE models — gemma4-26B-A4B (128 experts) and DeepSeek-V4 — on Blackwell (sm120), Hopper (sm90) and sm100, following axolotl's config-driven/plugin design (user intent in schema, process-global reset in the plugin, model specifics in adapters, execution in kernel libs).

Two expert quantization paths, both with a fused/low-overhead LoRA training kernel:

- **NVFP4 (modelopt checkpoint)** → fused 4-bit-read grouped GEMM: **Marlin on sm120**, DeepGEMM on sm90/sm100. Fastest at short context.
- **bnb-4bit (any bf16 checkpoint, `quantize_moe_experts`)** → 1-launch `parallel_linear` (scatter2scatter) path whose dequant'd bf16 is **recomputed in backward via a recipe** instead of saved — so it's fast *and* low-memory (no ~3x VRAM blowup). This is now the bnb default; a chunked `torch._grouped_mm` fallback (`moe_bnb_fast: false`) bounds the per-pass transient for large-expert models / tiny GPUs.

The other big throughput lever for gemma4 is **hybrid attention** (`gemma4_hybrid_attn_impl: true` + `attn_implementation: flash_attention_2`): FA2 on sliding-window layers + SDPA on global layers, via the `kernels` lib (no flash-attn source build). Plus a Triton head_dim=512 flash kernel and `sdpa_varlen` for packing.

## Benchmarks — gemma4-26B-A4B LoRA

Single **RTX PRO 6000 (sm120, 96 GB)**, `mb=1, ga=1, r=16`, sample-packing, gradient checkpointing, adamw_8bit. **True total tok/s** (CUDA-synced `perf_counter` over the steady window, identical methodology for axolotl and unsloth) · **peak VRAM active / reserved (GiB)** (peak reset at train start, so both exclude load transients).

| seq | axolotl **NVFP4** | axolotl **bnb** | unsloth (bnb) | axolotl speedup |
|-----|-------------------|-----------------|---------------|-----------------|
| 4k  | **2945** · 22.2 / 26.4 | 2254 · **16.8** / 21.1 | 1391 · 23.4 / 24.9 | **2.1×** |
[...]

PR #3753, opened 2026-06-24
feat: add nvfp4 adapter merge

NanoCode012:
# Description

Open draft to discuss whether this path for merging is correct (dequant -> merge -> quant). This is based off the #3747 . Once that lands, this can be iterated upon, then cherry picked onto a new rebase.

## How has this been tested?

Untested.


thad0ctor:
Opt-in to save a NVFP4 would seem appropriate,  NVFP4 is getting more common for rough CPT passes and having a f16 output would enable easier follow on runs at f16,

NanoCode012:
This got superseded in PR #3771

NanoCode012:
> Opt-in to save a NVFP4 would seem appropriate, NVFP4 is getting more common for rough CPT passes and having a f16 output would enable easier follow on runs at f16,

@thad0ctor , interesting point, we'll see if this becomes requested. A one shot external script can also do this for time being.

NanoCode012:
Update: I think we do have a `--dequant` flag, but I haven't verified

Issue #3754, opened 2026-06-25
`chat_template` Strategy Never Trains the Gemma Turn Terminator


thad0ctor:
Credit: @ferno9 for discovering the issue and diagnosing 

### Please check that this issue hasn't been reported before.

- [x] I searched previous [Bug Reports](https://github.com/axolotl-ai-cloud/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

When fine-tuning any **Gemma** model (1 / 2 / 3 / 3n / 4) with `type: chat_template`, the per-turn stop token that the Gemma chat template actually emits — `<end_of_turn>` (id **107** on Gemma 1/2, id **106** on Gemma 3/3n) or `<turn|>` (id **106** on Gemma 4) — should be **trained** at the end of each assistant turn (it should receive a real label, honoring `train_on_eos`/`train_on_eot`), so the model learns to stop. This already happens by default for other families (Qwen `<|im_end|>`, Llama-3 `<|eot_id|>`, Mistral `{{ eos_token }}`, …).

### Current behaviour

For Gemma, the turn terminator is **silently masked (`-100`) by default**, so the model is never taught to stop. After fine-tuning, the model produces the correct response. Loss/eval curves look perfectly healthy because the only thing never learned — stopping — is never in the loss.

**Root cause:** `ChatTemplateStrategy` defaults `eot_tokens` to `[tokenizer.eos_token]`, and the EOS/EOT finders key off `tokenizer.eos_token_id`. For Gemma that is `<eos>` (id **1**), but `<eos>` **never appears in the rendered chat** (turns end with `<end_of_turn>`/`<turn|>`), so neither finder matches the turn boundary → the terminator's label stays `-100`.

A warning *is* emitted but is easy to miss and not an error:

```
[WARNING] [axolotl.prompt_strategies.chat_template] EOS token '<eos>' not found in chat_template. Please check if your template/EOS token is correct.
```
The default gemma configs (e.g. Gemma 4) don't include the config flags needed to train the real stop token (the most likely fine-tuning scenario for non-base variants). The override works however:

```
  special_tokens:
    eos_token: "<turn|>"     # "<end_of_turn>" for gemma 2/3/3n
  train_on_eos: turn

```

[...]

PR #3756, opened 2026-06-26
fix: add EOS token to gemma configs

thad0ctor:
# Description

- Set eot_tokens on every Gemma chat_template example that lacked it (gemma2, gemma4, gemma4-unified, qat_nvfp4 Math-Gemma3); existing gemma3/3n examples already set it. reward-model.yaml is skipped (the Bradley-Terry strategy emits scalar labels, so the terminator is moot).
-  Strengthen warning when now EOS present

## Motivation and Context

Gemma keeps tokenizer.eos_token = <eos>, which never appears in the rendered chat (turns end with <end_of_turn> on Gemma 1/2/3/3n or <turn|> on Gemma 4). The default eot_tokens = [eos_token] therefore misses the real terminator and it is never trained, so the model never learns to stop. See GH issue #3754.

Credit to @ferno9 for discovering and troubleshooting

## How has this been tested?

See GH issue #3754.

## Summary by CodeRabbit

* **New Features**
  * Added explicit end-of-turn token settings to several Gemma and Math-Gemma example configurations for more consistent chat and training behavior.

* **Bug Fixes**
  * Updated the chat template warning to give clearer guidance when the end-of-sequence token is not found, including what token to set for turn endings.

### DIFF
--- a/src/axolotl/prompt_strategies/chat_template.py
+++ b/src/axolotl/prompt_strategies/chat_template.py
@@ -355,7 +355,8 @@ class ChatTemplateStrategy(PromptTokenizingStrategy):
                 and "eos_token" not in self.prompter.chat_template
             ):
                 LOG.warning(
-                    f"EOS token '{self.tokenizer.eos_token}' not found in chat_template. Please check if your template/EOS token is correct."
+                    f"EOS token '{self.tokenizer.eos_token}' not found in chat_template; the turn "
+                    "terminator won't be trained. Set `eot_tokens` to your template's turn-ending token."
                 )
             return
 
--- a/examples/gemma2/qlora.yml
+++ b/examples/gemma2/qlora.yml
@@ -10,6 +10,8 @@ load_in_4bit: true
 
 # huggingface repo
 chat_template: gemma
+eot_tokens:
+  - <end_of_turn>
 datasets:
   - path: cgato/SlimOrcaDedupCleaned
     type: chat_template
--- a/examples/gemma4-unified/12b-text-lora.yaml
+++ b/examples/gemma4-unified/12b-text-lora.yaml
@@ -9,6 +9,8 @@ liger_layer_norm: true
 strict: false
 
 chat_template: gemma4_unified
+eot_tokens:
+  - "<turn|>"
 datasets:
   - path: mlabonne/FineTome-100k
     type: chat_template
--- a/examples/gemma4-unified/12b-vision-lora.yaml
+++ b/examples/gemma4-unified/12b-vision-lora.yaml
@@ -16,6 +16,8 @@ remove_unused_columns: false
 sample_packing: false
 
 chat_template: gemma4_unified
+eot_tokens:
+  - "<turn|>"
 datasets:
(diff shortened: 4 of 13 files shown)


src/axolotl/prompt_strategies/chat_template.py
```diff
@@ -355,7 +355,8 @@ def _validate_eot_and_eos_tokens(self):
                 and "eos_token" not in self.prompter.chat_template
             ):
                 LOG.warning(
-                    f"EOS token '{self.tokenizer.eos_token}' not found in chat_template. Please check if your template/EOS token is correct."
+                    f"EOS token '{self.tokenizer.eos_token}' not found in chat_template; the turn "
+                    "terminator won't be trained. Set `eot_tokens` to your template's turn-ending token."
```
NanoCode012, line 359: I think this is not the same error that you were pointing to? This is about EOT token right? This section is on EOS token

NanoCode012 (commented):
Does the issue occur even as far back as regular Gemma template?

src/axolotl/prompt_strategies/chat_template.py
thad0ctor: That branch only runs when eot_tokens defaulted to [eos_token] (the bug). In the default config the EOT is the EOS token, so "EOS not found in template" and "turn terminator won't be trained" are the same event, which is why I reworded it here. I see how this could be confusing though.

From main:
```
        # If the EOT token is the same as the EOS token, we need to check differently
        if len(self.eot_tokens) == 1 and self.eot_tokens[0] == self.tokenizer.eos_token:
            # Check if the eos_token is in the chat_template or as a variable `eos_token`
            # Note: we check for `eos_token` in the string, but it could possibly not be a variable
            if (
                self.tokenizer.eos_token not in self.prompter.chat_template
                and "eos_token" not in self.prompter.chat_template
            ):
```

thad0ctor:
> Does the issue occur even as far back as regular Gemma template?

Yes, it affects the regular gemma template (used by both Gemma 1 and Gemma 2 via chat_template: gemma. 

The template closes every turn with <end_of_turn> and never emits <eos>, while the tokenizer keeps eos_token = <eos>. So the default eot_tokens = [eos_token] misses the real terminator.

Pre-tokenized check on unsloth/gemma-2b-it (Gemma 1) and tiny-gemma2-137m, identical result:

        DEFAULT (eot_tokens unset → ['<eos>']):  '<end_of_turn>' NOT trained  → 
          terminators trained: NONE

Gemma3 wasn't broken in tests but I added it for consistency.

winglian (approved):

winglian:
@thad0ctor is this related to https://github.com/axolotl-ai-cloud/axolotl/pull/3762 ?

thad0ctor:
> @thad0ctor is this related to #3762 ?

@winglian Yes, I went for the immediate quick fix (config) pending the discussion here https://github.com/axolotl-ai-cloud/axolotl/issues/3754 , that PR was submitted after

kallewoof:
Gemma 4 31B training experiments showed near-2-digit losses for me. This PR pulled them down to healthy levels.

PR #3758, opened 2026-06-26
FIX:gemma gibberish

ved1beta:
# Description

for now only fixes total param metadata,
when the merged model is re-sharded, the index is regenerated with the tensor count written into the total_size field instead of the byte size
so , total_tensors is incremented as total_tensors += len(shard_tensors) 

## Motivation and Context
https://discord.com/channels/1104757954588196865/1110594519226925137/1519645544211939430

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the accuracy of model merge index metadata for sharded outputs.
  * Generated index files now report both total parameter count and total size, helping ensure downstream tools read the correct model metadata.
  * Added validation to confirm sharded merge outputs produce the expected metadata values.

### DIFF
--- a/src/axolotl/cli/utils/lora_merge.py
+++ b/src/axolotl/cli/utils/lora_merge.py
@@ -1108,6 +1108,9 @@ def merge_lora_sharded_efficient(
     total_tensors = 0
     # Track weight_map for index regeneration: {tensor_key: shard_filename}
     weight_map: Dict[str, str] = {}
+    # Accumulate real byte size and parameter count for the index metadata.
+    total_size_bytes = 0
+    total_parameters = 0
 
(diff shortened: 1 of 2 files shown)


src/axolotl/cli/utils/lora_merge.py
```diff
@@ -1206,7 +1211,10 @@ def merge_lora_sharded_efficient(
             else "pytorch_model.bin.index.json"
         )
         index = {
-            "metadata": {"total_size": total_tensors},
```
NanoCode012, line 1209: What happens to total_tensor? Is it the same as `total_size_bytes`?

NanoCode012 (approved):

PR #3761, opened 2026-06-28
build and test only torch 2.11.0+ w cu130

winglian:
# Description

## Summary by CodeRabbit

* **New Features**
  * Updated build and test coverage to use newer CUDA, Python, and PyTorch combinations.
  * Added support for additional PyTorch build targets in nightly Docker image workflows.

* **Bug Fixes**
  * Adjusted test matrices to remove older or redundant version combinations, helping keep CI aligned with current supported environments.
  * Updated GPU end-to-end test runs to match the latest build configurations.


.github/workflows/multi-gpu-e2e.yml
```diff
@@ -30,26 +30,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        include:
```
NanoCode012, line 33: I'm not sure if this row is accidentally removed?

NanoCode012 (commented):
We should update the docker.qmd as well to use these newer versions/tags

NanoCode012:
🚨 This PR changes tokenization hash to be more accurate, so tokenized datasets from this PR onwards will get new hash.

PR #3762, opened 2026-06-28
fix: auto-detect EOT token from chat template when eos_token is absent (Gemma)

### LINKED ISSUE #3754: `chat_template` Strategy Never Trains the Gemma Turn Terminator

JSap0914:
## Description

Auto-detect the actual end-of-turn (EOT) token from the chat template when
`tokenizer.eos_token` is absent from it. This fixes Gemma models where the
turn terminator (`<end_of_turn>` / `<turn|>`) was silently masked and never
trained.

## Motivation and Context

Closes #3754

For every Gemma generation (1/2/3/3n/4), `tokenizer.eos_token = "<eos>"` (id 1),
but the chat template emits `<end_of_turn>` (Gemma 1/2/3/3n, id 107/106) or
`<turn|>` (Gemma 4, id 106) as the per-turn stop marker.

The old code defaulted `eot_tokens = [tokenizer.eos_token]` and cached token
id 1. `find_first_eot_token` never found a match because id 1 never appears in
rendered Gemma sequences, leaving every turn terminator with label -100. The
model was never trained to stop, but loss/eval curves look fine because the
untrained token never contributed to the loss.

A warning was already emitted, but not acted upon.

PR #3756 (open) adds an `eot_tokens` workaround to example YAML configs and
improves the warning. This PR fixes the root cause in `chat_template.py` so
Gemma works out of the box without any config change.

### `src/axolotl/prompt_strategies/chat_template.py`

- When `_validate_eot_and_eos_tokens` detects that `eos_token` is absent from
  the template, call `_detect_eot_from_template()` instead of just warning.
- `_detect_eot_from_template()`: renders the template with a sentinel assistant
  turn, locates the sentinel token-ids in the output, then collects the
  non-whitespace non-EOS tokens that immediately follow. These are the actual
  turn terminator(s). Updates `self.eot_tokens` with them.
- If detection fails, clears `self.eot_tokens` (correct: stop searching for the
  wrong id) and logs a warning with explicit instructions.
- Explicit `eot_tokens` set by the user always win; the new logic only triggers
  when the default `[eos_token]` is in use and that token is absent from the
  template.
- Templates that reference `eos_token` as a Jinja variable (Llama-3, ChatML,
  Mistral, …) are completely unaffected.

## Verification

```
pytest tests/prompt_strategies/test_gemma_eot_autodetect.py -v
# 5 passed
```

All tests use a mock tokenizer and are fully offline (no GPU / model download).
No existing test was edited.

```
pytest tests/prompt_strategies/test_chat_template_utils.py \
       tests/prompt_strategies/test_jinja_template_analyzer.py \
       tests/prompt_strategies/test_synthetic.py \
       tests/integrations/test_kd_chat_template.py -v
# 33 passed
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of chat turn terminators when the template uses a different end marker than the tokenizer’s EOS token.
  * Added fallback behavior that clearly warns when a terminator can’t be identified, preventing incorrect training on turn endings.
  * Preserved user-provided terminator settings so they aren’t overwritten by automatic detection.

* **Tests**
  * Added coverage for Gemma-like chat templates, EOS mismatch cases, explicit terminator overrides, and terminator position detection.

### DIFF
--- a/src/axolotl/prompt_strategies/chat_template.py
+++ b/src/axolotl/prompt_strategies/chat_template.py
@@ -355,8 +355,28 @@ class ChatTemplateStrategy(PromptTokenizingStrategy):
                 and "eos_token" not in self.prompter.chat_template
             ):
                 LOG.warning(
-                    f"EOS token '{self.tokenizer.eos_token}' not found in chat_template. Please check if your template/EOS token is correct."
+                    f"EOS token '{self.tokenizer.eos_token}' not found in "
+                    "chat_template; the turn terminator won't be trained with the "
+                    "default eot_tokens. Attempting to auto-detect the actual EOT "
+                    "token from the rendered template. Set `eot_tokens` explicitly "
+                    "to suppress this detection and control the behaviour."
                 )
+                detected = self._detect_eot_from_template()
+                if detected:
+                    LOG.info(
+                        f"Auto-detected EOT tokens from template rendering: {detected}. "
+                        "These will be used instead of eos_token. Set `eot_tokens` "
+                        "explicitly if you want different behaviour."
+                    )
+                    self.eot_tokens = detected
+                else:
+                    LOG.warning(
+                        "Could not auto-detect EOT token from template. "
+                        "The turn terminator will not be trained. "
+                        "Set `eot_tokens` to the template's turn-ending token "
+                        "(e.g. '<end_of_turn>' for Gemma)."
+                    )
+                    self.eot_tokens = []
             return
 
         # Create a new list to store tokens that should be kept
@@ -397,6 +417,57 @@ class ChatTemplateStrategy(PromptTokenizingStrategy):
                 f"eot_tokens: {self.eot_tokens}"
                 f"eos_token: {self.tokenizer.eos_token}"
             )
+
+    def _detect_eot_from_template(self) -> list[str]:
+        """
+        Render the chat template with a sentinel assistant turn and find tokens
+        that appear immediately after the assistant content but are not the EOS
+        token. These are candidates for the real end-of-turn (EOT) token.
+
+        Returns a list of detected EOT token strings (may be empty if detection
+        fails or no suitable token is found).
+        """
+        _SENTINEL = "SENTINEL_XYZ_DO_NOT_USE"
+        dummy_conv = [
+            {"role": "user", "content": "x"},
+            {"role": "assistant", "content": _SENTINEL},
+        ]
+        try:
+            full_ids = list(
+                self.tokenizer.apply_chat_template(
+                    dummy_conv,
+                    tokenize=True,
+                    add_generation_prompt=False,
+                    chat_template=self.prompter.chat_template,
+                )
+            )
+            sentinel_ids = list(
+                self.tokenizer.encode(_SENTINEL, add_special_tokens=False)
+            )
+            if not sentinel_ids:
+                return []
+
+            sentinel_len = len(sentinel_ids)
+            for start_pos in range(len(full_ids) - sentinel_len + 1):
+                if full_ids[start_pos : start_pos + sentinel_len] == sentinel_ids:
+                    eot_tokens = []
+                    for token_id in full_ids[start_pos + sentinel_len :]:
+                        if token_id == self.tokenizer.eos_token_id:
+                            break
+                        token_str = self.tokenizer.decode(
+                            [token_id], skip_special_tokens=False
+                        )
+                        # Stop at pure-whitespace tokens (e.g. '\n' after <end_of_turn>)
+                        if not token_str.strip():
+                            break
+                        eot_tokens.append(token_str)
+                    return eot_tokens
+        except Exception:  # pylint: disable=broad-except
+            LOG.debug(
+                "Failed to auto-detect EOT token from template rendering",
+                exc_info=True,
+            )
+        return []
 
     @property
     def supports_batched(self) -> bool:
--- a/tests/prompt_strategies/test_gemma_eot_autodetect.py
+++ b/tests/prompt_strategies/test_gemma_eot_autodetect.py
@@ -1,0 +1,198 @@
+"""
+Tests for auto-detection of the actual EOT token for Gemma-like models
+where tokenizer.eos_token != the chat template turn terminator.
+
+Covers issue #3754: chat_template strategy never trains the Gemma turn
+terminator because <eos> (id=1) is used as the EOT token, but Gemma
+templates emit <end_of_turn> (id=107) / <turn|> (id=106).
+"""
+from unittest.mock import MagicMock
+
+import pytest
+
+from axolotl.prompt_strategies.chat_template import ChatTemplateStrategy
+
+
+# Gemma 2 chat template - uses <end_of_turn>, NOT eos_token variable
+GEMMA2_TEMPLATE = (
+    "{{ bos_token }}"
+    "{% for message in messages %}"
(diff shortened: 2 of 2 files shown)


thad0ctor:
I'd recommend double checking your transformers version meets or exceeds the pinned version in your environment.

```python
full_ids = list(self.tokenizer.apply_chat_template(..., tokenize=True))
```

This assumes a list of ids, but on transformers v5 `return_dict` defaults to `True`, so `apply_chat_template` returns a `BatchEncoding` and `list(...)` gives the keys instead:

```
['input_ids', 'attention_mask']
```

This failed silently when I tested

NanoCode012:
I think I mentioned this elsewhere. There's no specific code that you can gate for Gemma chat template (besides to update the yaml which was already done), so this change impacts all chat template parsing.

As some templates do not have EOT, this can be breaking change, so holding off on this. Did you verify with other templates?

PR #3765, opened 2026-06-30
fix(cli/vllm_serve): let explicit CLI False override config-enabled booleans

### LINKED ISSUE #3764: vLLM serve CLI false booleans are ignored when config enables them
## Bug

When `axolotl vllm-serve` builds `ScriptArguments`, boolean CLI overrides for `enable_prefix_caching` and `enable_reasoning` are combined with config values using `or`. If the YAML config enables one of these options, passing the CLI value as `False` cannot disable it because `False or True` evaluates to `True`.

## Reproduction

Construct a vLLM config with `enable_prefix_caching: true` and `enable_reasoning: true`, then call `do_vllm_serve` with `cli_args={"enable_prefix_caching": False, "enable_reasoning": False}`. The generated script arguments still have both fields set to `True`.

## Expected

Explicit CLI `False` values should override config `True` values, matching the existing `enforce_eager` precedence behavior.

## Verification

Added a focused unit test covering CLI `False` overriding config `True`: `.venv/bin/python -m pytest tests/test_vllm_serve.py -q`.

Anai-Guo:
## Summary

Fixes #3764.

When `do_vllm_serve` builds the script arguments, the boolean CLI overrides for `enable_prefix_caching` and `enable_reasoning` were combined with the config values using `cli or cfg`:

```python
enable_prefix_caching = (
    cli_args.get("enable_prefix_caching") or cfg.vllm.enable_prefix_caching
)
enable_reasoning = (
    cli_args.get("enable_reasoning") or cfg.vllm.enable_reasoning or False
)
```

If the YAML config enables one of these options, passing the CLI value as `False` cannot disable it, because `False or True` evaluates to `True`. The CLI silently has no effect.

## Fix

Use explicit `None` checks, matching the precedence already used a few lines below for `enforce_eager`, so an explicit CLI value (including `False`) always wins, and an omitted CLI value (`None`) falls back to the config:

```python
cli_enable_prefix_caching = cli_args.get("enable_prefix_caching")
enable_prefix_caching = (
    cfg.vllm.enable_prefix_caching
    if cli_enable_prefix_caching is None
    else cli_enable_prefix_caching
)
```

(Same treatment for `enable_reasoning`, with the existing `None -> False` fallback preserved.)

## Tests

Added `tests/cli/test_vllm_serve.py` with two focused cases:

- CLI `False` overrides config-enabled `enable_prefix_caching` / `enable_reasoning`.
- Omitted CLI value (`None`) falls back to the config value.

```
tests/cli/test_vllm_serve.py ..                                          [100%]
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed CLI boolean settings so explicit `False` values are respected instead of being overridden by saved settings.
  * Improved fallback behavior when options are left unset, ensuring defaults are applied consistently.

* **Tests**
  * Added regression coverage for CLI/config precedence.
  * Verified both explicit `False` overrides and unset values falling back to saved settings.


src/axolotl/cli/vllm_serve.py
```diff
@@ -65,15 +65,26 @@ def do_vllm_serve(
     )
     dtype = cli_args.get("dtype") or cfg.vllm.dtype
     max_model_len = cli_args.get("max_model_len") or cfg.vllm.max_model_len
+    # Booleans use explicit None checks (like enforce_eager below) so an explicit
+    # CLI False can disable an option the config enabled. Using `cli or cfg` would
+    # let the config value win whenever the CLI value is falsy.
+    cli_enable_prefix_caching = cli_args.get("enable_prefix_caching")
```
ved1beta: trim

tests/cli/test_vllm_serve.py
```diff
@@ -0,0 +1,91 @@
...
+    try:
+        with patch("axolotl.cli.vllm_serve.load_cfg", return_value=cfg):
+            do_vllm_serve("config.yml", cli_args)
+    finally:
+        del sys.modules[serve_module_name]
+
+    stub.main.assert_called_once()
+    script_args = stub.main.call_args.args[0]
+    assert script_args.enable_prefix_caching is True
+    assert script_args.enable_reasoning is True
```
ved1beta: simplify the test feels like bloat

ved1beta (changes requested):

Anai-Guo:
Thanks @ved1beta — addressed both:

- **Trim**: folded the `enable_reasoning` `None`→`False` fallback into a trailing `or False` and shortened the comment.
- **Test**: collapsed the two precedence tests into a single parametrized `test_vllm_serve_bool_precedence` covering CLI-`False`-overrides-config and CLI-`None`-falls-back-to-config.

For coderabbit's note on empty-string `reasoning_parser`: left as-is since `reasoning_parser` is a value option (not a boolean flag) and `""`/unset are equivalent there — happy to switch it to a `None` check too if you'd prefer.

ved1beta (approved):

Anai-Guo:
Friendly ping @ved1beta — this is approved and CI is green. Is there anything else needed before it can be merged?

PR #3768, opened 2026-06-30
fix: bound chat template nesting depth in the jinja analyzer

eharris128:
# Description

`JinjaTemplateAnalyzer` parses a chat template and walks its AST at prompter
init (`ChatTemplatePrompter.__init__` -> `get_chat_template_msg_variables` ->
`JinjaTemplateAnalyzer`). Both steps recurse once per nesting level with no
bound:

- `env.parse()` uses jinja2's recursive-descent parser, which raises
  `RecursionError` on a deeply nested expression (for example many nested
  parentheses) or deeply nested block tags (`{% if %}` ... `{% endif %}`).
- The analyzer's own `_visit_node` walk recurses over the parsed AST, so a deep
  attribute chain (for example `messages.a.a.a...`) parses fine but then
  overflows the stack during analysis.

In both cases the analyzer raises an uncaught `RecursionError`, which aborts
`axolotl preprocess` / `axolotl train` before training starts. With the default
`chat_template: tokenizer_default`, the analyzed template is the base model's
`tokenizer.chat_template` (read from the downloaded model's
`tokenizer_config.json`), so a malformed or over-nested template can come in
with the model rather than from the user's own config.

This change rejects an over-nested template with a clear `ValueError` instead of
letting it crash with a stack-overflow traceback. The bound (`MAX_TEMPLATE_DEPTH
= 200`) sits far above any real template: the deepest chat template bundled in
this repo is about 21 AST levels deep, so well-formed templates are unaffected.

## Motivation and Context

Analyzing a chat template should fail cleanly on malformed input rather than
exhaust the interpreter stack. The fix turns an uncaught `RecursionError` into a
clear, actionable error and bounds the recursion the analyzer performs. This is
a defensive robustness change: the impact is a single local preprocessing or
training process aborting, not remote code execution or anything cross-process.

## How has this been tested?

- Added unit tests in `tests/prompt_strategies/test_jinja_template_analyzer.py`:
  a deeply nested expression, deeply nested block tags, and a deep attribute
  chain are each rejected with a `ValueError`, and a normally nested template
  still analyzes and returns its variables.
- Confirmed the patched analyzer returns identical results to the previous
  behavior on the existing test templates, so real templates are unchanged.
- `ruff format` and `ruff check` clean with the repo config; `mypy` clean on the
  changed module.

Logs:

```
4 passed, 10 deselected
```

(The 10 deselected tests are the existing analyzer tests that depend on
`transformers` / `datasets` fixtures; they were not run in my offline
environment. The 4 selected are the new regression tests.)

## AI Usage Disclaimer

Yes. This change was prepared with assistance from Claude (Anthropic's Claude
Code). The depth-guard implementation, the regression tests, and this
description were AI-assisted, then reviewed, validated, and verified by me before
submission.

## Types of changes

- [x] Bug fix / hardening (non-breaking change which fixes an issue)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved template analysis to fail gracefully when templates are too deeply nested.
  * Added clearer validation for overly complex templates, preventing unexpected crashes during parsing.
  * Ensured valid nested templates still analyze correctly and continue to return expected variables.

### DIFF
--- a/src/axolotl/prompt_strategies/jinja_template_analyzer.py
+++ b/src/axolotl/prompt_strategies/jinja_template_analyzer.py
@@ -5,6 +5,26 @@ from typing import Dict, Optional, Set, TypedDict, Union
 from jinja2 import Environment, meta, nodes
 from jinja2.ext import Extension
 
+# Bound nesting below where jinja2's parser or our AST walk overflow the stack.
+MAX_TEMPLATE_DEPTH = 200
+
+
+def _assert_within_depth(node: nodes.Node, max_depth: int) -> None:
+    """Reject an AST nested deeper than ``max_depth``.
+
+    The check itself walks iteratively so it cannot exhaust the stack on the very
+    input it is meant to guard against.
+    """
+    stack = [(node, 1)]
+    while stack:
+        current, depth = stack.pop()
+        if depth > max_depth:
+            raise ValueError(
+                f"chat template nesting depth exceeds the maximum of {max_depth}"
+            )
+        for child in current.iter_child_nodes():
+            stack.append((child, depth + 1))
+
 
 class JinjaTemplateAnalysis(TypedDict):
     """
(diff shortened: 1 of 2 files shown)


NanoCode012:
Thanks for the PR, this looks like an interesting catch. Which template caused this issue for you?

eharris128:
> Thanks for the PR, this looks like an interesting catch. Which template caused this issue for you?

No template in particular - I was just red teaming your code for fun.

PR #3769, opened 2026-07-01
fix: respect false vllm serve cli booleans

### LINKED ISSUE #3764: vLLM serve CLI false booleans are ignored when config enables them

JSap0914:
## Summary

- respect explicit `False` CLI values for `enable_prefix_caching` and `enable_reasoning` in `axolotl vllm-serve`
- add a focused unit test that captures the config-true/CLI-false precedence case

Closes #3764

## Tests

- `.venv/bin/python -m pytest tests/test_vllm_serve.py -q`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed CLI overrides so explicitly set `false` values are respected instead of falling back to saved defaults.
  * Improved handling for caching and reasoning options when launching the service.
* **Tests**
  * Added coverage to verify `false` overrides are passed through correctly from the command line.

### DIFF
--- a/src/axolotl/cli/vllm_serve.py
+++ b/src/axolotl/cli/vllm_serve.py
@@ -66,13 +66,17 @@ def do_vllm_serve(
     dtype = cli_args.get("dtype") or cfg.vllm.dtype
     max_model_len = cli_args.get("max_model_len") or cfg.vllm.max_model_len
     enable_prefix_caching = (
-        cli_args.get("enable_prefix_caching") or cfg.vllm.enable_prefix_caching
+        cli_args["enable_prefix_caching"]
+        if cli_args.get("enable_prefix_caching") is not None
+        else cfg.vllm.enable_prefix_caching
     )
     reasoning_parser = (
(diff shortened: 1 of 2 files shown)


NanoCode012:
Duplicate of https://github.com/axolotl-ai-cloud/axolotl/issues/3765

PR #3771, opened 2026-07-01
fix(merge): dequantize quantized bases before folding LoRA delta

winglian:
The memory-efficient (shard-by-shard) LoRA merge folded scaling*(B@A) into the raw base tensor. For a block-fp8 (FineGrainedFP8) checkpoint this read the e4m3 values without applying weight_scale_inv and re-rounded the sum to fp8, dropping the delta — a silently-wrong merge (weight rel up to 1.0).

Dequantize fused quantized weights to bf16 per shard before folding, then keep bf16 and strip quantization_config so the merged checkpoint loads correctly. Detection is by weight dtype + scale sibling:

  - block-fp8 : e4m3 + <k>_scale_inv (128x128)
  - mxfp8     : e4m3 + <k>_scale (e8m0/32)
  - nvfp4     : uint8 + <k>_scale (e4m3/16) [+ <k>_scale_2]  (single- or two-level)
  - mxfp4     : uint8 + <k>_scale (e8m0/32)

Covers 2D linears and fused-3D experts (target_parameters). The expert delta uses PEFT's ParamWrapper get_delta_weight, verified to match the ScatterMoE kernel's expert-major B layout exactly.

Also:
  - carry full-weight adapter overrides (resized embed_tokens/lm_head, modules_to_save) into the merge and bump config vocab_size
  - warn when a fused expert LoRA meets a per-expert-unfused base (the expert LoRA would otherwise be silently dropped)
  - warn when a LoRA folds into a still-quantized weight the merge did not dequantize (mxfp4/per-tensor-fp8/per-expert layouts)
  - ragged-block and swizzled-scale handling

Adds TestQuantizedBaseMerge (22 hermetic cases): analytic weight-level + forward-level end-to-end + regression guards.

# Description

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model merging for quantized checkpoints, reducing the risk of corrupted or incorrect merged weights.
  * Added safeguards and warnings when a merge would be unsafe for unsupported or incompatible weight formats.
  * Fixed handling for resized embeddings and output heads so merged models preserve the correct vocabulary size.
  * Improved support for expert-based model layouts to avoid silent loss of merged changes.

### DIFF
--- a/src/axolotl/cli/utils/lora_merge.py
+++ b/src/axolotl/cli/utils/lora_merge.py
@@ -650,82 +650,180 @@ def _dequant_nvfp4(w, scale, scale2, dev: str) -> torch.Tensor:
     return nv.dequantize(torch.bfloat16)
 
 
+def _detect_quant_format(key, w, shard_tensors, e8m0):
+    """Return (fmt, {suffix: scale_tensor}) for a quantized weight ``key``, else (None, {}).
+    fmt in {block_fp8, mxfp8, nvfp4, mxfp4}."""
+    si = shard_tensors.get(key + "_scale_inv")
+    sc = shard_tensors.get(key + "_scale")
+    sc2 = shard_tensors.get(key + "_scale_2")
+    is_e8m0 = sc is not None and (
+        sc.dtype == torch.uint8 or (e8m0 is not None and sc.dtype == e8m0)
+    )
+    if w.dtype == torch.float8_e4m3fn and si is not None:
+        return "block_fp8", {"_scale_inv": si}
+    if w.dtype == torch.float8_e4m3fn and is_e8m0:
+        return "mxfp8", {"_scale": sc}
+    if w.dtype == torch.uint8 and sc is not None and sc.dtype == torch.float8_e4m3fn:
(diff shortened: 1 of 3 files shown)


src/axolotl/cli/utils/lora_merge.py
```diff
@@ -544,6 +545,414 @@ def _find_dora_magnitude(
     return None
 
 
+def _dequant_block_fp8(w: torch.Tensor, si: torch.Tensor, dev: str) -> torch.Tensor:
+    """FineGrainedFP8 (block-fp8): ``W`` e4m3 * ``scale_inv`` fp32 (128x128 blocks). 2D or fused-3D
+    (block axes = last two dims)."""
+    wf = w.to(dev).float()
+    s = si.to(dev).float()
```
NanoCode012, line 552: Just quick look, but are we upscaling to fp32 here?

winglian: maintain as much accuracy as possible. lora merge only loads one shard at a time and then unloads, so the actual memory required is relatively low still

PR #3774, opened 2026-07-01
feat: add PaddleOCR-VL + model support registry 

thad0ctor:
## Description

Adds PaddleOCR-VL as a supported multimodal image-text model path in Axolotl.

https://huggingface.co/PaddlePaddle/PaddleOCR-VL-1.6

This includes:

- Mapping `paddleocr_vl` to `AutoModelForImageTextToText` in multimodal loader constants.
- A PaddleOCR-VL processing strategy that masks user prompts and image placeholder tokens for chat-template SFT.
- Config guards that reject unsupported PaddleOCR-VL combinations: sample packing, Liger, and Cut Cross Entropy.
- Validated QLoRA and full fine-tune examples 
- Focused tests for loader dispatch, processing-strategy masking, strategy dispatch, and unsupported config combinations.

## Motivation and Context

PaddleOCR-VL uses a document-oriented multimodal processor and a model-specific image placeholder / RoPE path. Without an explicit processing strategy and compatibility guards, it can fall back to generic multimodal behavior or allow config combinations that do not preserve the model's multimodal position handling.

## How has this been tested?

Focused unit coverage:

```bash
PYTHONPATH=src python -m pytest -q \
  tests/test_attn_implementation.py::TestSamplePackingValidation \
  tests/test_processing_strategies.py::test_paddleocr_vl_masks_user_prompt_and_image_tokens \
  tests/test_processing_strategies.py::test_dispatch_paddleocr_vl \
  tests/test_processing_strategies.py::test_dispatch_paddleocr_vl_chat_template_key \
  tests/test_loaders.py::test_paddleocr_vl_is_multimodal_image_text_model
```

Result: `17 passed`.

Real-model smoke testing against a local PaddleOCR-VL-1.6 checkpoint on CUDA:

- Loaded `PaddleOCRVLProcessor`; verified `image_token == "<|IMAGE_PLACEHOLDER|>"` and image token id `100295`.
- Direct bf16 model forward with `attn_implementation="flash_attention_2"`; Transformers resolved `_attn_implementation` to `kernels-community/flash-attn2`; forward produced logits shape `(1, 148, 103424)`.
- Direct 4-bit load with `BitsAndBytesConfig(load_in_4bit=True)` and `attn_implementation="flash_attention_2"`; forward produced logits shape `(1, 148, 103424)`.
- Full fine-tune gradient smoke with FA2: one image/text batch, loss backward succeeded, and gradients were observed in the vision encoder, multimodal projector, and language model. Peak allocated memory was about `3.46 GiB` for the tiny smoke batch.
- QLoRA gradient smoke with FA2 and the example's target regex: LoRA adapters attached to language decoder, vision encoder, and multimodal projector modules; loss backward succeeded; LoRA gradients were observed in all three areas. Peak allocated memory was about `2.59 GiB` for the tiny smoke batch.
- Multimodal collator smoke produced `pixel_values`, `image_grid_thw`, and `mm_token_type_ids`; label masking masked `144` image tokens.

Model verification harness runs (Pre-PR Testing Skill (not yet upstream): https://github.com/thad0ctor/axolotl/pull/63):

- Smoke gates `G1,G2,G4,G5,G8`: passed.
- Full `G1` config matrix: passed.
- `G3` preprocess gate: passed.

Future compatibility check:

- Applied this branch over the pending multimodal CPT dataset-pipeline branch https://github.com/axolotl-ai-cloud/axolotl/pull/3629 in a temporary worktree.
- Combined targeted CPT + PaddleOCR coverage passed; real PaddleOCR-VL CPT collator smoke produced pixel values/grid/token-type fields and masked image tokens correctly.

## AI Usage Disclaimer

OpenAI Codex 5.5 throughout

## Types of changes

- [x] New feature
- [x] Documentation
- [x] Tests

## Summary by CodeRabbit

* **New Features**
  * Added support and guidance for PaddleOCR-VL as a multimodal training option.
  * Included new example configurations for both QLoRA and full fine-tuning.

* **Bug Fixes**
  * Improved masking and labeling behavior for PaddleOCR-VL training inputs.
  * Added safeguards to prevent unsupported settings for this model.

* **Documentation**
  * Updated the overview, supported models list, and example docs with PaddleOCR-VL setup instructions and recommendations.


examples/paddleocr-vl/README.md
```diff
@@ -0,0 +1,42 @@
...
+    To use a local checkout of the model, replace `base_model` with your local path:
+
+    ```yaml
+    base_model: /path/to/local/PaddleOCR-VL-1.6
+    ```
+
+## Tips
+
+- The model uses its bundled chat template through `processor_type: AutoProcessor`; no explicit `chat_template` is needed.
+- Do not set `trust_remote_code` for this example; Axolotl's pinned Transformers version includes the PaddleOCR-VL model and processor implementation.
```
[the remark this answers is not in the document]
thad0ctor: Validated the CodeRabbit trust_remote_code note against the pinned dependency. In a temporary venv with transformers==5.12.1, AutoConfig, AutoProcessor, and AutoModelForImageTextToText all load PaddlePaddle/PaddleOCR-VL-1.6 / the local checkpoint with trust_remote_code=False and resolve to native transformers.models.paddleocr_vl classes. With trust_remote_code=True, config/processor switch to transformers_modules remote code. So the README guidance to omit trust_remote_code is intentional and I am leaving it unchanged.

docs/multimodal.qmd
```diff
@@ -235,6 +236,23 @@ base_model: Qwen/Qwen3.5-9B
 chat_template: qwen3_5
 ```
 
+### PaddleOCR-VL {#sec-paddleocr-vl}
```
NanoCode012, line 239: Let's cleanup this section. See other as ref. Only minimum necessary points.

NanoCode012: If more details needed, add it in the README for the examples folder.

examples/paddleocr-vl/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+    ```bash
+    axolotl train examples/paddleocr-vl/paddleocr-vl-1_6-full-finetune.yaml
+    ```
+
+    To use a local checkout of the model, replace `base_model` with your local path:
+
+    ```yaml
+    base_model: /path/to/local/PaddleOCR-VL-1.6
+    ```
```
NanoCode012: Not needed.
```suggestion
```

examples/paddleocr-vl/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Tips
+
+- The model uses its bundled chat template through `processor_type: AutoProcessor`; no explicit `chat_template` is needed.
+- Do not set `trust_remote_code` for this example; Axolotl's pinned Transformers version includes the PaddleOCR-VL model and processor implementation.
+- Keep `sample_packing: false`; PaddleOCR-VL uses 3D multimodal RoPE positions that are not compatible with Axolotl's packed 2D position IDs.
+- `attn_implementation: flash_attention_2` works with PaddleOCR-VL; use `sdpa` as a fallback if your environment does not have Flash Attention 2 kernels available.
```
NanoCode012:
```suggestion
```

Not needed, recommend best default. Try to keep info as simple for newcomers in general. You can apply this to current docs.

src/axolotl/utils/config/__init__.py
```diff
@@ -118,6 +118,44 @@ def _model_with_inherited_default_fallback(model_cls, data):
         return _model_validate_with_field_names(model_cls, data_with_defaults)
 
 
+def _check_model_config_constraints(cfg):
```
NanoCode012: We avoid to put checks here. Rather in the respective libs.

src/axolotl/processing_strategies.py
```diff
@@ -1478,6 +1509,18 @@ def get_processing_strategy(
         return PixtralProcessingStrategy(**processing_kwargs)
     if chat_template_type == "mistral_v7_tekken":
         return MistralV7TekkenProcessingStrategy(**processing_kwargs)
+    if chat_template_type == "paddleocr_vl":
+        return PaddleOCRVLProcessingStrategy(**processing_kwargs)
+
+    try:
+        from transformers.models.paddleocr_vl.processing_paddleocr_vl import (
+            PaddleOCRVLProcessor,
+        )
```
NanoCode012: Is this for new transformers version?

README.md
```diff
@@ -75,7 +75,7 @@ Axolotl is a free and open-source tool designed to streamline post-training and
 Features:
 
 - **Multiple Model Support**: Train various models like GPT-OSS, LLaMA, Mistral, Mixtral, Pythia, and many more models available on the Hugging Face Hub.
-- **Multimodal Training**: Fine-tune vision-language models (VLMs) including LLaMA-Vision, Qwen2-VL, Pixtral, LLaVA, SmolVLM2, GLM-4.6V, InternVL 3.5, Gemma 3n, and audio models like Voxtral with image, video, and audio support.
+- **Multimodal Training**: Fine-tune vision-language models (VLMs) including LLaMA-Vision, Qwen2-VL, Pixtral, LLaVA, SmolVLM2, GLM-4.6V, InternVL 3.5, Gemma 3n, [PaddleOCR-VL](https://huggingface.co/PaddlePaddle/PaddleOCR-VL-1.6), and audio models like Voxtral with image, video, and audio support.
```
NanoCode012: I don't think we need to add link here to HF.

tests/test_processing_strategies.py
```diff
@@ -668,6 +669,60 @@ def test_mistral_v7_tekken_train_on_eos_all_respects_user_include_end_false():
...
+# PaddleOCR-VL
+# --------------------------------------------------------------------------- #
+
+
+def _paddleocr_vl_tokenizer():
+    vocab = {
+        "User: ": [2969, 93963, 93919],
+        "Assistant:\n": [92267, 93963, 23],
+        "<|IMAGE_PLACEHOLDER|>": [100295],
+    }
```
NanoCode012: Make sure to verify these numbers are correct given it's multiple ids

NanoCode012: or use their proper template roles

docs/multimodal.qmd
thad0ctor: All set section is now just the YAML block plus one caveat line with link to readme

src/axolotl/utils/config/__init__.py
thad0ctor: Done, removed `_check_model_config_constraints` and relocated the guards

src/axolotl/processing_strategies.py
thad0ctor: The guard is for older transformers.

PaddleOCR-VL landed in transformers **v5.0.0** (PR huggingface/transformers#42178)). 

Since Axolotl pins `>=5.12.1`, the module always exists, so it's now a plain top-level import like the neighboring InternVL/SmolVLM/Voxtral processors.

README.md
thad0ctor: updated

tests/test_processing_strategies.py
thad0ctor: The tests now load the real `PaddlePaddle/PaddleOCR-VL-1.6` processor (module-scoped fixture, same pattern as the prompt_strategies tokenizer tests), render `apply_chat_template` on role messages, and assert the masking structurally: everything through `Assistant:\n` masked, the decoded unmasked span equals the assistant response + eos, and image placeholder positions masked.

thad0ctor:
Addressed merging conflicts in 224c2e0

thad0ctor:
Addressed merging conflicts [51b82f3](https://github.com/axolotl-ai-cloud/axolotl/pull/3774/commits/51b82f3cbec6375383e746f238384fec79766b61)

winglian:
thanks, I'm looking into whether we can normalize/package together how we support individual model architectures so we have less code sprawl.

thad0ctor:
> thanks, I'm looking into whether we can normalize/package together how we support individual model architectures so we have less code sprawl.

you're welcome and totally understood, sounds like some major work

winglian:
maybe something like this? https://github.com/axolotl-ai-cloud/axolotl/compare/rfc/model-support-registry?expand=1

thad0ctor:
> maybe something like this? https://github.com/axolotl-ai-cloud/axolotl/compare/rfc/model-support-registry?expand=1

@winglian I like the idea, you could even evolve it in the future:

>  Vanilla training pipeline
>       ↓
>   Reusable architecture templates
>       ↓
>   Per-model profile/config
>       ↓
>   Model-specific strategies and monkeypatches

Provides “template” means a reusable bundle of strategies:

  For example:

> model_support/
>    templates/
>      causal_lm.py
>      multimodal.py
>      hybrid_ssm.py
>      remote_code.py

>    paddleocr_vl/
>      profile.py
>      processing.py
>      patches.py

>    kimi_linear/
>      profile.py
>      modeling.py
>      patches.py

  A model profile could select shared behavior and override only the exceptions:

```
  PADDLE_OCR_VL = ModelProfile(
      model_types=("paddleocr_vl",),
      template=MULTIMODAL_SFT,
      model_loader=AutoModelForImageTextToText,
      processing_strategy=PaddleOCRVLProcessingStrategy,
      capabilities={
          "liger": Unsupported("No conditional-generation patch"),
          "cut_cross_entropy": Unsupported("No CCE forward implementation"),
          "sample_packing": Experimental(),
      },
      patches=[],
  )
```

Basically three kinds of customization:

  **1. Declarative selection**

     Choose standard components: causal-LM versus conditional-generation loader, tokenizer versus processor, text versus multimodal
     collator, supported attention paths, and so on.

  **2. Strategy replacement**

     Supply a custom processing strategy, batch preparation step, loss adapter, attention-mask builder, or PEFT target resolver. These
     are ordinary typed interfaces and are easier to test than monkeypatches.

  **3. Imperative escape hatches**

     Use phase-specific monkeypatches only where upstream code cannot be configured or subclassed:

     pre_config_load
     pre_tokenizer_load
     pre_model_load
     post_model_build
     pre_trainer_init

I tried out something similar (template-based) with an attention kernel sweep tool to essentially create a kernel via a config

winglian:
If you can cherry pick those last few commits for model support registry to this, I think we can get this merged.

thad0ctor:
Pushed one more fix found while e2e-validating the registry port (4763ed4): single-GPU lora/qlora runs auto-enable the fused LoRA kernels, and the QKV/O source rewrite can't resolve `PaddleOCRAttention` (`paddleocr_vl` isn't in the causal-LM mapping, and the forward doesn't match the rewrite patterns), so the bundled example configs crashed at model load. This pre-dates the registry commits — fixed the registry way: the descriptor now declares `lora_kernels: Unsupported(...)`, the auto-enable validator consults the registry, and explicit `lora_qkv_kernel`/`lora_o_kernel` opt-in fails fast with a clear message instead of the AttributeError.

Validated with a 10-step QLoRA smoke run of the bundled example on a 3090 — trains cleanly through the registry-dispatched processing strategy (loss 3.79→2.85), checkpoint save OK. `test_model_support` (+2 new tests) and `tests/patched/test_validation.py` pass, pre-commit clean.

thad0ctor:
@winglian did you want to look into this more?

https://github.com/axolotl-ai-cloud/axolotl/pull/3774#issuecomment-4950546622

I could expand this framework and update https://github.com/thad0ctor/axolotl/pull/63 (PR pending) and also add a model implementation and/or conversion skills to start porting in old models unless the desire to to just use a new schema moving forward

winglian:
let's do the framework expansion as a follow-up PR and keep this scoped to the new registry and new model

thad0ctor:
> let's do the framework expansion as a follow-up PR and keep this scoped to the new registry and new model

agreed I should have clarified that , I'll look into it and also making a skill or framework to assist with backwards looking refactoring and new model additions

tests/test_processing_strategies.py
```diff
@@ -668,6 +669,65 @@ def test_mistral_v7_tekken_train_on_eos_all_respects_user_include_end_false():
...
+# --------------------------------------------------------------------------- #
+# PaddleOCR-VL
+# --------------------------------------------------------------------------- #
+
+
+@pytest.fixture(scope="module", name="paddleocr_vl_processor")
+def fixture_paddleocr_vl_processor():
+    from transformers import AutoProcessor
+
+    return AutoProcessor.from_pretrained("PaddlePaddle/PaddleOCR-VL-1.6")
```
NanoCode012: You might want to mock this, we run test offline

src/axolotl/utils/schemas/config.py
```diff
@@ -1864,6 +1864,29 @@ def check_auto_enable_lora_kernels(cls, data):
...
+                    from transformers import AutoConfig
+
+                    model_type = getattr(
+                        AutoConfig.from_pretrained(
+                            data["base_model"], trust_remote_code=False
+                        ),
+                        "model_type",
+                        None,
+                    )
+                except Exception:  # pylint: disable=broad-exception-caught
```
NanoCode012: This runs load AutoConfig, a network call here. I'm not sure it's the best way to handle it in validator.

Claude gave rec to run in `normalize_config` where we already get the model_type

src/axolotl/model_support/paddleocr_vl/processing.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def _build_role_boundaries(self) -> list[RoleBoundary]:
+        tok = self.processor.tokenizer
+        assistant_start = _encode_markers(tok, ["Assistant:\n"])
+        eos = getattr(tok, "eos_token_id", None)
+        if not assistant_start or eos is None:
+            return []
+
+        boundaries = []
+        user_start = _encode_markers(tok, ["User: "])
```
NanoCode012: There's a trailing space here.  How does the template actually mark user? Would this be two / three separate tokens?

NanoCode012 (commented):
First review pass on main points but agreed with Wing's point. Let's keep this small and focused on model only so we can get that part in.

tests/test_processing_strategies.py
thad0ctor: Good call — replaced with the file's fake-tokenizer pattern (ids captured from the real tokenizer), fully offline now in 15768b2fa.

src/axolotl/utils/schemas/config.py
thad0ctor: Agreed — dropped the AutoConfig fallback and moved the check to normalize_config where model_type is already resolved (warns + disables the auto-enabled kernels) in 15768b2fa.

src/axolotl/model_support/paddleocr_vl/processing.py
thad0ctor: Great catch, this was a live bug: the template emits literal "User: " and the trailing space only stays a standalone token when an image special token follows — on text-only turns it BPE-merges into the first content word, so the marker never matched; fixed to the stable 2-token "User:" prefix (+ regression test) in 15768b2fa.

winglian (approved):

Issue #3077, opened 2025-08-17
fsdp with `cpu_ram_efficient_loading=false` results in NaN loss and gradient values


Co1lin:
### Please check that this issue hasn't been reported before.

- [x] I searched previous [Bug Reports](https://github.com/axolotl-ai-cloud/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

For the configuration below, only changing `cpu_ram_efficient_loading` to `false` leads to NaN loss and gradient values. While it is ok to use `cpu_ram_efficient_loading: true` with `bf16: true`, using `fp8: true` requires `cpu_ram_efficient_loading` to be `false` due to [this](https://github.com/huggingface/accelerate/blob/23cf4ef8a3b58f016f63eeb158b4aa2c3e79fe6f/src/accelerate/accelerator.py#L525).

```yaml
seed: 42
# Allow overwrite yml config using from cli
strict: false
# Resume from a specific checkpoint dir
# resume_from_checkpoint:
# This is the huggingface model that contains *.pt, *.safetensors, or *.bin files
# This can also be a relative path to a model on disk
base_model: Qwen/Qwen3-Coder-30B-A3B-Instruct
# base_model: Qwen/Qwen3-8B
trust_remote_code: true
# Where to save the full-finetuned model to
output_dir: train/outputs/sft_stack_test_8k

# wandb configuration if you're using it
# Make sure your `WANDB_API_KEY` environment variable is set (recommended) or you login to wandb with `wandb login`.
# wandb_mode: # "offline" to save run metadata locally and not sync to the server, "disabled" to turn off wandb
wandb_project: ar_sft # Your wandb project name
wandb_entity: cu_colin_org # A wandb Team name if using a Team
# wandb_watch:
wandb_name: sft_stack_30k # Set the name of your wandb run
wandb_run_id: # Set the ID of your wandb run
# wandb_log_model: # "checkpoint" to log model to wandb Artifacts every `save_steps` or "end" to log only at the end of training

# Training hyperparameters
sequence_len: 20480
# dp_replicate_size: 1
# Total: dp_shard_size × tensor_parallel_size × context_parallel_size = #GPUs on a single node
# dp_shard_size: 8         # FSDP across ? GPUs
# tensor_parallel_size: 1  # TP across ? GPUs
# context_parallel_size: 1 # CP across ? GPUs
flash_attention: true
sample_packing: true
eval_sample_packing: true
pad_to_sequence_len: true

[...]

PR #3791, opened 2026-07-03
fix fp8: exclude MoE router linears from torchao fp8 conversion

ved1beta:
# Description
fp8 training converted every nn.Linear except first/last to Float8Linear, including the MoE router gate tensorwise-fp8 router logits destabilize softmax/topk routing and diverge to NaN. This patches accelerate's fp8 conversion to skip router/gate linears while keeping experts and attention in fp8.
## Motivation and Context
#3077

## How has this been tested?
runs fine without nan 

## AI Usage Disclaimer
fable 5 helped T_T" 

## Summary by CodeRabbit

* **New Features**
  * FP8 training now better supports Mixture-of-Experts models by automatically excluding router/gate layers from FP8 conversion.

* **Bug Fixes**
  * Improved FP8 handling to reduce the risk of numerical issues or unexpected behavior when using expert-routing architectures.
  * FP8 setup now applies the router exclusion consistently alongside existing FP8 configuration.

### DIFF
--- a/src/axolotl/loaders/patch_manager.py
+++ b/src/axolotl/loaders/patch_manager.py
@@ -627,6 +627,9 @@ class PatchManager:
     def _apply_fp8_patches(self):
         """Apply patches for FP8 support."""
         if self.cfg.fp8:
+            from axolotl.monkeypatch.accelerate.float8_moe_filter import (
+                patch_fp8_exclude_moe_router,
+            )
             from axolotl.monkeypatch.trainer_accelerator_args import (
                 patch_create_accelerate_code_for_fp8,
             )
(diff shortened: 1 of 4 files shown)


src/axolotl/loaders/patch_manager.py
```diff
@@ -632,13 +632,17 @@ def _fix_nemotron_h_conversion_mapping():
     def _apply_fp8_patches(self):
         """Apply patches for FP8 support."""
         if self.cfg.fp8:
+            from axolotl.monkeypatch.accelerate.float8_moe_filter import (
```
winglian, line 630: was this tested? this import is incorrect.

ved1beta: renamed/ moved the file after changes , that caused mismatch  (100% tested 🫡 )

Issue #3787, opened 2026-07-02
Expert-granularity CPU offload for quantized MoE: 30B-A3B QLoRA in ~7 GB (single resident expert layer, extends the layer_offloading idea)


pjordanandrsn:
## Use case

`quantize_moe_experts` (#3439) got fused-expert 4-bit working, and #3747's selective dequant lowered
the ceiling further — but the floor for a 30B-class MoE QLoRA is still ~17–24 GiB reserved, so
12–16 GB cards are out. Demand for exactly that exists in the tracker (#3530 OOMs a 32 GB card on
gpt-oss-20b; #3402 asks for lower-VRAM MoE training generally).

The missing piece is granularity: `layer_offloading` (#3512) already proves the pinned-CPU +
prefetch-stream pattern in-tree, but at whole-decoder-layer scope its measured savings are modest
(~3 GB on Qwen3.5-35B) because attention/router params are small — in a fine-tuned MoE, ~90 % of
frozen bytes are the experts. Offloading **only the quantized expert tensors, with a single
resident expert layer on GPU** and prefetch of layer N+1 during layer N's compute, moves almost all
of those bytes off-device while touching nothing else.

## Evidence it works (measured, single GPU)

I've implemented this out of tree ([experts4bit-qlora](https://github.com/pjordanandrsn/experts4bit-qlora),
MIT) against transformers-v5 fused experts, and measured on a 12 GB RTX A2000:

| model | peak VRAM with expert offload | without |
|---|---|---|
| Qwen3-30B-A3B QLoRA | **7.16 GB** | OOM |
| Gemma-4-26B-A4B QLoRA | **8.47 GB** | OOM |
| OLMoE-1B-7B QLoRA | 2.57 GB (from 5.97) | 5.97 GB |

Cost is ~+11 % s/step (OLMoE; pinned-memory H2D copies overlap compute for the larger models).
Correctness: the frozen 4-bit forward is bit-identical with offload on/off, including through
`use_reentrant=False` gradient-checkpoint recompute (the recompute re-fetches the same quantized
bytes; adapters stay resident and trainable).

## Proposed shape

An expert-granularity mode that composes with `quantize_moe_experts` — conceptually
`layer_offloading` restricted to the expert parametrizations: pin the packed 4-bit expert storage
in CPU RAM at load, keep one layer's experts on GPU, prefetch the next layer's on a side stream
[...]

PR #3797, opened 2026-07-04
feat(integrations): expert_offload — single-GPU CPU offload of frozen 4-bit MoE experts

pjordanandrsn:
For #3787 (as requested there — a demonstrating PR).

Adds an `axolotl.integrations.expert_offload` plugin that streams a MoE's **frozen 4-bit experts** from pinned CPU RAM to the GPU one block at a time, so a model whose experts exceed VRAM can QLoRA-train on a single small card. Only the experts move — attention, router/gate, norms and the trainable LoRA adapters stay GPU-resident, so per-step PCIe traffic is limited to the experts (the bulk of a MoE's parameters, and the reason it doesn't fit).

### Measured — three arms, one card (RTX 4090, seq 2048)

Re-run together rather than appended, because `layer_offloading` only exists in the axolotl
path and the A2000 numbers below came from a different harness on a different card. One
session, one commit. OLMoE-1B-7B-0924, `load_in_4bit` + `quantize_moe_experts`, qlora r8 on
q/k/v/o, 150 steps, seq 2048, sample_packing, seed 42, checkpointing with `use_reentrant: false`.

| config | loaded GPU | peak GPU | peak vs resident | median s/step | held-out eval |
|---|--:|--:|--:|--:|--:|
| experts resident | 5.18 GB | 5.84 GB | — | 1.62 | 5.453 → 3.648 → 2.491 → 2.354 |
| `expert_offload` | 2.80 GB | **4.11 GB** | **−29.6%** | 3.36 | 5.453 → 3.685 → 2.516 → 2.374 |
| `layer_offloading`, no plugin | 3.36 GB | 5.79 GB | −0.8% | 4.73 | 5.453 → 3.640 → 2.498 → 2.366 |

`layer_offloading` lowers the *loaded* footprint but restages a layer on every forward, so its
peak sits at resident; the single-resident-slot policy over experts is what brings peak down,
and it costs less per step than whole-layer offload. All three converge together, so the memory
rows are comparable rather than an artifact of one arm training differently. axolotl's own
per-step accounting agrees with the sampler (`memory/max_active` peaks 5.06 / 2.25 / 2.63 GiB).

### Measured — the original two-arm A/B (12 GB RTX A2000, seq 256)

Kept rather than replaced: the saving is **regime-dependent**, and the gap between these two
tables is the useful part. Short sequences are weight-dominated, long ones transfer-dominated,
so a headline number only means something with its regime attached.

| config | loaded GPU | peak GPU | held-out eval (before → after) |
|---|--:|--:|--:|
| experts resident | 4.70 GB | 6.00 GB | 1.6448 → 1.2213 |
| experts offloaded | **1.08 GB** | **2.60 GB** | 1.6448 → 1.2270 |

Peak VRAM **−57%**, load footprint **−77%**; both arms share a seed (identical data order), and the loss curves overlay — convergence is preserved. Throughput cost is the per-block H2D copy, ~+11% s/step uncontended on this card. Full telemetry (per-step JSONL, `wandb sync`-able offline runs, charts, the A/B runner + methodology) is in [`ab-telemetry/`](https://github.com/pjordanandrsn/experts4bit-qlora/tree/main/ab-telemetry).

loss
vram

### How it works

Each expert is a bitsandbytes `Linear4bit` whose big tensor is the packed `weight.data` (the `quant_state` scales are ~1/32 the size and stay resident). Every offloaded block's `weight.data` is homed in pinned CPU RAM; a forward **pre-hook** on the MoE block stages its experts just before it runs, and eviction is a **single-resident-slot** policy — staging a block evicts the previously-staged one; there is deliberately no evict post-hook. Under `use_reentrant=False` gradient checkpointing, the backward recompute re-runs the pre-hook, so exactly one block's experts are GPU-resident in forward *and* backward — without depending on when PyTorch stops a recompute.

Gradient checkpointing is **required and enforced** (not just recommended): `matmul_4bit`'s backward re-reads the packed weight via `save_for_backward`, and eviction repoints `weight.data` at a 0-element placeholder. The recompute is what makes eviction both correct and actually memory-freeing (instead of pinning every staged weight alive as a saved tensor). While evicted, modules hold only the 0-element placeholder, so a stray `.to()` or checkpoint-save never drags the experts back to the GPU; a `state_dict` post-hook substitutes the CPU homes so full-model saves stay correct.

### Scope and guards

- Single-GPU QLoRA only — `pre_model_load` refuses under FSDP / DeepSpeed / `expert_parallel`, and requires `load_in_4bit` + `adapter: qlora` + `gradient_checkpointing: true` with `use_reentrant: false`, each with an actionable error message.
- Per-expert `Linear4bit` layouts (`experts` `ModuleList`): Mixtral, Qwen2/3-MoE, OLMoE, DeepSeek-MoE, Jamba, … PEFT `base_layer` wrappers are unwrapped. Fused-expert layouts (GPT-OSS, DBRX) aren't 4-bit-quantized per-expert and are left untouched.
- Complements `layer_offloading` (whole-layer): measured above, the built-in leaves peak at resident (−0.8%) on this model while this brings it down 29.6% for less per-step traffic. Orthogonal to `expert_parallel` (multi-GPU sharding — the opposite regime). Chose the standalone-plugin shape to keep it self-contained; happy to refold it into the `layer_offloading` mixin instead if you'd prefer.

### Tests

`tests/integrations/test_expert_offload.py` — 23 tests (21 CPU + 2 CUDA-gated):

- args + cross-field config validation (each rejected combination asserted)
- MoE-block discovery incl. PEFT `base_layer` unwrap, dedup, and non-4bit/singleton skip
- **the core proof on CPU**: with gradient checkpointing, offloaded training produces gradients identical to the non-offloaded reference (the recompute reads staged weights, not eviction placeholders); exactly one block resident at any instant; `state_dict` home-substitution
- CUDA (real bitsandbytes `Linear4bit`): install frees the experts' bytes off the GPU; offloaded grads match the non-offloaded reference through the real `matmul_4bit` backward

Run on a 12 GB RTX A2000: 23/23 pass. Example config in `src/axolotl/integrations/expert_offload/examples/`.

## Summary by CodeRabbit

* **New Features**
  * Added an Expert Offload integration for single-GPU MoE QLoRA training, helping move frozen 4-bit experts out of GPU memory while training continues.
  * Added a new configuration option to enable expert offloading and control CPU memory pinning.
  * Added an example setup for using the feature with a supported model and training configuration.

* **Bug Fixes**
  * Improved compatibility checks to prevent unsupported training setups from being used with expert offload.

* **Documentation**
  * Added usage notes, setup requirements, and expected performance impact details.

* **Tests**
  * Added coverage for configuration validation, offload behavior, gradient correctness, and state saving.

---

Developed with [Claude Code](https://claude.com/claude-code) (Anthropic) — implementation, tests, and the validation runs described in this PR were produced with AI assistance under the author's direction and review.

### DIFF
--- a/src/axolotl/integrations/expert_offload/args.py
+++ b/src/axolotl/integrations/expert_offload/args.py
@@ -8,14 +8,16 @@
 
 """Pydantic args for the expert-offload plugin."""
 
-from pydantic import BaseModel
+from pydantic import BaseModel, model_validator
 
 
 class ExpertOffloadArgs(BaseModel):
     """Input args for the expert_offload plugin. See the integration README.
 
-    Cross-field requirements (4-bit adapter, gradient checkpointing, single-GPU) depend on config
-    fields this model cannot see, so they are validated in the plugin's ``pre_model_load``.
+    Axolotl folds plugin args into the full config by inheritance
+    (``integrations/config.py::merge_input_args``), so the cross-field validator below runs on
+    the merged config and sees the base-config fields (``load_in_4bit``, ``fsdp``, ...); the
+    ``getattr`` defaults keep the model instantiable standalone as well.
     """
 
     expert_offload: bool = False
@@ -28,3 +30,55 @@ class ExpertOffloadArgs(BaseModel):
     """Home the offloaded expert weights in pinned CPU memory so the per-block host->device copy is
     truly asynchronous. Set false only if pinned memory is scarce (falls back to a correct but
     synchronous pageable copy)."""
+
+    @model_validator(mode="after")
+    def validate_expert_offload_requirements(self):
+        if not self.expert_offload:
+            return self
+
+        errors: list[str] = []
+
+        # 4-bit experts: the mechanism swaps ``Linear4bit.weight.data``.
+        if not getattr(self, "load_in_4bit", False):
+            errors.append(
+                "requires load_in_4bit: true (it offloads 4-bit Linear4bit experts)"
+            )
+        if getattr(self, "adapter", None) not in ("lora", "qlora"):
+            errors.append("requires adapter: qlora (or lora with load_in_4bit)")
+
+        # Gradient checkpointing (use_reentrant=False) is load-bearing: the backward recompute
+        # re-stages each block's experts, and it is what lets eviction actually free memory rather
+        # than pin every staged weight alive as a matmul_4bit saved-for-backward tensor.
+        # ``use_reentrant`` must be EXPLICITLY false: this validator runs at config-parse time,
+        # before ``normalize_config`` defaults omitted kwargs to ``{"use_reentrant": True}``.
+        if not getattr(self, "gradient_checkpointing", False):
+            errors.append(
+                "requires gradient_checkpointing: true (the backward recompute re-stages experts; "
+                "without it, offload is neither correct nor memory-saving)"
+            )
+        elif (getattr(self, "gradient_checkpointing_kwargs", None) or {}).get(
+            "use_reentrant"
+        ) is not False:
+            errors.append(
+                "requires an explicit gradient_checkpointing_kwargs.use_reentrant: false "
+                "(axolotl defaults it to true when omitted; reentrant checkpointing does not "
+                "re-run the block pre-hook on recompute)"
+            )
+
+        # Single GPU: FSDP / DeepSpeed / expert-parallel move or shard these same weights and
+        # would race the stage/evict swaps.
+        if getattr(self, "fsdp_config", None) or getattr(self, "fsdp", None):
+            errors.append("is single-GPU only and incompatible with FSDP")
+        if getattr(self, "deepspeed", None):
+            errors.append("is single-GPU only and incompatible with DeepSpeed")
+        if (getattr(self, "expert_parallel_size", None) or 1) > 1:
+            errors.append(
+                "is incompatible with expert_parallel (both manage the expert weights)"
+            )
+
+        if errors:
+            raise ValueError(
+                "expert_offload is enabled but the config is incompatible:\n  - "
+                + "\n  - ".join(errors)
+            )
+        return self
--- a/src/axolotl/integrations/expert_offload/args.py
+++ b/src/axolotl/integrations/expert_offload/args.py
@@ -14,25 +14,18 @@ from pydantic import BaseModel, model_validator
 class ExpertOffloadArgs(BaseModel):
     """Input args for the expert_offload plugin. See the integration README.
 
-    Axolotl folds plugin args into the full config by inheritance
-    (``integrations/config.py::merge_input_args``), so the cross-field validator below runs on
-    the merged config and sees the base-config fields (``load_in_4bit``, ``fsdp``, ...); the
-    ``getattr`` defaults keep the model instantiable standalone as well.
+    ``merge_input_args`` folds plugin args into the full config, so the validator below sees the
+    merged base-config fields; the ``getattr`` defaults keep the model instantiable standalone.
     """
 
     expert_offload: bool = False
-    """Keep frozen 4-bit MoE experts in CPU RAM and stream one block's experts to the GPU at a time,
-    lowering peak VRAM so MoE models whose experts exceed VRAM can QLoRA-train on a small GPU.
-    Requires a 4-bit adapter (``load_in_4bit`` + ``adapter: qlora``) and
-    ``gradient_checkpointing: true`` with ``use_reentrant: false``. One GPU per replica: single-GPU
-    and plain DDP (multi-GPU data parallel) are supported — under DDP each rank homes its own pinned
-    copy of the experts, so CPU RAM cost scales with world size. FSDP / DeepSpeed / expert-parallel
-    shard or move the same weights and are refused."""
+    """Stream frozen 4-bit MoE experts from pinned CPU RAM to the GPU one block at a time, cutting
+    peak VRAM. Requires ``load_in_4bit`` + ``adapter: qlora`` and non-reentrant gradient
+    checkpointing; single-GPU or plain DDP. See the integration README."""
 
     expert_offload_pin_memory: bool = True
-    """Home the offloaded expert weights in pinned CPU memory so the per-block host->device copy is
-    truly asynchronous. Set false only if pinned memory is scarce (falls back to a correct but
-    synchronous pageable copy)."""
+    """Home offloaded experts in pinned CPU memory for async H2D copies; set false only if pinned
+    memory is scarce."""
 
     @model_validator(mode="after")
     def validate_expert_offload_requirements(self):
@@ -41,7 +34,7 @@ class ExpertOffloadArgs(BaseModel):
 
         errors: list[str] = []
 
-        # 4-bit experts: the mechanism swaps ``Linear4bit.weight.data``.
+        # The mechanism swaps packed 4-bit ``weight.data`` — 4-bit only.
         if not getattr(self, "load_in_4bit", False):
             errors.append(
                 "requires load_in_4bit: true (it offloads 4-bit Linear4bit experts)"
@@ -49,11 +42,8 @@ class ExpertOffloadArgs(BaseModel):
         if getattr(self, "adapter", None) not in ("lora", "qlora"):
             errors.append("requires adapter: qlora (or lora with load_in_4bit)")
 
-        # Gradient checkpointing (use_reentrant=False) is load-bearing: the backward recompute
-        # re-stages each block's experts, and it is what lets eviction actually free memory rather
-        # than pin every staged weight alive as a matmul_4bit saved-for-backward tensor.
-        # ``use_reentrant`` must be EXPLICITLY false: this validator runs at config-parse time,
-        # before ``normalize_config`` defaults omitted kwargs to ``{"use_reentrant": True}``.
+        # The recompute is what makes eviction correct and memory-freeing. ``use_reentrant`` must
+        # be EXPLICITLY false: this runs before ``normalize_config`` defaults it to true.
         if not getattr(self, "gradient_checkpointing", False):
             errors.append(
                 "requires gradient_checkpointing: true (the backward recompute re-stages experts; "
@@ -68,13 +58,13 @@ class ExpertOffloadArgs(BaseModel):
                 "re-run the block pre-hook on recompute)"
             )
 
-        # One GPU per replica: plain DDP is fine (per-process replicas; the offloaded weights go
-        # on DDP's ignore list at install), but FSDP / DeepSpeed / expert-parallel move or shard
-        # these same weights and would race the stage/evict swaps.
+        # FSDP / DeepSpeed / expert-parallel move or shard the same weights (plain DDP is fine).
         if getattr(self, "fsdp_config", None) or getattr(self, "fsdp", None):
             errors.append("is incompatible with FSDP (use single-GPU or plain DDP)")
         if getattr(self, "deepspeed", None):
-            errors.append("is incompatible with DeepSpeed (use single-GPU or plain DDP)")
+            errors.append(
+                "is incompatible with DeepSpeed (use single-GPU or plain DDP)"
+            )
         if (getattr(self, "expert_parallel_size", None) or 1) > 1:
             errors.append(
                 "is incompatible with expert_parallel (both manage the expert weights)"
--- a/src/axolotl/integrations/expert_offload/args.py
+++ b/src/axolotl/integrations/expert_offload/args.py
@@ -1,0 +1,30 @@
+# Copyright 2026 Axolotl AI. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+"""Pydantic args for the expert-offload plugin."""
+
+from pydantic import BaseModel
+
+
+class ExpertOffloadArgs(BaseModel):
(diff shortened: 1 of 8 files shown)


src/axolotl/integrations/expert_offload/plugin.py
```diff
@@ -0,0 +1,92 @@
...
+        # Single GPU: FSDP / DeepSpeed / expert-parallel move or shard these same weights and would
+        # race the stage/evict swaps.
+        if getattr(cfg, "fsdp_config", None) or getattr(cfg, "fsdp", None):
+            errors.append("is single-GPU only and incompatible with FSDP")
+        if getattr(cfg, "deepspeed", None):
+            errors.append("is single-GPU only and incompatible with DeepSpeed")
+        if (getattr(cfg, "expert_parallel_size", None) or 1) > 1:
+            errors.append(
+                "is incompatible with expert_parallel (both manage the expert weights)"
+            )
```
winglian: these could be moved to `args.py` and use the native pydantic schema validation

pjordanandrsn: Done in fa42e43 — moved the whole cross-field check into a `model_validator(mode="after")` on `ExpertOffloadArgs`. Since `merge_input_args` folds plugin args into the config by inheritance, the validator sees the merged config (same pattern as liger's `tensor_parallel_size` check), so it now fires at config-parse time — before the model download — and the plugin's `pre_model_load` is gone entirely.

One tightening that fell out: `use_reentrant` must now be **explicitly** `false`. Parse-time validation runs before `normalize_config` defaults omitted `gradient_checkpointing_kwargs` to `{"use_reentrant": True}`, so an omitted value can't be trusted at that point — the validator demands the explicit setting, with the reason in the error message.

Verified end-to-end through the real `merge_input_args` (bare enable rejected with all requirements listed; omitted kwargs rejected; valid config accepted; plugin-off config unaffected), plus the unit tests now assert `ValidationError` through a merged-shape model instead of calling a plugin helper.

src/axolotl/integrations/expert_offload/offload.py
```diff
@@ -0,0 +1,270 @@
...
+    if no offloadable MoE block was found).
+    """
+    blocks = find_moe_expert_blocks(model)
+    if not blocks:
+        LOG.warning(
+            "expert_offload: no 4-bit MoE expert blocks found. This integration offloads "
+            "per-expert bitsandbytes Linear4bit weights (Mixtral / Qwen-MoE / OLMoE / DeepSeek "
+            "style). Fused-expert or non-quantized models are unaffected."
+        )
+        return []
```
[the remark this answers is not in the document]
pjordanandrsn: Done in fa42e43 — `install_expert_offload` now raises `RuntimeError` when no 4-bit MoE expert blocks are found (the message names the supported layouts and the fix), with a regression test (`test_install_raises_when_no_offloadable_blocks`).

pjordanandrsn:
Review addressed in fa42e43:

- **validation → `args.py`** as a pydantic `model_validator` (@winglian's comment — details in that thread); the plugin now has no `pre_model_load` at all
- **fail loudly when nothing to offload**: `install_expert_offload` raises `RuntimeError` instead of warn + silent no-op, + regression test
- **`pytest.importorskip("bitsandbytes")`** in the CUDA test class (clean skip when CUDA is present but bnb isn't)
- **narrowed `_is_pinned`'s except** to `(RuntimeError, AssertionError)`, matching `_to_home`

Declining one nit — `ConfigDict(extra="forbid")` on `ExpertOffloadArgs`: plugin args become a *parent class* of the merged `AxolotlInputConfig` (`merge_input_args`), so a `model_config` set here merges into the full config's `model_config` and could change unknown-key behavior for the entire config, not just these two fields. The base schema deliberately leaves pydantic's default `extra="ignore"`, and none of the existing integrations set `ConfigDict` — happy to revisit if there's a scoped way to do it.

Suite after the changes: 25 tests (23 CPU + 2 CUDA against real bitsandbytes `Linear4bit`), all green on a 12 GB RTX A2000.

pjordanandrsn:
Ran the benchmarks through `axolotl train` on this branch—no out-of-tree runner this time. Four arms, one pod, one session: single/DDP × resident/offload, seed-matched.

| arm | loaded GPU | peak GPU | s/it (steady) | eval loss (step 150) |
|---|--:|--:|--:|--:|
| single resident | 5.03 GB | 17.9 GB | 2.62 | 2.359 |
| single offload | **1.21 GB** | 15.4 GB | 3.55 | 2.365 |
| DDP resident, per rank | 5.19 GB | 17.9 GB | 2.73 | 2.214 |
| DDP offload, per rank | **1.36 GB** | 15.7 GB | 3.66 | 2.213 |

Host: 2×RTX 3090 24 GB, fresh `pip install -e .` of the branch—torch 2.12.1+cu130, transformers 5.12.1, peft 0.19.1, bitsandbytes 0.49.1, accelerate 1.13.0. Deltas from the shipped example, identical across all arms: `seed: 42`, `max_steps: 150`, `eval_steps: 50`, `save_strategy: "no"`, `wandb_mode: offline`, `flash_attention: false`. DDP arms differ from single arms only by visible-GPU count (effective batch 8 vs 4). Full configs, 2s-interval VRAM traces, and raw logs: https://github.com/pjordanandrsn/experts4bit-qlora/tree/main/ab-telemetry/axolotl-ab

Running it through your trainer caught something the unit suite couldn't—current transformers stores OLMoE's experts as fused 3D stacks, so plain `load_in_4bit` leaves them in bf16: resident OOMs a 24 GB card during PEFT prep and the plugin has nothing 4-bit to offload. Two commits on the branch fix it: the example now sets `quantize_moe_experts: true`, and the plugin discovers/offloads the parametrized stacks that flag produces (the packed `parametrizations.<p>.original` is homed exactly like a `Linear4bit` `weight.data`; its `quant_state` stays resident). Also on the branch since your comment: plain-DDP support—per-rank pinned homes, offloaded weights on DDP's parameter ignore list—measured above.

Numbers worth being straight about: loaded footprint drops 76%, but steady-state s/it costs +35% in this setup—the parametrized layout dequantizes the full 64-expert stack per access and the training path stages synchronously by design at seq 2048. The PR body's ~11% figure came from the per-expert `Linear4bit` path at seq 256 over PCIe 3.0—different regime, both numbers are real. Convergence is unaffected either way: seed-matched resident/offload eval curves overlay within noise (table).

Happy to run Qwen3-30B-A3B through the trainer on a bigger-RAM host as a follow-up if that's useful for the review. No rush—branch is stable as of `fd12f92`.

pjordanandrsn:
Rebased onto current main (230fe761) and folded in the bitsandbytes parametrize-cache workaround: `use_reentrant=False` checkpointing's early stop skips bnb's cache-disable forward hook, leaking the global parametrize cache and retaining every dequantized expert (~4x the packed bytes; ~53 GB on Qwen3-30B). The installer now strips the hook pair from offloaded expert modules — their params are single-access per forward, so the cache buys them nothing — with `AXOLOTL_EXPERT_OFFLOAD_KEEP_BNB_CACHE_HOOKS=1` as the escape hatch, plus a bnb-free regression test. Upstream fix filed as bitsandbytes-foundation/bitsandbytes#1999; the workaround stays correct either way.

Re-validated post-rebase on GPU (RTX A5000): the full `test_expert_offload.py` + leak-test suite passes (40 tests), and a 15-step OLMoE off-vs-on A/B reproduces the bit-identical step-0 loss (0.9636 == 0.9636) with final-loss drift 0.010 (the usual `index_add_` atomics floor).

src/axolotl/integrations/expert_offload/args.py
```diff
@@ -0,0 +1,88 @@
...
+    """
+
+    expert_offload: bool = False
+    """Keep frozen 4-bit MoE experts in CPU RAM and stream one block's experts to the GPU at a time,
+    lowering peak VRAM so MoE models whose experts exceed VRAM can QLoRA-train on a small GPU.
+    Requires a 4-bit adapter (``load_in_4bit`` + ``adapter: qlora``) and
+    ``gradient_checkpointing: true`` with ``use_reentrant: false``. One GPU per replica: single-GPU
+    and plain DDP (multi-GPU data parallel) are supported — under DDP each rank homes its own pinned
+    copy of the experts, so CPU RAM cost scales with world size. FSDP / DeepSpeed / expert-parallel
+    shard or move the same weights and are refused."""
```
ved1beta: trim discription

src/axolotl/integrations/expert_offload/offload.py
```diff
@@ -0,0 +1,452 @@
...
+tensor. ``gradient_checkpointing: true`` with an explicit ``use_reentrant: false`` is enforced at
+config validation (the ``ExpertOffloadArgs`` schema validator).
+
+One GPU **per replica**: plain DDP (multi-GPU data parallel) is supported — each rank is its own
+process with a full replica, so each rank homes its own pinned copy of the experts (CPU RAM cost
+scales with world size) and stages to its own device; the offloaded weights are registered on
+DDP's ignore list so the initial module-state sync never touches the 0-element placeholders.
+FSDP / DeepSpeed / expert-parallel move or shard these same weights and would race the
+stage/evict swaps; the config schema refuses to enable under any of them.
+"""
```
ved1beta, line 20: trim

ved1beta (changes requested):
comments look over explained

src/axolotl/integrations/expert_offload/args.py
pjordanandrsn: Trimmed in 4be9846 — field docstrings are down to 2–3 lines (requirements + a pointer to the integration README).

src/axolotl/integrations/expert_offload/offload.py
pjordanandrsn: Trimmed in 4be9846, and took the pass over the whole plugin rather than just this spot: module docstring 50→11 lines, every helper docstring cut to the non-obvious why, −109 comment lines total. The mechanism detail (layouts, DDP semantics, why non-reentrant checkpointing is required) stays in the integration README. Comments-only change — code AST is identical.

pjordanandrsn:
@ved1beta thanks — over-explaining trimmed in 4be9846: module docstring 50→11 lines, field docstrings down to 2–3, −109 comment lines total (mechanism detail moved to the integration README). Both threads addressed above — mind taking another look? Happy to cut further.

pjordanandrsn:
@winglian @ved1beta — ready for another look whenever you have a moment.

Both review points are addressed and those threads are resolved: the cross-field validation moved into `args.py` as a pydantic `model_validator(mode="after")` (fa42e43), and the over-explained comments/docstrings trimmed (4be9846, −109 comment lines).

The one thing I can't do from my side is CI — `Tests`, `docker-e2e-tests` and `Preview` are all sitting at `action_required` on the current head (4be9846), so no CI has actually run on this branch yet. Could one of you approve the workflows? Happy to rebase onto current main first if it's drifted.

pjordanandrsn:
@ved1beta both items landed in 4be9846—the review's still marked CHANGES_REQUESTED, which is what's holding this at BLOCKED.

@winglian the workflows have never run; they're at action_required on the fork gate. Mind approving them? Happy to rebase onto current main first.

ved1beta (approved):
LGTM

pjordanandrsn:
@NanoCode012 @ved1beta this is `APPROVED` and `CLEAN` with CI green as of `83f9f062`, and all review threads are now resolved. Ready to merge whenever one of you has a moment.

pjordanandrsn:
@NanoCode012 with #3915 merged and the point release still ahead, could #3797 ride along? It's approved (ved1beta, Aug 4), CLEAN, CI green on 83f9f062, no open threads; only the July `hold` is left. With #3915 in main the installer's own hook-stripping is belt-and-braces, so I'm happy to drop it in a follow-up or here if you'd rather.

ved1beta (changes requested):
Please can you run your OLMoE A/B a third time with layer_offloading: true (no plugin) and add the row to the table.
note:  the bnb cache-hook code is dead against 0.50.2

pjordanandrsn:
Third arm added, and I re-ran all three together rather than appending a row: `layer_offloading` only exists in the axolotl path, and the existing table is a different harness on a different card. One 4090, one commit. OLMoE-1B-7B-0924, 150 steps, seq 2048, packing, seed 42.

| config | loaded | peak | vs resident | s/step | final eval |
|---|--:|--:|--:|--:|--:|
| experts resident | 5.18 GB | 5.84 GB | — | 1.62 | 2.354 |
| `expert_offload` | 2.80 GB | **4.11 GB** | **−29.6%** | 3.36 | 2.374 |
| `layer_offloading`, no plugin | 3.36 GB | 5.79 GB | −0.8% | 4.73 | 2.366 |

`layer_offloading` lowers the loaded footprint but streams a layer back each forward, so its peak sits at resident. The single-resident-slot policy over experts is what moves peak, and it is cheaper per step as well.

I kept the A2000 seq-256 A/B in the body next to this one, because the gap between them is the useful part: −57% at ~+11% s/step there, −29.6% at ~2× here. Short sequences are weight-dominated, long ones transfer-dominated, so the headline numbers only mean something with the regime attached — which is why both are now labelled rather than one replacing the other.

Hooks removed, along with the test that only pinned them. They weren't dead on the branch, though: it pinned `bitsandbytes==0.49.1`, which still registers the pair without `always_call=True`. main pins `0.50.2`, so I rebased onto main — the removal is correct there, and the numbers above come from that environment.

One thing this makes worth deciding before I write it rather than after: now that the built-in's peak behaviour on a MoE is measured, would you rather expert granularity landed *inside* the `layer_offloading` mixin instead of beside it as a plugin? That is a different PR and I would rather you picked the shape.

PR #2772, opened 2025-06-09
Fix: lora kernel pre-patch applied despite post-patch not applied

NanoCode012:
# Description

Fix #2770 

## How has this been tested?

Added test and manual run before/after to ensure error appears and got fixed

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved handling of LoRA dropout by ensuring kernel patching is skipped when dropout is non-zero, with a warning logged for clarity.

- **Tests**
	- Added a new test to verify that self-attention kernel patching does not occur when LoRA dropout is set above zero.

PR #3806, opened 2026-07-08
Fix/lora attn kernel dropout gate

thad0ctor:
# Description

Removes the stale `lora_dropout == 0` gate in `PatchManager._apply_self_attention_lora_patch` so the fused QKV/O self-attention source rewrite (`patch_self_attn_lora`) runs when `lora_qkv_kernel`/`lora_o_kernel` are enabled with non-zero LoRA dropout.

## Motivation and Context

The gate dates to #2772 (2025-06), when the fused Functions could not handle dropout. Since #3528 (2026-03) the fused QKV/QK Functions do support dropout — applied outside the `autograd.Function` with a shared QKV mask — and the MLP path (`apply_lora_kernel_patches`) already patches unconditionally. The attention gate was left behind, so with `lora_qkv_kernel: true` / `lora_o_kernel: true` and any `lora_dropout > 0`, the QKV/O source rewrite was silently skipped and the attention forward kept calling the unfused projections. `patch_self_attn_lora` itself has no dropout dependency.

Out of scope for #3800; flagged in https://github.com/axolotl-ai-cloud/axolotl/pull/3800#issuecomment-4913292978.

## How has this been tested?

Extended `tests/e2e/patched/lora_kernels/test_lora_kernel_patching.py::test_kernel_training_integration_dropout_non_zero` to set `lora_qkv_kernel`/`lora_o_kernel: true` (the flags that actually reach the gate) with `lora_dropout: 0.1`, and to assert the source rewrite ran via `type(self_attn)._original_forward`. The test fails on `main` (gate present) and passes with this change. Pinned `ruff` check + format clean.

## AI Usage Disclaimer

Opus 4.8 used throughout

## Types of changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Self-attention LoRA kernel patching now applies even when dropout is enabled, improving compatibility for more training configurations.
  * Integration coverage was strengthened to verify kernel patching behavior in dropout-enabled runs and confirm the attention module is rewritten as expected.


thad0ctor:
@coderabbitai review

winglian (approved):

PR #3807, opened 2026-07-08
test: mark flaky VRAM-leak regression test as non-strict xfail

winglian:
# Description

Marks `test_no_vram_leak_regression` in `tests/e2e/test_activation_offloading.py` with `@pytest.mark.xfail(reason="flaky", strict=False)`.

## Motivation and Context

The test is flaky in CI. With a non-strict xfail it still runs every cycle for signal — a failure is reported as XFAIL in the summary instead of failing the job, and a pass shows as XPASS. Same convention already used in `tests/e2e/integrations/test_scattermoe_lora_kernels.py` and `tests/e2e/solo/test_trainer_loss_calc.py`.

## How has this been tested?

Marker-only change, identical to the existing flaky-test pattern in the repo; syntax verified locally.

## Types of changes

Test-only change (no runtime code affected)

## Summary by CodeRabbit

* **Tests**
  * Marked an unstable end-to-end regression test as an expected failure, so intermittent failures will no longer break the test suite.

### DIFF
--- a/tests/e2e/test_activation_offloading.py
+++ b/tests/e2e/test_activation_offloading.py
@@ -327,6 +327,7 @@ class TestActivationOffloading:
             f"grad diverged: max|d|={(grad_ref - grad_off).abs().max().item()}"
         )
 
+    @pytest.mark.xfail(reason="flaky", strict=False)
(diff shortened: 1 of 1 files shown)


NanoCode012 (approved):
Also saw this happening too

PR #3825, opened 2026-07-10
feat(dpo): support tools and tool_calls in chat_template strategy

### LINKED ISSUE #3824: DPO chat_template Strategy Drops tool_calls and Never Passes tools to the Chat Template

### Expected Behavior

Using `rl: dpo` with `type: chat_template.default` on an OpenAI-format tool-calling dataset (`tools` + `messages` with `tool_calls` / `tool` roles + `chosen` / `rejected`) — the common format for **multi-step tool-use preference data**, where the shared context contains assistant `tool_calls` turns and `tool` result turns — should render the same way SFT's `chat_template` strategy does:

- the `tools` list is passed to `apply_chat_template` so tool definitions are rendered into the system prompt
- assistant `tool_calls` turns are preserved in the rendered prompt
- messages with `content: null` (standard for OpenAI-format tool call turns) are handled
- a chosen/rejected response that is itself a tool call is extracted correctly
- `reasoning_content` in chosen/rejected renders into the template's thinking block

### Current behaviour

I hit this while training tool calling with DPO on my own OpenAI-format dataset. `src/axolotl/prompt_strategies/dpo/chat_template.py` copies only `role` and `content` from each message, so:

1. **assistant `tool_calls` are silently dropped** — the prompt renders an empty assistant turn, and the model trains on corrupted context (no error, easy to miss)
2. **`tools` is never passed to `apply_chat_template`** — tool definitions never reach the system prompt
3. **`content: null` crashes while rendering the chat template** — with `qwen3_5`: `TypeError: 'NoneType' object is not iterable` (the template iterates non-string content); other templates fail similarly (e.g. `can only concatenate str (not "NoneType") to str`)
4. if `chosen`/`rejected` is itself a tool call (empty content), the response extraction (`result["chosen"].find(chosen["content"])`) fails and **leaks the internal `[[dummy_message]]` prompt into the training text**
5. similarly, `reasoning_content` in a chosen/rejected response is silently dropped by the content-based extraction — for thinking templates (e.g. `qwen3_5`) the response should continue the `<think>` block that the generation prompt opens

Repro rendering (`qwen3_5` template, `content: ""` to avoid the crash):

```
<|im_start|>system
You are a helpful weather assistant.<|im_end|>      <-- tools never rendered
<|im_start|>user
What's the weather in Paris?<|im_end|>
<|im_start|>assistant
<|im_end|>                                          <-- tool_calls silently dropped
<|im_start|>user
<tool_response>
22C, sunny
</tool_response><|im_end|>
<|im_start|>assistant
<think>
```

Related: #3217 fixed the `KeyError: 'tool'` for tool *roles*, and #3228 adds `tool` to `message_property_mappings` — but neither preserves `tool_calls` nor renders `tools`. The SFT `chat_template` strategy already supports all of this (`ChatTemplateStrategy._get_tools`, `transform_message`); the DPO strategy is missing feature parity.

### Steps to reproduce

1. Create `data.jsonl` (OpenAI chat format). This is the typical shape when building DPO pairs for multi-step tool use — e.g. from agent trajectories, or by converting SFT tool-calling datasets (ToolACE, hermes-function-calling) into preference pairs; same shape as the dataset in #3217:

```json
{
  "tools": [{"type": "function", "function": {"name": "get_weather", "description": "Get current weather for a city", "parameters": {"type": "object", "properties": {"city": {"type": "string"}}, "required": ["city"]}}}],
  "messages": [
    {"role": "system", "content": "You are a helpful weather assistant."},
    {"role": "user", "content": "What's the weather in Paris?"},
    {"role": "assistant", "content": null, "tool_calls": [{"id": "call_1", "type": "function", "function": {"name": "get_weather", "arguments": "{\"city\": \"Paris\"}"}}]},
    {"role": "tool", "tool_call_id": "call_1", "content": "22C, sunny"}
  ],
  "chosen": {"role": "assistant", "content": "It's 22C and sunny in Paris.", "reasoning_content": "The tool returned 22C and sunny, so I should report that."},
  "rejected": {"role": "assistant", "content": "I cannot check the weather.", "reasoning_content": "I don't have access to weather tools."}
}
```

2. Run `axolotl preprocess config.yaml --debug` with the config below
3. Observe the crash during "Mapping RL Dataset":

```
File "/axolotl/src/axolotl/prompt_strategies/dpo/chat_template.py", line 96, in transform_fn
  result["prompt"] = tokenizer.apply_chat_template(
...
File "<template>", line 85, in top-level template code
TypeError: 'NoneType' object is not iterable
```

4. With `"content": ""` instead of `null`, preprocessing succeeds but silently produces the corrupted prompt shown above (no tools, empty assistant turn, no reasoning in responses)

### Config yaml

```yaml
base_model: Qwen/Qwen3.5-27B
rl: dpo
chat_template: qwen3_5
datasets:
  - path: data.jsonl
    type: chat_template.default
micro_batch_size: 1
learning_rate: 5e-6
output_dir: ./outputs/dpo-out
```

### Possible solution

Mirror the SFT `chat_template` strategy in `dpo/chat_template.py`:

- preserve the message properties the chat template actually uses (`tool_calls`, `tool_call_id`, `name`, `reasoning_content`, …) via `JinjaTemplateAnalyzer`, skipping `None` values; union in the standard OpenAI message keys since some templates access properties via `message.get(...)` which template analysis can't see (e.g. `gemma4`)
- read `field_tools` (default `"tools"`) from the sample, decode JSON-encoded strings, and pass `tools=` to `apply_chat_template`
- decode JSON-string `tool_call.function.arguments` into dicts (templates apply `| tojson`)
- extract chosen/rejected by stripping the rendered dummy-user prompt prefix (longest common prefix) instead of searching for the response content, so contentless tool-call responses and `reasoning_content` responses extract cleanly
- pass the config's `chat_template_kwargs` (e.g. `enable_thinking`) to `apply_chat_template`, matching SFT

I have a working patch with tests, verified against `qwen_25` / `qwen3` / `qwen3_5` / `gemma4` / `exaone4` / `llama4` / `llama3_2_vision` / `command_a_tool_use` / `jamba` — PR incoming.

- [x] Linux
- [x] macOS
- [ ] Windows

### Python Version

3.12

### axolotl branch-commit

main/6401a353

### Acknowledgements

inoutro:
# Description

The DPO `chat_template` strategy only copied `role`/`content` from messages, breaking OpenAI-format tool-calling datasets. This PR brings it to parity with the SFT `chat_template` strategy:

- Preserve message properties the chat template uses (`tool_calls`, `tool_call_id`, `name`, `reasoning_content`, …) via `JinjaTemplateAnalyzer`, skipping `None` values. The standard OpenAI message keys are always unioned in, since some templates access properties via `message.get(...)` which template analysis can't see (e.g. `gemma4`).
- Read `field_tools` (default `"tools"`) from the sample, decode JSON-encoded strings (both a whole-list JSON string and per-tool JSON strings), and pass `tools=` to `apply_chat_template`.
- Decode JSON-string `tool_call.function.arguments` into dicts, matching SFT behavior (templates apply `| tojson`).
- Extract chosen/rejected by stripping the longest common prefix with the rendered dummy-user prompt instead of searching for the response content, so contentless tool-call responses and `reasoning_content` responses extract cleanly (thinking templates open `<think>` blocks with slightly different whitespace than completed renderings).
- Pass the config's `chat_template_kwargs` (e.g. `enable_thinking`) to `apply_chat_template`, matching SFT.

Applies to both `chat_template.default` and `chat_template.argilla_chat`.

## Motivation and Context

Fixes #3824. Training tool calling with DPO on OpenAI-format data currently either crashes (`content: null`) or silently corrupts the training prompt (dropped `tool_calls`, missing tool definitions). See the issue for full failure modes and repro.

## How has this been tested?

- New unit tests in `tests/prompt_strategies/test_dpo_chat_templates.py`: tools rendering into the system prompt, `tool_calls` preservation with JSON-string argument decoding, tool-call chosen/rejected extraction (no dummy-message leakage), tools as a list of JSON strings, `reasoning_content` extraction, and `chat_template_kwargs` passthrough
- Full existing test file passes: `pytest tests/prompt_strategies/test_dpo_chat_templates.py` (13 passed, 2 xpassed, no regressions)
- Manually verified rendering across 9 tool-capable registered templates: `qwen_25`, `qwen3`, `qwen3_5`, `gemma4`, `exaone4`, `llama4`, `llama3_2_vision`, `command_a_tool_use`, `jamba`
- End-to-end `axolotl preprocess --debug` with `rl: dpo` on an OpenAI-format tool-calling jsonl (crash before, correct rendering after)
- `pre-commit run` (ruff, mypy, bandit) passes on changed files

## AI Usage Disclaimer

Yes — implemented with Claude Code (analysis, implementation, and tests), reviewed and verified by me.

## Types of changes

## Summary by CodeRabbit

* **New Features**
  * Added support for tool definitions and tool calls in DPO and Argilla chat template rendering.
  * Tool data can be provided as structured lists or JSON-encoded strings.
  * Preserves additional message fields referenced by chat templates.
  * Supports reasoning content and configurable thinking behavior during response generation.

* **Bug Fixes**
  * Improved extraction of chosen and rejected responses from rendered templates.
  * Correctly handles tool-call responses and JSON-encoded function arguments.

### DIFF
--- a/src/axolotl/prompt_strategies/chat_template_utils.py
+++ b/src/axolotl/prompt_strategies/chat_template_utils.py
@@ -1,0 +1,198 @@
+"""
+Shared helpers for rendering OpenAI-format preference (chosen/rejected) messages
+via tokenizer chat templates.
+
+Reused across RL prompt strategies (DPO, ORPO, Bradley-Terry) so tool_calls,
+tools, and reasoning_content handling only needs to be implemented once.
+"""
+
+import json
+
+from axolotl.prompt_strategies.jinja_template_analyzer import JinjaTemplateAnalyzer
+from axolotl.utils.logging import get_logger
+
+LOG = get_logger(__name__)
+
+
+def parse_tools(tools):
+    """Parse tools into a list of dicts, decoding JSON-encoded strings."""
+    if tools is None:
+        return None
+
+    if isinstance(tools, str):
+        try:
+            tools = json.loads(tools)
+        except json.JSONDecodeError as e:
+            LOG.error(f"Error parsing tools as JSON. Error: {e}")
+            raise
+
+    if isinstance(tools, list):
+        parsed_tools = []
+        for tool in tools:
+            # some datasets store each tool as a JSON-encoded string
+            if isinstance(tool, str):
+                try:
+                    tool = json.loads(tool)
+                except json.JSONDecodeError as e:
+                    LOG.error(f"Error parsing tool as JSON. Tool: {tool!r}, Error: {e}")
+                    raise
+            if isinstance(tool, dict) and "function" in tool:
+                function = tool["function"]
+                params = function.get("parameters")
+                if isinstance(params, str):
+                    try:
+                        function["parameters"] = json.loads(params)
+                    except json.JSONDecodeError as e:
+                        LOG.error(
+                            f"Error parsing tool parameters as JSON. "
+                            f"Function: {function.get('name', 'unknown')}, "
+                            f"Parameters string: {params!r}, "
+                            f"Error: {e}"
+                        )
+                        raise
+            parsed_tools.append(tool)
+        return parsed_tools
+
+    raise ValueError(
+        "Unknown tools format. Please convert it into a list[dict].\n"
+        f"Current format: {type(tools)}"
+    )
+
+
+def parse_tool_call_arguments(message):
+    """Decode JSON-encoded tool call arguments so templates receive dicts."""
+    for tool_call in message.get("tool_calls") or []:
+        if "function" in tool_call and "arguments" in tool_call["function"]:
+            args = tool_call["function"]["arguments"]
+            if isinstance(args, str):
+                try:
+                    tool_call["function"]["arguments"] = json.loads(args)
+                except json.JSONDecodeError as e:
+                    LOG.error(
+                        f"Error parsing tool_calls arguments as JSON. "
+                        f"Function: {tool_call.get('function', {}).get('name', 'unknown')}, "
+                        f"Arguments string: {args!r}, "
+                        f"Error: {e}"
+                    )
+                    raise
+
+
+def build_message_transform(message_property_mappings, role_map):
+    """Build a function that maps a raw dataset message to a chat template message,
+    preserving any extra properties the chat template uses (e.g. tool_calls)."""
+
+    def transform_message(message, msg_variables):
+        """Map a raw dataset message to a chat template message."""
+        transformed = {}
+        for target, source in message_property_mappings.items():
+            value = message.get(source)
+            if value is not None:
+                transformed[target] = value
+
+        if "role" in transformed:
+            transformed["role"] = role_map.get(transformed["role"], transformed["role"])
+
+        mapped_sources = set(message_property_mappings.values())
+        for key in msg_variables - mapped_sources:
+            value = message.get(key)
+            if value is not None:
+                transformed[key] = value
+
+        parse_tool_call_arguments(transformed)
+        return transformed
+
+    return transform_message
+
+
+# always preserved: template analysis misses properties accessed via
+# `message.get(...)` (e.g. gemma4), so OpenAI message keys are unioned in
+BASE_MSG_VARIABLES = frozenset(
+    ["tool_calls", "tool_call_id", "name", "reasoning_content", "reasoning"]
+)
+
+
+def make_msg_variables_getter():
+    """Cache chat template message variable analysis per template string."""
+    cache = {}
+
+    def get_msg_variables(chat_template_string):
+        """Return the message properties used by the chat template."""
+        if chat_template_string not in cache:
+            cache[chat_template_string] = (
+                JinjaTemplateAnalyzer(chat_template_string).get_message_vars("messages")
+                | BASE_MSG_VARIABLES
+            )
+        return cache[chat_template_string]
+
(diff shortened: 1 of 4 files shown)


NanoCode012 (commented):
I think I like the direction, but if we're going to bring to parity, we should figure out a way to centralize the changes, otherwise, we'd have to do one for each RL method which would have redundant code

inoutro:
Thanks for the feedback! You're right that fixing this per RL method would get redundant fast.

I moved the tools/tool_calls/reasoning_content handling that was in `dpo/chat_template.py` into a new shared module, `src/axolotl/prompt_strategies/preference_chat_template.py` (message property mapping + role mapping, tools parsing, chat-template variable detection, and prompt/chosen/rejected extraction). `dpo/chat_template.py` now just wires dataset config into it.

This keeps the current PR scoped to the DPO parity fix from #3824, but other RL methods that share the same `prompt/chosen/rejected` string contract (e.g. ORPO, Bradley-Terry) can pick up tools/tool_calls/reasoning_content support later by importing from this module instead of reimplementing it.

While doing this I also noticed the SFT `chat_template.py` strategy already has its own similar (but not identical) tools-parsing and tool_calls-decoding logic. I left that untouched here since it's a different output contract (token-level label masking vs. rendered prompt/chosen/rejected strings) and didn't want to expand this PR's blast radius — happy to look at consolidating that separately if it's useful.

PR #3832, opened 2026-07-11
nightlies fix for upstream indexer rope change in glm isa

winglian:
transformers main switched the GlmMoeDsaIndexer to interleaved rope and removed apply_rotary_pos_emb from the module; select the rope matching the installed indexer via the symbol's presence. Also cast weights_proj output to fp32 before the n_heads**-0.5 scale as the stock indexer does: bf16 rounding there perturbs near-tied index scores enough to flip top-k selections against the eager reference.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility across supported GLM-MoE-DSA implementations by selecting the appropriate rotary-position embedding behavior automatically.
  * Improved the accuracy and consistency of top-k selection by preserving higher numerical precision during indexer weight calculations.


winglian:
kicked off nightlies GHA manually too @ https://github.com/axolotl-ai-cloud/axolotl/actions/runs/29166405488 against this PR

PR #3837, opened 2026-07-12
fix(schemas): honor micro_batch_size/gradient_accumulation_steps defaults in check_batch_size_fields

### LINKED ISSUE #3836: check_batch_size_fields rejects configs that rely on the documented micro_batch_size/gradient_accumulation_steps defaults
`check_batch_size_fields` in `src/axolotl/utils/schemas/validation.py` requires at least two of `micro_batch_size`, `gradient_accumulation_steps`, `batch_size` to be present:

```python
@classmethod
def check_batch_size_fields(cls, data):
    fields = ("micro_batch_size", "gradient_accumulation_steps", "batch_size")
    non_empty_count = sum(1 for field in fields if data.get(field))

    if non_empty_count < 2:
        raise ValueError(f"At least two of {', '.join(fields)} must be set")
    return data
```

Both `micro_batch_size` and `gradient_accumulation_steps` have a documented default of `1` in `training.py`:

```python
micro_batch_size: int | None = Field(default=1, ...)
gradient_accumulation_steps: int | None = Field(default=1, ...)
```

But `check_batch_size_fields` runs in `mode="before"`, so it only sees whatever keys the user actually put in their YAML - the field defaults haven't been applied yet. A config that sets only `gradient_accumulation_steps` and leaves `micro_batch_size` to its documented default of `1` has exactly one of the three fields present, so `non_empty_count == 1` and the validator rejects it outright. Same thing in reverse for a config that only sets `micro_batch_size`.

Minimal repro against current `main`:

```python
from axolotl.utils.schemas.config import AxolotlInputConfig

AxolotlInputConfig(
    base_model="HuggingFaceH4/tiny-random-LlamaForCausalLM",
    datasets=[{"path": "yahma/alpaca-cleaned", "type": "alpaca"}],
    val_set_size=0.05,
    output_dir="./out",
    sequence_len=512,
    gradient_accumulation_steps=4,
    num_epochs=1,
    learning_rate=2e-4,
)
```

```
pydantic_core._pydantic_core.ValidationError: 1 validation error for AxolotlInputConfig
  Value error, At least two of micro_batch_size, gradient_accumulation_steps, batch_size must be set [type=value_error, ...]
```

Dropping `gradient_accumulation_steps` and setting only `micro_batch_size=4` instead hits the exact same error. Either one, by itself, is a perfectly normal config - the whole point of the documented default is that you shouldn't have to set both.

I checked the example configs under `examples/` and 246 of 248 happen to set two of the three fields already, which is presumably why this hasn't shown up in CI - and `tests/utils/schemas/validation/test_config_validators.py`'s GRPO batch-size tests always pass both `micro_batch_size` and `gradient_accumulation_steps` explicitly, so this particular validator has no direct test coverage of the single-field case either way.

I'm not sure what the intent behind requiring two fields was - possibly to catch the case where none of the three are set at all (which would silently fall back to defaults without the user realizing gradient accumulation is off)? If so the check should probably be `non_empty_count == 0` combined with a check against the actual resolved defaults, or the validator should run in `mode="after"` where the real field values (including defaults) are visible, rather than inspecting the pre-default raw dict. Happy to send a PR once there's agreement on which of these was intended - didn't want to guess and paper over what might be a deliberate (if surprising) design choice.

Anai-Guo:
### Problem

`check_batch_size_fields` (in `src/axolotl/utils/schemas/validation.py`) requires at least two of `micro_batch_size`, `gradient_accumulation_steps`, `batch_size` to be present:

```python
non_empty_count = sum(1 for field in fields if data.get(field))
if non_empty_count < 2:
    raise ValueError(f"At least two of {', '.join(fields)} must be set")
```

But it runs as a `@model_validator(mode="before")`, so it only sees the keys the user actually put in their YAML — field defaults have not been applied yet. Both `micro_batch_size` and `gradient_accumulation_steps` carry a **documented default of `1`** (`schemas/training.py`). So a config that sets only one of the pair and relies on the other's default has `non_empty_count == 1` and is rejected outright:

```python
AxolotlInputConfig(
    base_model="HuggingFaceTB/SmolLM2-135M",
    datasets=[{"path": "mhenrichsen/alpaca_2k_test", "type": "alpaca"}],
    gradient_accumulation_steps=4,
)
# ValueError: At least two of micro_batch_size, gradient_accumulation_steps, batch_size must be set
```

Setting only `micro_batch_size` and relying on the `gradient_accumulation_steps` default hits the same error. Either one alone is a perfectly normal config — the point of the documented default is that you shouldn't have to set both. This has not surfaced in CI because the example configs and the `min_base_cfg` fixture happen to set both fields explicitly.

### Fix

Treat an explicit `micro_batch_size` **or** `gradient_accumulation_steps` as sufficient, since whichever one is omitted resolves to its documented default of `1`.

The deprecated `batch_size` is deliberately **not** given the same treatment: it has to be paired with an explicit `micro_batch_size` or `gradient_accumulation_steps` so that it can be split, and on its own it stays ambiguous. This keeps the existing `tests/patched/test_validation.py::test_batch_size_more_params` assertion (lone `batch_size` must raise `"At least two of ..."`) green — a naive "count the declared defaults" fix would have silently regressed it.

### Tests

Adds `TestBatchSizeFieldsValidator` covering the three previously-broken cases: only `gradient_accumulation_steps` set, only `micro_batch_size` set, and neither set (relying on both defaults). It builds its own base config rather than using `min_base_cfg`, because that fixture sets both batch fields explicitly — i.e. exactly the case that already worked.

Fixes #3836

### DIFF
--- a/src/axolotl/utils/schemas/validation.py
+++ b/src/axolotl/utils/schemas/validation.py
@@ -219,13 +219,6 @@ class TrainingValidationMixin:
     @classmethod
     def check_batch_size_fields(cls, data):
         fields = ("micro_batch_size", "gradient_accumulation_steps", "batch_size")
-        # This runs in mode="before", so field defaults have not been applied
-        # yet. Both `micro_batch_size` and `gradient_accumulation_steps` carry a
-        # documented default of 1, so a config that explicitly sets either one
-        # (relying on the other's default) is fully specified and must not be
-        # rejected. The deprecated `batch_size`, however, still has to be paired
-        # with an explicit `micro_batch_size` or `gradient_accumulation_steps`
-        # so that it can be split; on its own it stays ambiguous.
         if data.get("micro_batch_size") or data.get("gradient_accumulation_steps"):
             return data
 
--- a/src/axolotl/utils/schemas/validation.py
+++ b/src/axolotl/utils/schemas/validation.py
@@ -219,9 +219,17 @@ class TrainingValidationMixin:
     @classmethod
     def check_batch_size_fields(cls, data):
         fields = ("micro_batch_size", "gradient_accumulation_steps", "batch_size")
-        non_empty_count = sum(1 for field in fields if data.get(field))
+        # This runs in mode="before", so field defaults have not been applied
+        # yet. Both `micro_batch_size` and `gradient_accumulation_steps` carry a
+        # documented default of 1, so a config that explicitly sets either one
+        # (relying on the other's default) is fully specified and must not be
+        # rejected. The deprecated `batch_size`, however, still has to be paired
+        # with an explicit `micro_batch_size` or `gradient_accumulation_steps`
+        # so that it can be split; on its own it stays ambiguous.
+        if data.get("micro_batch_size") or data.get("gradient_accumulation_steps"):
+            return data
 
-        if non_empty_count < 2:
+        if data.get("batch_size"):
             raise ValueError(f"At least two of {', '.join(fields)} must be set")
         return data
 
--- a/tests/utils/schemas/validation/test_config_validators.py
+++ b/tests/utils/schemas/validation/test_config_validators.py
@@ -252,3 +252,46 @@ class TestGRPOBatchSizeValidator:
         }
         out = self._check(data)
         assert out["gradient_accumulation_steps"] == 8
+
+
+class TestBatchSizeFieldsValidator:
+    """``micro_batch_size`` and ``gradient_accumulation_steps`` both default to
+    1, so a config that sets only one of them (or neither) is still fully
+    specified and must not be rejected by ``check_batch_size_fields``.
+
+    ``check_batch_size_fields`` runs in ``mode="before"`` (before field defaults
+    are applied), so these guard against a regression where relying on the
+    documented defaults raised "At least two of ... must be set".
+    """
+
+    @staticmethod
+    def _base():
+        """``min_base_cfg`` sets both batch fields explicitly, which is exactly
+        the case that already worked; build a config without them instead."""
+        return DictDefault(
+            base_model="HuggingFaceTB/SmolLM2-135M",
+            learning_rate=1e-3,
+            datasets=[
+                {
+                    "path": "mhenrichsen/alpaca_2k_test",
+                    "type": "alpaca",
+                },
+            ],
+        )
+
+    def test_only_gradient_accumulation_steps_passes(self):
+        cfg = self._base() | DictDefault(gradient_accumulation_steps=4)
+        validated = validate_config(cfg)
+        assert validated.gradient_accumulation_steps == 4
+        assert validated.micro_batch_size == 1
+
+    def test_only_micro_batch_size_passes(self):
+        cfg = self._base() | DictDefault(micro_batch_size=4)
+        validated = validate_config(cfg)
+        assert validated.micro_batch_size == 4
+        assert validated.gradient_accumulation_steps == 1
+
+    def test_neither_set_relies_on_defaults(self):
+        validated = validate_config(self._base())
+        assert validated.micro_batch_size == 1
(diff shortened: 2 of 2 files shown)


NanoCode012 (commented):
I think I found 1-2 bugs with this in claude. However, looking at this at a higher level, I'm wondering whether we should deprecate `batch_size` itself. It's not something we recommend anyways.

Anai-Guo:
Force-pushed a rebuilt, rebased single commit — sorry for the churn.

The previous head had two problems that made it unreviewable, both my fault:

- the branch copy of `validation.py` had been written back with **CRLF line endings**, so GitHub rendered the change as a whole-file `+1852/-1844` diff;
- that same write had mangled the file's 8 non-ASCII em-dashes into U+FFFD replacement characters, and it was based on a stale checkout, so the diff also reverted unrelated work that had landed since (the `selective_checkpointing` / `check_hidden_states_offloading` validators, the GLM DSA context-parallel handling, the FSDP1 default pin, and the `attn_supports_packing` rename).

None of that was intended and none of it should ever have been in this PR. The branch is now a single commit on top of `main` containing only the intended change: **+10/-2** in `validation.py` and **+43/-0** in the test file, LF throughout, no character corruption, and the PR is mergeable again.

The fix logic is unchanged from the second commit of the old head — an explicit `micro_batch_size` **or** `gradient_accumulation_steps` is sufficient, while a lone deprecated `batch_size` still raises, which keeps `tests/patched/test_validation.py::test_batch_size_more_params` green. I have updated the PR description, which had still been describing the earlier "count the declared defaults" approach that regressed that test.

winglian:
I'm for deprecating batch size as a yaml schema and making it an internal schema as a helper only.

Anai-Guo:
Thanks @winglian — that makes sense as the end state, and I'm not attached to keeping `batch_size` in the public YAML schema.

Worth separating the two though: this PR doesn't add or bless the field, it just stops the current schema from silently producing a wrong effective batch size (and restores the deprecation warning path). If the deprecation lands soon, feel free to close this in favour of it — the fix only matters for the window where the field still exists in user configs.

If it'd help, I'm happy to rework this into the "internal helper only" shape instead: keep `batch_size` accepted at parse time, translate it into `micro_batch_size` × `gradient_accumulation_steps` in one place, and drop it from the exported schema. Just say which direction you want and I'll follow it.

src/axolotl/utils/schemas/validation.py
```diff
@@ -219,9 +219,17 @@ class TrainingValidationMixin:
...
     def check_batch_size_fields(cls, data):
         fields = ("micro_batch_size", "gradient_accumulation_steps", "batch_size")
-        non_empty_count = sum(1 for field in fields if data.get(field))
+        # This runs in mode="before", so field defaults have not been applied
+        # yet. Both `micro_batch_size` and `gradient_accumulation_steps` carry a
+        # documented default of 1, so a config that explicitly sets either one
+        # (relying on the other's default) is fully specified and must not be
+        # rejected. The deprecated `batch_size`, however, still has to be paired
+        # with an explicit `micro_batch_size` or `gradient_accumulation_steps`
+        # so that it can be split; on its own it stays ambiguous.
```
NanoCode012: I don't think we need this chunk.
```suggestion
```

NanoCode012 (commented):
Thanks, I agree let's fix the underlying issue then we can have follow up batch_size deprecation. One nit left.

Also, you caught the earlier bugs I mentioned, so all good there.

Anai-Guo:
Done — dropped the comment block in fb14541. Thanks for the review!

ved1beta (approved):
LGTM

Anai-Guo:
Friendly ping @winglian — this one is approved and `MERGEABLE`, and the requested change (dropping the comment block) landed in `fb14541` on Aug 4. Is there anything else needed before it can be merged?

PR #3840, opened 2026-07-15
feat: register plugin CLI commands via entry points

winglian:
A subcommand has to be discoverable before any config is read, so it cannot come from the `plugins:` list the way training hooks do. Resolve subcommands instead from a builtin registry plus the `axolotl.cli_commands` entry point group, so an installed package can contribute `axolotl <cmd>` without changes here.

Commands resolve lazily through a stand-in that carries the summary line, so rendering `axolotl --help` no longer imports an integration (and transitively torch) just to read its docstring. Commands declared on the group win, so a plugin cannot shadow a core command.

Bundled integrations stay in the registry rather than using entry points, since entry point metadata is written at install time and would otherwise need a reinstall to pick up a new command in a source checkout.

- lm-eval moves off its hand-written wrapper onto the registry

# Description

## Summary by CodeRabbit

* **New Features**
  * Added support for integrations to provide custom `axolotl` CLI subcommands through Python packaging entry points.
  * Added lazy loading for integration commands, helping the CLI remain usable even when an optional command cannot load.
  * Built-in commands take precedence over integration-provided commands with the same name.

* **Documentation**
  * Added setup guidance, Click examples, and operational notes for registering custom CLI commands.

### DIFF
--- a/docs/custom_integrations.qmd
+++ b/docs/custom_integrations.qmd
@@ -119,3 +119,54 @@ It is not necessary to place your integration in the `integrations` folder. It c
 See this repo for an example: [https://github.com/axolotl-ai-cloud/diff-transformer](https://github.com/axolotl-ai-cloud/diff-transformer)
 
 :::
+
+## Adding a CLI command
+
+An integration can contribute its own `axolotl` subcommand. Unlike plugin hooks, which are
+loaded from the `plugins:` list of a config file, a subcommand has to be discoverable before
+any config is read, so it is registered through a packaging entry point instead.
+
+Define a regular `click` command in your package:
+
+```python
+# my_package/cli.py
+import click
+
+@click.command()
+@click.argument("config", type=click.Path(exists=True, path_type=str))
+@click.option("--rank-ratio", type=float, default=0.01)
+def my_command(config: str, rank_ratio: float):
+    """one-line summary shown in `axolotl --help`"""
+    from my_package.core import run  # keep heavy imports inside the function
+
+    run(config, rank_ratio=rank_ratio)
+```
+
+and advertise it under the `axolotl.cli_commands` entry point group:
+
+```toml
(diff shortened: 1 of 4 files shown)


thad0ctor:
@winglian while you're at it, would you be interested in expanding the plugin system like this?

https://github.com/thad0ctor/axolotl/pull/44

I was thinking of something like this and maybe qmd with, or a link to a discussion, issue, etc with community plugins

winglian:
> @winglian while you're at it, would you be interested in expanding the plugin system like this?
> 
> [thad0ctor#44](https://github.com/thad0ctor/axolotl/pull/44)
> 
> I was thinking of something like this and maybe qmd with, or a link to a discussion, issue, etc with community plugins

yes, def interested in expanding the community plugins and eventually having some sort of "marketplace". The main issue that I'm trying to solve with this is how do we register plugins without a yaml file, since in my example, we want to add subcommands later that don't use yaml files, or need to show up in the cli via `axolotl --help`

winglian:
@thad0ctor Additionally, having it defined as repo, and then having the cli clone and install it is likely a security minefield and it would be better to opt to have users explicitly install packages/repos instead.

thad0ctor:
> @thad0ctor Additionally, having it defined as repo, and then having the cli clone and install it is likely a security minefield and it would be better to opt to have users explicitly install packages/repos instead.

@winglian Let's me think though how to go about improving the install more securely

Regarding the market place, maybe a .py script that has an interactive checkbox format that pulls from a json (in repo or another repo) that has whatever pointers in needs for installation (pip, etc) .  

As part of getting on the list perhaps a defined PR structure that has a basic format for plugin capabilities, and a means for security scans (user uploads proof of scans or perhaps even a CI flow against the target).  The challenge is getting enough trust in the quality of the actual plugin and its security without having to do a full up code review of it

PR #3822, opened 2026-07-10
feat: sonicmoe nvfp4 lora merge aware

NanoCode012:
# Description

Previously, the LoRA would be trained in BF16, and when merged to NVFP4 base in #3780 , it would lose its accuracy.

This PR introduces a different path, fixes #3803 , we train the LoRA fake-quantized, so that when merged, it's the same weights as in during training.

How to use: see NVFP4_LORA.qmd attached / updated example ymls. `nvfp4_merge_aware: true`

Drawback: An adapter trained this way can NOT be used loaded ontop in realtime for inference. If you want to use adapter in runtime, please use the prior training method, do not enable merge-aware.

```
┌─────────────────────────────────────────┬─────────────────────┐
  │                 Forward                 │         NLL         │
  ├─────────────────────────────────────────┼─────────────────────┤
  │ Base model                              │ 0.7164              │
  ├─────────────────────────────────────────┼─────────────────────┤
  │ Snapped Q(W_eff) = what training        │ 0.5588              │
  │ optimized = merged                      │                     │
  ├─────────────────────────────────────────┼─────────────────────┤
  │ Raw unsnapped W_eff = unmerged adapter  │ 0.7354 — worse than │
  │ at inference                            │  base               │
  └─────────────────────────────────────────┴─────────────────────┘
```

Also:
- update transformer to 5.14.1 , kernels to 0.15.x
- added support for qwen, deepseek, and other arch supported by scattermoe
- updated kernel upstream to follow latest kernel builder format

## How has this been tested?

Validated on 1xB200
- inference check
- forward pass nll check (for merge)
- cpu tests
- weights format check

## AI Usage Disclaimer

Claude

## Summary by CodeRabbit

* **New Features**
  * Added merge-aware NVFP4 MoE LoRA for SonicMoE, including bitwise-consistent baking of trained adapters into standalone NVFP4 checkpoints.
  * Added configurable merge-aware activation timing and automated adapter metadata stamping during training/saving.
  * Enhanced `merge-lora` with `--dequant` and `--override-quantizer` support, plus NVFP4 merge-aware quantization with Triton acceleration where available.
* **Documentation**
  * Added a dedicated NVFP4 MoE LoRA guide, updated navigation, support/optimization sections, and example workflows/commands.
* **Tests**
  * Added extensive merge-aware validation and correctness tests (CPU and CUDA where applicable).

PR #3841, opened 2026-07-16
fix(docker): preinstall kernels-data before the editable axolotl build

winglian:
## Description

Main's multi-arch `ci-cd` docker build fails on arm64 since the `kernels==0.15.2` bump (#3822): `build_editable` dies with `ModuleNotFoundError: No module named 'kernels_data'` ([failing job](https://github.com/axolotl-ai-cloud/axolotl/actions/runs/29530215378/job/87728473226)).

`kernels>=0.14` registers an `egg_info.writers` setuptools entry point (`kernels.lock = kernels.lockfile:write_egg_lockfile`) whose module imports `kernels_data` at load time. setuptools loads every `egg_info.writers` entry point during any editable build, and the image build runs `uv pip install --no-build-isolation -e .` — so when the kernels 0.15.2 upgrade lands in the env mid-transaction before `kernels-data` does, the editable build of axolotl itself crashes. amd64 has been surviving on cache ordering; the qemu-emulated arm64 build loses the race.

Preinstalling `kernels-data` alongside `huggingface_hub` (both in `docker/Dockerfile-uv` and `cicd/Dockerfile-uv.jinja`) makes the entry point importable regardless of transaction ordering.

## Motivation and Context

Unblocks docker image publishing on main (failed on the #3822 merge commit).

## How has this been tested?

Root-caused from the buildkit log (uv transcript shows `Installed 143 packages` → `Building axolotl` interleaved with remaining downloads). The entry-point → `from kernels_data import Metadata` chain verified against the installed kernels 0.15.2 wheel.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved container builds by ensuring required package metadata is available before the editable installation step.
  * Prevented installation failures caused by inconsistent package entry-point registration.

### DIFF
--- a/cicd/Dockerfile-uv.jinja
+++ b/cicd/Dockerfile-uv.jinja
@@ -21,7 +21,7 @@ RUN git fetch origin +$GITHUB_REF && \
 
 RUN uv pip install packaging==26.0 setuptools==78.1.1
 RUN uv pip uninstall causal_conv1d
-RUN uv pip install "huggingface_hub>=1.5.0" "httpx<1"
+RUN uv pip install "huggingface_hub>=1.5.0" "httpx<1" "kernels-data"
 RUN if [ "$AXOLOTL_EXTRAS" != "" ] ; then \
         uv pip install --no-build-isolation -e .[deepspeed,optimizers,ray,$AXOLOTL_EXTRAS] $AXOLOTL_ARGS; \
     else \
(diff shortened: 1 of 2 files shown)


winglian:
Validated the mechanism and the fix locally — it's deterministic, not a flake (the rerun of the main build failed identically):

**Repro** (kernels installed without kernels_data, as the arm64 image build transaction transiently is):
```
uv venv kdtest && uv pip install setuptools wheel packaging huggingface_hub pyyaml tomlkit httpx
uv pip install --no-deps kernels==0.15.2
uv pip install --no-build-isolation --no-deps -e .
# × Failed to build axolotl
# ╰─▶ Call to setuptools.build_meta.prepare_metadata_for_build_editable
#         from kernels_data import Metadata
#     ModuleNotFoundError: No module named 'kernels_data'
```

**With the fix** (kernels-data preinstalled, everything else identical):
```
uv pip install kernels-data
uv pip install --no-build-isolation --no-deps -e .
# + axolotl==0.17.0.dev0 (from file:///workspace/wing/axolotl)
```

PR #3855, opened 2026-07-18
Read dataclass field help from the "help" metadata key

vineethsaivs:
# Description

`add_options_from_dataclass` built every Click option with `help=field.metadata.get("description")`, but the CLI dataclasses document their fields with `metadata={"help": ...}`:

```python
# src/axolotl/cli/args.py
host: Optional[str] = field(
    default=None,  # nosec B104
    metadata={"help": "Host address to run the server on."},
)
```

`"help"` is the `dataclasses` convention (and the key `HfArgumentParser` reads), so the `"description"` lookup returned `None` for every field and all option descriptions were silently dropped from `--help`.

`"description"` is the *Pydantic* key, used correctly a few lines further down in the same module by `_get_field_description` (`field.description` / `json_schema_extra["description"]`) for the `add_options_from_config` path. It looks like it was carried over into the dataclass path by mistake. No dataclass field in the repo uses a `"description"` metadata key, so reading `"help"` is a complete fix rather than a partial one.

Both call sites were affected (the `bool` branch and the general branch), so this touches every command using the decorator: `preprocess`, `train`, `evaluate`, `vllm-serve`, and `quantize`.

## Motivation and Context

The most visible case is `axolotl vllm-serve --help`, where all nine documented options rendered with no description at all.

Before:

```
Options:
  --tensor-parallel-size INTEGER
  --data-parallel-size INTEGER
  --host TEXT
  --port INTEGER
  --gpu-memory-utilization FLOAT
  ...
```

After:

```
Options:
  --tensor-parallel-size INTEGER  Number of tensor parallel workers to use.
  --data-parallel-size INTEGER    Number of data parallel workers to use for
                                  vLLM serving. This controls how many model
                                  replicas are used for parallel inference.
  --host TEXT                     Host address to run the server on.
  --port INTEGER                  Port to run the server on.
  --gpu-memory-utilization FLOAT  Ratio (between 0 and 1) of GPU memory to
                                  reserve for the model weights, activations,
                                  ...
```

The deprecation notice on `axolotl preprocess --iterable` was invisible for the same reason.

## How has this been tested?

Added `tests/cli/test_dataclass_options.py`, mirroring the existing `tests/cli/test_nested_options.py` style (Click `CliRunner` against a small local dataclass, covering both the `bool` branch and the general branch, plus a field with no help metadata so bare options keep working).

The new test fails on `main` and passes with this change:

```
>       assert "--tensor-parallel-size INTEGER Number of workers." in output
E       AssertionError: assert '--tensor-parallel-size INTEGER Number of workers.' in
E         'Usage: cmd [OPTIONS] Options: --verbose / --no-verbose
E          --tensor-parallel-size INTEGER --undocumented TEXT --help ...'
```

I also confirmed the real `VllmServeCliArgs` help output before and after (quoted above). `ruff check` and `ruff format --check` are clean on both files.

The generated `config_options.py` is unaffected: `generate_config_options.py` builds from the Pydantic `AxolotlInputConfig` through its own `_get_field_description`, not through `add_options_from_dataclass`, so the `check-cli-config-options` hook stays green.

## Types of changes

## Summary by CodeRabbit

* **Bug Fixes**
  * CLI option help text now correctly uses each field’s help metadata.
  * Boolean and non-boolean options display their configured descriptions consistently.
  * Options without help text remain available without displaying unintended descriptions.

* **Tests**
  * Added coverage verifying CLI help output for documented and undocumented options.


tests/cli/test_dataclass_options.py
```diff
@@ -0,0 +1,61 @@
+"""Tests for dataclass-based CLI option handling."""
```
ved1beta: move the test to test_nested_options.py

ved1beta (approved):
LGTM

tests/cli/test_dataclass_options.py
vineethsaivs: Done, moved both tests into `test_nested_options.py` and dropped the separate file

PR #3858, opened 2026-07-20
fix: handle trust_remote_code when packing properly

NanoCode012:
# Description

## Social Handles (Optional)

Reported by `Gregory from u9up`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of remote-code permissions when the setting is not explicitly configured.
  * Fixed sensitive path redaction for values nested within lists, improving telemetry privacy.


winglian:
should we add a section for quack under `project.optional-dependencies` in `pyprojecy.toml`?

PR #3863, opened 2026-07-21
fix: multipack drop_last total_slots stale reference

### LINKED ISSUE #3848: MultipackBatchSampler.generate_batches() stale-reference bug corrupts total_token_slots and can crash with IndexError
### Expected Behavior

When `drop_last=True` and sample packing produces an incomplete final batch, `MultipackBatchSampler.generate_batches()` should drop that batch *and* correctly reduce `total_slots` by exactly the dropped batch's capacity, so `total_token_slots` (and therefore `sampler.efficiency()` / `cfg.sample_packing_eff_est`) stays accurate.

### Current behaviour

`generate_batches()` (`src/axolotl/utils/samplers/multipack.py`, lines 362-372) has a stale-reference bug:

```python
# Group bins into batches (each batch contains batch_size bins)
batches = [
    bins[i : i + self.batch_size] for i in range(0, len(bins), self.batch_size)
]

# Drop last batch if requested and it's incomplete
if self.drop_last and len(batches[-1]) < self.batch_size:
    batches = batches[:-1]
    # Adjust total_slots if we dropped a batch
    if not self.sequential:
        total_slots -= (self.batch_size - len(batches[-1])) * self.batch_max_len
```

Line 368 checks `batches[-1]`, the last chunk, confirmed incomplete. Line 369 reassigns `batches = batches[:-1]`, which drops that chunk. Line 372 then reads `batches[-1]` again, but `batches` has already been resliced, so this now refers to the *previous* chunk instead of the one that was just dropped. Since `bins[i:i+batch_size]` chunking guarantees every chunk except the true last one is exactly `self.batch_size` long, this second `batches[-1]` is a full batch in the common case, so `(self.batch_size - len(batches[-1]))` evaluates to `0` and the "adjustment" on line 372 is a no-op. `total_slots` is left at its full, pre-drop value even though bins were actually dropped.

There's also a hard crash in the edge case where packing produces exactly one batch and it's the incomplete one: `batches = batches[:-1]` leaves `batches == []`, and the second `batches[-1]` on line 372 raises `IndexError: list index out of range`.

`generate_batches()` is the sampler actually wired into training (`src/axolotl/core/trainers/base.py`, which uses `drop_last=True`), and its `efficiency()` output feeds `cfg.sample_packing_eff_est`, which `calculate_total_num_steps` (`src/axolotl/utils/trainer.py`) uses to size `total_num_steps` for the LR scheduler. So this silently inflates `total_token_slots` whenever the packed bins don't divide evenly by `batch_size`, which is the common case, corrupting the packing-efficiency estimate used for step counting.

### Steps to reproduce

Isolated repro of just the buggy logic (matches the real code exactly, no GPU/model needed):

```python
def simulate_generate_batches(bins, batch_size, batch_max_len, drop_last, sequential, total_used, total_slots):
    batches = [bins[i : i + batch_size] for i in range(0, len(bins), batch_size)]
    if drop_last and len(batches[-1]) < batch_size:
        batches = batches[:-1]
        if not sequential:
            total_slots -= (batch_size - len(batches[-1])) * batch_max_len
    return batches, total_used, total_slots

# 13 bins, batch_size=5 -> chunks [5, 5, 3], last (3) gets dropped
n_bins, batch_size, batch_max_len = 13, 5, 10
bins = [[i] for i in range(n_bins)]
batches, _, total_slots = simulate_generate_batches(
    bins, batch_size, batch_max_len, drop_last=True, sequential=False,
    total_used=n_bins * batch_max_len, total_slots=n_bins * batch_max_len,
)
n_bins_kept = sum(len(b) for b in batches)
print("bins kept:", n_bins_kept, "  total_slots reported:", total_slots,
      "  total_slots that should be reported:", n_bins_kept * batch_max_len)
```

Output:
```
bins kept: 10   total_slots reported: 130   total_slots that should be reported: 100
```

`total_slots` stayed at 130 (the full 13-bin value) instead of reflecting the 10 bins actually kept, confirming the adjustment never happens.

Crash case (3 bins, `batch_size=8`, one incomplete chunk):
```python
simulate_generate_batches([[0], [1], [2]], batch_size=8, batch_max_len=10,
                           drop_last=True, sequential=False, total_used=30, total_slots=30)
```
```
IndexError: list index out of range
```

I ran both of these against the isolated logic above (which matches `src/axolotl/utils/samplers/multipack.py` lines 362-372 verbatim) and also confirmed the source lines are current by fetching them directly from `main`.

### Config yaml

Not applicable, this reproduces at the sampler level directly, independent of any specific training config; it triggers whenever `sample_packing: true` (which sets `drop_last=True` in the trainer) and the packed bins for an epoch don't divide evenly by the per-step batch size, which is the normal case.

### Possible solution

Capture the dropped batch's length before reslicing `batches`, instead of reading `batches[-1]` a second time after the slice:

```python
if self.drop_last and len(batches[-1]) < self.batch_size:
    dropped_len = len(batches[-1])
    batches = batches[:-1]
    if not self.sequential:
        total_slots -= dropped_len * self.batch_max_len
```

This also naturally avoids the `IndexError`, since `dropped_len` is captured while `batches` still contains the chunk being dropped, no second index into the (possibly now-empty) resliced list needed.

### Related to #3152

This looks like the root cause of #3152 ("total_num_steps calculation is incorrect with sample_packing_eff_est"), which is still open. The reporter there saw `sample_packing_eff_est: 1.0`, and a maintainer commented "I'm not sure reaching 1.0 is possible" — that's consistent with this bug's failure mode, since a corrupted `total_token_slots` that never gets reduced by dropped bins pushes `efficiency() = total_tokens_used / total_token_slots` toward 1.0 more often than it should. I don't think #3152 is a duplicate of this since it reports the symptom without the code-level cause, I'm filing this separately since it's a concrete, independently reproducible bug in specific lines of code, but wanted to flag the connection.

- [x] Windows (where I reproduced the isolated logic)
- Note: this is a pure Python/CPU logic bug in the sampler, not OS or hardware dependent

### Python Version

3.12.10

### axolotl branch-commit

main (verified against the current `main` branch source of `src/axolotl/utils/samplers/multipack.py` at the time of filing)

### Acknowledgements

Solaris-star:
# Description

Fixes a stale-reference bug in `MultipackBatchSampler.generate_batches()` when `drop_last=True` drops an incomplete final batch.

Previously the code did:

```python
if self.drop_last and len(batches[-1]) < self.batch_size:
    batches = batches[:-1]
    if not self.sequential:
        total_slots -= (self.batch_size - len(batches[-1])) * self.batch_max_len
```

After `batches = batches[:-1]`, `batches[-1]` no longer refers to the dropped incomplete batch — it refers to the previous full batch. That made `(batch_size - len(batches[-1]))` evaluate to `0` in the common case, so `total_token_slots` stayed inflated. In the edge case where packing produces exactly one incomplete batch, the re-slice left `batches == []` and the next `batches[-1]` raised `IndexError`.

This PR captures `dropped_bins = len(batches[-1])` before the re-slice, subtracts `dropped_bins * batch_max_len`, and guards empty `batches`.

Fixes #3848

## Motivation and Context

`total_token_slots` feeds packing efficiency estimates (`sampler.efficiency()` / `cfg.sample_packing_eff_est`). Leaving it inflated after `drop_last` silently misreports packing efficiency. The empty-batch path can also crash training startup.

## How has this been tested?

- Added unit tests in `tests/utils/samplers/test_multipack_drop_last_stats.py`
- Ran locally:
  ```bash
  PYTHONPATH=src python3 -m pytest tests/utils/samplers/test_multipack_drop_last_stats.py -q --noconftest
  # 3 passed
  ```

## AI Usage Disclaimer

Yes — assisted with drafting the patch and unit tests (Hermes/agent). Logic and review were checked against the issue repro.

N/A

## Types of changes

## Social Handles (Optional)

N/A

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected token-slot accounting when incomplete final batches are dropped.
  * Prevented errors when dropping the only incomplete batch.
  * Ensured reported capacity reflects the remaining batches accurately.

* **Tests**
  * Added coverage for dropped-batch accounting and empty-batch scenarios.

### DIFF
--- a/src/axolotl/utils/samplers/multipack.py
+++ b/src/axolotl/utils/samplers/multipack.py
@@ -364,12 +364,17 @@ class MultipackBatchSampler(BatchSampler):
             bins[i : i + self.batch_size] for i in range(0, len(bins), self.batch_size)
         ]
 
-        # Drop last batch if requested and it's incomplete
-        if self.drop_last and len(batches[-1]) < self.batch_size:
+        # Drop last batch if requested and it's incomplete.
+        # Capture the dropped batch length *before* re-slicing: after
+        # `batches = batches[:-1]`, `batches[-1]` refers to the previous (full)
+        # batch, so the old adjustment became a no-op and left total_token_slots
+        # inflated. Also guard empty batches to avoid IndexError when packing
+        # yields a single incomplete batch that is then dropped (#3848).
+        if self.drop_last and batches and len(batches[-1]) < self.batch_size:
+            dropped_bins = len(batches[-1])
             batches = batches[:-1]
-            # Adjust total_slots if we dropped a batch
             if not self.sequential:
-                total_slots -= (self.batch_size - len(batches[-1])) * self.batch_max_len
+                total_slots -= dropped_bins * self.batch_max_len
 
         # Update statistics if requested
         if set_stats:
--- a/tests/utils/samplers/test_multipack_drop_last_stats.py
+++ b/tests/utils/samplers/test_multipack_drop_last_stats.py
@@ -1,0 +1,96 @@
+"""Unit tests for MultipackBatchSampler drop_last total_slots accounting."""
+
+from __future__ import annotations
+
+import numpy as np
+from torch.utils.data import SequentialSampler
+
+from axolotl.utils.samplers.multipack import MultipackBatchSampler
+
+
+class _ListDataset:
+    def __init__(self, n: int):
+        self.n = n
+
+    def __len__(self) -> int:
+        return self.n
+
+
+def _make_sampler(
+    lengths: list[int],
+    *,
+    batch_size: int,
+    batch_max_len: int,
+    drop_last: bool,
+    sequential: bool = False,
+) -> MultipackBatchSampler:
+    dataset = _ListDataset(len(lengths))
+    return MultipackBatchSampler(
+        sampler=SequentialSampler(dataset),
+        lengths=np.array(lengths, dtype=np.int32),
+        batch_size=batch_size,
+        batch_max_len=batch_max_len,
+        bin_size=max(1, len(lengths)),
+        group_size=max(1, len(lengths)),
+        sequential=sequential,
+        drop_last=drop_last,
+        num_count_samples=1,
+        num_processes=1,
+        safe_mode=True,
+    )
+
+
+def test_drop_last_reduces_total_token_slots_by_dropped_bins():
+    # lengths=4, capacity=8 => two sequences per bin via first-fit.
+    # 5 sequences => bins [[0,1],[2,3],[4]] (3 bins).
+    # batch_size=2 => batches [full, incomplete]; incomplete last batch dropped.
+    lengths = [4] * 5
+    batch_size = 2
+    batch_max_len = 8
+
+    sampler = _make_sampler(
+        lengths,
+        batch_size=batch_size,
+        batch_max_len=batch_max_len,
+        drop_last=True,
+        sequential=False,
+    )
+    batches = sampler.generate_batches(set_stats=True)
+
+    assert len(batches) == 1
+    assert len(batches[0]) == batch_size
+
+    # Kept bins: 2; dropped bins: 1. total_token_slots must shrink by dropped bins.
+    assert sampler.total_token_slots == 2 * batch_max_len
+
+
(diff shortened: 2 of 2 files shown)


NanoCode012:
Duplicate #3849

PR #3864, opened 2026-07-21
fix: drop samples with no trainable tokens (all labels -100)

vineethsaivs:
### Summary

`process_datasets_for_packing` runs a `drop_no_trainable_tokens` filter that is meant to drop training samples whose labels are entirely `-100` (i.e. zero trainable tokens). The filter is currently a no-op: it never drops anything.

### Root cause

The filter compares the raw label sequence to `-100` and wraps it in `np.any`:

```python
if isinstance(labels[0], int):
    return np.any(labels != -100)
results = [np.any(row_labels != -100) for row_labels in labels]
```

When `datasets.Dataset.filter(..., batched=True)` calls this, `sample["labels"]` is a Python `list` (a list of lists in batched mode), not a NumPy array. `some_list != -100` is a plain Python object comparison that always evaluates to the scalar `True`, so `np.any(True)` is always `True`. Every sample is kept regardless of its labels, so fully-masked samples (all `-100`) are never removed. The single-example branch has the same problem.

### Fix

Coerce the labels to a NumPy array before comparing, so `!= -100` is element-wise as intended:

```python
if isinstance(labels[0], int):
    return bool(np.any(np.asarray(labels) != -100))
results = [
    bool(np.any(np.asarray(row_labels) != -100)) for row_labels in labels
]
```

### Verification

Reproduced against a real `datasets.Dataset` with a batched filter: with a batch `[[-100, -100, -100], [-100, 7, -100]]` the current code keeps both rows; after the fix it keeps only the second (the one with a trainable token). Fully-masked samples are now dropped, samples with any trainable token are kept.

Added `TestDropNoTrainableTokens` in `tests/test_data.py`, driving `process_datasets_for_packing` end to end and asserting that fully-masked samples are dropped, all-masked batches drop to empty, and samples with trainable tokens are preserved. These tests fail on the current code and pass with the fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of masked training examples by consistently identifying samples with no trainable tokens.
  * Ensured fully masked samples are excluded, while samples containing trainable tokens are retained.

* **Tests**
  * Added coverage for single samples, fully masked batches, and mixed trainable data scenarios.

### DIFF
--- a/src/axolotl/utils/data/sft.py
+++ b/src/axolotl/utils/data/sft.py
@@ -351,7 +351,7 @@ def _load_raw_datasets(
         if (split == "train" and cfg.sample_packing) or (
             split == "test" and cfg.eval_sample_packing
         ):
-            dataset, _ = process_datasets_for_packing(cfg, dataset, None, split=split)
+            dataset, _ = process_datasets_for_packing(cfg, dataset, None)
 
         # Deduplicate before saving so the saved dataset is already de-duplicated
         if cfg.dataset_exact_deduplication:
--- a/src/axolotl/utils/data/sft.py
+++ b/src/axolotl/utils/data/sft.py
@@ -351,7 +351,7 @@ def _load_raw_datasets(
         if (split == "train" and cfg.sample_packing) or (
             split == "test" and cfg.eval_sample_packing
         ):
-            dataset, _ = process_datasets_for_packing(cfg, dataset, None)
+            dataset, _ = process_datasets_for_packing(cfg, dataset, None, split=split)
 
         # Deduplicate before saving so the saved dataset is already de-duplicated
         if cfg.dataset_exact_deduplication:
--- a/src/axolotl/utils/trainer.py
+++ b/src/axolotl/utils/trainer.py
@@ -271,11 +271,13 @@ def process_datasets_for_packing(cfg, train_dataset, eval_dataset):
         # If it's a list, we assume we're dealing with a batch
         if isinstance(labels[0], int):
             # Single example: return a single bool
-            return np.any(labels != -100)
+            return bool(np.any(np.asarray(labels) != -100))
 
         # Batched: 'labels' is a list of lists
         # Return a list of booleans, one per sub-list
-        results = [np.any(row_labels != -100) for row_labels in labels]
+        results = [
+            bool(np.any(np.asarray(row_labels) != -100)) for row_labels in labels
+        ]
         return results
 
     try:
--- a/src/axolotl/utils/trainer.py
+++ b/src/axolotl/utils/trainer.py
@@ -241,7 +241,7 @@ def filter_sequences_by_length(
     return results
 
 
-def process_datasets_for_packing(cfg, train_dataset, eval_dataset):
+def process_datasets_for_packing(cfg, train_dataset, eval_dataset, split="train"):
     drop_attn_mask = cfg.model_config_type in ["mamba", "gemma3"]
     if drop_attn_mask:
         LOG.info("dropping attention_mask column")
@@ -307,9 +307,9 @@ def process_datasets_for_packing(cfg, train_dataset, eval_dataset):
         dropped = prior_len - len(train_dataset)
         if dropped:
             LOG.warning(
-                f"Dropped {dropped} samples with no trainable tokens from train dataset"
+                f"Dropped {dropped} samples with no trainable tokens from {split} dataset"
             )
-        raise_if_empty(train_dataset, "train")
+        raise_if_empty(train_dataset, split)
 
     if eval_dataset:
         try:
--- a/src/axolotl/utils/trainer.py
+++ b/src/axolotl/utils/trainer.py
@@ -241,7 +241,7 @@ def filter_sequences_by_length(
     return results
 
 
-def process_datasets_for_packing(cfg, train_dataset, eval_dataset, split="train"):
+def process_datasets_for_packing(cfg, train_dataset, eval_dataset):
     drop_attn_mask = cfg.model_config_type in ["mamba", "gemma3"]
     if drop_attn_mask:
         LOG.info("dropping attention_mask column")
@@ -307,9 +307,9 @@ def process_datasets_for_packing(cfg, train_dataset, eval_dataset, split="train"
         dropped = prior_len - len(train_dataset)
         if dropped:
             LOG.warning(
-                f"Dropped {dropped} samples with no trainable tokens from {split} dataset"
+                f"Dropped {dropped} samples with no trainable tokens from train dataset"
             )
-        raise_if_empty(train_dataset, split)
+        raise_if_empty(train_dataset, "train")
 
     if eval_dataset:
         try:
--- a/src/axolotl/utils/trainer.py
+++ b/src/axolotl/utils/trainer.py
@@ -280,8 +280,8 @@ def process_datasets_for_packing(cfg, train_dataset, eval_dataset):
             raise ValueError(
                 f"The {split} dataset has no samples left after dropping samples with "
                 "no trainable tokens. Every sample had all of its labels masked to "
-                "-100, so there is nothing to train on. Check `train_on_inputs` and "
-                "your prompt strategy / chat template."
+                "-100, so there is nothing to train on. Check `train_on_inputs`, "
+                "`roles_to_train`, and your prompt strategy / chat template."
             )
 
     try:
--- a/src/axolotl/utils/trainer.py
+++ b/src/axolotl/utils/trainer.py
@@ -303,7 +303,7 @@ def process_datasets_for_packing(cfg, train_dataset, eval_dataset):
         **filter_map_kwargs,
         **drop_long_kwargs,
     )
-    if prior_len is not None:
+    if prior_len:
         dropped = prior_len - len(train_dataset)
         if dropped:
             LOG.warning(
@@ -322,7 +322,7 @@ def process_datasets_for_packing(cfg, train_dataset, eval_dataset):
             **filter_map_kwargs,
             **drop_long_kwargs,
         )
-        if prior_len is not None:
+        if prior_len:
             dropped = prior_len - len(eval_dataset)
             if dropped:
                 LOG.warning(
--- a/src/axolotl/utils/trainer.py
+++ b/src/axolotl/utils/trainer.py
@@ -275,11 +275,11 @@ def process_datasets_for_packing(cfg, train_dataset, eval_dataset):
         results = [any(v != -100 for v in row_labels) for row_labels in labels]
         return results
 
-    def raise_if_empty(dataset, split):
+    def raise_if_empty(dataset):
         if len(dataset) == 0:
             raise ValueError(
-                f"The {split} dataset has no samples left after dropping samples with "
-                "no trainable tokens. Every sample had all of its labels masked to "
+                "The dataset has no samples left after dropping samples with no "
+                "trainable tokens. Every sample had all of its labels masked to "
                 "-100, so there is nothing to train on. Check `train_on_inputs`, "
                 "`roles_to_train`, and your prompt strategy / chat template."
             )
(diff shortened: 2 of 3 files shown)


vineethsaivs:
On the bot's suggestion to drop the comments around `trainer.py` lines 276-277: those comments are pre-existing on main, not added here. This PR only changes the two expressions (`np.any(labels != -100)` to `bool(np.any(np.asarray(labels) != -100))`), so removing them would be an unrelated change to code the fix does not otherwise touch. Happy to strip them in a follow-up if you would rather they go.

The one comment this PR does add is the single line in `test_fully_masked_sample_is_dropped`, which explains why an all -100 sample is the case under test.

NanoCode012:
Interesting catch. What's the perf cost to this? Any other alternative?

I guess folks usually don't have fully -100 unless user configured incorrectly.

tests/test_data.py
```diff
@@ -162,5 +166,31 @@ def test_excess_length_strategy(self):
...
 
+class TestDropNoTrainableTokens(unittest.TestCase):
+    """
+    test that process_datasets_for_packing drops samples with no trainable tokens
+    """
+
+    def _process(self, labels):
+        cfg = DictDefault({"dataset_num_proc": 1, "is_preprocess": True})
+        dataset = Dataset.from_dict({"labels": labels})
+        train_dataset, _ = process_datasets_for_packing(cfg, dataset, None)
```
NanoCode012: We should also add `dataset` for the eval_dataset, so we can also test that path please.

src/axolotl/utils/trainer.py
```diff
@@ -271,11 +271,13 @@ def drop_no_trainable_tokens(sample):
...
             # Single example: return a single bool
-            return np.any(labels != -100)
+            return bool(np.any(np.asarray(labels) != -100))
 
         # Batched: 'labels' is a list of lists
         # Return a list of booleans, one per sub-list
-        results = [np.any(row_labels != -100) for row_labels in labels]
+        results = [
+            bool(np.any(np.asarray(row_labels) != -100)) for row_labels in labels
+        ]
```
NanoCode012: We don't need to convert to np array
```
any(v != -100 for v in row_labels)
```

works.

Small test: `0.0815s for np.asarray vs 0.0040s for above on 2000×2048 rows`

src/axolotl/utils/trainer.py
NanoCode012, line 311: For both train_dataset and eval, we should check if `len(train_dataset) == 0`, raise. This is done for eval in `data/sft.py:105-111`, but we should consolidate it here.

NanoCode012 (commented):
> re: coderrabbit comments

You can apply the L277 and test L181 fix from CR. Test L170 should be kept.

vineethsaivs:
Thanks for the review. Numbers first, since the perf question decides the shape of the fix.

**Cost of the predicate itself**, 2000 rows x 2048 tokens, best of 5:

| labels in the row | `np.asarray` (what this PR had) | `any(v != -100 ...)` |
|---|---|---|
| all masked | 89 ms | 80 ms |
| prompt masked, response trainable | 89 ms | 41 ms |
| nothing masked | 91 ms | 0.5 ms |

Your version wins because it short-circuits on the first trainable token, and the more normal the data the earlier that happens. `np.asarray` always materialises the whole row first. Switched to it in 4f59f2e.

**Cost in context**, which is the more useful number. Driving a real `Dataset.filter(..., batched=True)` over 10k samples of 2048 tokens, prompt masked and response trainable:

| | wall clock |
|---|---|
| `filter` with a predicate that just returns `True` | 3.54 s |
| current main (the no-op comparison) | 3.57 s |
| `np.asarray` | 4.11 s |
| `any(v != -100 ...)` | 3.75 s |

So the fix costs about 0.2 s per 10k samples on top of a filter pass that already runs and already costs 3.5 s. It happens once at preprocess, under `dataset_num_proc`, and it is linear in tokens.

**Other alternatives.** The predicate is not really the cost here, the Arrow to Python conversion inside `filter` is. Computing the keep-mask in pyarrow instead, never building Python lists, is 0.04 s for the same 10k samples, roughly 90x faster than the whole filter pass:

```python
col = dataset.data.column("labels").combine_chunks()
agg = (
    pa.table({"i": pc.list_parent_indices(col), "t": pc.not_equal(col.values, -100)})
    .group_by("i")
    .aggregate([("t", "any")])
)
```

I kept it out of this PR because it is a different change with its own edges: it bypasses `filter`, so `dataset_num_proc`, caching and the progress description all need handling; it does not apply to `IterableDataset`; and the group-by silently drops rows whose label list is empty, so the mask has to be re-indexed against `range(len(dataset))` or it misaligns by one row onward. The current code deliberately keeps empty rows (`if not labels: return True`). Happy to do it as a follow-up if you want that preprocessing time back.

On folks not usually having fully masked samples unless something is configured wrong: agreed, and I think that argues for the raise you asked for rather than against the fix. Today a misconfigured run trains on samples with zero trainable tokens and just shows a flat loss. After this they get dropped, and if that empties a split it now says so instead of handing an empty dataset to the trainer.

Also applied: the trainer L277 and test L181 comment removals from CodeRabbit, keeping the test class docstring as you said. `import numpy as np` is gone from `trainer.py` too, those two lines were its only users.

src/axolotl/utils/trainer.py
vineethsaivs: Done in 4f59f2e, both branches. Reproduced your result, and it holds up even better on data that is not fully masked, since `any` stops at the first trainable token. 2000x2048 rows, best of 5: `np.asarray` is 89 ms regardless, `any(...)` is 80 ms all-masked, 41 ms prompt-masked, 0.5 ms unmasked.

`import numpy as np` came out of the module with it, those two lines were its only users.

src/axolotl/utils/trainer.py
vineethsaivs: Done in 4f59f2e. `raise_if_empty` is called for both splits right after each filter, and only when the dataset is sized, since an `IterableDataset` has no `len` and the existing code already tracks that as `prior_len is None`. I also changed the two guards from `if prior_len:` to `if prior_len is not None:`, otherwise a split that arrived already empty would skip the check.

One thing I did not do, so you can tell me if you want it: I left the `data/sft.py:105-111` check where it is rather than folding it in, because it is not quite the same condition. It fires when `calculate_total_num_steps` is 0 under `sample_packing`, which a non-empty eval split can still hit. `len == 0` implies that, but not the reverse. So this new check catches the empty case earlier and with a message that names the cause, and that one still covers "non-empty but too small to pack". Happy to move the packing check here too and drop it from `sft.py` if you would rather have one place.

tests/test_data.py
vineethsaivs: Done in 4f59f2e. `_process` now passes a second dataset as `eval_dataset` and returns both, and every assertion covers both splits.

That path is worth having for a second reason: the eval filter is called without `batched=True`, so it exercises the single-example branch of `drop_no_trainable_tokens` while the train filter exercises the batched one. Confirmed the eval split was equally broken before the fix, nothing was dropped there either. Also added `test_all_eval_samples_masked_raises` for the eval side of the new guard.

vineethsaivs:
Thanks for pushing 0872725, 925d6a3 and d4596ea. I ran all three rather than reading them, and they are right, including the one that corrects me.

**0872725 is a real bug in my version.** I had changed the guards to `prior_len is not None` so a dataset that filtered down to zero would still be caught. That also caught a dataset that arrived empty, and then told the user something false:

```
4f59f2e  process_datasets_for_packing(cfg, Dataset.from_dict({"labels": []}), None)
  -> ValueError: The train dataset has no samples left after dropping samples with
     no trainable tokens. Every sample had all of its labels masked to -100 ...

d4596ea  same call
  -> ok, len=0
```

Nothing was dropped there, so the diagnosis was wrong. Going back to `if prior_len:` is the correct guard: it still fires when a non-empty dataset filters to zero, which is the case worth reporting. Your `test_already_empty_dataset_does_not_raise` pins exactly that.

The other two check out too, driving the shipped function:

```
split="test", all masked   -> ValueError: The test dataset has no samples left ...
split default, all masked  -> ValueError: The train dataset has no samples left ...
already-empty train        -> ok, len=0
already-empty eval         -> ok, train kept
```

## On the two red e2e checks

Neither is reachable from this diff.

`docker-e2e-kernel-tests` is not an independent failure. It was cancelled at 13:09:26, after `docker-e2e-tests` failed at 13:02:07.

The one real failure is `tests/e2e/integrations/test_kd.py::TestKnowledgeDistillation::test_llama_lora_kd[True]`, `Train Loss (1.4765625) is too high` against a 1.2 threshold. Two independent reasons it is not this PR:

1. `kd_min_cfg` sets `skip_prepare_dataset: True`. `process_datasets_for_packing` has exactly one call site in the repo, `utils/data/sft.py:354`, and it sits inside `if not cfg.skip_prepare_dataset and not streaming:`. The KD configs never reach the filter at all.
2. `test_llama_kd` and `test_llama_lora_kd[False]` passed in the same run, on the same fixture and the same dataset. Only `load_in_8bit=True` failed. A dataset-level change cannot be selective by quantization mode.

Worth flagging honestly while we are looking at e2e, because it does change the blast radius: on `main` this filter never fires. `np.any(labels != -100)` compares a Python list against an int, which is `True`, so `np.any(True)` is `True` for every sample, batched or not:

```
main today, all four fixtures:  nothing dropped, including [[-100, -100], [-100, -100, -100]]
this PR:                        drops as intended, raises when a non-empty split filters to zero
```

So this is the first time the filter actually does anything, and any e2e config that uses packing without `skip_prepare_dataset` is where a genuine regression would show. Nothing else in that run went red.

## coderabbit's comment

Two of the three landed in 4f59f2e: the `# Batched: 'labels' is a list of lists` pair in `trainer.py` and the `# a sample whose labels are all -100 ...` line in the test are both gone. I left the `TestDropNoTrainableTokens` class docstring, since `AGENTS.md`'s rule is about comments and the only other class in `tests/test_data.py`, `TestEncodePretraining`, carries the same one-line docstring. Happy to drop it if you would rather it go.

src/axolotl/utils/data/sft.py
```diff
@@ -351,7 +351,7 @@ def _load_raw_datasets(
         if (split == "train" and cfg.sample_packing) or (
             split == "test" and cfg.eval_sample_packing
         ):
-            dataset, _ = process_datasets_for_packing(cfg, dataset, None)
+            dataset, _ = process_datasets_for_packing(cfg, dataset, None, split=split)
```
winglian: this is unnecessary, see above in line 335, we essentially iterate over all the dataset configurations and grab whichever split is requested and "hoist" it out so there should no longer be a split

NanoCode012: Split here was just to be used for logging purposes. I reverted this api change and fixed the incorrect messaging instead then.

NanoCode012 (approved):

src/axolotl/utils/data/sft.py
vineethsaivs: Fair call on reverting the signature, and thanks for the approval. Your version is better than mine: `split` was only ever reaching `raise_if_empty` to decorate a string, so threading a new parameter through a public function to carry it was the wrong trade. Making the message not name a split at all gets the same result without touching the API.

Ran your final head to confirm nothing regressed in the revert, all six cases behave:

```
fully masked row dropped              train=[[-100, 7, -100]]
all rows masked                       ValueError: The dataset has no samples left ...
trainable rows kept                   train=[[1, 2, 3], [-100, 4, -100]]
eval all masked                       ValueError (and the warning still says "from eval dataset")
already-empty train                   train=[]        no false raise
already-empty eval                    train=[[1, 2]]  no false raise
```

The asymmetry in the two log lines reads deliberate to me and I would leave it: the eval branch can safely say "from eval dataset" because it only ever handles eval, while the other call site is reached for both train and test, so a generic message is the honest one there.

PR #3867, opened 2026-07-22
fix(sequence_parallel): reject batch_ring with pre-packed multi-doc sequences

### LINKED ISSUE #3608: Ring Attention w/ document packing produces different results

### Expected Behavior

When using document packing, ring attention should produce similar or bitwise identical outputs to varlen_llama3 or padded batches.

### Current behaviour

When using specifically document packing with ring attention, the loss on a given step is much higher and different than would otherwise be expected. Ring attention on a single non-packed document works. Document packing with varlen_llama3 attention works.

**Multi-doc packed data** (`data/packed.pt`, pack 0 = 5 docs with
position_id resets at each boundary):

| config                                | step-0 loss | ratio to cp=1      |
|---------------------------------------|-------------|--------------------|
| `cp=1` (baseline)                     | `3.171875`  | 1.000×             |
| `cp=2, ring_attn_func=batch_ring`     | `5.843750`  | **1.843×** (bug)   |
| `cp=2, ring_attn_func=varlen_llama3`  | `3.156250`  | 0.995× (fix)       |

**Single-doc control** (`data/packed_single.pt`, one 1024-token doc,
`position_ids` 0..1023 contiguous, zero resets). Every step is
bitwise-identical across all three configs:

| config                                | loss (all steps) |
|---------------------------------------|------------------|
| `cp=1` (baseline)                     | `3.031250`       |
| `cp=2, ring_attn_func=batch_ring`     | `3.031250`       |
| `cp=2, ring_attn_func=varlen_llama3`  | `3.031250`       |

### Steps to reproduce

I don't have the skill set to debug exactly what is wrong with it, but Claude was able to produce a minimal reproduction of the issue which can be found here: https://github.com/Ueaj-Kerman/axolotl-cp-repro

### Config yaml

```yaml

```

### Possible solution

I'd bet the document ids aren't being rotated along the ring, causing the issue

- [x] Linux
- [ ] macOS
- [ ] Windows

### Python Version

3.11.15

### axolotl branch-commit

08fc7de87e79f38c367f6776c5111b40a914062e

### Acknowledgements

Hetens:
## Summary

Fixes #3608. `ring_attn_func: batch_ring` silently inflates training loss (~1.8×) on pre-packed multi-document sequences under context parallelism. This turns that silent corruption into a loud, immediate error.

Full diagnosis and reproduction in [the issue thread](https://github.com/axolotl-ai-cloud/axolotl/issues/3608#issuecomment-5040210074) — reproduced on `main` (v0.18.0), 2× GPU, SmolLM2-135M:

| config | step-0 loss | ratio |
|---|---|---|
| multi-doc `cp=1` baseline | 3.171875 | 1.000× |
| multi-doc `cp=2, batch_ring` | 5.812500 | **1.833×** |
| multi-doc `cp=2, varlen_llama3` | 3.171875 | 1.000× |
| single-doc control (all three configs) | 3.031250 | 1.000× (bitwise-identical) |

## Root cause

`batch_ring`'s adapter (`monkeypatch/ring_attn/adapters/batch.py`) calls the ring-flash-attn **batch** API, which has no `cu_seqlens` channel — it accepts `position_ids`/`cu_seq_lens_*` and discards them (each is documented `"Not used in this implementation."`). So it treats a packed sequence as one contiguous document and rotates K/V across every internal boundary; queries in document B attend to keys in document A, inflating the loss. `varlen_llama3` derives `cu_seqlens` and is unaffected — hence it lands exactly on baseline.

The selector in `validation.py` only routes to `varlen_llama3` when the **`sample_packing` flag** is set:

```python
self.ring_attn_func = RingAttnFunc.VARLEN_LLAMA3 if sample_packing else RingAttnFunc.BATCH_RING
```

Externally pre-packed data (`position_ids` reset per document, `sample_packing: false`) falls through to `batch_ring`. The existing decontamination check (`validation.py`) keys on `attn_implementation`, not `ring_attn_func`, so nothing catches it.

## The change

Document boundaries can't be seen at config-validation time (they live in the data), so the guard goes where `position_ids` are available — `apply_sequence_parallelism`, which already receives both the batch and `ring_attn_func` (previously unused there). When `batch_ring` meets `position_ids` containing resets, it raises with a message pointing to `varlen_llama3`. Boundary detection mirrors `get_cu_seqlens_from_pos_ids` (reset-to-0 after stripping right-side padding), so a single padded document is **not** mistaken for a pack — the legitimate `batch_ring` case is untouched.

This is the **safety-net** direction: it converts silently-wrong training into a config error and can't regress any working setup. An alternative is to **auto-route** to `varlen_llama3` when `position_ids` contain resets — more convenient but more opinionated (behavior change for existing configs, detection semantics). Happy to switch to that or add it on top — deferring to maintainers on scope.

## Test

`tests/utils/ctx_managers/test_sequence_parallel_doc_boundaries.py` (CPU-only):
- `_has_multiple_documents` across single-doc, multi-doc, and padded variants.
- `apply_sequence_parallelism` **raises** for `batch_ring` + multi-doc, and **does not** for the single-doc case.

Verified against a `git worktree` on unfixed `main`: the multi-doc test fails there with `DID NOT RAISE ValueError` (the function returns normally — the silent path), and passes on this branch. Pinned `ruff`/`ruff-format`/`bandit` clean; `mypy` scope unchanged.

## Draft

Opening as draft to confirm the direction (guard vs. auto-route) before finalizing. The second, separate observation from the investigation — that `batch_ring`'s adapter doesn't import against the pinned `transformers==5.14.1` (two flash-attn helpers moved modules upstream) — I'll file as its own issue rather than fold in here.

### DIFF
--- a/src/axolotl/utils/ctx_managers/sequence_parallel.py
+++ b/src/axolotl/utils/ctx_managers/sequence_parallel.py
@@ -19,6 +19,24 @@ from axolotl.monkeypatch.ring_attn import (
 from axolotl.utils.schemas.enums import RingAttnFunc
 
 
+def _has_multiple_documents(position_ids: torch.Tensor) -> bool:
+    """Whether any row packs more than one document.
+
+    A document boundary shows up as ``position_ids`` resetting to 0 mid-sequence.
+    Mirrors the boundary detection in ``get_cu_seqlens_from_pos_ids`` (reset to 0
+    starts a new document) after stripping right-side zero padding, so single
+    padded documents are not mistaken for packs.
+    """
+    if position_ids.dim() == 1:
+        position_ids = position_ids.unsqueeze(0)
+    for row in position_ids:
+        padding_length = (row == 0).int().flip(dims=[0]).cumprod(dim=0).sum().item()
+        adjusted_row = row[:-padding_length] if padding_length else row
+        if adjusted_row.numel() > 1 and bool((adjusted_row[1:] == 0).any()):
+            return True
+    return False
+
+
 # TODO(djsaunde): implement zigzag, stripe patterns here (and elsewhere) in this
 # module. Currently, we just focus on batch ring and varlen llama3 for simplicity.
 def apply_sequence_parallelism(
@@ -39,8 +57,9 @@ def apply_sequence_parallelism(
         local_rank: Local rank in the sequence parallel group.
         local_world_size: World size of the sequence parallel group.
         gradient_accumulation_steps: Number of steps to accumulate gradients over.
-        ring_attn_func: Which ring attention function to use. Currently unused, but
-            related to above TODO.
+        ring_attn_func: Which ring attention function to use. ``batch_ring`` does not
+            respect document boundaries and is rejected for pre-packed multi-document
+            sequences.
 
     Returns:
         tuple of:
@@ -52,6 +71,18 @@ def apply_sequence_parallelism(
 
     # Update ring attention params if needed
     if batch.get("position_ids") is not None and batch_size == 1:
+        # batch_ring ignores cu_seqlens: it rotates K/V across position_id resets, so
+        # tokens attend across documents and the loss is silently wrong. Only varlen
+        # respects boundaries within a packed sequence.
+        if ring_attn_func is RingAttnFunc.BATCH_RING and _has_multiple_documents(
+            batch["position_ids"]
+        ):
+            raise ValueError(
+                "ring_attn_func='batch_ring' does not respect document boundaries in a "
+                "packed sequence: it rotates K/V across position_id resets, so queries "
+                "attend across documents and the training loss is silently wrong. Set "
+                "ring_attn_func: varlen_llama3 for pre-packed multi-document data."
+            )
         update_ring_attn_params(position_ids=batch["position_ids"])
     else:
         # If position_ids aren't already in the batch, create them
--- a/tests/utils/ctx_managers/test_sequence_parallel_doc_boundaries.py
+++ b/tests/utils/ctx_managers/test_sequence_parallel_doc_boundaries.py
@@ -1,0 +1,83 @@
+"""Guard against ``batch_ring`` corrupting pre-packed multi-document sequences.
+
+``batch_ring`` uses the ring-flash-attn *batch* API, which has no ``cu_seqlens``
+channel, so it rotates K/V across document boundaries inside a packed sequence and
+silently inflates the loss (axolotl-ai-cloud/axolotl#3608). ``varlen_llama3`` derives
+``cu_seqlens`` from ``position_ids`` and is unaffected. These tests pin the config
+that reproduces the bug to a loud error rather than a wrong number.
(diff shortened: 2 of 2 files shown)


winglian:
Hi @Hetens , thanks for the PR, we are doing a refactor of CP/SP in https://github.com/axolotl-ai-cloud/axolotl/pull/3844, would you mind taking a look at that and seeing how that works for you?

Hetens:
Thanks @winglian — took a good look at #3844. It supersedes this PR cleanly: it deletes `monkeypatch/ring_attn/adapters/batch.py` and `utils/ctx_managers/sequence_parallel.py` (the two files this PR touches) and deprecates `ring_attn_func` entirely, so the `batch_ring` mechanism behind #3608 is gone by construction. I'll **close this PR** in favor of the refactor — no point guarding code that's being removed.

One thing worth flagging for #3844 before it lands, though, since it's the same failure class:

The new validation rejects `sample_packing: true` under CP:
```python
if self.sample_packing:
    raise ValueError("sample_packing is not yet supported with ringmaster context parallelism; ...")
```
and the plugin README notes *"Packing/varlen is v2."* That's the right call. But the #3608 repro didn't use `sample_packing: true` — it used **`sample_packing: false` with externally pre-packed multi-document data** (position_ids resetting per document, the NeMo/iclmax pretraining-packing convention). That shape slips past the `sample_packing` guard: the flag is false, yet the sequence still contains document boundaries.

So until packing/varlen v2, that case is unguarded. Whether it's actually unsafe depends on the strategy — under Ulysses (all-to-all) the HF kernel sees a full sequence and may respect `position_ids`, but under Ring it needs varlen-aware masking that "v2" implies isn't there yet — so it could silently contaminate across documents the same way, now with no `varlen_llama3` escape hatch to fall back to.

Two ways I'd be happy to help close that gap against the #3844 branch:

1. **Safe-by-refusal guard** — extend `check_context_parallel_size` to also reject sequences whose `position_ids` contain resets (not just `sample_packing: true`) until v2 lands. It's a few lines; I already have the boundary detector (`_has_multiple_documents`, reset-to-0 with padding stripped) and a CPU regression test from this PR that port straight over.
2. **Empirical check** — I have the #3608 repro + a working 2-GPU setup; I can run it against `feat/context-parallel-ringmaster` (adapted to the plugin flow + the `ringmaster` package) and report whether the multi-doc contamination reproduces under Ring/Ulysses, so the guard is driven by a measurement rather than my reading of it.

Happy to do either or both — just let me know which is useful and I'll open it against your branch.

Hetens:
Closing as superseded by #3844, which removes the batch_ring / ring_flash_attn path this PR guarded. Follow-up on the residual pre-packed (sample_packing:false + position_id resets) gap will go against the #3844 branch — see comment above.

winglian:
Thanks @Hetens , if you can tackle the follow-ups after https://github.com/axolotl-ai-cloud/axolotl/pull/3844 is merged, that would be much appreciated.

PR #3869, opened 2026-07-24
feat: model support system (staged templates) refactor

thad0ctor:
# Description

Adds an immutable staged model-support layer on top of the per-architecture registry (merged upstream in axolotl-ai-cloud#3774)

- reusable family templates (`VANILLA_CAUSAL_LM`, `IMAGE_TEXT_TO_TEXT`) plus declarative per-model `ModelProfile`s that layer capability, component-strategy, matcher, and hook differences over a family
- typed lazy component providers (auto-model class, processing strategy), explicit lifecycle hook phases (`BEFORE_CONFIG_LOAD` → `AFTER_ADAPTER_LOAD`), phase replacement, and localized idempotent monkeypatch escape hatches
- legacy descriptor attributes/methods are adapted into the staged lifecycle, so existing descriptors keep working during migration; a contextvar guard prevents `super()`/redispatch double-runs
- hardened registry: fail-fast rejection of non-`ModelSupport` registrations, deterministic plugin-over-builtin override ordering, retryable builtin loading, and ambiguity errors for overlapping matchers instead of import-order-dependent first-match
- Kimi Linear and PaddleOCR-VL migrated to profiles; processor and run-mode context threaded through model loading (including the RL reference-model loader); extension contract documented in `docs/agents/new_model_support.md`

Profile resolution order: family template → per-model declaration → explicit legacy override. Omitted strategy fields inherit, explicit `None` restores the generic fallback, and profile hooks append unless a phase is explicitly replaced. Unprofiled and unregistered models keep the existing neutral/generic fallbacks; there are no YAML/config-schema changes.

### Structure

```
src/axolotl/model_support/
  base.py       ModelSupport descriptor base, Capability value objects
                (Supported / Unsupported / Experimental), check_capability
  profile.py    ModelProfile + ModelFamilyTemplate + typed parts
                (ModelStrategies, ModelMatchers, ModelHooks, ModelHookContext),
                resolve_model_support() resolution, run_model_support_hooks() dispatch
  templates.py  reusable family templates: VANILLA_CAUSAL_LM, IMAGE_TEXT_TO_TEXT
  registry.py   model_type -> descriptor registry (@register_model_support,
                get_model_support / _for_cfg / _for_processor)
  kimi_linear/    profile descriptor using phase hooks (remote-code patching)
  paddleocr_vl/   profile descriptor for multimodal (strategy + capabilities)
```

Resolution layers values as family template → per-model profile → explicitly declared legacy attributes/methods, producing an immutable `ResolvedModelProfile`. The pipeline consumes it at fixed lifecycle boundaries:

| Phase | Dispatched from |
|-------|-----------------|
| `BEFORE_CONFIG_LOAD` | `normalize_config` → `PatchManager.apply_pre_config_load_patches` (matcher-based discovery) |
| `CONFIGURE_RUN` | `normalize_config`, once `model_type` is known |
| `BEFORE_TOKENIZER_LOAD` | `load_tokenizer` → `PatchManager.apply_pre_tokenizer_load_patches` |
| `BEFORE_MODEL_BUILD` | `ModelLoader.load()` pre-load patch slot |
| `AFTER_BASE_MODEL_BUILD` | right after the base model is built, before adapters |
| `AFTER_ADAPTER_LOAD` | after adapter load and final setup, before generic/plugin post-load hooks |

Component strategies plug into `ModelLoader._set_auto_model_loader` (auto-model class) and `get_processing_strategy` (multimodal collator strategy); `check_capability` enforces declared capabilities wherever features consult them.

### Usage example

```python
# src/axolotl/model_support/my_model/__init__.py  (or any imported plugin module)
from axolotl.model_support import (
    IMAGE_TEXT_TO_TEXT, Experimental, ModelHookPhase, ModelHooks, ModelMatchers,
    ModelProfile, ModelStrategyOverrides, ModelSupport, Unsupported,
    register_model_support,
)


def _processing_strategy_cls():          # lazy: imported only when needed
    from .processing import MyModelProcessingStrategy
    return MyModelProcessingStrategy


def _matches_processor(processor):
    from transformers import MyModelProcessor
    return isinstance(processor, MyModelProcessor)


def _before_model_build(context):        # ModelHookContext: cfg, tokenizer, ...
    from .patches import patch_my_model
    patch_my_model(context.cfg)


@register_model_support
class MyModelSupport(ModelSupport):
    model_types = ("my_model",)
    profile = ModelProfile(
        family=IMAGE_TEXT_TO_TEXT,       # inherits is_multimodal + auto-model class
        capabilities={
            "cut_cross_entropy": Unsupported("No CCE forward for this arch."),
            "sample_packing": Experimental("Verify loss parity vs unpacked."),
        },
        strategies=ModelStrategyOverrides(
            processing_strategy_cls=_processing_strategy_cls,
        ),
        matchers=ModelMatchers(processor=_matches_processor),
        hooks=ModelHooks({ModelHookPhase.BEFORE_MODEL_BUILD: (_before_model_build,)}),
    )
```

Omitted fields inherit from the family; explicit `None` restores the generic fallback; hooks append unless the phase is listed in `ModelHooks(replace_phases=...)`. Full contract in `docs/agents/new_model_support.md`.

## Motivation and Context

Supporting a new architecture today means edits scattered across loaders, integrations, and `processing_strategies.py`, with implicit ordering between them. A declarative profile keeps the generic loader as the vanilla path: a model selects a family template, declares only its differences, and attaches model-local hooks at stable boundaries when the vanilla path needs help.

Reference discussion for follow-up improvements here https://github.com/axolotl-ai-cloud/axolotl/pull/3774#issuecomment-4955010998

## How has this been tested?

- **GPU e2e smoke test (PaddleOCR-VL)**: 5-step QLoRA run on a local PaddleOCR-VL-1.6 checkpoint (RTX 3090, real image-chat data from `HuggingFaceH4/llava-instruct-mix-vsft`, flash-attention-2, 4-bit base) after rebasing onto upstream `main`. The profile machinery drove the full run: `is_multimodal` derived from the `IMAGE_TEXT_TO_TEXT` family template, `PaddleOCRVLProcessingStrategy` selected via the profile's strategy provider with assistant-only masking producing non-zero trainable tokens each step, and the `lora_kernels: Unsupported` capability correctly auto-disabling the fused LoRA kernels through the resolved profile. Loss 4.45 → 3.81 over 5 steps; adapter checkpoint saved cleanly. (Kimi Linear smoke skipped: 48B exceeds local hardware.)
- 66 focused offline model-support tests (resolution, registry, compatibility, lifecycle ordering, processor forwarding, fallback) plus the 159 loader/normalize-config/processing-strategy CPU tests, run on torch 2.13 / transformers 5.12
- full CI PyTest matrix green (Python 3.12/3.14 × torch 2.11/2.12), pinned pre-commit suite (ruff, ruff-format, mypy, bandit) passing

## Deferred / Topics for Conversation

Open questions for maintainers:

1) Is there a desire to migrate already implemented models to the new system?
2) If yes to 1 (above) there could be an agent skill to assist with migration and validation
3) model validation harness (agent skill), I have a draft here but will need to be updated once the PR lands - https://github.com/thad0ctor/axolotl/pull/63
4) additional templates (deferred for now to reduce PR scope)
5) transformers registry seams — declarative checkpoint weight conversions (`WeightConverter`, including the save-time reversal) and `register_patch_mapping` — deferred to a follow-up PR from [`feat/model-support-transformers-registries`](https://github.com/axolotl-ai-cloud/axolotl/compare/feat/model-support-transformers-registries?expand=1) per https://github.com/axolotl-ai-cloud/axolotl/pull/3869#issuecomment-5089156392 (supersedes my draft in thad0ctor#80)
6) `ModelHookContext` currently exposes both `tokenizer` and `processor`; transformers v5 is moving processor-first, so deprecating the tokenizer field toward `processor` is deferred until hook consumers migrate (from https://github.com/axolotl-ai-cloud/axolotl/pull/3869#discussion_r3656775715)

The migration question in (1) extends beyond individual models: several whole subsystems also stay on their existing code paths in this PR, each a candidate to become a typed profile strategy once its seam stabilizes. The table below lays them out in recommended implementation order — the early rows already have a ready seam and grow the descriptor inventory that the later rows depend on.

| # | Deferred subsystem | Where it lives today | Becomes a strategy when |
|---|------------------|----------------------|-------------------------|
| 1 | Remaining `model_type` branches | `loaders/patch_manager.py`, `monkeypatch/` | each branch migrates into a phase hook on its architecture's descriptor (seam ships in this PR; incremental and independently testable) |
| 2 | Loss adaptation | CCE / Liger integrations, loss patches | loss-patch dispatch consults capabilities rather than per-arch patch lists (removes the silent-drift class) |
| 3 | Batch preparation | collators in `utils/collators/`, packing utils | collator choice is profile-driven alongside the already-typed processing strategy |
| 4 | Adapter loading | `loaders/adapter.py` | PEFT wrapping exposes a stable pre/post boundary; near-term win is per-arch `lora_target_modules` defaults |
| 5 | Optimizer construction | `core/trainers/mixins/optimizer.py` + builders | optimizer factory selection is decoupled from trainer mixin wiring (rarely arch-specific today) |
| 6 | Trainer selection | `core/builders/` (`causal.py`, `rl.py`) | builder choice has a generic seam instead of RL-type/arch branching (hardest seam; few arch-specific consumers yet) |

## AI Usage Disclaimer

Fable and Opus 4.8 throughout

## Types of changes

- [x] Documentation update

## Summary by CodeRabbit

* **New Features**
  * Added declarative model-support profiles and reusable family templates for model architectures.
  * Improved multimodal model loading by making processors available throughout model and reference-model setup.
  * Added centralized lifecycle hooks for configuration, tokenizer, model, and adapter loading.
  * Expanded model-support exports and API reference documentation.

* **Bug Fixes**
  * Improved model-support selection, plugin registration, precedence, and ambiguity handling.
  * Preserved compatibility with legacy model-support implementations.

* **Documentation**
  * Added guidance for creating and migrating model-support profiles.


winglian (commented):
I think there are a couple of pieces still neded for proper model support. In transformers, see `WeightConverter(WeightTransform)` (for both model loading and saving) as well as `register_patch_mapping`. There are probably other important modification sites, but these are the primary ones I'm aware of.

winglian:
@thad0ctor actually, let's merge this as is, and I'll do a follow up PR from https://github.com/axolotl-ai-cloud/axolotl/compare/feat/model-support-transformers-registries?expand=1

src/axolotl/loaders/model.py
```diff
@@ -106,6 +111,7 @@ def __init__(
         cfg: DictDefault,
         tokenizer: PreTrainedTokenizerBase,
         *,
+        processor: ProcessorMixin | None = None,
```
NanoCode012, line 114: Nice catch, thanks

src/axolotl/model_support/base.py
```diff
@@ -71,37 +106,55 @@ def check_capability(
...
-    capabilities: ClassVar[dict[str, Capability]] = {}
+    profile: ClassVar[ModelProfile | None] = None
+    is_multimodal: ClassVar[bool] = cast(
+        bool,
+        _ProfileProjection("is_multimodal", False),
+    )
+    capabilities: ClassVar[Mapping[str, Capability]] = cast(
+        Mapping[str, Capability],
+        _ProfileProjection("capabilities", {}),
+    )
```
NanoCode012, line 136: Claude pointed this out, but I'm not sure if it's correct.
```
  ▎ capabilities resolves differently depending on which
  ▎ reader you go through, and this projection is the only
  ▎ reason for the divergence. _ProfileProjection is a
  ▎ non-data descriptor, so a legacy class-level capabilities
  ▎ = {...} shadows it entirely, while resolve_model_support()
  ▎ merges family + profile + legacy:
  ▎ class Hybrid(ModelSupport):
  ▎     model_types = ("hybrid_arch",)
  ▎     profile = ModelProfile(family=VANILLA_CAUSAL_LM,
  ▎ capabilities={"liger": Unsupported()})
  ▎     capabilities = {"cut_cross_entropy": Unsupported()}
  ▎
  ▎ Hybrid().capabilities                         #
  ▎ {'cut_cross_entropy'}          <- profile/family keys lost
  ▎ resolve_model_support(Hybrid()).capabilities  #
  ▎ {'liger','cut_cross_entropy'}
  ▎ The remaining consumer on the shadowed path is
  ▎ src/axolotl/utils/schemas/config.py:1936
  ▎ (support.capabilities.get("lora_kernels")), not touched by
  ▎ this PR.
```

tests/test_model_support_callsites.py
```diff
@@ -0,0 +1,521 @@
...
+        model_loader_module.ModelLoader,
+        "nullable_loader_test",
+        is_multimodal=True,
+    )
+    monkeypatch.setattr(model_loader_module, "get_model_support", lambda _type: support)
+    monkeypatch.setattr(
+        model_loader_module,
+        "MULTIMODAL_AUTO_MODEL_MAPPING",
+        {"nullable_loader_test": _MappedMultimodalAutoModel},
+    )
```
NanoCode012, line 161: We'd need to add a test for str mapping to cover this check `isinstance(auto_model_loader, str)` in `src/axolotl/loaders/model.py:592-593`

tests/test_model_support.py
NanoCode012: I think this test is failing now, you can refer to `capture_axolotl_warnings` fn to check this?

src/axolotl/model_support/profile.py
```diff
@@ -0,0 +1,480 @@
...
+
+    return hook
+
+
+# Declarative resolution depends only on the immutable class-level ``profile`` and
+# ``model_types``, so the result is memoized per descriptor class. Ephemeral classes
+# (e.g. per-test descriptors) are evicted once garbage-collected.
+_DECLARATIVE_CACHE: WeakKeyDictionary[type[ModelSupport], ResolvedModelProfile] = (
+    WeakKeyDictionary()
+)
```
NanoCode012: What if we change, such as in a later plugin on runtime, some var like `is_multimodal`? Would cache become stale, and would our ref still be incorrect?

src/axolotl/loaders/model.py
```diff
@@ -558,16 +568,27 @@ def _set_parallel_config(self):
...
+        )
+        auto_model_loader = (
+            auto_model_provider() if auto_model_provider is not None else None
+        )
+        if auto_model_loader is not None:
+            self.auto_model_loader = auto_model_loader
+        elif self.cfg.is_multimodal:
+            auto_model_loader = MULTIMODAL_AUTO_MODEL_MAPPING.get(
+                self.model_config.model_type, AutoModelForImageTextToText
+            )
```
NanoCode012, line 606: I think order of checks is swapped here. See Claude's points:
```
  ▎now a profile
  ▎ provider wins unconditionally
  ▎ future VANILLA_CAUSAL_LM registration for a model_type in
  ▎ MODEL_FOR_IMAGE_TEXT_TO_TEXT_MAPPING_NAMES would silently
  ▎ load via AutoModelForCausalLM
```

src/axolotl/model_support/registry.py
```diff
@@ -17,15 +19,73 @@
...
+        if _builtins_loaded or _loading_builtins:
+            return
+
+        _loading_builtins = True
+        try:
+            for module in _BUILTIN_MODULES:
+                importlib.import_module(module)
+        except Exception:
+            # Leave partial registrations intact so the failed import can be retried.
+            raise
```
NanoCode012: Is this no-op as we have `else` below?

src/axolotl/model_support/registry.py
```diff
@@ -17,15 +19,73 @@
...
+        yield support
+
+
+def _one_match(matches: list[ModelSupport], subject: str) -> ModelSupport | None:
+    if not matches:
+        return None
+    if len(matches) == 1:
+        return matches[0]
+    names = ", ".join(type(support).__name__ for support in matches)
+    raise ValueError(f"Ambiguous model support for {subject}: {names}")
```
NanoCode012: Let's also add a "Set `cfg.model_type = ...` to one of the options? instead of just error, so users know what to do.

src/axolotl/model_support/registry.py
```diff
@@ -17,15 +19,73 @@
...
+    if not (isinstance(support_cls, type) and issubclass(support_cls, ModelSupport)):
+        raise TypeError(
+            f"register_model_support requires a ModelSupport subclass, "
+            f"got {support_cls!r}"
+        )
+    model_types = support_cls.model_types
+    if not isinstance(model_types, tuple) or not model_types:
+        raise ValueError(
+            f"{support_cls.__name__}.model_types must be a non-empty tuple"
+        )
```
NanoCode012, line 62: Claude noticed this was prev a `list` and now it's a `tuple`. If anything checks this at any specific point, we should rectify those path/docs.

src/axolotl/model_support/profile.py
```diff
@@ -0,0 +1,480 @@
...
+    def with_overrides(self, overrides: ModelMatchers) -> ModelMatchers:
+        return ModelMatchers(
+            cfg=overrides.cfg if overrides.cfg is not None else self.cfg,
+            processor=(
+                overrides.processor
+                if overrides.processor is not None
+                else self.processor
+            ),
+        )
+
```
NanoCode012, line 144: There's a discrepancy with terminology here.
```
ModelMatchers.with_overrides treats None as "inherit"
ModelStrategies.with_overrides (:91) treats it as "remove"
```

src/axolotl/model_support/profile.py
```diff
@@ -0,0 +1,480 @@
...
+    """Model-specific hooks grouped by explicit lifecycle phase.
+
+    Hooks append to inherited family hooks unless their phase is included in
+    ``replace_phases``. Replacing a phase with an empty tuple suppresses the family
+    hooks for that phase.
+    """
+
+    by_phase: Mapping[ModelHookPhase, tuple[ModelHook, ...]] = field(
+        default_factory=dict
+    )
```
NanoCode012, line 157: Similar to above:
```
Matcher scan is now exhaustive (base returned on first

  ModelStrategies.with_overrides (:91) treats it as "remove" — same name,
  opposite semantics, only the strategy behavior documented.
```

src/axolotl/model_support/registry.py
```diff
@@ -17,15 +19,73 @@
...
-        importlib.import_module(module)
+    global _builtins_loaded, _loading_builtins  # pylint: disable=global-statement
+    with _builtins_lock:
+        if _builtins_loaded or _loading_builtins:
+            return
+
+        _loading_builtins = True
+        try:
+            for module in _BUILTIN_MODULES:
+                importlib.import_module(module)
```
NanoCode012: Claude also picked up potential lock conflict. Python already handles idempotent , this lock will then clash with anything later down like Kimi-linear lock. See:
```
 _ensure_builtins holds L for the entire import loop
  (registry.py:28-42), so it acquires L then wants I:

  Thread A: get_model_support("llama")
              -> _ensure_builtins()
              -> acquires L
              -> import_module("axolotl.model_support.kimi_linear")
   # wants I

  Thread B: already mid-import of kimi_linear (it got there first,
  e.g. a plugin
            module doing `from axolotl.model_support.kimi_linear
  import ...`)
              -> holds I, executing the module body
              -> hits @register_model_support
              -> register_model_support() -> _ensure_builtins()
  # wants L

  A holds L wants I; B holds I wants L. Hang. The RLock doesn't help
  — re-entrancy is per-thread, and these are two threads.
```

src/axolotl/model_support/registry.py
```diff
@@ -61,20 +124,30 @@ def get_model_support(model_type: str | None) -> ModelSupport | None:
...
 
-    Used for pre-config/pre-tokenizer patching, where ``model_type`` is not
-    yet available and descriptors match on the config (e.g. model name).
+    An exact resolved ``model_config_type`` wins. Before that is available,
+    descriptors can match on config fields such as the model name.
     """
-    _ensure_builtins()
-    for support in dict.fromkeys(_REGISTRY.values()):
-        if support.matches_cfg(cfg):
+    model_type = getattr(cfg, "model_config_type", None)
```
NanoCode012: cosmetic, we can do `cfg.model_config_type` as we already default to None

src/axolotl/utils/config/__init__.py
```diff
@@ -312,7 +322,15 @@ def normalize_config(cfg):
...
                 cfg[k] = False
 
-        model_support.validate_cfg(cfg)
+    run_model_support_hooks(
+        model_support,
+        ModelHookPhase.CONFIGURE_RUN,
+        ModelHookContext(
+            cfg=cfg,
+            model_config=model_config,
+            inference=getattr(cfg, "inference", None),
```
NanoCode012: cosmetic, we can do `inference` as we already default to None

docs/agents/new_model_support.md
```diff
@@ -4,29 +4,146 @@ Guide for debugging and adding support for new model architectures in axolotl. B
...
+The context fields available at each phase are:
+
+| Phase | `model_config` | `tokenizer` | `processor` | `model` | `inference` / `reference_model` |
+|-------|----------------|-------------|-------------|---------|---------------------------------|
+| `BEFORE_CONFIG_LOAD` | — | — | — | — | `inference` from `cfg` when set; `reference_model` unavailable |
+| `CONFIGURE_RUN` | Available | — | — | — | `inference` from `cfg` when set; `reference_model` unavailable |
+| `BEFORE_TOKENIZER_LOAD` | — | — | — | — | `inference` from `cfg` when set; `reference_model` unavailable |
+| `BEFORE_MODEL_BUILD` | Available | Available | Optional | — | Available |
+| `AFTER_BASE_MODEL_BUILD` | Available | Available | Optional | Raw base model | Available |
+| `AFTER_ADAPTER_LOAD` | Available | Available | Optional | Final loaded model | Available |
```
NanoCode012, line 132: cosmetic: can you swap these empty field for regular dash?

docs/agents/new_model_support.md
```diff
@@ -49,15 +166,10 @@ The fastest way to isolate a loss issue — bypass the trainer entirely:
...
-from axolotl.loaders.tokenizer import load_tokenizer
-from axolotl.loaders.model import ModelLoader
+from axolotl.cli.utils.load import load_model_and_tokenizer
 
 cfg = load_cfg("your_config.yaml")
-normalize_config(cfg)
-prepare_plugins(cfg)
-tokenizer = load_tokenizer(cfg)
-model, _ = ModelLoader(cfg, tokenizer).load()
+model, tokenizer, processor = load_model_and_tokenizer(cfg=cfg)
```
NanoCode012, line 172: out of scope: but I'm wonder whether we should even return tokenizer. Transformers have been moving to processor by default in v5.

NanoCode012 (commented):
Looks like a good first step. Added comments below.

> re: the conversion mapping

I think there are some handling here, but missing the handling during save?
```
  ▎profile.py:58 can't
  ▎ express transformers' scoped @contextmanager
  ▎ apply_patches() / unregister_patch_mapping, so a
  ▎ register_patch_mapping from a hook leaks globally with no
  ▎ teardown
```

> re: agent to repopulate in followup PR

Yeah, I think we'd need to do this later to fill in old Model Support. Relying on example configs was very brittle.

src/axolotl/model_support/base.py
thad0ctor: Confirmed — the schema validator was reading the shadowed attribute; it now goes through resolve_model_support(), with a hybrid-descriptor test that fails on the old path (f9eb774ca).

tests/test_model_support_callsites.py
thad0ctor: Added — test_model_loader_resolves_string_multimodal_mapping_to_generic_loader covers the str fallback.

tests/test_model_support.py
thad0ctor: Good catch — it fails once anything has called configure_logging() (propagate=False blinds caplog); switched to capture_axolotl_warnings.

src/axolotl/model_support/profile.py
thad0ctor: It would have gone stale — cache entries now validate against the current profile/model_types on lookup (regression test added); legacy attrs like is_multimodal were already read uncached.

src/axolotl/loaders/model.py
thad0ctor: Fixed — a non-multimodal profile's provider no longer shadows the multimodal mapping (warns and falls back), tests cover both directions.

src/axolotl/model_support/registry.py
thad0ctor: It was — gone in the _ensure_builtins restructure.

src/axolotl/model_support/registry.py
thad0ctor: Done — the error now suggests narrowing the matchers or registering under an exact model_type (model_config_type is derived from the HF config, so there's no user-settable key to point at).

src/axolotl/model_support/registry.py
thad0ctor: Lists are now accepted and normalized to the documented tuple at registration, so pre-refactor descriptors keep working.

src/axolotl/model_support/profile.py
thad0ctor: Intentional (matchers have no removal form, only narrower overrides) but it was underdocumented — both docstrings and the docs now spell out the difference.

src/axolotl/model_support/profile.py
thad0ctor: Same as above — documented; the exhaustive scan is deliberate so ambiguity is detected instead of depending on registration order.

src/axolotl/model_support/registry.py
thad0ctor: Confirmed real — builtins now import outside the registry lock; the new two-thread regression test deadlocks on the old code and passes now.

src/axolotl/model_support/registry.py
thad0ctor: Done.

src/axolotl/utils/config/__init__.py
thad0ctor: Done — aligned the two identical sites in patch_manager too.

docs/agents/new_model_support.md
thad0ctor: Agreed on the direction — kept both for now since pre-build hooks still consume the tokenizer, but deprecating toward processor in the migration follow-up makes sense.

thad0ctor:
@NanoCode012 all inline comments addressed in f9eb774ca. Thanks! 

One extra find while validating: the lora_kernels auto-enable tests were vacuous (the validator lives on AxolotlConfigWCapabilities, which bare validate_config never instantiates) — they now pass capabilities so it actually runs.

thad0ctor:
> @thad0ctor actually, let's merge this as is, and I'll do a follow up PR from https://github.com/axolotl-ai-cloud/axolotl/compare/feat/model-support-transformers-registries?expand=1

sounds good, I'll update the deferred in the PR description for visibility

src/axolotl/loaders/model.py
thad0ctor: you're welcome and thanks for the feedback!

docs/agents/new_model_support.md
thad0ctor: All set

winglian (approved):

PR #3880, opened 2026-07-27
gemma skip

ved1beta:
# Description
Skip the sdpa override when the call is building a sliding mask — those requests carry and_mask_function
```
if kwargs.get("and_mask_function") is not None:
    return original(config, *args, **kwargs)

```

## Motivation and Context
https://discord.com/channels/1104757954588196865/1111279858136383509/1530217278501552128

## How has this been tested?

config runs fine 

## AI Usage Disclaimer
NA

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Gemma 4 vision attention mask handling to prevent FlashAttention indexing and unpadding errors.
  * Preserved sliding-window attention behavior while maintaining the correct mask format for global attention.

### DIFF
--- a/src/axolotl/monkeypatch/gemma4_hybrid_mask.py
+++ b/src/axolotl/monkeypatch/gemma4_hybrid_mask.py
@@ -27,6 +27,24 @@ text backbone's ``forward`` is always 4D/SDPA-compatible.
 ``create_sliding_window_causal_mask`` is left alone, so sliding-window
 layers continue to receive FA2-format masks.
 
+That last invariant needs one extra guard. ``create_masks_for_vision_model``
+— taken whenever ``mm_token_type_ids`` is supplied and the text config sets
+``use_bidirectional_attention="vision"``, which is every Gemma 4 batch under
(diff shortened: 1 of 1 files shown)


winglian:
@thad0ctor iirc you use gemma4 and are working on ocr/multimodal. Have you run into this?>

thad0ctor:
> @thad0ctor iirc you use gemma4 and are working on ocr/multimodal. Have you run into this?>

@winglian 

  Haven't hit it, but only because of a version pin, my gemma4 OCR runs have all been on transformers ≤5.12.1,
  where Gemma4Model.forward still builds the mapping the old way (create_causal_mask for global,
  create_sliding_window_causal_mask for sliding). The wrapper only ever touches the global mask there, so it's
  fine.

  create_masks_for_vision_model shows up in 5.13 it seems, and that's where it breaks: both masks go through
  create_causal_mask, with the sliding one tagged and_mask_function=sliding_window_overlay(...). Since main is
  on 5.14.1 now this makes sense

NanoCode012 (approved):

PR #3882, opened 2026-07-27
perf(quantize): pass raw ptrs/ints to bnb dequant when argtypes are set

NanoCode012:
# Description

A bnb dev commented on discord about our access to the low level C types and about potential optim in the 0.50.0 release. This PR checks if those new types are available and uses them, otherwise, falls back.

Ref: https://discord.com/channels/1104757954588196865/1104758010959634503/1529861378934833162

## How has this been tested?

We do have quantize test to check it, waiting on CI.

## Summary by CodeRabbit

  * Improved NF4 quantization performance by reducing overhead during dequantization.
  * Added compatibility handling for different bitsandbytes versions while preserving existing behavior.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -78,7 +78,7 @@ dependencies = [
     "mistral-common==1.11.5",
 
     # Platform-specific (Linux only)
-    "bitsandbytes==0.49.1 ; sys_platform != 'darwin'",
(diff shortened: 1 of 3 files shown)


NanoCode012:
The gains should only be seen once we upgrade to 0.50.0. This is just an optim for later

winglian (approved):

PR #3884, opened 2026-07-28
feat(model_support): declarative weight-conversion and patch-mapping registrations

winglian:
# Description

Follow-up to #3869, addressing the review comment about the missing transformers modification sites: `WeightConverter` (for both model loading and saving) and `register_patch_mapping`.

## `ModelRegistrations`: typed seams for transformers' own registries

Profiles (and family templates) gain a `registrations` part with two lazy providers:

- `weight_conversions` — maps a `model_type` or model class name to `WeightTransform` entries (`WeightConverter`, `WeightRenaming`, ...), registered via `transformers.conversion_mapping.register_checkpoint_conversion_mapping`. Class-name keys take priority inside transformers, and remote-code models only receive conversion mappings registered through this user path (`USER_REGISTERED_MAPPINGS`).
- `patch_mappings` — maps class names (or regex patterns) to replacement `nn.Module` subclasses, registered via `transformers.monkey_patching.register_patch_mapping`; transformers applies them as an init-context during `from_pretrained`/`from_config` module construction.

Both are applied idempotently (`overwrite=True`) at the `BEFORE_MODEL_BUILD` boundary in `PatchManager`, before checkpoint construction. They are declared as one part deliberately: a patch mapping that changes a module's checkpoint layout must ship matching weight conversions — transformers' own kernel-fusion machinery (`fusion_mapping.py`, `hub_kernels.py`) always registers the two together.

**The save side is handled up front.** `save_pretrained(save_original_format=True)` — the default — reverses registered conversions via each op's `reverse_op`, and raises for transforms carrying a `quantization_operation`. Registration therefore warns immediately when a declared transform cannot round-trip, instead of failing at save/merge time. (The reverse pass is skipped while a PEFT config is loaded, so merged and full-parameter saves are the affected path.)

Resolution follows the existing strategy semantics: family → profile, omitted field inherits, explicit `None` clears; there is no legacy-method equivalent.

In-tree precedent that wants this seam once migrated: the NVFP4 expert converters in `integrations/kernels/`, the `nemotron_h` conversion-mapping fix in `patch_manager.py`, and `lora_merge`'s fuse/unfuse handling.

## Review fix

- **Kimi dynamic-module redirect anchored to exact module stems.** The substring match (`"modeling_kimi" in module_path`) also captured sibling remote-code modules such as `modeling_kimi_vl` (Kimi-VL), redirecting them to the in-tree kimi-linear copies and crashing class resolution. Matching is now on the exact module stem, tolerant of both path and dotted module forms.

## Motivation and Context

See the review discussion on #3869: the profile system covered auto-model class, processing strategy, capabilities, matchers, and lifecycle hooks, but had no seam for the transformers-side registration tables a new architecture needs — most immediately checkpoint weight conversion (load **and** save) and sanctioned module class replacement. This PR adds those two as a paired, typed declaration; further registries (attention/mask interfaces, experts interface, quantizers, loss) can follow the same shape.

## How has this been tested?

- New `tests/test_model_support_registrations.py`: resolution semantics (inherit/override/clear), real registration into transformers' registries through the `PatchManager` boundary with idempotent re-registration, the irreversibility warning, and no-op behavior for registration-free/unregistered models.
- New regression test for the kimi stem anchoring.
- Full model-support suite (77 tests, including the tests added in the final #3869 revisions) on torch 2.x / transformers 5.14.1, rebased on `main` at the #3869 merge.

## Types of changes

- [x] Documentation update

## Summary by CodeRabbit

* **New Features**
  * Added declarative model-support registrations for checkpoint weight conversions and patch mappings.
  * Registrations are applied automatically before model loading and can inherit or override profile settings.
  * Added public model-registration descriptors for extension and integration support.
  * Added warnings when weight conversions may prevent restoring the original checkpoint format.

* **Bug Fixes**
  * Improved Kimi-Linear module matching to avoid redirecting similarly named modules.

* **Documentation**
  * Documented registration precedence, transformer registry declarations, conversion behavior, and patch mappings.


src/axolotl/loaders/patch_manager.py
```diff
@@ -200,6 +225,33 @@ def apply_post_model_build_patches(self, model: PreTrainedModel):
...
+        )
+        if conversions:
+            from transformers.conversion_mapping import (
+                register_checkpoint_conversion_mapping,
+            )
+
+            for key, entries in conversions.items():
+                entries = list(entries)
+                register_checkpoint_conversion_mapping(key, entries, overwrite=True)
+                _warn_irreversible_weight_transforms(key, entries)
```
NanoCode012: I think we should merge instead of overwrite.
```
NVFP4 experts on qwen3_moe/glm4_moe, both wiped 2→1
```

What do you think? If we merge, we need to do read->merge style like in nemotron. I can make this change if needed

PR #3885, opened 2026-07-28
fix(attention): don't route fp32/CPU QKV into the sdpa varlen flash kernel

winglian:
The sdpa_varlen fast path guarded on mask/dropout/head_dim/scaling but not on dtype or device, so sdpa + sample_packing with fp32 (or CPU) tensors fed torch.nn.attention.varlen.varlen_attn, whose backing flash kernel only supports CUDA fp16/bf16 — crashing with 'FlashAttention only support fp16 and bf16 data type' on torch 2.12.1. Such rows now fall back to stock SDPA with the rebuilt block-diagonal mask (documents stay isolated).

# Description

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved attention compatibility by limiting the optimized variable-length path to supported CUDA half-precision formats.
  * Ensured float32 attention uses the standard implementation for correct results.

* **Tests**
  * Added regression coverage validating accurate float32 attention behavior in packed, multi-document inputs.

### DIFF
--- a/src/axolotl/loaders/patch_manager.py
+++ b/src/axolotl/loaders/patch_manager.py
@@ -748,6 +748,16 @@ class PatchManager:
         if not varlen_available():
             return  # torch < 2.10; block-diagonal packing path is correct
 
+        # Every call would fall back anyway, and patching drops the shared 4D mask.
+        half = self.cfg.torch_dtype in (torch.float16, torch.bfloat16)
+        if not (half and torch.cuda.is_available()):
+            if explicit:
+                LOG.info(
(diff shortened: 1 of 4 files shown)


NanoCode012:
Fixed the test to move to cpu checks and added cuda guard

NanoCode012 (commented):
>  sdpa + sample_packing with fp32 (or CPU) tensors fed torch.nn.attention.varlen.varlen_attn

when does this happen though? act checkpointing or with mixed precision?

PR #3887, opened 2026-07-28
fix(ebft): ebft config validator bypassed

NanoCode012:
# Description

From reviewing some other PRs, found that EBFT configs may be silently ignored. This PR fixes it.

## How has this been tested?

Ongoing (will add CI if needed)

## AI Usage Disclaimer

Claude

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved configuration validation when EBFT is explicitly set to `None`.
  * Prevented validation errors and incorrect warnings across torch compilation, gradient checkpointing, activation offloading, and sequence-length settings.


winglian (approved):

ErenAta16 (commented):
The fix is right — I reproduced the failure and the two families of it are genuinely different, which is worth pinning down because the description only describes one of them.

Splitting the four validators by mode:

```
check_ebft_torch_compile                       mode="before"  (dict)
check_ebft_strided_sequence_len                mode="before"  (dict)
check_ebft_gradient_checkpointing_reentrant    mode="after"   (model)
check_ebft_activation_offloading               mode="after"   (model)
```

The two `mode="after"` ones were calling `.get("mode")` on `self.ebft`, which is an `EBFTConfig`, not a dict:

```python
class EBFTConfig(BaseModel):   # plain BaseModel, no .get, no __getitem__
ebft: EBFTConfig | None = Field(...)
```

A Pydantic model is always truthy, so `(self.ebft or {})` yields the model itself and `.get` doesn't exist on it. Reproduced against a stand-in with the same shape:

| config | `main` |
|---|---|
| `rl: ebft`, no `ebft:` block, dangerous combo | guard silently skipped, no error, no warning |
| `rl: ebft`, `ebft:` block set, dangerous combo | `AttributeError: 'EBFTConfig' object has no attribute 'get'` |
| `rl: ebft`, `ebft:` block set, nothing wrong | `AttributeError` — the same crash on a perfectly valid config |
| `rl: dpo` | unaffected |

The short-circuit doesn't save it either: `RLType` is a `str` `Enum`, so `RLType.EBFT == "ebft"` is `True` and evaluation reaches the `.get` on every EBFT run.

So "silently ignored" is accurate for row 1 — omit the `ebft:` block and the guard quietly never fires — but rows 2 and 3 are a hard failure, and row 3 is the one I'd call out: setting an `ebft:` block at all is currently fatal at config validation, whether or not the dangerous combination is present. Someone reading the PR title will go looking for a missing warning and won't connect it to a crash they may have already worked around by deleting the block.

The two `mode="before"` ones are a separate defect with the same shape:

```python
data.get("ebft", {}).get("mode")
```

The `{}` default only applies when the key is *absent*. An explicit `ebft:` with nothing under it parses to `None`, so this becomes `None.get("mode")`. `(data.get("ebft") or {})` is the right shape there, and keeping dict access in the before-validators while moving the after-validators to attribute access is the correct split rather than an inconsistency.

Two smaller things:

`transform(cfg, *args, **kwargs)` → `transform(cfg, **kwargs)` in `ebft_pretrain.py` is a real narrowing, not cosmetic — any caller passing a second positional now raises `TypeError` instead of silently discarding it. Worth a line in the description since the file also moves out of `examples/` into `src/` in the same commit, and the two changes read as one rename in the diff.

Moving `ebft_pretrain.py` under `src/axolotl/prompt_strategies/ebft/` makes it importable by path, which is presumably the point, but it also means the `examples/ebft/*.yaml` that referenced it by file path need to be pointing at the new location. `llama-3b-ebft-strided-fft.yaml` loses a line here; worth confirming the other example configs in that directory don't still name the old path.

PR #3888, opened 2026-07-28
fix(kernels): resolve the NF4 dequant stream per call

NanoCode012:
# Description

Followup to #3878 . Applies the same optimization to the `quantize.py`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved NF4 dequantization stream handling for more reliable CUDA execution.
  * Reduced potential issues caused by stale or incorrectly selected device streams.

### DIFF
--- a/src/axolotl/kernels/quantize.py
+++ b/src/axolotl/kernels/quantize.py
@@ -5,7 +5,7 @@ from typing import List
 
 import bitsandbytes as bnb
 import torch
-from bitsandbytes.functional import QuantState, _get_tensor_stream, get_ptr
+from bitsandbytes.functional import QuantState, get_ptr
 
 cdequantize_blockwise_fp32 = bnb.functional.lib.cdequantize_blockwise_fp32
 cdequantize_blockwise_fp16_nf4 = bnb.functional.lib.cdequantize_blockwise_fp16_nf4
@@ -40,8 +40,8 @@ def _ctypes_nf4_dequant(
         n_elements_absmax, dtype=torch.float32, device=target_device
     )
 
-    # raw pointer: the live stream without building a Python Stream on this hot path
-    stream = _get_tensor_stream(W)
+    # c_void_p is required: bnb sets no argtypes, so a bare int truncates to a C int.
+    stream = ctypes.c_void_p(torch._C._cuda_getCurrentRawStream(W.device.index))
 
     cdequantize_blockwise_fp32(
         get_ptr(code2),
--- a/src/axolotl/kernels/quantize.py
+++ b/src/axolotl/kernels/quantize.py
@@ -5,7 +5,7 @@ from typing import List
 
 import bitsandbytes as bnb
 import torch
-from bitsandbytes.functional import QuantState, get_ptr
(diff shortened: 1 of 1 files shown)


winglian (approved):

matthewdouglas:
I'm not sure I would recommend doing this. It's likely I'd remove `bitsandbytes.functional._get_tensor_stream` in the future. If you wanted to simplify I suppose `stream = torch._C._cuda_getCurrentRawStream(W.device.index)`. This is approximately what we do under the hood now.

NanoCode012:
Thank you for pointing that out. I’ll make that change

ErenAta16 (commented):
@matthewdouglas's point stands and the swap to `torch._C._cuda_getCurrentRawStream(W.device.index)` is the right landing spot. One thing worth adding before it goes in, because the PR is selling itself short.

**This is a correctness fix, not only an optimization.**

The description says "applies the same optimization", and the removed comment frames the dict purely as a speed measure:

```python
# Cached per-device: per-call current_stream() measurably slows this hot path.
CUDA_STREAM: dict[torch.device, torch.cuda.Stream] = {}
```

But the cache is populated once per device and never refreshed:

```python
stream = CUDA_STREAM.get(target_device)
if stream is None:
    stream = CUDA_STREAM.setdefault(target_device, torch.cuda.current_stream(target_device))
```

`torch.cuda.current_stream()` is not a property of the device, it's a property of the *calling context*. Inside `with torch.cuda.stream(s):`, or during graph capture, it returns something different. So on `main` the first dequant on a device freezes whichever stream happened to be current at that moment, and every later call — including ones made under a different stream — enqueues onto the frozen one.

Both `cdequantize_blockwise_fp32` and the NF4 kernel take that pointer, and `out` / `out_absmax` are allocated in the caller's context. So a caller that switches streams gets its kernels launched on a stream it isn't ordering against, with no error: just a read of memory whose producing kernel hasn't necessarily completed. That's the shape of a bug that shows up as rare, load-dependent garbage rather than a crash, and only on multi-stream or graph-capture paths — which is presumably why nobody hit it.

Resolving per call removes that entirely. Worth saying so in the description, both so it can be backported deliberately and so anyone bisecting an intermittent corruption report on the NF4 path lands here.

**Two small things on the replacement itself.**

`target_device` is `W.device` in this function, and `_get_tensor_stream(W)` / `_cuda_getCurrentRawStream(W.device.index)` both key off `W`, so there's no device mismatch to worry about — the buffers are allocated on the same device the stream is read from. Confirmed by reading the head of `_ctypes_nf4_dequant`; just noting it since a stream/allocation device split would be the obvious way for this kind of change to go wrong.

Both spellings are private APIs, so this trades one underscore for another. The argument for the torch one isn't that it's public, it's that removing it would break every out-of-tree kernel that launches into the current stream, which is a much larger blast radius than a bnb internal. That reasoning is worth a comment on the line, otherwise the next person sees `torch._C._` and reaches for `torch.cuda.current_stream(...).cuda_stream` — which is public, returns the same pointer, and reintroduces exactly the Python `Stream` construction this is avoiding.

PR #3889, opened 2026-07-28
feat(model_support): expand profile seams across transformers registries and the save boundary

winglian:
# Description

Follow-up to #3884 completing the deferred transformers-seam list, so the model-support surface stabilizes before any model migrations.

## `ModelRegistrations`: six new registry seams

Same contract as the existing fields (lazy zero-argument providers; family → profile resolution; omitted inherits, explicit `None` clears; applied idempotently at the `BEFORE_MODEL_BUILD` boundary). `with_overrides` is now field-generic so future registries are one-line additions.

- `attention_functions` / `attention_mask_functions` → `ALL_ATTENTION_FUNCTIONS` / `ALL_MASK_ATTENTION_FUNCTIONS`. Declared as a pair for the same reason `weight_conversions`/`patch_mappings` are: every attention implementation needs a matching mask entry.
- `experts_functions` → `ALL_EXPERTS_FUNCTIONS` (selected via `config._experts_implementation`).
- `quantizers` → `QuantizerRegistration(quantizer_cls, config_cls)` written to `AUTO_QUANTIZER_MAPPING` / `AUTO_QUANTIZATION_CONFIG_MAPPING`. Written directly because `register_quantizer()` raises on any existing key: re-applying the same classes is a no-op, a different existing registration is overwritten with a warning. (The in-tree NVFP4 quantizer's `AUTO_QUANTIZER_MAPPING["fp8"]` dict-swap can migrate onto this.)
- `auto_classes` → `AutoClassRegistration` entries wiring a config class plus model / tokenizer / processor classes through `AutoConfig.register`, `AutoModelFor*.register`, `AutoTokenizer.register`, `AutoProcessor.register` with `exist_ok`. Unknown auto-class names raise. (Gives kimi-style in-tree modeling code a sanctioned registration path instead of dynamic-module interception.)
- `model_class_attrs` → plain `setattr` overrides for the class-level knobs that have no registration API: `_no_split_modules`, `_keep_in_fp32_modules(_strict)`, `_tied_weights_keys`, `supports_gradient_checkpointing`, and config-side `base_model_tp_plan` / `base_model_ep_plan` / `base_model_fsdp_plan` class vars.

## Per-instance loss strategy

`ModelStrategyOverrides(loss_function=provider)` sets `model.loss_function` (transformers' per-instance override, honored by its loss dispatch) right after base-model build — a per-architecture loss no longer requires mutating the global `LOSS_MAPPING`.

## `BEFORE_SAVE` hook phase

New lifecycle phase dispatched right before the trained model is saved (`save_trained_model`, which now receives tokenizer/processor) and before the legacy merge-lora path writes the merged model. Use it to undo or finalize model state for serialization. The memory-efficient merge never materializes a model instance and does not dispatch it (documented). No legacy-method equivalent.

## Processor-first hooks

Per the review discussion on #3869: hook docs now steer consumers to `context.processor` (transformers v5 direction), with `context.tokenizer` retained for text-only runs and compatibility; processor is threaded through every dispatch site including the new save phase.

Model migrations (kimi_linear, paddleocr_vl, nemotron_h branch, NVFP4 converters) are deliberately deferred until this API settles.

## How has this been tested?

- `tests/test_model_support_registrations.py` grown to cover every new seam: interface registration (idempotent re-apply), quantizer no-op/override-with-warning semantics, auto-class wiring incl. unknown-name error, `model_class_attrs` application, per-instance loss strategy, and `BEFORE_SAVE` dispatch from `save_trained_model`.
- Full model-support suite (84 tests) plus the loader/normalize-config/processing-strategy CPU tests, rebased on `main` at the #3884 merge (including its review tweaks: staticmethod warn helper, `capture_axolotl_warnings`).

## Types of changes

- [x] Documentation update

## Summary by CodeRabbit

* **New Features**
  * Added a pre-save lifecycle hook for trained and merged models.
  * Expanded model support to configure attention, masking, expert functions, quantizers, automatic class mappings, model attributes, and per-model loss functions.
  * Added public APIs for declaring these model integrations.
* **Documentation**
  * Clarified hook inheritance, save lifecycle behavior, registry integrations, and loss-function configuration.
* **Bug Fixes**
  * Ensured pre-save hooks receive the finalized model and are not dispatched redundantly.

### DIFF
--- a/docs/agents/new_model_support.md
+++ b/docs/agents/new_model_support.md
@@ -107,11 +107,11 @@ Move one concern at a time using this mapping:
 | `pre_config_load()` / `pre_tokenizer_load()` / `pre_model_load()` | `BEFORE_CONFIG_LOAD` / `BEFORE_TOKENIZER_LOAD` / `BEFORE_MODEL_BUILD` hook |
 | `post_model_load()` | `AFTER_ADAPTER_LOAD` hook |
 
-Unless a profile phase explicitly replaces its family hooks, hooks from the family, profile, and legacy method are additive. Remove a legacy method once its behavior moves into a profile hook, or the old and new implementations both run. The compatibility guard prevents a legacy method's `super()` call from re-entering the declarative hook, but it cannot identify duplicate logic implemented in both places. `AFTER_BASE_MODEL_BUILD` has no legacy-method equivalent.
+Unless a profile phase explicitly replaces its family hooks, hooks from the family, profile, and legacy method are additive. Remove a legacy method once its behavior moves into a profile hook, or the old and new implementations both run. The compatibility guard prevents a legacy method's `super()` call from re-entering the declarative hook, but it cannot identify duplicate logic implemented in both places. `AFTER_BASE_MODEL_BUILD` and `BEFORE_SAVE` have no legacy-method equivalent.
 
 ### Hook phases
 
-Hooks receive an immutable `ModelHookContext` containing the run config and the objects available at that phase. The `tokenizer` and `processor` fields are optional because early phases do not construct them and direct `ModelLoader` callers can omit a processor; hooks must handle `None` unless their phase and caller guarantee the object. Register hooks under `ModelHookPhase` rather than relying on an implied ordering.
+Hooks receive an immutable `ModelHookContext` containing the run config and the objects available at that phase. The `tokenizer` and `processor` fields are optional because early phases do not construct them and direct `ModelLoader` callers can omit a processor; hooks must handle `None` unless their phase and caller guarantee the object. Prefer consuming `context.processor` when it is set — transformers v5 is converging on the processor as the primary preprocessing object — and fall back to `context.tokenizer`; the tokenizer field remains for text-only runs and compatibility. Register hooks under `ModelHookPhase` rather than relying on an implied ordering.
 
 | Phase | Meaning |
 |-------|---------|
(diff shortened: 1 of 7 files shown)


ErenAta16 (commented):
The sentinel design is the right one and worth saying explicitly: `_INHERIT_STRATEGY` as the default rather than `None` is what makes "omitted inherits, explicit `None` clears" expressible at all. A plain `None` default cannot distinguish the two, and this is the kind of contract that gets quietly broken later by someone "simplifying" the default. The docstrings state it on both override classes, which should keep that from happening.

One gap in the field-generic claim.

## `with_overrides` is generic on one side only

```python
for registry_field in fields(self)        # ModelRegistrations
    ... getattr(overrides, registry_field.name)
```

It iterates the fields of `ModelRegistrations` and reads them off `ModelRegistrationOverrides`. So the description's

> `with_overrides` is now field-generic so future registries are one-line additions

is one line in `ModelRegistrations` and **a second line in `ModelRegistrationOverrides`**. Miss the second and it doesn't fail at import or at construction — it fails when someone actually layers an override:

```
Regs      fields: ['a', 'b', 'c']
Overrides fields: ['a', 'b']
Regs(a=1,b=2,c=3).with_overrides(Overrides(a=9))
  -> AttributeError: 'Overrides' object has no attribute 'c'
```

Both pairs are in sync today — I checked, `ModelRegistrations`/`ModelRegistrationOverrides` are 8/8 and `ModelStrategies`/`ModelStrategyOverrides` are 3/3 — so this is about the next registry rather than this PR.

## Nothing currently guards it

`tests/test_model_support_registrations.py` is 495 lines and 11 tests, and none of them compare the two field sets:

```
occurrences of "fields(" in the test file : 0
a sync / parity / mirror test             : none
```

Given the PR's stated goal is that "the model-support surface stabilizes before any model migrations", the cheapest thing that makes the next addition safe is one assertion per pair:

```python
def test_overrides_mirror_registrations():
    assert {f.name for f in fields(ModelRegistrations)} == {
        f.name for f in fields(ModelRegistrationOverrides)
    }
```

That turns a runtime `AttributeError` on whichever model first uses the new registry into a red test at the moment the field is added. Six new seams landed here; the seventh is where this bites.

## Smaller

The `ModelRegistrations` docstring carries two real constraints that only exist in prose:

> A patch mapping that changes a module's checkpoint layout must ship matching weight conversions, or loading and saving break.

> Every attention implementation needs a matching mask entry.

Both are pairwise invariants between fields that are declared independently. The attention one in particular looks checkable — if `attention_functions` returns a dict of implementation names, asserting the same key set appears in `attention_mask_functions` at the `BEFORE_MODEL_BUILD` boundary would catch a half-registered attention implementation at registration time rather than at the first forward pass. Declaring them "as a pair" in the docstring is a good start; enforcing it is what stops the pairing from drifting.

Not asking for either in this PR — the seam expansion is coherent on its own, and both notes are about what makes the next model migration land cleanly, which is what this is groundwork for.

PR #3891, opened 2026-07-28
feat(model_support): migrate nemotron_h and kimi_linear onto the profile seams

winglian:
# Description

> **Stacked on #3889** — the first three commits are that PR; only the last commit (`feat(model_support): migrate nemotron_h and kimi_linear onto the profile seams`) is new here. Opened early so CI can run and the migration shape is visible; will be rebased once #3889 merges.

First migration wave onto the profile seams, proving the surface on real architectures before wider migration.

## nemotron_h → builtin descriptor

New `model_support/nemotron_h/` descriptor replaces both `model_type` branches in `PatchManager`:

- The cfg-conditional SSM packing patch (`sample_packing` / context parallelism) plus the `supports_gradient_checkpointing` enable moves to a `BEFORE_MODEL_BUILD` hook.
- The conversion-mapping fix (strip the spurious `embedding→embeddings` `WeightRenaming` that breaks merged saves) moves to `AFTER_BASE_MODEL_BUILD` — the phase exists precisely because `NemotronHForCausalLM.__init__` re-registers its mapping with `overwrite=True`.

`falcon_h1` / `granitemoehybrid` have the same SSM-hybrid shape and can follow the identical pattern in a later pass.

## kimi_linear → `auto_classes`

The in-tree `KimiLinearConfig` / `KimiLinearForCausalLM` / `TikTokenTokenizer` now register through the `auto_classes` seam, so kimi runs work **without `trust_remote_code`**. The stem-anchored dynamic-module redirect stays for configs that still set it (remote resolution bypasses the registered classes in that case).

Two real pre-existing breaks surfaced by the new test and fixed here: the vendored kimi modules did not import under the pinned transformers 5.x at all (`OutputRecorder` moved to `utils.output_capturing`, `bytes_to_unicode` moved out of the gpt2 tokenizer module); compat imports added.

## Enabling change: registrations apply at descriptor discovery

`auto_classes` must exist before `AutoConfig.from_pretrained` resolves, which happens well before the pre-model-build slot. Registrations are now applied as soon as the run's descriptor is discovered: at the pre-config boundary (matcher or exact lookup) and again by exact `model_type` before model build. They are idempotent by contract, so the double application is safe.

## Mixed quantization (fp8 attention + nvfp4 experts)

Checked whether the seam surface covers modelopt-style MIXED_PRECISION checkpoints: it does, as a composition — `quantizers` registers the mixed-precision quantizer plus its `QuantizationConfigMixin` (per-module routing lives inside the quantizer via `param_needs_quantization`), `weight_conversions` carries the expert fuse/unfuse transforms, and `HfQuantizer.update_weight_conversions` composes the two inside transformers' loading pipeline. One deliberate caveat, documented: transformers keys quantizer resolution on the *method* name globally (`AUTO_QUANTIZER_MAPPING`), so a profile's quantizer override is scoped per-run-by-discovery, not per-architecture — a multi-model process loading two different fp8 models after a profile override sees the overridden quantizer for both. The in-tree NVFP4 quantizer swap is **not** migrated here to avoid colliding with in-flight NVFP4 branches; it is the intended first user of this composition.

## Deliberately not migrated

- NVFP4/fp8 quantizer swap (above) and the chunked-CE `LOSS_MAPPING` patch — the latter is feature-scoped (config flag), not architecture-scoped, so the per-arch `loss_function` strategy is the wrong home for it.
- paddleocr_vl is already fully declarative; no further change needed.

## How has this been tested?

- New tests: nemotron_h descriptor resolution + both lifecycle hooks, the conversion-mapping fix against the real transformers registry (restoring state), the packing hook's no-op path, kimi `auto_classes` provider wiring (guarded by `importorskip` on `fla`/`tiktoken`), and pre-config registration application.
- Full model-support suite (89 tests) plus loader/normalize-config/processing-strategy CPU suites (unchanged vs baseline), torch 2.x / transformers 5.14.1.

## Types of changes

- [x] Documentation update

## Summary by CodeRabbit

* **New Features**
  * Added a pre-save model-support lifecycle step for trained and merged models.
  * Expanded model integration support for attention functions, quantizers, auto classes, expert functions, and model attributes.
  * Added support for Nemotron-H models and improved Kimi-Linear loading compatibility.
  * Added configurable per-model loss-function integration.

* **Bug Fixes**
  * Improved model registration timing and compatibility across supported Transformers versions.
  * Prevented unintended parameter renames during Nemotron-H adapter merging.

* **Documentation**
  * Updated model-support lifecycle, registration, and hook guidance.


ErenAta16 (commented):
Reviewed this as the first real user of the seams from #3889, since a migration is where a surface design either holds or shows its edges. It mostly holds — and the one place it doesn't is documented honestly in a comment rather than papered over, which I'd rather see.

## The first migration already has to route around the `weight_conversions` seam

`nemotron_h` doesn't register its conversions through the seam. It fixes them in an `AFTER_BASE_MODEL_BUILD` hook with a direct call:

```python
def _after_base_model_build(_context: ModelHookContext) -> None:
    # Must run after model build because NemotronHForCausalLM.__init__
    # calls register_nemotron_h_conversion_mapping() with overwrite=True,
    # which would clobber any earlier fix.
    fix_nemotron_h_conversion_mapping()
```

The reason is real and the comment states it plainly. But it means the seam's contract — registrations applied at the `BEFORE_MODEL_BUILD` boundary — doesn't survive a model whose `__init__` re-registers its own conversions with `overwrite=True`, and the very first architecture migrated is one of those.

Two things I'd want out of that:

**Say it in the seam's own docs.** `ModelRegistrations`' docstring describes `weight_conversions` without noting that a self-registering model will clobber it. The next person migrating an architecture will reach for the seam, get silently overwritten during build, and have to rediscover this. One sentence in `profile.py` pointing at `nemotron_h` as the worked example would save that.

**Is nemotron_h the only one?** `register_nemotron_h_conversion_mapping()` being called from `__init__` looks like a transformers-side pattern rather than a one-off. If other hybrid models do the same, the `AFTER_BASE_MODEL_BUILD` workaround becomes the norm and the seam covers fewer models than it appears to.

## A collision waiting for #3893

`fix_nemotron_h_conversion_mapping` calls `register_checkpoint_conversion_mapping(..., overwrite=True)` directly. That's fine today — `patch_manager.py:225` does the same, and `axolotl/utils/weight_conversions.py` isn't on this branch.

Once #3893 lands, though, the seam moves to `register_weight_conversions`, which merges with what's already registered and tracks merge-mode registrations in `_MERGE_REGISTERED_MODEL_TYPES` to refuse an unsafe replace. This post-build fix would still be a raw overwrite, invisible to that bookkeeping: a profile registering merge-mode conversions for `nemotron_h` would have them dropped here, and the guard #3893 adds cannot fire because this path never reaches it.

Not a change to make now — the two stacks land independently — but worth a note on whichever merges second, because the interaction is silent in exactly the way #3893's guard exists to prevent.

## The filter is exact-match with no negative signal

```python
if not (
    isinstance(entry, WeightRenaming)
    and entry.source_patterns == ["embedding.weight"]
    and entry.target_patterns == ["embeddings.weight"]
)
```

and the log only fires on success:

```python
if len(filtered) != len(mapping):
    register_checkpoint_conversion_mapping(...)
    LOG.info("Removed embedding→embeddings WeightRenaming ...")
```

If upstream renames the pattern, wraps it differently, or ships it as something other than a bare `WeightRenaming`, the filter matches nothing, nothing is logged, and the silent `backbone.embeddings.weight` → `backbone.embedding.weight` rename on LoRA merge comes back. Given the whole point is that the corruption is silent, an `else: LOG.warning(...)` when the expected entry isn't found would turn an upstream change into a visible signal instead of a returning bug.

## Smaller

`_needs_ssm_hybrid_patch` gates on `sample_packing or context_parallel_size > 1`, and `supports_gradient_checkpointing = True` is set only inside that branch. So a run with neither keeps upstream's `False` — correct, since the GC-safe block forward isn't installed. Worth a word in the comment that the flag intentionally tracks the patch rather than the model, because as written it reads like the patch enables a capability the model always had.

PR #3893, opened 2026-07-29
fix: merge builtin checkpoint conversions

NanoCode012:
# Description

Followup to #3889 , this ensures that conversions are not overrode when we add a future model_type conversion. It does read-merge-write now.

Also, this puts a future notice for potential conflict between (nvfp4 conversion <=> custom model_type conversion), which we should fix when it happens / user reported.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved model weight-conversion registration for more reliable checkpoint loading.
  * Preserved fallback conversion rules while allowing specialized model conversions to take precedence.
  * Prevented conflicting conversion registrations from silently replacing previously merged rules.

* **Compatibility**
  * Improved support for NVFP4 expert and other model-specific weight conversions.

### DIFF
--- a/src/axolotl/integrations/kernels/libs/scattermoe_lora/nvfp4_weight_converter.py
+++ b/src/axolotl/integrations/kernels/libs/scattermoe_lora/nvfp4_weight_converter.py
@@ -485,18 +485,16 @@ def register_nvfp4_expert_converters(
     detected index layout). The only per-model knob is which ``model_type`` the loader looks the
     mapping up under. Safe to call repeatedly (idempotent via overwrite on re-entry).
     """
-    from transformers.conversion_mapping import register_checkpoint_conversion_mapping
+    from axolotl.utils.weight_conversions import register_weight_conversions
 
     converters = (nvfp4_experts_weight_converters() if include_routed else []) + list(
         extra or []
     )
     if not converters:
         return
-    try:
-        register_checkpoint_conversion_mapping(model_type, converters)
-    except ValueError:
-        # Already registered; overwrite to keep converters fresh.
-        register_checkpoint_conversion_mapping(model_type, converters, overwrite=True)
+    # replace_existing: the direct-expert-load fast path (include_routed=False) relies on
+    # the stock fp16 fusion converters being absent, so these must be the only converters.
+    register_weight_conversions(model_type, converters, replace_existing=True)
 
     LOG.info(
         "Registered %s NVFP4 WeightConverters (%d) in transformers conversion_mapping",
--- a/src/axolotl/loaders/patch_manager.py
+++ b/src/axolotl/loaders/patch_manager.py
@@ -213,13 +213,11 @@ class PatchManager:
             conversions_provider() if conversions_provider is not None else None
         )
         if conversions:
-            from transformers.conversion_mapping import (
-                register_checkpoint_conversion_mapping,
-            )
+            from axolotl.utils.weight_conversions import register_weight_conversions
 
             for key, entries in conversions.items():
                 entries = list(entries)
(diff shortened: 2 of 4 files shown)


ErenAta16 (commented):
Read-merge-write is the right shape, and the docstrings do a lot of work here — the note about `WeightRenaming`s chaining in list order while `WeightConverter`s take first-match is exactly the trap this could have walked into. I simulated the registry with transformers' actual semantics to check the guard, and it holds in one direction only.

## The `NotImplementedError` guard is asymmetric, and the unguarded order does the thing it was written to prevent

`_MERGE_REGISTERED_MODEL_TYPES` is consulted by the `replace_existing` branch, but `merge_weight_conversions` never checks whether a replace-mode registration already happened. So:

```
merge(profile) then replace(NVFP4)   ->  NotImplementedError   (guard fires)
replace(NVFP4) then merge(profile)   ->  registry = [PROFILE]
```

In the second order the NVFP4 entry is not merely reordered behind the profile — it is **gone**. `merge_weight_conversions` drops any existing entry whose `source_patterns` tuple matches an incoming one, and two converters claiming the same expert keys will have the same source patterns by construction. So the merge silently removes the replace-mode consumer's converters.

That's the same harm the error message describes, with the roles swapped:

> replacing would silently drop them

and no error is raised. If the source patterns differ rather than match exactly, it degrades to an ordering problem instead: the profile leads, wins first-match, and the NVFP4 fast path is shadowed by converters it was specifically registered to exclude.

Whether the reverse order is reachable depends on registration sequencing I can't see from outside — `_apply_model_support_registrations` versus when `register_nvfp4_expert_converters` runs. If it isn't reachable today, the guard is still one-sided in a way that will bite whenever that ordering changes, and the cheap fix is a symmetric `_REPLACE_REGISTERED_MODEL_TYPES` checked from the merge branch.

## The guard is process-global and outlives the registry

`_MERGE_REGISTERED_MODEL_TYPES` is a module-level set with no reset. Clearing or overwriting the conversion mapping does not clear it:

```
register("llama", profile)            # merge mode
<registry cleared / model discarded>
register("llama", nvfp4, replace_existing=True)
    -> NotImplementedError, with the registry empty
```

For a single training run that's fine. It matters for anything loading more than one model in a process — sweeps, multi-model serving, and the test suite, where test ordering starts deciding whether a `replace_existing` call raises. A module-level `_reset_weight_conversion_state()` for tests to call in a fixture would cover it without changing behaviour.

## The merged registry holds copies of transformers' built-ins

`get_checkpoint_conversion_mapping` returns `deepcopy(...)` of the cached mapping. So the fallbacks that survive a merge are copies, and every merge-mode registration replaces them with fresh copies — including their `operations`:

```
after 1st merge registration:  original converter identity preserved 0/3, op identity 0/3
after 5th:                     0/3, 0/3   (list length stable at 4 throughout)
```

The list does **not** grow, so the dedup is doing its job and there's no leak — I'd checked for that first and it's clean. What's left is that the registry no longer contains transformers' own objects, and `ConversionOps` instances get deep-copied on each registration. Harmless if ops are pure descriptions of a transform; worth a second look if any op holds a buffer, a compiled artefact, or anything else that shouldn't be duplicated per registration.

## Smaller

`_source_key` falls back to `()` for an entry without `source_patterns`, which would make all such entries collide on one key. Both `WeightConverter` and `WeightRenaming` inherit `source_patterns` from `WeightTransform`, so nothing in transformers hits that today — but the `getattr(..., ()) or ()` reads like it's defending against something, and if that something ever arrives, one incoming entry would evict every existing entry that also lacks the attribute. Either drop the defensive default so it fails loudly, or key unknown shapes on `id(entry)` so they never collide.

PR #3896, opened 2026-07-29
chore(liger): bump liger-kernel to 0.8.1 — native gemma4, kernel-impl selector, DPO loss guards

thad0ctor:
# Description

Bumps `liger-kernel` to `0.8.1` and picks up what the release offers:

- **Drops the `gemma4` hand-patch elif.** 0.8.1 ships native `apply_liger_kernel_to_gemma4` (linkedin/Liger-Kernel#1203), which covers everything the hand-patch did *plus* the multimodal FLCE forward it could only warn about — the elif's own comment marked it as a stopgap for exactly this release. gemma4 now routes through the generic native dispatch like gemma4_text already does. Native `deepseek_v4` also comes along for free.
- **New `liger_kernel_impl` option** exposing 0.8.1's opt-in kernel backends (`cutile` / `cutedsl`, the latter tuned for Blackwell). liger reads `LIGER_KERNEL_IMPL` exactly once at first import, so the plugin sets it from the `register` hook (before the pre-model-load patches that import liger), with a hard error if something already imported liger on a different backend. FLCE stays Triton on both backends — noted in the docs.
- **DPO guards**: 0.8.1's new fused-DPO loss types (`hinge`, `bco_pair`, `robust`, `exo_pair`, `discopop`) already flow through `dpo_use_liger_kernel` + `dpo_loss_type` untouched, but `ipo` as the active loss now errors at config time (liger raises `NotImplementedError` mid-run otherwise) and multiple loss types warn that only the first reaches liger.

Also fixes a small gap along the way: `axolotl.utils.config.prepare_plugins` never called `plugin.register(cfg)` like the CLI version does — it does now.

## Motivation and Context

Keeps the pin current and the gemma4 path on the (better) native kernels. Upstream-sync audit against 0.8.1 is clean: no shadowed hand-patches, no stale override entries, no signature drift.

## How has this been tested?

- CPU suite: 3217 passed (only reds were environment noise on my box — port conflicts from an unrelated training run and parallel-run timing flakes; all pass standalone).
- GPU (RTX 5090): the three `tests/e2e/integrations/test_liger.py` training runs pass; gemma4 dispatch check confirms the native multimodal FLCE forward is installed with finite fwd/bwd on a tiny model; `cutedsl` validated end to end through the plugin flow (RMSNorm bit-exact, CE matches torch to ≤1e-6 up to V=262k, e2e train passes with the backend active); all five new DPO loss types run fwd/bwd with finite grads.
- New tests cover the backend selector (env timing, invalid values deferring to schema validation) and the DPO guards (ipo rejection, multi-loss warning, set coercion).

## AI Usage Disclaimer

Yes — Fable/Opus throughout

N/A

## Types of changes

- [x] Test coverage added

## Summary by CodeRabbit

* **New Features**
  * Added support for selecting Liger Kernel backends, including `cutile`, `cutedsl`, and `ascend`.
  * Added configuration support for running DPO loss with Liger’s fused implementation.
  * Added validation for supported backends and compatible DPO loss types.

* **Documentation**
  * Expanded Liger configuration, supported-model, DPO, and compatibility documentation.
  * Clarified backend coverage and Triton usage for fused linear cross-entropy.

* **Bug Fixes**
  * Updated Liger Kernel to version 0.8.1.
  * Prevented backend changes after Liger has already been loaded.


src/axolotl/integrations/liger/args.py
```diff
@@ -16,12 +16,16 @@
...
 
+from typing import Literal
+
 from pydantic import BaseModel, Field, model_validator
 
 from axolotl.utils.logging import get_logger
 
 LOG = get_logger(__name__)
 
+LIGER_KERNEL_IMPLS = ("cutile", "cutedsl", "ascend")
```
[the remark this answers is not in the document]
thad0ctor: Checked this against the liger v0.8.1 source before changing anything — `ascend` **is** a registered `LIGER_KERNEL_IMPL` value: `ops/backends/_ascend/__init__.py` calls `register_impl(ImplInfo(name="ascend", devices=("npu",), default_devices=("npu",), ...))`, and `select_impl` accepts it explicitly from the env var (raising a clear device-mismatch error on non-NPU). So it's not a separate backend path, and removing it would block a legitimate NPU value (the upstream source even anticipates future explicit ascend variants). Runtime device/dependency fitness is the same story for `cutedsl` without `nvidia-cutlass-dsl` — liger owns those errors and they're actionable. Kept the value; clarified the field description in 36bbca1ae to note ascend is NPU-only and auto-selected there.

src/axolotl/integrations/liger/args.py
```diff
@@ -62,6 +66,21 @@ class LigerArgs(BaseModel):
             )
         },
     )
+    liger_kernel_impl: Literal["cutile", "cutedsl", "ascend"] | None = Field(
```
NanoCode012: Ooh, interesting that they got ascend. Is this compatible across all their patches?

src/axolotl/integrations/liger/plugin.py
```diff
@@ -11,6 +11,13 @@
 
 LOG = get_logger(__name__)
 
+_KERNEL_IMPL_ENV = "LIGER_KERNEL_IMPL"
+# env value before this plugin's last owned write, and what it last wrote — lets
+# a config that omits liger_kernel_impl undo a previous config's mutation (e.g.
+# one rejected by validation after register() already ran)
+_env_before_write: str | None = None
+_last_written: str | None = None
```
NanoCode012, line 19: What's this store for? I see the changes below but not really sure why

src/axolotl/integrations/liger/plugin.py
```diff
@@ -227,57 +309,6 @@ def patched_init(self, *args, **kwargs):
...
-                    "Liger fused linear cross entropy for multimodal gemma4 is not in "
-                    "liger-kernel 0.8.0 (added upstream in PR #1203, post-0.8.0). Skipping; "
-                    "the gemma4_text language model gets FLCE via liger's native path."
-                )
-            LOG.info(
-                f"Applied Liger kernels for gemma4: "
-                f"rms_norm={cfg.liger_rms_norm}, glu={cfg.liger_glu_activation}, "
-                f"rope=False (incompatible), layer_norm={cfg.liger_layer_norm}"
-            )
         elif cfg.model_config_type in ("gemma4_unified", "gemma4_unified_text"):
```
NanoCode012, line 333: I assume they didn't add support for the unified models yet?

NanoCode012 (commented):
Good additions. From a quick read, this seems quite large for just a liger kernel bump though?

src/axolotl/integrations/liger/plugin.py
thad0ctor: Correct, MODEL_TYPE_TO_APPLY_LIGER_FN in 0.8.1 has gemma4_text and gemma4 but nothing for gemma4_unified

src/axolotl/integrations/liger/plugin.py
thad0ctor: Env-var ownership across repeated `load_cfg()` calls in the same process. liger reads `LIGER_KERNEL_IMPL` once at first import, so the plugin has to set it in `register()` — which means it also needs to be able to take it back:

- config A sets `cutedsl`, config B omits it → B undoes A's write instead of silently inheriting it
- config A gets rejected by validation after `register()` already ran → `on_config_validation_error` undoes it
- you exported `LIGER_KERNEL_IMPL` yourself → we put *your* value back rather than deleting it, and we back off entirely if something else overwrites ours

That last case is why it's two variables and not one: `_last_written` is how we notice a foreign overwrite, `_env_before_write` is what to restore.

Behavior's covered by `TestKernelImplEnv::test_omitted_option_undoes_previous_config_write`, `test_omitted_option_keeps_foreign_env_value`, and `test_foreign_overwrite_relinquishes_ownership`.

src/axolotl/integrations/liger/args.py
thad0ctor: Short answer: yes. Picking a backend doesn't replace liger's kernels wholesale — it only swaps in the specific ops that backend implements, and anything it doesn't implement keeps using the normal Triton kernel. Liger's model patches all pull their ops from `liger_kernel.ops`, which is exactly what gets swapped, so the choice reaches every patched path and nothing ends up half-patched. A backend can only override or add, never remove. That's also why FLCE stays on Triton with `cutile`/`cutedsl` — neither implements it.

On `ascend`: it's the Huawei NPU backend, and it only runs on NPU. Liger registers it as supporting NPU and nothing else, *and* as the automatic default there — so on an NPU box it's already active whether or not you set anything, and on an NVIDIA box liger just errors out (`select_impl()` raises "supports devices (npu), but the current device is 'cuda'"). Setting it can never do anything useful.

So I dropped it from the `Literal`. To be clear, that doesn't remove NPU support — NPU users still get ascend automatically from liger, exactly as before. It only removes a config value that was either a no-op or a crash depending on your hardware. `cutile` and `cutedsl` are the two that actually change behavior.

thad0ctor:
> Good additions. From a quick read, this seems quite large for just a liger kernel bump though?

@winglian 

Yes, I added some of the extra functionality from the latest release (cutedsl kernels) , associated tests doc changes to reflect those features, config flags and validation mostly

docs/rlhf.qmd
```diff
@@ -51,6 +51,8 @@ datasets:
     type: chatml
 ```
 
+To run the DPO loss through liger-kernel's fused (chunked) implementation, set `dpo_use_liger_kernel: true`. The fused path supports `dpo_loss_type` values `sigmoid` (default), `hinge`, `exo_pair`, `nca_pair`, `robust`, `bco_pair`, `sppo_hard`, `apo_zero`, `apo_down`, and `discopop` (liger-kernel >= 0.8.1); `ipo` is not supported on the fused path, and only the first entry of `dpo_loss_type` is used.
```
NanoCode012, line 54: more of a user validation: if we set liger FLCE but forgot to set `dpo_use_liger_kernel`, what would be the expected behavior? Should we do a `if rl is dpo and liger_flce is true and not dpo_use_liger_kernel: raise error`?

NanoCode012: For this segment, also good to add it inside a `note codeblock`

src/axolotl/cli/config.py
```diff
@@ -308,11 +308,20 @@ def load_cfg(
...
+            env_capabilities=env_capabilities,
+        )
+    except Exception:
+        # a rejected config must not leave register()-time side effects (e.g.
+        # LIGER_KERNEL_IMPL) behind for the next config in this process
+        if cfg.get("plugins"):
+            from axolotl.integrations.base import PluginManager
+
+            PluginManager.get_instance().on_config_validation_error(cfg)
+        raise
```
NanoCode012, line 324: Do you have flow of how this can be hit?

src/axolotl/integrations/liger/plugin.py
```diff
@@ -12,6 +12,13 @@
 
 LOG = get_logger(__name__)
 
+_KERNEL_IMPL_ENV = "LIGER_KERNEL_IMPL"
```
NanoCode012, line 15: Any reason we need to use env specifically and not regular var?

Also, would you need to check for unlikely chance of potential conflict (set by user)?

pyproject.toml
```diff
@@ -81,7 +81,7 @@ dependencies = [
     "bitsandbytes==0.50.2 ; sys_platform != 'darwin'",
     "triton>=3.4.0 ; sys_platform != 'darwin'",
     "xformers>=0.0.33.post2 ; sys_platform != 'darwin' and platform_machine != 'aarch64'",
-    "liger-kernel==0.8.0 ; sys_platform != 'darwin'",
+    "liger-kernel==0.8.1 ; sys_platform != 'darwin'",
```
NanoCode012, line 84:
```suggestion
    "liger-kernel==0.8.2 ; sys_platform != 'darwin'",
```
> Liger-Kernel v0.8.2 delivers performance, correctness, and numerical-precision improvements across cross entropy, MoE, RMSNorm, JSD, and other operators. The release also improves Megatron integration, testing, and overall kernel robustness.
> This release further expands experimental backends and features, including CuTe DSL, cuTile, and LigerCute, with new kernels and optimizations targeting Hopper and Blackwell GPUs.

NanoCode012:
just minor nits , rest looks good

NanoCode012 (approved):

Issue #2975, opened 2025-07-24
Loading Large Long-Context SFT Dataset leads to RAM OOM


PranjalChitale:
### Please check that this issue hasn't been reported before.

- [x] I searched previous [Bug Reports](https://github.com/axolotl-ai-cloud/axolotl/labels/bug) didn't find any similar reports.

### Expected Behavior

The dataset is loaded appropriately and training commences.

### Current behaviour

``[2025-07-24 10:42:41,253] [INFO] [axolotl.utils.data.shared.load_preprocessed_dataset:470] [PID:126508] [RANK:0] Loading prepared dataset from disk at ../[path]``

The dataset loading freezes here, and if we monitor htop, the RAM usage shoots upto almost 1.3T, causing RAM OOMs, resulting in a complete system failure.

### Steps to reproduce

Dataset used : [Dataset](https://huggingface.co/datasets/VarunGumma/updesh-beta-merged)
Size : ~9.5M samples

The dataset is already preprocessed. tokenized and saved to disk and consumes around 552GB on the disk.
The context length of the dataset ranges from ~1K to ~65K tokens.

The training is carried out on 8 nodes each containing 8 MI300X GPUs (Total 64 GPUs with 192 GB VRAM each).
Each machine has ~1.5T of RAM.

### Config yaml

```yaml
datasets:
- path: data.parquet
  type: chat_template
  field_messages: messages
  message_property_mappings:
    role: role
    content: value
train_on_eos: turn
train_on_eot: turn
roles_to_train:
- assistant
dataset_processes: 88
dataset_keep_in_memory: false
dataset_prefetch_factor: 2
dataset_num_workers: 88
dataset_prepared_path: /scratch/last_run_prepared
num_dataset_shards_to_save: 10

sequence_len: 65536
sample_packing: true
pad_to_sequence_len: true
include_tokens_per_second: true
gradient_accumulation_steps: 16
micro_batch_size: 1
num_epochs: 3
group_by_length: false
```

### Which Operating Systems are you using?

- [x] Linux
- [ ] macOS
- [ ] Windows

### Python Version

3.12.10

### axolotl branch-commit

main/9f2bb188a4e21b72802807164f451a311e4d581b

### Acknowledgements

- [x] My issue title is concise, descriptive, and in title casing.
- [x] I have searched the existing issues to make sure this bug has not been reported yet.
- [x] I am using the latest version of axolotl.
[...]

PR #3898, opened 2026-07-29
fix(multimodal): stop pixel_values RAM blowup during pre-tokenization

thad0ctor:
# Description

Pre-tokenizing image datasets exhausts host RAM (`Tokenizing Prompts` gets OOM-killed) and produces doubled Arrow caches. Measured trigger: Qwen3-VL page-scale scans (1700x2200 -> 14,628 patches x 1536 float32 ≈ 90 MB of `pixel_values` per row) on a 377 GB / 48-core machine, dying at `dataset_num_proc: 8` and even `2`. Three compounding causes, fixed together:

1. **`build_prompt` converted `pixel_values` tensors to Python lists.** `.tolist()` boxes every float (~8x the tensor's memory: 90 MB -> 720 MB/row measured) and Arrow then infers float64, doubling the on-disk cache at precision the data never had. The conversion also dominates tokenization time. Now kept as the processor's numpy float32 array; rows read back from the Arrow cache are unchanged in structure (nested lists), just float32.
2. **Unbounded map buffers.** The tokenization map hardcoded `batch_size=1000` and left `writer_batch_size` at the HF default of 1000, so each worker buffered up to 2000 rows of pixel data (~186 GB/worker at page scale, pre-fix). Both are now configurable (`dataset_map_batch_size` / `dataset_writer_batch_size`) and default to 32 for multimodal runs; text-only runs keep today's defaults exactly. The limits apply across the whole prepared pipeline — tokenization, `add_position_ids`/`add_length`/PoSE maps, zero-trainable-token filters, truncation, and the `TokenizedChatDataset` path — via a shared `dataset_map_buffer_kwargs` helper.
3. **Mixed image+text datasets crashed the batched tokenizer.** Rows without images emit no `pixel_values`, so a mixed dataset died with `pyarrow.lib.ArrowInvalid` (ragged columns within a batch) — or, with single-modality batches, with `ValueError: Schema and number of arrays unequal` / `KeyError: 'pixel_values'` because the Arrow writer locks its schema on the first batch. Missing keys are now padded with `None` within a batch, and mixed datasets are detected up front via an Arrow-level presence check (no image decode), tokenized as two homogeneous maps, concatenated, and restored to the original row order.

**Measured effect** (microbench on the exact Qwen3-VL page shape): 720 -> 90 MB per row in the map worker, ~79x faster map throughput (1.3 -> 103 examples/s at 1/10 scale), half the cache size.

## Example usage

Nothing required — multimodal runs get the safe defaults automatically. To tune:

```yaml
processor_type: AutoProcessor
skip_prepare_dataset: false   # pre-tokenize (e.g. for multimodal sample packing)

# rows per batched tokenization call / rows buffered before each Arrow flush.
# defaults: 32 for multimodal, 1000 otherwise. Lower -> less RAM, more flushes.
dataset_map_batch_size: 32
dataset_writer_batch_size: 32

dataset_num_proc: 32
```

Peak preprocessing RAM ≈ `dataset_num_proc x (map_batch + writer_batch) x bytes_per_row`. For 90 MB/row page scans:

| workers | buffers | approx peak |
|---|---|---|
| 32 | 32 + 32 | ~185 GB |
| 32 | 8 + 8 | ~46 GB |
| 8 | 1000 + 1000 (old default, pre-fix rows) | ~11.5 TB (OOM) |

Docs: new "Controlling preprocessing RAM" section in `docs/dataset_preprocessing.qmd`; `examples/qwen2-vl/lora-7b.yaml` and `examples/qwen2_5-vl/lora-7b.yaml` carry commented knobs.

## Smoke tests (real processors, not stubs)

- **Mixed-modality pre-tokenization (Qwen/Qwen2-VL-2B-Instruct processor, `num_proc=2`)**: 12-row dataset interleaving 1700x2200 page scans with text-only rows through `TokenizedPromptDataset`. Row order preserved exactly, Arrow cache stores `List(List(Value('float32')))`, page rows are 19,276 x 1176 patches = 90.7 MB float32 (matches the report's math), text rows carry `pixel_values: None`. The same dataset crashes with `ValueError: Schema and number of arrays unequal` without this PR.
- **Text regression (SmolLM2-135M-Instruct, `axolotl preprocess` CLI, this branch vs upstream main)**: 24-row chat dataset — `input_ids`/`labels`/`attention_mask` contents are identical (row order differs only via the pre-existing unseeded merge shuffle, which is nondeterministic on main too).
- **CLI preprocess (`axolotl preprocess`) of the mixed dataset**: completes on this branch, peak RSS 1.35 GB at `num_proc=2`.

One caveat surfaced while smoking: on current main the image-emitting pre-tokenize branch is effectively unreachable from YAML (the chat_template strategy only calls the processor with images when `roles_to_train` is empty *and* `train_on_eos` is null, and config validation drops explicit nulls so the `"turn"` default always wins). The fixes here are exercised today by direct API use and by the multimodal sample-packing work (#3817), which routes prepared MM tokenization through this path.

## Relationship to other work

- **Designed for multimodal sample packing (#3817).** That PR is what makes the prepared multimodal path reachable from YAML, and its headline prepared-packing mode is exactly the pipeline that OOMs without these fixes. The intended order is: this PR merges first, #3817 rebases (one small conflict in the `build_prompt` hunk) and inherits the numpy fix, the buffer limits, and the mixed-dataset handling. The two are coordinated, not competing.
- **Resolution levers (#3899).** The companion PR: this one fixes the per-row buffering/dtype cost, #3899 provides the levers that shrink `bytes_per_row` itself (`processor_kwargs.max_pixels`, a working `image_size` on the pre-tokenize path) and makes them bust the prepared-dataset cache hash. Both edit `build_prompt` in adjacent hunks; whichever merges second rebases trivially.
- **Standalone value on text today (#2975).** The buffer limits are not multimodal-only: long-context text rows (e.g. 65k tokens) cost several MB each while boxed inside a map worker, so the hardcoded 1000-row map batch + 1000-row writer buffer at high `dataset_num_proc` produces exactly the load-time host-RAM exhaustion reported in #2975 ("RAM usage shoots up to almost 1.3T"). `dataset_writer_batch_size` / `dataset_map_batch_size` give those users a direct mitigation with no behavior change for anyone else.

## How has this been tested?

- `tests/prompt_strategies/test_chat_template_multimodal_cache.py` (22 tests): numpy/float32 retention through `build_prompt` and into an Arrow cache, ragged-batch None-padding, cross-batch single-modality splits incl. exact order restoration (also when rows are dropped mid-tokenization), `keep_in_memory` forwarding, empty-row filtering, buffer-kwarg plumbing (map + filter + iterable), multimodal-vs-text wrapper defaults, and helper unit tests.
- `tests/prompt_strategies` (219), `tests/utils/data` + validation (137), `tests/test_packed_batch_sampler.py` (16) all green.
- Ruff (pinned 0.15.8) check + format clean.

## AI Usage Disclaimer

Claude (Fable 5) used throughout.


thad0ctor:
Synced with main and resolved the conflict in `a79499465`.

Only `chat_template.py`, and it was adjacency: #3750 added the `bisect` imports on the same line this branch removed the now-unused `defaultdict` import. Kept both intents. `config.py` merged cleanly — the two new dataset-buffer fields and main's `attn_impl_base` changes are in different parts of the file.

Full CPU suite: 27 failed / 3324 passed, and the same 27 fail identically on clean main in this environment (27 failed / 3299 passed — the +25 is this branch's new tests, all green). Diffing the failure sets gives zero regressions and zero fixes; they're all NVFP4/triton kernel tests unrelated to this change.

PR #3901, opened 2026-07-30
perf: Apply numpy optimization for label masking

n-dlms:
Replace list multiplication with `np.full` for faster label masking during tokenization.

**Changes:**
- Use `np.full` instead of `[IGNORE_INDEX] * user_prompt_len` for constructing label arrays
- Removes TODO comment about potential numpy optimization

## Summary by CodeRabbit

* **Performance**
  * Updated internal prompt-label preparation to use a NumPy-based operation while preserving existing output behavior.
* **Compatibility**
  * No changes to exported interfaces or user-visible functionality.


NanoCode012:
Thanks for the PR. Need benchmark, not sure if converting np array to list just negates perf boost

PR #3902, opened 2026-07-30
fix: Handle non-text content types in chat message tokenization

n-dlms:
Fix crash when chat messages contain non-text content types (e.g., images, audio) during tokenization by checking for `.get("text", "")` instead of assuming `.get("text")` always returns a string.

**Changes:**
- Add fallback to empty string when message content is not plain text

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved message tokenization safety when text or tool content is missing.
  * Non-text content, such as images and audio, is now explicitly skipped during tokenization.

### DIFF
--- a/src/axolotl/core/chat/messages.py
+++ b/src/axolotl/core/chat/messages.py
@@ -142,13 +142,14 @@ class Messages(BaseModel):
         pending_weight = self.weight
         running_content = ""
         for _, msg_content in enumerate(self.content):
-            # TODO also handle non-text content types
+            # Handle text/tool content; skip non-text (image/audio) for tokenization
             if msg_content.type in [
                 MessageContentTypes.text.value,
                 MessageContentTypes.tool_call.value,
                 MessageContentTypes.tool_response.value,
             ]:
-                running_content += str(msg_content)
+                content_str = str(msg_content) if msg_content.content is not None else ""
+                running_content += content_str
                 tok_results = tokenizer(running_content, add_special_tokens=False)
                 tok_input_ids = tok_results["input_ids"]
                 if pending_input_ids:


NanoCode012 (commented):
This does not look right. You may be accessing a prop that doesn’t exist

ErenAta16 (commented):
Please hold this one. As written it raises on every message it touches, so it would break tokenization rather than harden it.

`msg_content` here is a `MessageContents`, and that model has no `content` field:

```
MessageContents fields: ['eoc', 'has_newline', 'meta', 'type', 'value', 'weight']

mc = MessageContents(type="text", value="hello")
str(mc)      -> 'hello'
mc.content   -> AttributeError: 'MessageContents' object has no attribute 'content'
```

The new guard evaluates `msg_content.content is not None` before the `str()` call, and it sits inside the branch that runs for `text`, `tool_call` and `tool_response`, which is to say the branch that runs for essentially all training data. Every one of those raises `AttributeError` on this branch. The attribute you want is `value`.

Two other things do not line up, and I think they point at the same underlying confusion rather than three separate slips.

**The type filter already exists.** The title says non-text content types crash tokenization, but `main` already skips them:

```python
if msg_content.type in [
    MessageContentTypes.text.value,
    MessageContentTypes.tool_call.value,
    MessageContentTypes.tool_response.value,
]:
```

Images and audio never reach the `str()` call, so there is no crash there to fix. Replacing the `# TODO also handle non-text content types` comment with `# Handle text/tool content; skip non-text (image/audio) for tokenization` also removes a genuine TODO and replaces it with a description of what the line above already did.

**The description describes different code.** It talks about `.get("text", "")` versus `.get("text")`, but nothing in the diff uses `.get`, and `MessageContents` is a pydantic model rather than a dict.

**On the underlying premise.** `value` is declared `Union[str, ToolCallContents, ToolResponseContents, SpecialToken]`, with no default and no `None` member, so pydantic rejects `value=None` at construction. I could not find a path that produces a `MessageContents` with a null value, which suggests there may be nothing to guard here at all. If you have a repro where tokenization fails on real data, that would be worth pasting, because the fix should follow from it and right now the change and the reported symptom are describing different problems.

Happy to look again if there is a concrete failing input behind this. I just do not think it should merge in its current form.

PR #3903, opened 2026-07-30
feat: Convert tool_calls to ToolCallContents in dataset transforms

n-dlms:
Add conversion of raw `tool_calls` dicts to `ToolCallContent` objects in the chat builder dataset transform, enabling downstream processors to handle tool calls uniformly.

**Changes:**
- Map `tool_calls` entries to `ToolCallContent` instances during chat dataset transformation

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized chat message content into consistently typed content items.
  * Improved handling of tool calls, lists, strings, and other content values.
  * Converted unsupported content formats into readable text instead of passing through inconsistent structures.

### DIFF
--- a/src/axolotl/core/datasets/transforms/chat_builder.py
+++ b/src/axolotl/core/datasets/transforms/chat_builder.py
@@ -123,28 +123,22 @@ def chat_message_transform_builder(
                 else role_default_weights_mappings[role]
             )
 
-            # TODO if "tool_calls" in message[message_content_field]: then convert tool call to ToolCallContents
-            if isinstance(message[message_content_field], str):
-                messages.append(
-                    {
-                        "role": role,
-                        "content": [
-                            {
-                                "type": "text",
-                                "value": message[message_content_field],
-                            }
-                        ],
-                        "weight": weight,
-                    }
-                )
+            # Convert tool calls to ToolCallContents when present
+            content_data = message[message_content_field]
+            if isinstance(content_data, str):
+                content_items = [{"type": "text", "value": content_data}]
             else:
-                messages.append(
-                    {
-                        "role": role,
-                        "content": message[message_content_field],
-                        "weight": weight,
-                    }
-                )
+                if isinstance(content_data, dict) and "tool_calls" in content_data:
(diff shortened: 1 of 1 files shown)


ErenAta16 (commented):
This fails validation on exactly the input it exists to handle. `tool_calls` in a chat message is a list, and the PR assigns that list straight into `value`:

```python
content_items = [{"type": "tool_call", "value": content_data["tool_calls"]}]
```

but `MessageContents.value` does not accept a list:

```
MessageContents.value annotation:
    Union[str, ToolCallContents, ToolResponseContents, SpecialToken]

MessageContents(type="tool_call", value=[{"name": "get_weather", "arguments": {...}}])
    ValidationError: 4 validation errors for MessageContents

MessageContents(type="tool_call", value={"name": "get_weather", "arguments": {...}})
    accepted -> value type ToolCallContents
```

So a message carrying one tool call raises, and a message carrying several raises the same way. The `"tool_call"` discriminator itself is correct, `MessageContentTypes.tool_call.value` is `"tool_call"`, so the type string is not the problem.

The shape the model wants is one `MessageContents` per tool call:

```python
content_items = [
    {"type": "tool_call", "value": call} for call in content_data["tool_calls"]
]
```

which also matches how the rest of `content` works, a flat list of single-valued items rather than one item holding a collection.

Two smaller things.

The title says the conversion produces `ToolCallContents`, but the code emits plain dicts and relies on pydantic to coerce them at construction. That does work, as the single-dict result above shows, but only if these dicts are actually fed through `MessageContents`. Worth confirming the caller does that rather than passing the raw list on.

The `else` branch changes behaviour for non-str, non-tool-call content:

```python
content_items = content_data if isinstance(content_data, list) else [{"type": "text", "value": str(content_data)}]
```

Previously a non-str `content` was passed through unchanged. Now anything that is not a list gets `str()`-ified into a text item, so a dict without `tool_calls` becomes the literal text `"{'foo': 'bar'}"` in the training data rather than failing or being handled. Silently stringifying structured content into the token stream is worse than the `TODO` this replaces; I would raise there instead.

Happy to look again once the list case is fixed.

PR #3907, opened 2026-07-30
upgrade flash-linear-attention package for bugfixes, etc

winglian:
## Summary by CodeRabbit

* **Chores**
  * Updated compatibility requirements for non-ARM64 systems.
  * Enables use of newer versions of the `fla-core` and `flash-linear-attention` packages (version 0.5.2 or later).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -85,8 +85,8 @@ dependencies = [
     "torchao==0.17.0 ; sys_platform != 'darwin' and platform_machine != 'aarch64'",
 
     # Architecture-specific
-    "fla-core==0.4.1 ; platform_machine != 'aarch64'",
-    "flash-linear-attention==0.4.1 ; platform_machine != 'aarch64'",
+    "fla-core>=0.5.2 ; platform_machine != 'aarch64'",
+    "flash-linear-attention>=0.5.2 ; platform_machine != 'aarch64'",
 ]
 
 [project.optional-dependencies]


ErenAta16 (commented):
Small thing, but the diff changes two things at once and only one is in the title. Alongside the version move it switches these from an exact pin to a lower bound:

```diff
-"fla-core==0.4.1 ; platform_machine != 'aarch64'",
-"flash-linear-attention==0.4.1 ; platform_machine != 'aarch64'",
+"fla-core>=0.5.2 ; platform_machine != 'aarch64'",
+"flash-linear-attention>=0.5.2 ; platform_machine != 'aarch64'",
```

That makes them outliers against the rest of the file. Counting the runtime dependency table, 48 are pinned with `==` and 10 use `>=`, and the immediate neighbours here are all exact:

```
"liger-kernel==0.8.0 ; sys_platform != 'darwin'",
"torchao==0.17.0 ; sys_platform != 'darwin' and platform_machine != 'aarch64'",
"flash-attn==2.8.3"
```

which is the pattern I would expect, since those are the packages whose kernels can move numerics between releases. With `>=`, the next `fla-core` release lands in a fresh install without anything changing in this repo, and if it shifts a kernel result the bisect starts in the wrong place.

If the intent was just to pick up the 0.5.2 bugfixes, `==0.5.2` gets that without opening the upper end. If the intent really is to float, it would be worth saying so in the description, because it is a different decision from "upgrade for bugfixes" and it is the one someone will want to find later.

Also worth confirming the two stay locked to each other. `fla-core` and `flash-linear-attention` are versioned in lockstep upstream, and independent lower bounds let a resolver land mismatched versions of the two if their release cadence ever diverges.

Nothing blocking, and if the pinning change is deliberate then ignore all of the above.

NanoCode012:
Superseded by #3980

PR #3130, opened 2025-09-04
setup env vars for ray train for FSDP

winglian:
# Description

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved intermittent failures and misconfigurations during distributed training with Ray by initializing the optimization environment earlier in the worker lifecycle. This ensures consistent configuration handling across workers and reduces setup-related errors. Users should experience smoother job launches and fewer retries under distributed workloads.
* **Performance**
  * Minor improvements to startup reliability and consistency for Ray-based training runs due to a more predictable environment setup.

PR #3909, opened 2026-07-30
fix: set up tracking env vars on Ray Train worker

### LINKED ISSUE #3908: MLflow/WandB/Comet/Trackio env vars not set on Ray Train worker (use_ray: true)
## Bug

When `use_ray: true`, tracking integrations (MLflow, WandB, Comet, Trackio) silently fail to log metrics because their environment variables are never set on the Ray Train worker process.

## Root cause

`load_cfg()` in `cli/config.py` (lines 324-327) calls:
```python
setup_wandb_env_vars(cfg)
setup_mlflow_env_vars(cfg)
setup_comet_env_vars(cfg)
setup_trackio_env_vars(cfg)
```

These functions convert YAML config fields (`mlflow_tracking_uri`, `mlflow_experiment_name`, `wandb_api_key`, etc.) into environment variables that the HuggingFace callbacks (`MLflowCallback`, `WandbCallback`, etc.) read via `os.getenv()`.

However, `load_cfg()` runs on the **driver process**. When `use_ray: true`, the actual training runs on a **separate Ray Train worker process** via `ray_train_func()` in `cli/train.py`. The worker reconstructs the config from the serialized dict but **never calls** these setup functions. The HF callbacks read env vars that don't exist, so tracking silently fails.

Without `use_ray` (direct `accelerate`/`torchrun` path), `load_cfg` and training run in the same process, so it works.

## Impact

- `mlflow_tracking_uri`, `mlflow_experiment_name` from YAML are ignored → MLflow logs to "Default" experiment or doesn't log at all
- Same for `wandb_*`, `comet_*`, `trackio_*` config fields
- No error is raised — tracking just silently doesn't work

## Prior context

This was flagged during review of PR #3130 by @NanoCode012:
> https://github.com/axolotl-ai-cloud/axolotl/pull/3130#discussion_r2329564517
>
> Do we need any of the others too?
> ```python
> setup_wandb_env_vars(cfg)
> setup_mlflow_env_vars(cfg)
> setup_comet_env_vars(cfg)
> ```

The question was not addressed and PR #3130 was merged without adding these calls.

## Fix

Add the four setup calls in `ray_train_func()` in `cli/train.py`, mirroring what `load_cfg()` does on the driver. See PR: https://github.com/axolotl-ai-cloud/axolotl/pull/3909

fialhocoelho:
## Summary

- Fixes #3908

When `use_ray: true`, tracking integrations (MLflow, WandB, Comet, Trackio) silently fail because their env-var setup functions are called only in `load_cfg()` (driver process), not in `ray_train_func()` (worker process). The HuggingFace callbacks read tracking URIs/experiment names from env vars, not TrainingArguments.

## Root cause

`load_cfg()` in `cli/config.py:324-327` calls:
```python
setup_wandb_env_vars(cfg)
setup_mlflow_env_vars(cfg)
setup_comet_env_vars(cfg)
setup_trackio_env_vars(cfg)
```

These convert YAML fields (`mlflow_tracking_uri`, `mlflow_experiment_name`, etc.) into env vars. But `load_cfg` runs on the driver — the Ray Train worker is a separate process that doesn't inherit them.

This was flagged during review of #3130 but not addressed:
https://github.com/axolotl-ai-cloud/axolotl/pull/3130#discussion_r2329564517

## Changes

- Add `setup_wandb_env_vars`, `setup_mlflow_env_vars`, `setup_comet_env_vars`, `setup_trackio_env_vars` to `_LAZY_IMPORTS` in `cli/train.py`
- Call them in `ray_train_func()` after `resolve_dtype(cfg)`, mirroring the order in `load_cfg()`
- Update existing tests to mock the new calls
- Add regression test `test_ray_train_func_sets_up_tracking_env_vars`

## Test plan

- [x] `ruff check` + `ruff format --check` pass
- [x] New regression test passes (verified locally — existing tests require HF model downloads that fail offline)
- [ ] CI passes on push

## Summary by CodeRabbit

* **Bug Fixes**
  * Tracking integrations now initialize their environment settings correctly when training runs on Ray workers.
  * Improved consistency for Weights & Biases, MLflow, Comet, and Trackio during distributed training.

* **Tests**
  * Added coverage to verify that all supported tracking environments are configured before training begins.

### DIFF
--- a/src/axolotl/cli/train.py
+++ b/src/axolotl/cli/train.py
@@ -23,6 +23,10 @@ _LAZY_IMPORTS = {
     "prepare_optim_env": "axolotl.utils.trainer",
     "prepare_plugins": "axolotl.cli.config",
     "resolve_dtype": "axolotl.utils.config",
+    "setup_comet_env_vars": "axolotl.utils.comet_",
+    "setup_mlflow_env_vars": "axolotl.utils.mlflow_",
+    "setup_trackio_env_vars": "axolotl.utils.trackio_",
+    "setup_wandb_env_vars": "axolotl.utils.wandb_",
     "train": "axolotl.train",
     "validate_config": "axolotl.utils.config",
 }
@@ -184,6 +188,19 @@ def ray_train_func(kwargs: dict):
     prepare_optim_env_fn(cfg)
     normalize_config_fn(cfg)
     resolve_dtype_fn(cfg)
+
+    # Tracking env vars must be set on the worker process — the HuggingFace
+    # MLflowCallback / WandbCallback read tracking URIs and experiment names
+    # from env vars, not TrainingArguments.  The driver sets these in load_cfg
+    # but that runs in a separate process; Ray Train workers do not inherit them.
+    setup_wandb_env_vars_fn: Any = _lazy_attr("setup_wandb_env_vars")
+    setup_mlflow_env_vars_fn: Any = _lazy_attr("setup_mlflow_env_vars")
+    setup_comet_env_vars_fn: Any = _lazy_attr("setup_comet_env_vars")
+    setup_trackio_env_vars_fn: Any = _lazy_attr("setup_trackio_env_vars")
+    setup_wandb_env_vars_fn(cfg)
+    setup_mlflow_env_vars_fn(cfg)
+    setup_comet_env_vars_fn(cfg)
+    setup_trackio_env_vars_fn(cfg)
 
     # ray serializing objects gets rid of frozen attribute - HF expects dict not DefaultDict
     if cfg.deepspeed and hasattr(cfg.deepspeed, "to_dict"):
--- a/src/axolotl/cli/train.py
+++ b/src/axolotl/cli/train.py
@@ -189,10 +189,7 @@ def ray_train_func(kwargs: dict):
     normalize_config_fn(cfg)
     resolve_dtype_fn(cfg)
 
-    # Tracking env vars must be set on the worker process — the HuggingFace
-    # MLflowCallback / WandbCallback read tracking URIs and experiment names
-    # from env vars, not TrainingArguments.  The driver sets these in load_cfg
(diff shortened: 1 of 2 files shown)


ErenAta16 (commented):
Checked the two things that would have made this quietly wrong, and both hold.

The trailing-underscore module paths are right, not typos:

```
axolotl.utils.comet_     EXISTS
axolotl.utils.mlflow_    EXISTS
axolotl.utils.trackio_   EXISTS
axolotl.utils.wandb_     EXISTS
```

which is the naming this repo uses to avoid shadowing the third-party packages, so the lazy map resolves.

The monkeypatches in the test are effective, which was my other concern given the call goes through `_lazy_attr` rather than a direct import:

```python
def _lazy_attr(name: str) -> Any:
    return globals().get(name) or __getattr__(name)
```

`monkeypatch.setattr("axolotl.cli.train.setup_wandb_env_vars", ...)` writes into this module's globals, and `globals().get(name)` is consulted first, so the patch wins and the test does not reach the real tracker setup. Worth saying out loud because a reader could reasonably assume the lazy path bypasses it.

The comment explaining *why* this has to happen on the worker is the most useful line in the diff. "HF callbacks read them from `os.environ`, not `TrainingArguments`" is the non-obvious fact that makes the whole change necessary, and it is the thing someone would otherwise undo while tidying up.

**One latent trap, not introduced here.** `globals().get(name) or __getattr__(name)` uses `or`, so a falsy override falls through to the real attribute. Patching with `lambda *_: None` is truthy and fine, but a future test that patches one of these with `None` to mean "disabled" would silently get the real tracker setup instead, and the failure would look like a network or credentials problem rather than a test-wiring one. `globals()[name] if name in globals() else __getattr__(name)` closes it. Unrelated to this PR beyond being the mechanism it now depends on.

Calling all four unconditionally matches how they behave elsewhere, each one no-ops when its config section is absent, so there is no need to gate them on which tracker is configured.

PR #3911, opened 2026-07-31
fix: use safe dict access for quant_method

lxcxjxhx:
Fix KeyError when quant_method key is missing from quantization_config by using .get() instead of direct dictionary access. Changes in src/axolotl/loaders/model.py lines 732, 735, 739, 744.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility when loading adapter-based GPTQ, AWQ, and bitsandbytes models with incomplete quantization configuration data.
  * Preserved existing quantization behavior while preventing configuration lookup errors.


NanoCode012:
When does this raise an error?

PR #3916, opened 2026-08-02
perf(chat): tokenize messages once via offset mapping

n-dlms:
## Description

Rewrites `Messages.tokenized` so the concatenated content is tokenized **once** instead of once per content item. Tokens are attributed to their `MessageContents` via `return_offsets_mapping` char offsets, and label masking follows the per-item weights.

## Motivation and Context

The previous implementation tokenized the running concatenated string at every content item (O(n²)), and maintained an incremental "pending window" that compared the tail of the new tokenization against previously emitted tokens. That window had a correctness flaw: **tokens produced by earlier items were frozen and never re-examined, so any token whose value changed after later content was appended was silently dropped** — the returned `input_ids` were not the tokenization of the final string. This corrupts the training signal: labels were masked/emitted for tokens that never appear in the actual sequence, and the model trains on label indices that point past the input.

The new implementation:

- Tokenizes the full string exactly once — `input_ids` are, by construction, equal to the tokenization of the final string.
- Is O(n) instead of O(n²): measured **4.9x faster at 4 items, 12.9x at 16, 30.8x at 64, and 116.4x at 256 items** (wall-clock per call, old vs new, llama-3 tokenizer).
- Works for both fast and slow (`TokenizersBackend`) tokenizers — `return_offsets_mapping` is supported by both.
- Handles empty / non-text-only content with an early return.

## How has this been tested?

- All 5 existing `test_messages.py` fixtures produce identical output (byte-for-byte on the final string).
- New `tests/core/chat/test_messages_fuzz.py`:
  - 200 randomized trials asserting output equals a naive full-string tokenization oracle;
  - 200 property trials asserting `input_ids` equal a single tokenization of the final string, with matching mask/label lengths;
  - 4 hardcoded regression cases for boundary merging across content items, including the exact case where the old code dropped tokens.
- Full suite: **2,004 tests passed** with `AXOLOTL_FUZZ_ITERATIONS=1000` (404 new + 5 existing at default iterations).
- `ruff`, `ruff-format`, `mypy`, `bandit` clean.

## AI Usage Disclaimer

## Types of changes

- [x] Performance improvement
- [x] New tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved message tokenization for mixed and weighted content.
  * Preserved accurate labels and attention masks across content boundaries.
  * Correctly handles messages containing only unsupported or non-text content.

* **Tests**
  * Added broad regression and fuzz coverage for tokenization, boundary merges, weighting, and single-item messages.

### DIFF
--- a/src/axolotl/core/chat/messages.py
+++ b/src/axolotl/core/chat/messages.py
@@ -134,41 +134,43 @@ class Messages(BaseModel):
     def tokenized(
         self, tokenizer: PreTrainedTokenizer, ignore_index=-100
     ) -> dict[str, List[int]]:
-        # iterate over the contents, tokenizing the concatenated string values up to the current MessageContents
-        # returns a dictionary mapping w input_ids, attention_mask, and labels
-        input_ids: List[int] = []
-        labels: List[int] = []
-        pending_input_ids: List[int] = []
-        pending_weight = self.weight
-        running_content = ""
-        for _, msg_content in enumerate(self.content):
+        # tokenize the concatenated content values once, attributing tokens to
+        # their content item by char offset; avoids re-tokenizing the running
+        # string per item (O(n) instead of O(n^2)) and guarantees the output
+        # equals the tokenization of the final string
+        content_str = ""
+        spans: List[tuple[int, int, MessageContents]] = []
+        for msg_content in self.content:
             # TODO also handle non-text content types
             if msg_content.type in [
                 MessageContentTypes.text.value,
                 MessageContentTypes.tool_call.value,
                 MessageContentTypes.tool_response.value,
             ]:
-                running_content += str(msg_content)
-                tok_results = tokenizer(running_content, add_special_tokens=False)
-                tok_input_ids = tok_results["input_ids"]
-                if pending_input_ids:
-                    new_pending_inputs = tok_input_ids[
-                        len(input_ids) : len(input_ids) + len(pending_input_ids)
-                    ]
-                    if new_pending_inputs != pending_input_ids:
-                        pending_input_ids = new_pending_inputs
-                    input_ids.extend(pending_input_ids)
-                    if pending_weight:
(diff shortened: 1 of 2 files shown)


n-dlms:
Verified empirically against the installed transformers 5.14.1, not just the docs:

- `AutoTokenizer.from_pretrained("huggyllama/llama-7b", use_fast=False)` (slow `LlamaTokenizer`): `tokenizer(..., return_offsets_mapping=True)` returns correct `offset_mapping` — no `NotImplementedError`.
- `AutoTokenizer.from_pretrained("NousResearch/Meta-Llama-3-8B")` (fast `TokenizersBackend`): same.
- `Messages.tokenized` output equals a single `tokenizer(<final string>)` call on both.

The slow-tokenizer `NotImplementedError` you referenced applies to the `PreTrainedTokenizer` base in older releases (and historically to `return_offsets_mapping` combined with `is_split_into_words`); in 5.14.1 the slow path computes offsets in `encode_plus`/`_encode_plus` directly. The new `test_messages_fuzz.py` also runs its oracle against whichever tokenizer is injected, and the suite passes with both backends.

src/axolotl/core/chat/messages.py
```diff
@@ -134,41 +134,43 @@ def __str__(self) -> str:
...
+            return {
+                "input_ids": [],
+                "attention_mask": [],
+                "labels": [],
+            }
+        tok_results = tokenizer(
+            content_str, add_special_tokens=False, return_offsets_mapping=True
+        )
+        input_ids = tok_results["input_ids"]
+        offsets = tok_results["offset_mapping"]
```
ved1beta, line 163: Either check tokenizer.is_fast and fall back, or fail loudly

ved1beta (changes requested):

ved1beta:
closing as stale , reopen if not

PR #3920, opened 2026-08-02
fix(grpo): seed dataloader workers by SP group

n-dlms:
## Description

In `_prepare_dataloader`, worker RNG is now seeded by **SP group** (`process_index // context_parallel_size`) when `context_parallel_size > 1`, instead of by global `process_index`. Also replaces the speculative `TODO(djsaunde)` about using accelerate's dataloader preparation with the verified invariant that documents why the raw dataloader must be returned.

## Motivation and Context

Sequence parallelism requires every rank in an SP group to receive **identical** full batches — the sequence dimension is split at attention time by ring attention, not by the dataloader (confirmed against transformers' own `_prepare_context_parallel_inputs`, which also splits at forward time). The old worker seeding used the global rank, so ranks within a group had different worker RNG; any collate that draws randomness (shuffling, random packing, augmentation) would produce different batches across the group, and ring attention would exchange KV shards across mismatched sequences — silently corrupting training. Seeding by SP group makes collate behavior deterministic and identical within a group (distinct across groups).

The `TODO` proposed `dispatch_batches` + `slice_fn_for_dispatch` for the dataloader; this cannot work: accelerate's dispatcher broadcasts from process 0 only and slices by global rank, which collapses every SP group onto group 0's data. The comment now states this invariant so the bypass is not "simplified away" later.

## How has this been tested?

- New `tests/core/test_grpo_dataloader.py` (6 tests): raw dataloader returned (prepare not called) when `context_parallel_size > 1`; prepared otherwise; worker seed rank equals SP group for CP runs, equals process index otherwise; identical seeds across ranks in the same group; eval has no worker init fn.
- `ruff`, `ruff-format`, `mypy`, `bandit` clean; full offline suite passes.

## AI Usage Disclaimer

## Types of changes

- [x] New tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sequence-parallel data loading so workers within the same parallel group receive consistent random seeds.
  * Preserved complete, identical batches across sequence-parallel workers by avoiding incompatible data distribution.
  * Ensured evaluation data loaders do not apply training-only worker initialization.

* **Tests**
  * Added coverage for parallel and non-parallel data-loader preparation, worker seed consistency, and evaluation behavior.

### DIFF
--- a/src/axolotl/core/trainers/grpo/trainer.py
+++ b/src/axolotl/core/trainers/grpo/trainer.py
@@ -236,10 +236,18 @@ class AxolotlGRPOSequenceParallelTrainer(AxolotlGRPOTrainer):
                 dataloader_params["drop_last"] = self.args.dataloader_drop_last
 
             if not is_eval:
+                # Seed workers by SP group so every rank in a group gets identical
+                # worker RNG and therefore identical batches (required by ring
+                # attention, which splits the sequence at forward time).
+                worker_seed_rank = (
+                    self.args.process_index // self.args.context_parallel_size
+                    if self.args.context_parallel_size > 1
+                    else self.args.process_index
+                )
                 dataloader_params["worker_init_fn"] = partial(
                     seed_worker,
                     num_workers=self.args.dataloader_num_workers,
-                    rank=self.args.process_index,
+                    rank=worker_seed_rank,
                 )
 
         # Create the dataloader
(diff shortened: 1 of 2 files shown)


src/axolotl/core/trainers/grpo/trainer.py
```diff
@@ -236,10 +236,18 @@ def _prepare_dataloader(
...
 
             if not is_eval:
+                # Seed workers by SP group so every rank in a group gets identical
+                # worker RNG and therefore identical batches (required by ring
+                # attention, which splits the sequence at forward time).
+                worker_seed_rank = (
+                    self.args.process_index // self.args.context_parallel_size
+                    if self.args.context_parallel_size > 1
+                    else self.args.process_index
+                )
```
winglian, line 246: could we make this a private method in `DistributedParallelMixin` that way we can reuse it across trainers?

n-dlms: @winglian Good point thanks, I'll move the SP-aware worker seed rank calculation into a private method on DistributedParallelMixin so it can be reused across trainers.

NanoCode012:
> that draws randomness (shuffling, random packing, augmentation) would produce different batches across the group, and ring attention would exchange KV shards across mismatched sequences

this is a concerning claim. Do you have results to show this? We never augment.

ved1beta:
closing as stale

PR #3921, opened 2026-08-02
chore(prompt): remove stale numpy slicing TODOs

n-dlms:
## Description

Removes three stale `# TODO this could be sped up using numpy array slicing` comments in the prompt tokenizers.

## Motivation and Context

Benchmarked before removing: `np.full(...).tolist()` is **6-9x slower** than the existing `[IGNORE_INDEX] * n` list multiplication at all sizes, and a profile of the tokenizing strategies shows ~86% of runtime is the tokenizer's `encode_batch` while the Python list operations account for ~3%. The TODOs suggest an optimization that would measurably regress performance, so they are removed to avoid misleading future contributors. No behavior change.

## How has this been tested?

- Existing prompt-tokenizer tests pass unchanged (3 tests, llama-7b tokenizer).
- `ruff`, `ruff-format`, `mypy`, `bandit` clean.

## AI Usage Disclaimer

## Types of changes

- [x] Chore/cleanup (no behavior change)

## Summary by CodeRabbit

* **Chores**
  * Removed obsolete internal TODO comments.
  * No user-visible functionality or behavior changes.

### DIFF
--- a/src/axolotl/prompt_strategies/alpaca_w_system.py
+++ b/src/axolotl/prompt_strategies/alpaca_w_system.py
@@ -40,7 +40,6 @@ class InstructionWSystemPromptTokenizingStrategy(PromptTokenizingStrategy):
         tokenized_prompt = self._tokenize(user_prompt, add_eos_token=False)
         if not self.train_on_inputs:
             user_prompt_len = len(tokenized_prompt["input_ids"])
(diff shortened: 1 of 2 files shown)


NanoCode012:
Thanks for PR. At what N, did you run the comparison at?

ved1beta (approved):
verified looks good enough

PR #3847, opened 2026-07-17
fix: FA4 quack0.6 with latest transformers support

NanoCode012:
# Description

FA4 incompatibility with quack 0.5, so this updates the doc and fix transformers compatibility.

## Summary by CodeRabbit

* **New Features**
  * Added native Flash Attention 4 support for compatible GPUs and model head dimensions.
  * Added automatic upgrades from Flash Attention 2 or 3 when FA4 is available.
  * Added `flash_attention_4` as a supported attention backend and packing option.
  * Added compatibility checks and warnings for outdated kernel packages.

* **Documentation**
  * Updated attention backend requirements, configuration examples, limitations, and troubleshooting guidance.
  * Clarified NVFP4 adapter merging terminology.

PR #3928, opened 2026-08-10
fix(attention): don't auto-upgrade to FA4 under context parallelism

thad0ctor:
# Description

`_resolve_flash_attention_4` silently upgrades a configured `flash_attention_2` / `flash_attention_3` to `flash_attention_4` whenever FA4 is usable on the host. Ring attention only substitutes the `flash_attention_2` dispatch key — both `ring_flash_attn`'s `hf_adapter` and `ring_attn/adapters/batch.py:196` write `ALL_ATTENTION_FUNCTIONS["flash_attention_2"]`, and nothing registers an FA4 ring variant. So under context parallelism the upgrade routes attention around ring communication and each rank attends over its own shard alone: wrong loss and gradients, no error.

Regression from #3847, which switched from patching `_lazy_imports` (dispatch key stayed `flash_attention_2`, so the ring forward still won) to rewriting the dispatch key.

Fix: skip the implicit upgrade when `context_parallel_size > 1`. An explicitly requested `flash_attention_4` is unaffected.

## Motivation and Context

Silent correctness bug on multi-GPU long-context runs — the setups context parallelism exists for. Training completes and losses look plausible.

## How has this been tested?

Two cases added to `TestResolveFlashAttention4`: FA2 and FA3 both stay put at `context_parallel_size=2`, and the upgrade still fires at 1. The shared `_loader()` helper now takes cfg kwargs, since the method reads `cfg.context_parallel_size`.

`tests/loaders/ tests/monkeypatch/ tests/utils/schemas/` — 389 passed, 1 xfailed. Pinned ruff clean.

Not validated on hardware: no multi-GPU SM90+ box with FA4 here, so the wrong-results path is traced from the dispatch code rather than reproduced.

## AI Usage Disclaimer

Yes — Claude Code was used to trace the dispatch path and draft the fix, tests, and description. Findings verified against the code by hand.

## Types of changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved the requested attention implementation when context parallelism is enabled.
  * Prevented unsupported Flash Attention 4 upgrades in ring-attention configurations.
  * Continued applying Flash Attention 4 automatically when context parallelism is not enabled.

### DIFF
--- a/src/axolotl/loaders/model.py
+++ b/src/axolotl/loaders/model.py
@@ -821,6 +821,16 @@ class ModelLoader:
         if hf_impl == "flash_attention_4":
             configure_fa4()
             return hf_impl
+
+        # Ring attention only substitutes the `flash_attention_2` dispatch key.
+        if (self.cfg.context_parallel_size or 1) > 1:
+            LOG.info(
+                "Not upgrading %s to Flash Attention 4: no ring-attention "
+                "implementation is registered for it under context parallelism.",
+                hf_impl,
+            )
+            return hf_impl
+
         if fa4_usable(self.model_config):
             configure_fa4()
             LOG.info("Flash Attention 4 enabled (upgraded from %s).", hf_impl)
--- a/src/axolotl/loaders/model.py
+++ b/src/axolotl/loaders/model.py
@@ -825,8 +825,8 @@ class ModelLoader:
         # Ring attention only substitutes the `flash_attention_2` dispatch key.
         if (self.cfg.context_parallel_size or 1) > 1:
             LOG.info(
-                "Not upgrading %s to Flash Attention 4: no ring-attention "
-                "implementation is registered for it under context parallelism.",
+                "Not upgrading %s to Flash Attention 4: ring attention only "
+                "supports the flash attention 2 backend.",
                 hf_impl,
             )
             return hf_impl
--- a/src/axolotl/utils/schemas/validation.py
+++ b/src/axolotl/utils/schemas/validation.py
@@ -16,6 +16,7 @@ from axolotl.utils.schemas.enums import (
     ChatTemplate,
     RingAttnFunc,
     RLType,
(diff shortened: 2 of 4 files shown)


thad0ctor:
@coderabbitai good catch — added in `995a37d50` as `test_explicit_fa4_still_configured_under_context_parallelism`.

Worth noting why it's more than coverage: the guard sits deliberately *after* the explicit-FA4 early return so an explicit request still reaches `configure_fa4()`. An earlier draft had it above, and nothing pinned that ordering — I verified the new test fails (and only it fails) when the guard is moved back above the early return.

One scoping note: explicit `flash_attention_4` + CP is only reachable with `use_glm_dsa_kernels`, since `flash_attention_4` isn't in `ATTN_IMPLS_USING_FLASH_LIB` and the CP validator otherwise rejects it. So the test pins this method's contract — never touch an explicit FA4 request — rather than asserting FA4+CP is safe; that remains the validator's job.

Also reworded the log line in the same commit. It previously read "Keeping flash_attention_3: ... upgrading it would break context parallelism", which implies FA3 is the CP-safe choice. It isn't: ring attention registers only under the `flash_attention_2` key, and FA2/FA3 start out sharing one dispatch function, so overriding FA2 leaves FA3 dispatching natively without ring communication. That's a separate pre-existing bug, not from #3847, and out of scope here — filed separately with a runnable repro.

NanoCode012:
Thanks for the catch. Is this CP only? Should we be able to patch it to work with FA4?

winglian:
@thad0ctor would this get solved by getting #3844 over the line?

NanoCode012 (approved):

thad0ctor:
> @thad0ctor would this get solved by getting #3844 over the line?

Sorry I was out of town and been tied up with personal stuff, glad you guys got it resolved @NanoCode012 @winglian

PR #3941, opened 2026-08-22
fix(callbacks): resolve the resume checkpoint after auto-resume runs (#3940)

### LINKED ISSUE #3940: TokensPerSecondCallback captures resume_from_checkpoint before auto-resume detection resolves it, silently zeroing token metrics
### Expected Behavior

When `include_tkps: true` and `auto_resume_from_checkpoints: true` are both set (no
explicit `resume_from_checkpoint` in the config), resuming a run should restore the
saved `tokens/total` and `tokens/trainable` counters from the checkpoint being
auto-resumed, the same way it does when `resume_from_checkpoint` is set explicitly
in the YAML.

### Current behaviour

On an auto-resumed run, `TokensPerSecondCallback.on_train_begin()` never restores
the saved token counters, so `tokens/total`, `tokens/trainable` and
`tokens/train_per_sec_per_gpu` reset to zero/spurious values, silently, with no
warning.

Root cause: `get_callbacks()` in `src/axolotl/core/builders/causal.py` constructs
`TokensPerSecondCallback(resume_from_checkpoint=self.cfg.resume_from_checkpoint)`
inside `setup_model_and_trainer()`, which `train.py` calls before
`determine_last_checkpoint(cfg)` resolves the auto-resume path. At construction
time `cfg.resume_from_checkpoint` is still `None`, and the callback stores that
value directly (`self.resume_from_checkpoint = resume_from_checkpoint`), so
`on_train_begin()`'s `isinstance(self.resume_from_checkpoint, str)` check fails and
the restore is skipped, even though `cfg.resume_from_checkpoint` gets the correct
path a few lines later in `train.py`. Passing `resume_from_checkpoint` explicitly in
the config works fine, since it's already a string at construction time.

I confirmed this by calling the real `determine_last_checkpoint()` and
`TokensPerSecondCallback` directly, in the same order `train.py` calls them: with a
checkpoint directory containing a `tokens_state.json`, constructing the callback
before calling `determine_last_checkpoint(cfg)` leaves `state.tokens` at `None`
after `on_train_begin()`; constructing it after (with the resolved path) restores
`state.tokens` correctly. I did not run a full GPU training job through an actual
resume cycle, only this direct call sequence against `main`.

### Steps to reproduce

1. Set `include_tkps: true` and `auto_resume_from_checkpoints: true`, with no
   `resume_from_checkpoint` in the config.
2. Train until a checkpoint is saved (which also writes `tokens_state.json`).
3. Restart the same run so it auto-resumes from that checkpoint.
4. Compare the first few `tokens/total` / `tokens/trainable` log values against the
   checkpoint's saved `tokens_state.json` - they don't carry over.

### Config yaml

```yaml
include_tkps: true
auto_resume_from_checkpoints: true
# resume_from_checkpoint intentionally omitted, relying on auto-resume
```

### Possible solution

Resolve `determine_last_checkpoint(cfg)` before `setup_model_and_trainer()` runs, or
have `get_callbacks()` re-read `self.cfg.resume_from_checkpoint` at
`on_train_begin()` time instead of at construction time. Happy to send a PR for
either, whichever fits the codebase better.

- [x] Linux

### Python Version

3.11 (bug is in the call ordering, not Python-version-dependent)

### axolotl branch-commit

main/e198b1d8ed1d36b8d732f67cbb1dda62f86d3986

### Acknowledgements

Anai-Guo:
# Description

`TokensPerSecondCallback` snapshots `resume_from_checkpoint` at construction time. On an auto-resumed run that value is still `None`, so the saved token counters are never restored.

The ordering in `train.py` is:

```python
(trainer, ...) = setup_model_and_trainer(cfg, dataset_meta)  # L647 -> builds the callbacks
...
resume_from_checkpoint = determine_last_checkpoint(cfg)      # L659 -> sets cfg.resume_from_checkpoint
execute_training(cfg, trainer, resume_from_checkpoint)
```

`get_callbacks()` (`core/builders/causal.py:91`) passes `self.cfg.resume_from_checkpoint` into the callback, but `determine_last_checkpoint()` — the only thing that resolves the auto-resume path (`utils/train.py:42`) — runs twelve lines later. So with

```yaml
include_tkps: true
auto_resume_from_checkpoints: true
# no explicit resume_from_checkpoint
```

`on_train_begin()`'s `isinstance(self.resume_from_checkpoint, str)` guard is `False`, the `tokens_state.json` restore is skipped, and `tokens/total`, `tokens/trainable` and `tokens/train_per_sec_per_gpu` silently restart from zero. No warning is emitted. Setting `resume_from_checkpoint` explicitly in the YAML works, because it is already a string by the time the builder runs.

The fix holds the config and re-reads it in `on_train_begin()` rather than snapshotting at construction. `cfg` is the same object throughout (`setup_trainer(cfg=cfg, ...)` → `HFCausalTrainerBuilder(cfg, ...)` → `self.cfg = cfg`), and `determine_last_checkpoint()` mutates it in place, so the late read sees the resolved path. An explicitly configured `resume_from_checkpoint` still takes precedence, and the constructor keeps working without a `cfg`.

## Motivation and Context

Fixes #3940.

Throughput accounting is the reason `include_tkps` exists; on a resumed run the cumulative counters are the part that has to survive. Today they only survive for users who spell the checkpoint out by hand, which is the opposite of what `auto_resume_from_checkpoints` is for — and the failure is silent, so the numbers just look wrong rather than reporting an error.

## How has this been tested?

Four tests added to `tests/utils/callbacks/test_tokens_per_second.py`, driving the **real** `determine_last_checkpoint()` in the same order `train.py` calls it (build the callback first, resolve the checkpoint second):

| case | expectation |
|---|---|
| auto-resume, no explicit path | counters restored from `tokens_state.json` |
| explicit `resume_from_checkpoint` | still restored (unchanged behaviour) |
| fresh run, nothing to resume | `state` never grows a `tokens` attribute |
| no `cfg` passed | falls back to the constructor value |

The third case asserts absence of the attribute rather than `None`, matching how the trainer gates on it (`core/trainers/base.py:401`, `if not hasattr(self.state, "tokens")`).

**Environment caveat, stated plainly:** I could not run the repo's pytest suite locally. This is a Windows box with torch 2.5.1, and the current `trl` needs `torch.distributed.fsdp.FSDPModule` (torch >= 2.6); separately, `pyarrow` hard-crashes the interpreter under pytest here. So instead of the suite I ran an equivalent old-vs-new harness that `exec`s the **actual shipped source files** — `utils/callbacks/tokens_per_second.py` and `utils/train.py`, not transcriptions of them — with only `axolotl.utils.logging` stubbed, and replicates the `causal.py` construction call verbatim:

```
########## BEFORE (unpatched main) ##########
FAIL  auto-resume restores counters                    (state.tokens['total']=None)
PASS  explicit resume_from_checkpoint still restores   (state.tokens['total']=42)
PASS  fresh run leaves counters untouched              (state.tokens['total']=None)

########## AFTER (patched) ##########
PASS  auto-resume restores counters                    (state.tokens['total']=1234)
PASS  explicit resume_from_checkpoint still restores   (state.tokens['total']=42)
PASS  fresh run leaves counters untouched              (state.tokens['total']=None)
```

Exactly one scenario changes state; the other two are byte-identical before and after. CI will be the first run of the pytest file itself — please flag anything it turns up and I'll fix it promptly.

`ruff check` and `ruff format --diff` pass at the pinned `v0.15.8` from `.pre-commit-config.yaml`, against the repo's own config.

Note this branch is based on an older `main`: syncing my fork trips GitHub's `workflow` scope guard on `.github/workflows/base.yml`, so I based on the fork tip and wrote `causal.py` from that same tree to keep unrelated upstream drift out of the diff. Happy to rebase if you'd prefer.

## AI Usage Disclaimer

Yes — AI-assisted (Claude). The diagnosis, patch, tests and the old-vs-new harness above were produced with AI assistance and reviewed by me before submitting; the before/after output quoted above is from an actual run, not generated text.

## Types of changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved training resume behavior by correctly restoring token-counting state from configured or explicitly selected checkpoints.
  * Preserved expected behavior for fresh training runs and configurations without a resume checkpoint.

### DIFF
--- a/src/axolotl/core/builders/causal.py
+++ b/src/axolotl/core/builders/causal.py
@@ -87,12 +87,7 @@ class HFCausalTrainerBuilder(TrainerBuilderBase):
             callbacks.append(QATCallback(self.cfg.qat))
 
         if self.cfg.include_tkps:
-            callbacks.append(
-                TokensPerSecondCallback(
-                    resume_from_checkpoint=self.cfg.resume_from_checkpoint,
-                    cfg=self.cfg,
-                )
-            )
+            callbacks.append(TokensPerSecondCallback(cfg=self.cfg))
         return callbacks
 
     def get_post_trainer_create_callbacks(self, trainer):
--- a/src/axolotl/core/builders/causal.py
+++ b/src/axolotl/core/builders/causal.py
@@ -90,6 +90,7 @@ class HFCausalTrainerBuilder(TrainerBuilderBase):
             callbacks.append(
                 TokensPerSecondCallback(
                     resume_from_checkpoint=self.cfg.resume_from_checkpoint,
+                    cfg=self.cfg,
                 )
             )
         return callbacks
--- a/src/axolotl/utils/callbacks/tokens_per_second.py
+++ b/src/axolotl/utils/callbacks/tokens_per_second.py
@@ -25,9 +25,19 @@ class TokensPerSecondCallback(TrainerCallback):
     gradient_accumulation_steps and logging_steps.
     """
 
-    def __init__(self, resume_from_checkpoint=None):
+    def __init__(self, resume_from_checkpoint=None, cfg=None):
         super().__init__()
         self.resume_from_checkpoint = resume_from_checkpoint
+        self.cfg = cfg
+
+    def _resolve_resume_from_checkpoint(self):
+        # auto-resume fills in cfg.resume_from_checkpoint only after the callbacks
+        # are built, so the config has to be re-read here rather than snapshotted
+        if isinstance(self.resume_from_checkpoint, str):
+            return self.resume_from_checkpoint
+        if self.cfg is not None:
+            return self.cfg.resume_from_checkpoint
+        return None
 
     def on_train_begin(
         self,
@@ -37,9 +47,10 @@ class TokensPerSecondCallback(TrainerCallback):
         **kwargs,
     ):  # pylint: disable=unused-argument
         """Restore total_tokens state when resuming from checkpoint."""
-        if not isinstance(self.resume_from_checkpoint, str):
+        resume_from_checkpoint = self._resolve_resume_from_checkpoint()
+        if not isinstance(resume_from_checkpoint, str):
             return
-        tokens_state_path = os.path.join(self.resume_from_checkpoint, TOKENS_STATE_FILE)
+        tokens_state_path = os.path.join(resume_from_checkpoint, TOKENS_STATE_FILE)
         if os.path.isfile(tokens_state_path):
             with open(tokens_state_path, "r", encoding="utf-8") as f:
                 tokens_state = json.load(f)
--- a/src/axolotl/utils/callbacks/tokens_per_second.py
+++ b/src/axolotl/utils/callbacks/tokens_per_second.py
@@ -52,8 +52,12 @@ class TokensPerSecondCallback(TrainerCallback):
             return
         tokens_state_path = os.path.join(resume_from_checkpoint, TOKENS_STATE_FILE)
         if os.path.isfile(tokens_state_path):
-            with open(tokens_state_path, "r", encoding="utf-8") as f:
-                tokens_state = json.load(f)
+            try:
+                with open(tokens_state_path, "r", encoding="utf-8") as f:
+                    tokens_state = json.load(f)
+            except (json.JSONDecodeError, OSError):
+                LOG.warning(f"Ignoring unreadable token state at {tokens_state_path}")
+                return
             state.tokens = {
                 "total": torch.tensor(tokens_state.get("total", 0)),
                 "trainable": torch.tensor(tokens_state.get("trainable", 0)),
--- a/tests/utils/callbacks/test_tokens_per_second.py
+++ b/tests/utils/callbacks/test_tokens_per_second.py
@@ -1,6 +1,16 @@
 """Tests for trainable-token throughput accounting."""
 
+import json
+import os
+from unittest.mock import MagicMock
+
+from transformers import TrainerState
+
+from axolotl.core.trainers.constants import TOKENS_STATE_FILE
 from axolotl.core.trainers.utils import trainable_tokens_per_sec_per_gpu
+from axolotl.utils.callbacks.tokens_per_second import TokensPerSecondCallback
+from axolotl.utils.dict import DictDefault
+from axolotl.utils.train import determine_last_checkpoint
 
 
 def _cumulative_after_window(microbatch_token_counts, world_size, start=0.0):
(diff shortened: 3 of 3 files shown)


src/axolotl/core/builders/causal.py
```diff
@@ -90,6 +90,7 @@ def get_callbacks(self):
             callbacks.append(
                 TokensPerSecondCallback(
                     resume_from_checkpoint=self.cfg.resume_from_checkpoint,
+                    cfg=self.cfg,
```
ved1beta:
```suggestion
```

src/axolotl/utils/callbacks/tokens_per_second.py
```diff
@@ -37,9 +47,10 @@ def on_train_begin(
...
         """Restore total_tokens state when resuming from checkpoint."""
-        if not isinstance(self.resume_from_checkpoint, str):
+        resume_from_checkpoint = self._resolve_resume_from_checkpoint()
+        if not isinstance(resume_from_checkpoint, str):
             return
-        tokens_state_path = os.path.join(self.resume_from_checkpoint, TOKENS_STATE_FILE)
+        tokens_state_path = os.path.join(resume_from_checkpoint, TOKENS_STATE_FILE)
         if os.path.isfile(tokens_state_path):
             with open(tokens_state_path, "r", encoding="utf-8") as f:
                 tokens_state = json.load(f)
```
ved1beta:
```suggestion
            try:
                with open(tokens_state_path, "r", encoding="utf-8") as f:
                    tokens_state = json.load(f)
            except (json.JSONDecodeError, OSError):
                LOG.warning(f"Ignoring unreadable token state at {tokens_state_path}")
                return

```

tests/utils/callbacks/test_tokens_per_second.py
```diff
@@ -48,3 +58,84 @@ def test_overhead_lowers_rate(self):
...
+
+
+def _restored_tokens(callback):
+    """Run on_train_begin and report the counters the trainer would see.
+
+    The trainer gates on ``hasattr(state, "tokens")``, so an untouched state
+    must not grow the attribute at all.
+    """
+    state = TrainerState()
+    callback.on_train_begin(MagicMock(), state, MagicMock())
```
ved1beta, line 81: The comment # the builder runs first, while the path is still unresolved names the real invariant worth asserting it through `HFCausalTrainerBuilder.get_callbacks()`

ved1beta (changes requested):

Anai-Guo:
Friendly ping @ved1beta — all three review points were addressed in `e983225` (try/except around the token-state read + a truncated-JSON test, the auto-resume tests now build the callback through `HFCausalTrainerBuilder.get_callbacks()`, and the naming nit). The branch is `MERGEABLE` and the only thing still holding it is the stale "changes requested" review. Could you take another look / dismiss when you get a chance? Happy to rebase if it has gone stale.

ved1beta (approved):

PR #3942, opened 2026-08-22
Fix SAC compatibility with saved tensor hooks

### LINKED ISSUE #3883: Opt out of upcoming SAC × saved_tensors_hooks change (pytorch/pytorch#190581)
[pytorch/pytorch#190581](https://github.com/pytorch/pytorch/pull/190581) makes tensors that selective activation checkpointing (SAC) saves flow through a surrounding `saved_tensors_hooks` context, via a new `checkpoint(..., respect_saved_tensors_hooks=...)` kwarg. The rollout is staged, so the unspecified default (`None`) is a moving target:

1. **Now:** `None` keeps legacy behavior and emits a `FutureWarning` when a user hook is in scope.
2. **Next release:** `None` becomes a **hard error** — you must choose explicitly.
3. **Later release:** the default flips to on (SAC-saved tensors go through the hooks).

Setting the kwarg explicitly now future-proofs you across all three stages.

For axolotl the affected combination is `activation_offloading: hidden_states` + `selective_checkpointing`: that would route SAC-saved tensors through the boundary-only `CheckpointHiddenStatesOffload` hook, but SAC-region offloading is already handled by the dispatch-mode offloader in `selective_checkpointing_offload.py`, so you want the legacy behavior here. Pin it explicitly (this silences the warning now and avoids the hard error in the transition release), feature-gated so it stays a no-op on older torch:

```diff
--- a/src/axolotl/monkeypatch/selective_checkpointing.py
+++ b/src/axolotl/monkeypatch/selective_checkpointing.py
@@
+import inspect
 from typing import Any, Callable
 
 import torch
 from torch.utils.checkpoint import (
     CheckpointPolicy,
+    checkpoint as _torch_checkpoint,
     create_selective_checkpoint_contexts,
 )
@@
 ATTENTION_GROUP = "attention"
+
+# pytorch/pytorch#190581; absent on older torch, so feature-detect the kwarg.
+_SUPPORTS_RESPECT_HOOKS = (
+    "respect_saved_tensors_hooks" in inspect.signature(_torch_checkpoint).parameters
+)
@@
     def enable_with_sac(gradient_checkpointing_kwargs=None, **kwargs):
         gc_kwargs = dict(gradient_checkpointing_kwargs or {})
         gc_kwargs["use_reentrant"] = False
         gc_kwargs["context_fn"] = context_fn
+        if _SUPPORTS_RESPECT_HOOKS:
+            gc_kwargs["respect_saved_tensors_hooks"] = False
         return orig_enable(gradient_checkpointing_kwargs=gc_kwargs, **kwargs)
```

<details>
<summary>Future cleanup (once your minimum supported torch includes #190581) — a task prompt</summary>

> With `respect_saved_tensors_hooks=True`, `selective_checkpointing_offload.py`'s hand-rolled dispatch modes are redundant — SAC now applies ambient hooks to the MUST_SAVE/PREFER_SAVE tensors itself. Refactor:
> - Keep `SacOffloadEngine`; expose it as a `torch.autograd.graph.saved_tensors_hooks` subclass (`pack` → `engine.pack`, `unpack` → `engine.restore`), installed around the forward like `CheckpointHiddenStatesOffload`.
> - For the offload path use the plain SAC `context_fn` (`build_sac_context_fn`) and pass `respect_saved_tensors_hooks=True`.
> - Delete `_OffloadCachingMode`, `_OffloadCachedMode`, `build_sac_offload_context_fn`, and the private imports (`_maybe_detach`/`_detach_helper`, `_VersionWrapper`, `_policy_from_bool`, `SAC_IGNORED_OPS`, `SelectiveCheckpointContext`) plus the `_maybe_detach` try/except shim.
> - Verify two things that differ from the dispatch mode: (1) the ambient hook now offloads a *superset* — boundary inputs too, overlapping the `hidden_states` path, so reconcile them; (2) cross-region prefetch (`_OffloadCachedMode.__enter__`'s `prefetch_region`) has no equivalent in the hook interface, so check for a backward-pass perf regression before deleting.
> - Only do this once the floor is past #190581's release; until then the dispatch mode stays as the old-torch path.

</details>

---

*Disclosure: this issue — the analysis, the diff, and the future-cleanup notes — was drafted with an AI assistant (Claude) and reviewed by a human before posting. The diff is illustrative and has not been run against your codebase (no GPU, no install of your repo), so please treat it as a starting point and validate before merging.*

Excelius-Wang:
# Description

Feature-detect PyTorch's `respect_saved_tensors_hooks` checkpoint
argument and explicitly disable it for Axolotl's selective activation
checkpointing wrapper.

Add regression tests for both supported and unsupported PyTorch APIs,
caller argument immutability, and conflicting caller input.

## Motivation and Context

PyTorch is introducing a staged behavior change for selective activation
checkpointing under ambient `saved_tensors_hooks`. Leaving the new argument
unspecified will first warn and later become an error.

Axolotl's hidden-state activation offloader is intentionally boundary-only,
while SAC-region offloading is handled separately. Passing
`respect_saved_tensors_hooks=False` preserves the current behavior and avoids
double-processing SAC-cached tensors.

The argument is feature-detected so currently supported PyTorch releases
continue to work.

Fixes #3883.

## How has this been tested?

- `pytest` on the three directly related SAC and activation-offload test files:
  25 passed, 3 CUDA-only tests skipped
- PyTorch 2.13 compatibility and argument-forwarding smoke test
- Repository pre-commit hooks on both changed files
- `python -m compileall`
- `git diff --check`

## AI Usage Disclaimer

Yes. OpenAI Codex assisted with issue analysis, implementation, and test
drafting. I reviewed the proposed diff and local verification results before
submission.

N/A

## Types of changes

- [x] Bug fix
- [x] Tests

### DIFF
--- a/src/axolotl/monkeypatch/selective_checkpointing.py
+++ b/src/axolotl/monkeypatch/selective_checkpointing.py
@@ -329,6 +329,8 @@ def apply_selective_checkpointing(
         gc_kwargs["context_fn"] = context_fn
         if _SUPPORTS_RESPECT_SAVED_TENSORS_HOOKS:
             gc_kwargs["respect_saved_tensors_hooks"] = False
+        else:
+            gc_kwargs.pop("respect_saved_tensors_hooks", None)
         return orig_enable(gradient_checkpointing_kwargs=gc_kwargs, **kwargs)
 
     enable_with_sac._axolotl_sac = True
--- a/src/axolotl/monkeypatch/selective_checkpointing.py
+++ b/src/axolotl/monkeypatch/selective_checkpointing.py
@@ -24,7 +24,7 @@ LOG = get_logger(__name__)
 
 ATTENTION_GROUP = "attention"
 
-# Older supported PyTorch releases reject this keyword.
+# PyTorch releases before 2.14 reject this keyword.
 _SUPPORTS_RESPECT_SAVED_TENSORS_HOOKS = (
     "respect_saved_tensors_hooks" in inspect.signature(_torch_checkpoint).parameters
 )
@@ -328,7 +328,7 @@ def apply_selective_checkpointing(
         gc_kwargs["use_reentrant"] = False
         gc_kwargs["context_fn"] = context_fn
         if _SUPPORTS_RESPECT_SAVED_TENSORS_HOOKS:
-            gc_kwargs["respect_saved_tensors_hooks"] = False
+            gc_kwargs.setdefault("respect_saved_tensors_hooks", False)
         else:
             gc_kwargs.pop("respect_saved_tensors_hooks", None)
         return orig_enable(gradient_checkpointing_kwargs=gc_kwargs, **kwargs)
--- a/src/axolotl/monkeypatch/selective_checkpointing.py
+++ b/src/axolotl/monkeypatch/selective_checkpointing.py
@@ -8,11 +8,13 @@ dispatcher-visible.
 
 from __future__ import annotations
 
+import inspect
 from typing import Any, Callable
 
 import torch
 from torch.utils.checkpoint import (
     CheckpointPolicy,
+    checkpoint as _torch_checkpoint,
     create_selective_checkpoint_contexts,
 )
 
@@ -21,6 +23,11 @@ from axolotl.utils.logging import get_logger
 LOG = get_logger(__name__)
 
 ATTENTION_GROUP = "attention"
+
+# Older supported PyTorch releases reject this keyword.
+_SUPPORTS_RESPECT_SAVED_TENSORS_HOOKS = (
+    "respect_saved_tensors_hooks" in inspect.signature(_torch_checkpoint).parameters
+)
 
 _ATEN_ATTENTION_PACKETS = (
     "_scaled_dot_product_flash_attention",
@@ -320,6 +327,8 @@ def apply_selective_checkpointing(
         gc_kwargs = dict(gradient_checkpointing_kwargs or {})
         gc_kwargs["use_reentrant"] = False
         gc_kwargs["context_fn"] = context_fn
+        if _SUPPORTS_RESPECT_SAVED_TENSORS_HOOKS:
+            gc_kwargs["respect_saved_tensors_hooks"] = False
         return orig_enable(gradient_checkpointing_kwargs=gc_kwargs, **kwargs)
 
     enable_with_sac._axolotl_sac = True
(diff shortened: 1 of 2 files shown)


src/axolotl/monkeypatch/selective_checkpointing.py
```diff
@@ -320,6 +327,10 @@ def enable_with_sac(gradient_checkpointing_kwargs=None, **kwargs):
         gc_kwargs = dict(gradient_checkpointing_kwargs or {})
         gc_kwargs["use_reentrant"] = False
         gc_kwargs["context_fn"] = context_fn
+        if _SUPPORTS_RESPECT_SAVED_TENSORS_HOOKS:
+            gc_kwargs["respect_saved_tensors_hooks"] = False
```
NanoCode012: I think this check should be: if this prop is not set, default to False, otherwise, keep user supplied value.

This is important we don't silently override.

tests/monkeypatch/test_selective_checkpointing.py
```diff
@@ -168,6 +169,50 @@ def test_injects_with_none_kwargs(self):
         assert model.seen_kwargs["use_reentrant"] is False
         assert callable(model.seen_kwargs["context_fn"])
 
+    @pytest.mark.parametrize("supports_kwarg", [False, True])
```
NanoCode012, line 172: It seems like both test could be collapsed to be less redundant (sample from claude):
```python
  @pytest.mark.parametrize("supports_kwarg", [False, True])
  @pytest.mark.parametrize("caller_kwargs", [{"preserve_rng_state": False},
  {"respect_saved_tensors_hooks": True}])
  def test_respect_saved_tensors_hooks_is_feature_gated(self, monkeypatch,
  supports_kwarg, caller_kwargs):
      monkeypatch.setattr(
          selective_checkpointing, "_SUPPORTS_RESPECT_SAVED_TENSORS_HOOKS",
  supports_kwarg
      )
      model = self._FakeModel()
      original = dict(caller_kwargs)
      apply_selective_checkpointing(model)
      model.gradient_checkpointing_enable(gradient_checkpointing_kwargs=call
  er_kwargs)

      assert caller_kwargs == original
      if supports_kwarg:
          assert model.seen_kwargs["respect_saved_tensors_hooks"] is False
      else:
          assert "respect_saved_tensors_hooks" not in model.seen_kwargs
```

src/axolotl/monkeypatch/selective_checkpointing.py
```diff
@@ -22,6 +24,11 @@
 
 ATTENTION_GROUP = "attention"
 
+# Older supported PyTorch releases reject this keyword.
```
NanoCode012: Would be worth to say which pytorch release I guess?

Per their tracking, it's unsupported on `< 2.14`

NanoCode012 (changes requested):
Thanks for the PR. Small nits

ved1beta (approved):

NanoCode012 (approved):

PR #3943, opened 2026-08-22
[Fix] Safely access model prefix attribute and guard bias extraction in get_lora_merged_state_dict

arnavahire19:
## Description
In `axolotl.utils.lora.get_lora_merged_state_dict`:
```python
key_list = [key for key, _ in model.named_modules() if model.prefix not in key]
```
If `model` is a standard `torch.nn.Module` or PEFT wrapper that does not define a `.prefix` attribute, this raises `AttributeError: '...' object has no attribute 'prefix'`.

In addition, `target.base_layer` and `ModulesToSaveWrapper` bias handling was omitted or unguarded when creating merged state dictionaries.

This PR:
1. Replaces direct attribute access with `getattr(model, "prefix", None)`.
2. Adds `weight` and `bias` key existence guards across `base_layer`, `ModulesToSaveWrapper`, and standard layers.
3. Adds unit test `test_get_lora_merged_state_dict_without_prefix` in `tests/utils/lora/test_merge_lora.py`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved LoRA state handling for models without a prefix.
  * Prevented errors when optional weight or bias parameters are unavailable.
  * Ensured parameter names and shapes are preserved during LoRA state extraction.

* **Tests**
  * Added regression coverage for models without a prefix attribute.

### DIFF
--- a/tests/utils/lora/test_merge_lora.py
+++ b/tests/utils/lora/test_merge_lora.py
@@ -2707,6 +2707,20 @@ class TestQuantizedBaseMerge:
         }
         out, did, left, _ = _dequantize_quantized_shard(shard, "cpu")
         assert did and left
-        assert out["a.weight"].dtype == torch.bfloat16  # dequantized
-        assert out["b.weight"].dtype == torch.float8_e4m3fn  # left as-is
         assert "b.weight_scale" in out  # its scale not dropped
+
+    def test_get_lora_merged_state_dict_without_prefix(self):
+        """Test get_lora_merged_state_dict on a model without prefix attribute."""
+        from axolotl.utils.lora import get_lora_merged_state_dict
+
+        class SimpleModel(torch.nn.Module):
+            def __init__(self):
+                super().__init__()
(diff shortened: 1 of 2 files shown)


tests/utils/lora/test_merge_lora.py
```diff
@@ -2707,6 +2707,20 @@ def test_partial_dequant_reports_left_quantized(self):
...
+        """Test get_lora_merged_state_dict on a model without prefix attribute."""
+        from axolotl.utils.lora import get_lora_merged_state_dict
+
+        class SimpleModel(torch.nn.Module):
+            def __init__(self):
+                super().__init__()
+                self.fc = torch.nn.Linear(10, 5)
+
+        model = SimpleModel()
+        state_dict = get_lora_merged_state_dict(model)
```
winglian: Why would you call `get_lora_merged_state_dict` on a model that hasn't been wrapped as a `LoraModel` yet by PEFT?

PR #3947, opened 2026-08-25
fix(sonicmoe): stop importing _sonicmoe_wrapper removed from transformers

OnePunchMonk:
## Summary

`sonicmoe_experts_forward_with_lora` currently does:

```python
from transformers.integrations.sonicmoe import _sonicmoe_wrapper
```

That private shim was **deleted upstream** in huggingface/transformers#47334 (`ebea912f`, 2026-07-28, "Kernels and loaders robustification"). On any transformers newer than that commit, every dense (non-NVFP4) sonicmoe expert forward raises `ImportError` at the call site. I hit this while trying to run the real `kernels-community/sonic-moe` kernel for the benchmark in #3936 — getting it to run at all required pinning transformers to a pre-`ebea912f` commit.

## What changed

Keep our own copy of the shim (`_sonicmoe_call`) in `experts.py` rather than reaching for a private upstream symbol:

- We can't just call upstream's public `sonicmoe_experts_forward`, because our whole reason for being here is calling the kernel with **LoRA-materialized** `W_eff = W + scaling * (B @ A)` weights (and our own layout permute), not the module's raw `gate_up_proj`/`down_proj`.
- The `@torch._dynamo.allow_in_graph` decorator is preserved and still applied at module load time. sonic-moe asserts `not torch.compiler.is_compiling()` internally, so the call must stay an opaque FX node — this is exactly why the upstream shim existed, and why inlining the kernel call into the compiled function would be wrong.
- Kernel loading goes through the current public `load_sonicmoe_kernel()`, falling back to the old private `_load_sonicmoe_kernel()` so **pre-#47334 transformers keeps working**. Both are exercised in tests.
- `ACT_MAP` is now local (3 entries, unchanged semantics). The lenient `.get(act_name, "swiglu")` fallback from the old shim is preserved deliberately — upstream's rewrite made unknown activations raise, but that would be a behavior change unrelated to fixing the import, so it's left alone here.

No behavior change on a working (pinned) transformers; this only restores the dense path on current transformers.

## Test plan

- [x] Added `TestSonicmoeKernelLoaderCompat` in `tests/integrations/test_sonicmoe.py` (5 tests, CPU-only, no CUDA / DeepEP / sonic-moe needed):
  - `_load_sonicmoe` resolves against **both** entry points (parametrized over public `load_sonicmoe_kernel` and legacy private `_load_sonicmoe_kernel`)
  - a regression guard that the source no longer references the removed `_sonicmoe_wrapper`
  - argument forwarding into `moe_general_routing_inputs`, including the easy-to-get-wrong positional order (`token_idx` before `expert_ids`, which is *not* the wrapper's keyword order) and that only the first element of the kernel's `(output, _)` tuple is returned
  - unknown-activation fallback to `SWIGLU`
- [x] `pytest tests/integrations/test_sonicmoe.py` — 27 passed
- [x] `ruff check` / `ruff format --check` / `mypy` clean on both changed files (the only mypy hit on `experts.py` is a pre-existing `peft.tuners.param_wrapper` stub warning)
- [x] **Verified on a real H100 against the real `kernels-community/sonic-moe` CUTLASS kernel** (details below).

### GPU verification

Ran on an **H100 80GB HBM3** (SM90) with **unpinned, current** dependencies — `torch 2.9.0+cu128`, `transformers 5.15.1`, `nvidia-cutlass-dsl 4.7.0`, `apache-tvm-ffi 0.1.13.post3`, `kernels 0.16.1` — driving the actual downloaded sonic-moe kernel, not a stand-in:

| # | check | result |
|---|---|---|
| 1 | `transformers 5.15.1` exports | `_sonicmoe_wrapper` **absent**; `load_sonicmoe_kernel` and `_load_sonicmoe_kernel` both present; `is_sonicmoe_loadable()` → `True` |
| 2 | importing the removed symbol | `ImportError: cannot import name '_sonicmoe_wrapper'` |
| 3 | **pre-fix** code (`upstream/main`, `2f16189e`) | forward dies with `ImportError` at `experts.py:125` — the reported breakage, reproduced |
| 4 | **post-fix** `_load_sonicmoe()` | resolves the real `SonicMoE` bundle (`enum ActivationType`, `moe_general_routing_inputs`) |
| 5 | end-to-end forward vs naive fp32 per-token/per-slot reference | `max_abs_diff = 4.88e-4` **without** LoRA and **with** LoRA — bf16 rounding level ✅ |
| 6 | LoRA is genuinely applied (not silently a no-op) | `max|out_lora − out_base| = 4.03e-2` ✅ |
| 7 | `torch.compile` vs eager | `max_abs_diff = 0.0` — confirms `allow_in_graph` still holds the kernel call as one opaque FX node ✅ |
| 8 | backward through the kernel | input grads finite and non-zero ✅ |

Checks 6–8 are there because they're the parts that could silently regress: 5 alone would still pass if LoRA were dropped on the floor, and the `allow_in_graph` placement is the subtle bit of this change.

The repo's own `tests/e2e/integrations/test_sonicmoe.py` still deserves a CI run on hardware, but note that suite currently can't run *at all* against unpinned transformers — which is the point of this PR.

Follow-up to the discussion in #3936.

### DIFF
--- a/src/axolotl/integrations/kernels/libs/sonicmoe/experts.py
+++ b/src/axolotl/integrations/kernels/libs/sonicmoe/experts.py
@@ -1,8 +1,8 @@
 """LoRA-aware sonicmoe experts forward for the transformers ExpertsInterface.
 
-Dense experts wrap upstream ``_sonicmoe_wrapper``, materializing expert LoRA via
-``MoELoRAMaterialize`` before the CUTLASS call. NVFP4 experts (which that kernel
-cannot read) take the grouped dequant path in ``nvfp4_lora`` instead.
+Dense experts call the sonic-moe CUTLASS kernel through ``_sonicmoe_call`` below,
+materializing expert LoRA via ``MoELoRAMaterialize`` first. NVFP4 experts (which that
+kernel cannot read) take the grouped dequant path in ``nvfp4_lora`` instead.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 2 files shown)


NanoCode012 (commented):
Thanks for the PR. I’m surprised we have to do this. Will need to check what’s a more long term solution than having to maintain this

PR #2295, opened 2025-01-28
refactor README; hardcode links to quarto docs; add additional quarto doc pages

djsaunde:
# Description

Title. Moved a lot of README content out into our quarto documentation and added links to it in the README.

Note: quarto doesn't seem to have native support for README.md, which would make this a bit neater w.r.t. linking out to quarto docs, but for now I have hardcoded links instead.

## Motivation and Context

README was getting quite large and a bit unfocused.

## How has this been tested?

N/A.

## Screenshots (if appropriate)

N/A.

## Types of changes

README updates, additional quarto documentation pages.

PR #3948, opened 2026-08-25
SkyPilot example and docs

alex000kim:
Docs only. Adds `docs/skypilot.qmd` (based on in the structure of `ray-integration.qmd`) covering single-node, multi-node, and spot, plus nav entry and cross-links. 
Re-adds what was dropped from the README in #2295 (originally #862), now in the docs format.

## Summary by CodeRabbit

* **Documentation**
  * Added a SkyPilot guide for launching Axolotl training on cloud accounts or Kubernetes clusters.
  * Documented single-node and multi-node training, spot instances, checkpoint mounting, networking, secrets, and dataset reuse.
  * Added SkyPilot references to the installation and multi-node documentation.
  * Added the new guide to the website’s Deployments section.

### DIFF
--- a/docs/installation.qmd
+++ b/docs/installation.qmd
@@ -102,6 +102,10 @@ For providers supporting Docker:
     - [JarvisLabs.ai](https://jarvislabs.ai/templates/axolotl)
     - [Latitude.sh](https://latitude.sh/blueprint/989e0e79-3bf6-41ea-a46b-1f246e309d5c)
 
+### SkyPilot {#sec-skypilot}
+
+To launch Axolotl on your own cloud accounts (AWS, GCP, Azure, etc.) or Kubernetes cluster with a single command, see the [SkyPilot guide](skypilot.qmd).
+
 ### Google Colab {#sec-colab}
 
 [![](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/axolotl-ai-cloud/axolotl/blob/main/examples/colab-notebooks/colab-axolotl-example.ipynb#scrollTo=msOCO4NRmRLa)
--- a/docs/installation.qmd
+++ b/docs/installation.qmd
@@ -102,13 +102,13 @@ For providers supporting Docker:
     - [JarvisLabs.ai](https://jarvislabs.ai/templates/axolotl)
     - [Latitude.sh](https://latitude.sh/blueprint/989e0e79-3bf6-41ea-a46b-1f246e309d5c)
 
-### SkyPilot {#sec-skypilot}
-
-To launch Axolotl on your own cloud accounts (AWS, GCP, Azure, etc.) or Kubernetes cluster with a single command, see the [SkyPilot guide](skypilot.qmd).
-
 ### Google Colab {#sec-colab}
 
 [![](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/axolotl-ai-cloud/axolotl/blob/main/examples/colab-notebooks/colab-axolotl-example.ipynb#scrollTo=msOCO4NRmRLa)
+
+### SkyPilot {#sec-skypilot}
+
+To launch Axolotl on your own cloud accounts (AWS, GCP, Azure, etc.) or Kubernetes cluster with a single command, see the [SkyPilot guide](skypilot.qmd).
 
 ## Platform-Specific Instructions {#sec-platform-specific}
 
--- a/docs/installation.qmd
+++ b/docs/installation.qmd
@@ -102,13 +102,13 @@ For providers supporting Docker:
     - [JarvisLabs.ai](https://jarvislabs.ai/templates/axolotl)
     - [Latitude.sh](https://latitude.sh/blueprint/989e0e79-3bf6-41ea-a46b-1f246e309d5c)
 
-### Google Colab {#sec-colab}
-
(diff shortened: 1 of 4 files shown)


docs/skypilot.qmd
```diff
@@ -0,0 +1,122 @@
...
+To run on spot (preemptible) instances, add `use_spot: true` under `resources` in your task file:
+
+```yaml
+resources:
+  accelerators: A100:1
+  image_id: docker:axolotlai/axolotl-uv:main-latest
+  use_spot: true
+```
+
+For long runs, use managed jobs instead of `sky launch`: SkyPilot then auto-recovers the job if the spot instance is preempted. If the task has a `workdir` or local `file_mounts`, managed jobs stage them through cloud object storage, which needs an object store (S3, GCS, etc.) enabled or `jobs.bucket` set in `~/.sky/config.yaml`.
```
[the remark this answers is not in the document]
alex000kim: reworded to make the bucket optional and mention the controller fallback.

docs/skypilot.qmd
```diff
@@ -0,0 +1,122 @@
...
+Mount a bucket for checkpoints so recovery resumes instead of restarting:
+
+```yaml
+file_mounts:
+  /checkpoints:
+    name: my-axolotl-checkpoints  # bucket is created if it doesn't exist
+    mode: MOUNT_CACHED
+```
+
+Point `output_dir` in your Axolotl config at `/checkpoints` and set `auto_resume_from_checkpoints: true` so a recovered job picks up where it left off. `MOUNT_CACHED` uploads asynchronously, so the newest checkpoint may be incomplete after an abrupt preemption.
```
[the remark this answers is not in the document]
alex000kim: softened the wording and documented deleting the incomplete checkpoint dir as the recovery path.

NanoCode012 (commented):
Thanks for the PR. Were you able to verify each command you’ve written to run successfully? Did you try multi-node as well?

docs/installation.qmd
```diff
@@ -102,6 +102,10 @@ For providers supporting Docker:
     - [JarvisLabs.ai](https://jarvislabs.ai/templates/axolotl)
     - [Latitude.sh](https://latitude.sh/blueprint/989e0e79-3bf6-41ea-a46b-1f246e309d5c)
 
+### SkyPilot {#sec-skypilot}
+
+To launch Axolotl on your own cloud accounts (AWS, GCP, Azure, etc.) or Kubernetes cluster with a single command, see the [SkyPilot guide](skypilot.qmd).
```
NanoCode012, line 107: We should also add this at the bottom, adding by chronological order

docs/multi-node.qmd
```diff
@@ -92,3 +92,7 @@ Please make sure to substitute the placeholder variables:
 The new CLI approach (Option 1) is recommended as it provides consistent argument handling and works seamlessly with other Axolotl CLI features.
 
 More info on the available configs can be found on the Pytorch docs [here](https://pytorch.org/docs/stable/elastic/run.html)
+
+::: {.callout-tip}
+[SkyPilot](skypilot.qmd) can provision the nodes on any cloud or Kubernetes and run this torchrun launcher on each of them automatically.
+:::
```
NanoCode012: I'm not sure it's related/needed here.
```suggestion

```

docs/skypilot.qmd
```diff
@@ -0,0 +1,126 @@
...
+```
+
+## Single node
+
+Create a SkyPilot task file, e.g. `axolotl-skypilot.yaml`:
+
+```yaml
+resources:
+  accelerators: A100:1  # or L40S:1, H100:8, ...
+  image_id: docker:axolotlai/axolotl-uv:main-latest
```
NanoCode012:
```suggestion
  image_id: docker:axolotlai/axolotl:main-latest
```

We recommend this tag instead. We migrated all to be uv based now and plan to deprecate the redundant `-uv` flag in the future.

This applies to other occurrence of this tag below

docs/skypilot.qmd
```diff
@@ -0,0 +1,126 @@
...
+
+```bash
+sky launch -c axolotl axolotl-skypilot.yaml
+sky logs axolotl
+sky down axolotl
+```
+
+Training outputs live on the cluster and disappear with `sky down`. The `run` block always executes in `~/sky_workdir`, so copy outputs off first (`rsync -Pavz axolotl:sky_workdir/outputs/ ./outputs/`) or mount a bucket as shown in the spot section below.
+
+The example configs referenced here use ungated models. For gated models, launch with `--secret HF_TOKEN`: the value is read from your local environment and exposed to the `run` commands.
```
NanoCode012: > launch with `--secret HF_TOKEN`

I'm a bit unfamiliar so may be a silly question. Is this supposed to be exact as written or supposed to be replaced with actual token?

Is there no alternative to pass the below in the yml
```yml
secret:
  HF_TOKEN: $HF_TOKEN
```
which it auto reads?

alex000kim: Exact as written, the bare name tells SkyPilot to read `$HF_TOKEN` from your local shell.

The YAML form doesn't auto-read, unfortunately: `HF_TOKEN: $HF_TOKEN` gets passed through as the **literal** string, so you end up with an opaque 401 rather than an error. 
You can declare it as an empty `secrets:` entry (plural), but the value still has to come from `--secret HF_TOKEN` at launch. Clarified this in the doc.

NanoCode012 (approved):
Thanks for the changes. I pushed some minor fixes and cleaned up the docs.

I think I gave an incorrect suggestion earlier which I reverted.

PR #3965, opened 2026-08-28
feat: add Qwen3.8-Flash-Next

NanoCode012:
# Description

Supersede https://github.com/axolotl-ai-cloud/axolotl/pull/3964

- Adds NF4 QLoRA support
  - [x] Verify merge & inference
- [x] (WIP) Adds NVFP4 LoRA support 
- [x] (WIP) Multimodal QLoRA
- Adds (optional) ngram offload
- Update CCE support
- [x] Cleanup docs
- [x] Add test where relevant

Note: We haven't done multi-gpu testing, so please let us know how it goes!

## How has this been tested?

Ran on 1xB300, e2e + profiling + forward pass & loss checks

## AI Usage Disclaimer

Claude

## Summary by CodeRabbit

* **New Features**
  * Added support for fine-tuning the Qwen3.8-Flash-Next multimodal MoE model.
  * Added QLoRA, vision QLoRA, and NVFP4 MoE-LoRA example configurations.
  * Added optional CPU offloading for selected model parameters to reduce VRAM usage.
  * Added support for packed multimodal training and optimized attention processing.

* **Documentation**
  * Added a model guide, setup instructions, limitations, and training examples.

* **Bug Fixes**
  * Improved offline checkpoint loading, mixed-precision LoRA gradients, and dependency installation reliability.

### DIFF
--- a/src/axolotl/loaders/model.py
+++ b/src/axolotl/loaders/model.py
@@ -539,6 +539,7 @@ class ModelLoader:
         That skip is accelerate 1.13.0's ``accelerator.py`` L1824 branch shadowing the
         ``model.to(self.device)`` at L1864, so an accelerate bump that reorders the two
         silently pulls the table back into VRAM.
+        ``tests/loaders/test_ple_offload_accelerate_contract.py`` fails if that happens.
         """
         suffixes = getattr(self.model, "_no_placement_params", None) or []
         targets = [
--- a/src/axolotl/loaders/model.py
+++ b/src/axolotl/loaders/model.py
@@ -535,6 +535,10 @@ class ModelLoader:
         ``Accelerator.prepare_model``, which skips its ``model.to(device)`` for a quantized
         model carrying an ``hf_device_map`` but does not get one from transformers when the
         map is a single device.
+
+        That skip is accelerate 1.13.0's ``accelerator.py`` L1824 branch shadowing the
+        ``model.to(self.device)`` at L1864, so an accelerate bump that reorders the two
+        silently pulls the table back into VRAM.
         """
         suffixes = getattr(self.model, "_no_placement_params", None) or []
         targets = [
--- a/src/axolotl/loaders/model.py
+++ b/src/axolotl/loaders/model.py
@@ -526,14 +526,48 @@ class ModelLoader:
 
         return lora_config
 
+    def _keep_no_placement_params_on_cpu(self):
+        """Stop the offloaded ``_no_placement_params`` being pulled back into VRAM.
+
+        These land in host RAM on their own (Qwen3.8-Flash-Next's n-gram table arrives as
+        128 checkpoint shards that transformers concatenates on CPU) and the model gathers
+        rows on whatever device the weight sits on. What drags them onto the accelerator is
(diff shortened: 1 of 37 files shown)


src/axolotl/loaders/model.py
```diff
@@ -526,14 +526,48 @@ def _load_adapters(self) -> PeftConfig | None:
...
+        if not targets:
+            LOG.warning(
+                "ple_cpu_offload is set but no `_no_placement_params` of %s are in host "
+                "RAM, so there is nothing to keep there.",
+                self.cfg.model_config_type,
+            )
+            return
+
+        if not hasattr(self.model, "hf_device_map"):
+            self.model.hf_device_map = {"": int(os.environ.get("LOCAL_RANK", 0))}
```
OnePunchMonk, line 555: This depends on a specific accelerate implementation detail: in accelerate 1.13.0's Accelerator.prepare_model (accelerator.py), the check `if (is_loaded_in_8bit or is_loaded_in_4bit) and hf_device_map` at line 1823 is what causes the `.to(self.device)` call at line 1864 to be skipped. Worth pinning that version/line reference in the docstring above, since this is the only thing keeping this correct, and an accelerate upgrade that reorders this logic will break it silently.

src/axolotl/processing_strategies.py
```diff
@@ -1495,8 +1496,12 @@ def get_processing_strategy(
         resolve_model_support,
     )
 
-    support = get_model_support(chat_template_type) or get_model_support_for_processor(
-        processor
+    # model_type first: architectures sharing a processor class with another
+    # family cannot be told apart by the processor matcher alone.
+    support = (
+        get_model_support(model_type)
```
OnePunchMonk, line 1502: model_type is only ever passed from causal.py:590 (`model_type=self.cfg.model_config_type`), so this reordering applies to every architecture resolved through CausalBuilder, not just qwen4_exp. Worth checking that no other registered model support entry currently relies on being matched via get_model_support_for_processor before a model_type match would now take precedence.

OnePunchMonk: Since model_type now comes before chat_template_type and processor for every model going through CausalBuilder, this needs a test showing the existing registered models still resolve to the same strategy as before. Right now nothing catches it if some other model's model_type happens to overlap with a processor or chat_template_type match, and it would silently pick the wrong strategy. Would rather see this scoped to qwen4_exp only, or have a test proving nothing else changes.

src/axolotl/loaders/model.py
OnePunchMonk: If accelerate ever changes prepare_model's placement logic, this breaks silently and badly. PLE is put on CPU on purpose here, expecting the 95GB table to stay off GPU. If a future accelerate version stops honoring hf_device_map the way 1.13.0 does, the whole table gets moved to GPU and the setup that's supposed to fit in ~111GB OOMs instead. I'd suggest either a pinned accelerate version, a real API for blocking placement instead of faking hf_device_map, or at least a test that checks the PLE params stay on CPU after Accelerator.prepare().

PR #3976, opened 2026-09-03
fix: match AsyncGRPOTrainer._get_per_token_logps_and_entropies to TRL…

### LINKED ISSUE #3971: Async GRPO fails with `num_tiles` when data producer and prefetch are enabled

### Expected Behavior

GRPO training should run when both data-producer mode and asynchronous rollout prefetching are enabled:

```yaml
trl:
  use_data_producer: true
  async_prefetch: true
```

`AsyncGRPOTrainer._get_per_token_logps_and_entropies` should remain compatible with the Axolotl’s pinned TRL version `1.9.0` (`main` branch).

To get around this bug, I am setting both the values to `false` for my experiments.

### Current behaviour

The first rollout fails during policy log-probability scoring:

```bash
File ".../axolotl/core/trainers/grpo/async_trainer.py", line 727, in produce
    output = self._trainer._generate_and_score_completions(inputs)
File ".../trl/trainer/grpo_trainer.py", line 2420, in _generate_and_score_completions
    old_per_token_logps, _, _ = self._get_per_token_logps_and_entropies(
TypeError: AsyncGRPOTrainer._get_per_token_logps_and_entropies()
got an unexpected keyword argument 'num_tiles'
```

### Steps to reproduce

Create `rewards.py`:

```python
def constant_reward(completions, **kwargs):
    return [0.0] * len(completions)
```

Create `data.jsonl`:

```json
{"prompt":[{"role":"user","content":"What is 2 + 2?"}]}
{"prompt":[{"role":"user","content":"What is 3 + 3?"}]}
```

Create `repro.yml` using the configuration below.

Start the LoRA-aware vLLM server on GPU 0:

```bash
CUDA_VISIBLE_DEVICES=0 axolotl vllm-serve repro.yml
```

Start training on GPU 1:

```bash
CUDA_VISIBLE_DEVICES=1 axolotl train repro.yml
```

The first rollout fails with the `num_tiles` `TypeError`.

The signature mismatch can also be confirmed without running GRPO:

```python
import inspect

from axolotl.core.trainers.grpo.async_trainer import AsyncGRPOTrainer
from trl import GRPOTrainer

upstream = inspect.signature(
    GRPOTrainer._get_per_token_logps_and_entropies
)
axolotl = inspect.signature(
    AsyncGRPOTrainer._get_per_token_logps_and_entropies
)

print(set(upstream.parameters) - set(axolotl.parameters))
```

Output on current `main`:

```text
{'compute_aux_loss', 'spatial_shapes', 'num_tiles', 'image_position_ids'}
```

### Config yaml

```yaml
base_model: Qwen/Qwen2.5-0.5B-Instruct

rl: grpo

datasets:
  - path: ./data.jsonl
    ds_type: json
    split: train

skip_prepare_dataset: true
remove_unused_columns: false
val_set_size: 0.0

output_dir: ./outputs/num-tiles-repro

sequence_len: 256
sample_packing: false
pad_to_sequence_len: false

micro_batch_size: 1
max_steps: 1

adapter: lora
lora_r: 8
lora_alpha: 16
lora_target_linear: true

bf16: true
gradient_checkpointing: false

vllm:
  host: 0.0.0.0
  port: 8000
  gpu_memory_utilization: 0.70
  max_model_len: 512

trl:
  use_vllm: true
  vllm_mode: server
  vllm_server_host: 127.0.0.1
  vllm_server_port: 8000
  vllm_server_timeout: 300
  vllm_lora_sync: true

  use_data_producer: true
  async_prefetch: true
  prefetch_depth: 1

  num_generations: 2
  generation_batch_size: 2
  max_completion_length: 64
  temperature: 0.7

  reward_funcs:
    - rewards.constant_reward
```

### Possible solution

Update [`AsyncGRPOTrainer._get_per_token_logps_and_entropies`](https://github.com/axolotl-ai-cloud/axolotl/blob/7d775808d0ff71f21f553bd7d761631d7fb089e6/src/axolotl/core/trainers/grpo/async_trainer.py#L2872) to match TRL 1.9’s [method](https://github.com/huggingface/trl/blob/885dd8c05e358dccdd80c7c49d08779cf24f8c2a/trl/trainer/grpo_trainer.py#L1447).

- [x] Linux
- [ ] macOS
- [ ] Windows

### Python Version

3.12.14

### axolotl branch-commit

main/6461c03b602bf0410f0388f1d029ba51e84aeaa2

### Acknowledgements

rickkosse:
… 1.9 signature (#3971)

# Description

`AsyncGRPOTrainer._get_per_token_logps_and_entropies` (`src/axolotl/core/trainers/grpo/async_trainer.py`) still had the pre-TRL-1.9 signature: 2 positional/keyword args short (`compute_aux_loss`, `spatial_shapes`, `num_tiles`, `image_position_ids`) and a 2-tuple return instead of a 3-tuple `(logps, entropies, aux_loss)`.

Updated the override to match `trl.GRPOTrainer._get_per_token_logps_and_entropies` (TRL 1.9.0) exactly: added the missing kwargs (including the tile-indexed LFM2-VL branch and MoE `output_router_logits`/`aux_loss` handling), and changed the return type to a 3-tuple. Updated all 8 internal call sites in `async_trainer.py`, `fast_async_trainer.py`, and `tests/e2e/solo/test_batch_flattening.py` to unpack 3 values instead of 2.

## Motivation and Context

Fixes #3971.

When both `use_data_producer: true` and `async_prefetch: true` are set, the first (synchronous warmup) rollout calls TRL's unmodified base `_generate_and_score_completions`, which polymorphically calls `self._get_per_token_logps_and_entropies(...)` with the new TRL 1.9 kwargs. Since axolotl's override didn't accept them, this crashed with:

TypeError: AsyncGRPOTrainer._get_per_token_logps_and_entropies() got an unexpected keyword argument 'num_tiles'

## How has this been tested?

- Verified the new override's signature matches `trl.GRPOTrainer._get_per_token_logps_and_entropies` exactly via `inspect.signature` (same parameter names/order, both return `tuple[Tensor, Tensor | None, Tensor | None]`).
- `pytest tests/core/test_async_grpo.py` and the broader `tests/core/` + `tests/integrations/test_nemo_gym.py` CPU suite: all pass except 2 pre-existing failures unrelated to this change (`ModuleNotFoundError: triton` in this environment, confirmed present on `main` before this change too).
- `tests/e2e/solo/test_batch_flattening.py` (GPU/flash-attn only, not runnable in this environment) collects cleanly (6 tests) after updating its call sites to the new 3-tuple return.
- Ran `pre-commit run --files` (ruff, ruff-format, mypy, bandit at CI-pinned versions) on all 3 changed files — all pass.
- Not tested against a live GRPO run with `use_data_producer` + `async_prefetch` + vLLM server (requires multi-GPU setup).

## AI Usage Disclaimer

Yes Claude Code was used to investigate the issue, implement the fix, and run the test/lint checks described above.

N/A

## Types of changes

## Summary by CodeRabbit

* **New Features**
  * Improved training support for multimodal inputs, including image positions and tile-indexed data.
  * Added optional auxiliary-loss handling for mixture-of-experts models.
  * Enhanced scoring and loss calculations to provide log-probabilities and entropy information.

* **Bug Fixes**
  * Updated asynchronous and replayed training flows to correctly process the expanded scoring results.

* **Tests**
  * Updated batch-flattening coverage for the enhanced training outputs.

### DIFF
--- a/src/axolotl/core/trainers/grpo/async_trainer.py
+++ b/src/axolotl/core/trainers/grpo/async_trainer.py
@@ -1598,6 +1598,21 @@ class AsyncGRPOTrainer(GRPOTrainer):
         else:
             forward_kwargs = {}
 
+        # Recover LFM2-VL tile counts; the full processor drops row/column metadata.
+        num_tiles = None
+        if images is not None and "spatial_shapes" in forward_kwargs:
+            image_info = self.processing_class.image_processor(
+                images=images, return_tensors="pt", return_row_col_info=True
+            )
+            tiles_per_image = image_info["image_rows"] * image_info["image_cols"]
+            if self.processing_class.image_processor.use_thumbnail:
+                tiles_per_image = tiles_per_image + (tiles_per_image > 1).to(
+                    tiles_per_image.dtype
+                )
+            num_tiles = [
+                group.sum().item() for group in torch.split(tiles_per_image, num_images)
+            ]
+
         # Extend token_type_ids / mm_token_type_ids for completion tokens
         for ttid_key in ("token_type_ids", "mm_token_type_ids"):
             if ttid_key in forward_kwargs:
@@ -1640,13 +1655,17 @@ class AsyncGRPOTrainer(GRPOTrainer):
             output["tool_mask"] = tool_mask
         if images is not None:
             output["num_images"] = num_images
+        if num_tiles is not None:
+            output["num_tiles"] = num_tiles
         for k in (
             "pixel_values",
             "image_grid_thw",
             "pixel_attention_mask",
+            "spatial_shapes",
             "image_sizes",
             "token_type_ids",
             "mm_token_type_ids",
+            "image_position_ids",
         ):
             if k in forward_kwargs:
                 output[k] = forward_kwargs[k]
@@ -1762,13 +1781,16 @@ class AsyncGRPOTrainer(GRPOTrainer):
             "pixel_values",
             "image_grid_thw",
             "pixel_attention_mask",
+            "spatial_shapes",
             "image_sizes",
             "token_type_ids",
             "mm_token_type_ids",
+            "image_position_ids",
         ):
             if key in data:
                 forward_kwargs[key] = data[key]
         num_images = data.get("num_images")
+        num_tiles = data.get("num_tiles")
 
         # --- Launch rewards in parallel with logprobs ---
         self._launch_reward_workers(inputs, prompts, completions, completion_ids_list)
@@ -1810,6 +1832,7 @@ class AsyncGRPOTrainer(GRPOTrainer):
                         logits_to_keep,
                         logprob_batch_size,
                         num_images=num_images,
+                        num_tiles=num_tiles,
                         **forward_kwargs,
                     )
                 data["old_per_token_logps"] = old_per_token_logps
@@ -1826,6 +1849,7 @@ class AsyncGRPOTrainer(GRPOTrainer):
                         logits_to_keep,
                         batch_size,
                         num_images=num_images,
+                        num_tiles=num_tiles,
                         **forward_kwargs,
                     )
                 else:
@@ -1844,6 +1868,7 @@ class AsyncGRPOTrainer(GRPOTrainer):
                             logits_to_keep,
                             batch_size,
                             num_images=num_images,
+                            num_tiles=num_tiles,
                             **forward_kwargs,
                         )
                 data["ref_per_token_logps"] = ref_logps
@@ -2146,9 +2171,11 @@ class AsyncGRPOTrainer(GRPOTrainer):
             "pixel_values",
             "image_grid_thw",
             "pixel_attention_mask",
+            "spatial_shapes",
             "image_sizes",
             "token_type_ids",
             "mm_token_type_ids",
+            "image_position_ids",
         ):
             if key in data:
                 val = data[key]
@@ -2167,6 +2194,13 @@ class AsyncGRPOTrainer(GRPOTrainer):
             and len(num_images) == len(data["prompt_ids"])
         ):
             num_images = num_images[s_start:s_end]
+        num_tiles = data.get("num_tiles")
+        if (
+            num_tiles is not None
+            and hasattr(num_tiles, "__getitem__")
+            and len(num_tiles) == len(data["prompt_ids"])
+        ):
+            num_tiles = num_tiles[s_start:s_end]
 
         # --- Launch rewards in parallel with logprobs ---
         self._launch_reward_workers(inputs, prompts, completions, completion_ids_list)
@@ -2203,6 +2237,7 @@ class AsyncGRPOTrainer(GRPOTrainer):
                         logits_to_keep,
                         logprob_batch_size,
                         num_images=num_images,
+                        num_tiles=num_tiles,
                         **forward_kwargs,
                     )
                 if "old_per_token_logps" not in data:
@@ -2253,6 +2288,7 @@ class AsyncGRPOTrainer(GRPOTrainer):
                         logits_to_keep,
                         batch_size,
                         num_images=num_images,
+                        num_tiles=num_tiles,
                         **forward_kwargs,
                     )
                 else:
@@ -2271,6 +2307,7 @@ class AsyncGRPOTrainer(GRPOTrainer):
                             logits_to_keep,
                             batch_size,
                             num_images=num_images,
+                            num_tiles=num_tiles,
                             **forward_kwargs,
                         )
                 if "ref_per_token_logps" not in data:
@@ -2924,8 +2961,9 @@ class AsyncGRPOTrainer(GRPOTrainer):
         all_aux_losses = []
         with autocast_ctx:
             for start in range(0, input_ids.size(0), batch_size):
-                input_ids_batch = input_ids[start : start + batch_size]
-                attention_mask_batch = attention_mask[start : start + batch_size]
+                end = min(start + batch_size, input_ids.size(0))
+                input_ids_batch = input_ids[start:end]
+                attention_mask_batch = attention_mask[start:end]
 
                 # Build model inputs
                 model_inputs = {
@@ -2944,15 +2982,15 @@ class AsyncGRPOTrainer(GRPOTrainer):
                     )
                     row_start, row_end = (
                         cum_rows[start].item(),
-                        cum_rows[start + batch_size].item(),
+                        cum_rows[end].item(),
                     )
                     model_inputs["pixel_values"] = pixel_values[row_start:row_end]
                     cum_imgs = torch.tensor([0] + num_images).cumsum(0)
-                    img_start, img_end = cum_imgs[start], cum_imgs[start + batch_size]
+                    img_start, img_end = cum_imgs[start], cum_imgs[end]
                     model_inputs["image_grid_thw"] = image_grid_thw[img_start:img_end]
                 elif image_position_ids is not None and pixel_values is not None:
                     cum_imgs = torch.tensor([0] + num_images).cumsum(0)
-                    img_start, img_end = cum_imgs[start], cum_imgs[start + batch_size]
+                    img_start, img_end = cum_imgs[start], cum_imgs[end]
                     model_inputs["pixel_values"] = pixel_values[img_start:img_end]
                     model_inputs["image_position_ids"] = image_position_ids[
                         img_start:img_end
@@ -2962,7 +3000,7 @@ class AsyncGRPOTrainer(GRPOTrainer):
                     cum_tiles = torch.tensor([0] + num_tiles).cumsum(0)
                     tile_start, tile_end = (
                         cum_tiles[start],
-                        cum_tiles[start + batch_size],
+                        cum_tiles[end],
                     )
                     model_inputs["pixel_values"] = pixel_values[tile_start:tile_end]
                     model_inputs["pixel_attention_mask"] = pixel_attention_mask[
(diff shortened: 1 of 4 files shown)


src/axolotl/core/trainers/grpo/async_trainer.py
```diff
@@ -2877,15 +2877,23 @@ def _get_per_token_logps_and_entropies(
...
         num_images=None,
         pixel_attention_mask=None,
+        spatial_shapes=None,
+        num_tiles=None,
         image_sizes=None,
         token_type_ids=None,
         mm_token_type_ids=None,
-    ) -> tuple[Any, torch.Tensor | None]:
-        """Compute log-probs and (optionally) entropies for each token.
+        image_position_ids=None,
```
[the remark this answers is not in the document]
ved1beta: @rickkosse please addres this

src/axolotl/core/trainers/grpo/fast_async_trainer.py
```diff
@@ -580,13 +580,16 @@ def _post_advantage_hook(
                                 "pixel_values",
                                 "image_grid_thw",
                                 "pixel_attention_mask",
+                                "spatial_shapes",
+                                "num_tiles",
                                 "image_sizes",
                                 "token_type_ids",
                                 "mm_token_type_ids",
+                                "image_position_ids",
```
ved1beta: r_fwd_kwargs `forwards image_position_ids` but omits `num_images`

src/axolotl/core/trainers/grpo/fast_async_trainer.py
```diff
@@ -580,13 +580,16 @@ def _post_advantage_hook(
                                 "pixel_values",
                                 "image_grid_thw",
                                 "pixel_attention_mask",
+                                "spatial_shapes",
+                                "num_tiles",
```
ved1beta: full-batch `num_tiles/pixel_values/spatial_shapes` passed alongside an r_start:r_end slice of r_ids, so   ` cum_tiles `starts at sample 0. For r_start > 0 the recomputed old_per_token_logps use the wrong images.

src/axolotl/core/trainers/grpo/async_trainer.py
```diff
@@ -2998,9 +3071,13 @@ def _get_per_token_logps_and_entropies(
...
                         all_entropies.append(entropies)
 
+                if compute_aux_loss:
+                    all_aux_losses.append(outputs.aux_loss)
+
         logps = torch.cat(all_logps, dim=0)
         entropies = torch.cat(all_entropies, dim=0) if compute_entropy else None
-        return logps, entropies
+        aux_loss = torch.stack(all_aux_losses).mean() if compute_aux_loss else None
+        return logps, entropies, aux_loss
```
ved1beta, line 3080: torch.stack(all_aux_losses) assumes outputs.aux_loss is a Tensor, but HF's load_balancing_loss_func is typed -> torch.Tensor | int and returns int 0 when router logits are absent:


`TypeError: expected Tensor as element 0 in argument 0, but got int`

ved1beta (changes requested):
ok everything looks good just a small bug 

A| Qwen2.5-0.5B, LoRA, beta=0.02 | ✅ 10/10 steps, KL flat at 2–3e-4
-- | -- | --
B | Qwen2.5-0.5B, full FT, streaming_partial_batch | ✅ 6/6 steps, KL 1.3e-4 → 2.8e-3
C | OLMoE-1B-7B, LoRA, batch_flattening | ✅ 4/4 steps, aux_loss 8.14–8.20

<p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; unicode-bidi: plaintext; color: rgb(191, 191, 191); font-family: -apple-system, &quot;system-ui&quot;, &quot;Segoe UI&quot;, Roboto, sans-serif; font-size: 13px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(18, 19, 20); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;">All four <code style="font-family: monospace; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">_get_per_token_logps_and_entropies</code> 3-tuple call sites executed: deferred scoring, streaming scoring, the <code style="font-family: monospace; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">use_adapter</code> ref branch (LoRA), and the <code style="font-family: monospace; color: rgb(140, 140, 140); background-color: rgb(38, 38, 38); padding: 2px 4px; border-radius: 3px; word-break: break-word; font-size: 0.9em;">ref_model is not None</code> branch (full FT).

NanoCode012:
@ved1beta , how's this PR looking? Is it good to go or need some minor touch ups?

ved1beta:
one review comment is pendng rest is goood , 
> torch.stack(all_aux_losses) assumes outputs.aux_loss is a Tensor, but HF's load_balancing_loss_func is typed -> torch.Tensor | int and returns int 0 when router logits are absent:
TypeError: expected Tensor as element 0 in argument 0, but got int

ved1beta (approved):
okayy , LGTM

linhongyu510:
Re-checked after `40a7767`. **The `int 0` case @ved1beta flagged is fixed** — I confirmed the claim against the real source and then verified the patch closes it:

```
transformers 5.16.1: load_balancing_loss_func
  signature -> (gate_logits, num_experts=None, top_k=2, attention_mask=None) -> torch.Tensor | int
  line 30-31:  if gate_logits is None or not isinstance(gate_logits, tuple):
                   return 0
```

Before the patch `torch.stack([0])` gives exactly the reported `TypeError: expected Tensor as element 0 in argument 0, but got int`; after it, `tensor(0.)`. `tests/core/test_async_grpo.py` on this head: **2 failed, 39 passed** (up from 37) — the 2 are the pre-existing `TestLoraFP8Guard` `ModuleNotFoundError: triton` ones, identical on `main`, since triton ships no macOS wheels. The `test_mixed_int_and_tensor_aux_loss_across_chunks_does_not_crash` case is a good one to have pinned.

### One residual gap: the attribute can be absent entirely

`outputs.aux_loss` on line 3075 is read *before* the `torch.is_tensor` check, so if the output class has no such field it raises before normalization ever runs. That isn't hypothetical — Llama4 satisfies TRL's MoE test but never populates it:

```
trl/trainer/grpo_trainer.py:779
  is_moe = getattr(text_config, "output_router_logits", None) is not None
  # Llama4 text_config.output_router_logits = False  -> is_moe = True  (attribute exists!)
  # router_aux_loss_coef = 0.001 (default, != 0)      -> aux_loss_enabled = True
```

```
Llama4ForCausalLM              sets aux_loss in forward: False
Llama4ForConditionalGeneration sets aux_loss in forward: False
Llama4CausalLMOutputWithPast fields:
  ['loss','logits','past_key_values','hidden_states','attentions','image_hidden_states']
                                                                    ^ no aux_loss
```

Driving the **real** method with that real HF output class:

```
AttributeError: 'Llama4CausalLMOutputWithPast' object has no attribute 'aux_loss'
  raised at async_trainer.py:3075  ->  batch_aux_loss = outputs.aux_loss
```

Separately, `float(None)` also raises `TypeError`, and Mixtral/Qwen2-MoE/Qwen3-MoE all do `aux_loss = None` when `output_router_logits` is falsy. That one needs the flag to not take effect, which the current code guards against by always passing it — I checked PEFT-wrapped Mixtral and `Qwen3VLMoe`/`Glm4vMoe`/`Qwen3_5Moe` and they all honour it (`Qwen3VLMoe` reads it out of `**kwargs`), so I could **not** construct a reachable `None` today. Worth covering anyway since it's one character, but I'd treat only the missing-attribute case as a live bug.

A one-line change covers all three:

```python
batch_aux_loss = getattr(outputs, "aux_loss", None)
...
if not torch.is_tensor(batch_aux_loss):
    batch_aux_loss = torch.tensor(
        float(batch_aux_loss or 0.0), device=input_ids.device, dtype=torch.float32
    )
```

Applied locally against the real method:

| `outputs.aux_loss` | current | with `getattr` + `or 0.0` |
|---|---|---|
| field absent (Llama4) | `AttributeError` | `0.000` ✅ |
| `None` | `TypeError` | `0.000` ✅ |
| `int 0` | `0.000` ✅ | `0.000` ✅ |
| `tensor(1.5)` | `1.500` ✅ | `1.500` ✅ |

Still **2 failed, 39 passed** with it applied — no regressions. Note the `torch.is_tensor` check must stay *before* the `or`, since `bool(tensor(0.))` is `False` and a genuine zero-valued aux_loss would otherwise be silently rebuilt; ordering already handles that, and I verified `tensor(0.)` passes through untouched.

Two things I'd flag rather than assert:

- **This is inherited from TRL, not introduced here.** Upstream `grpo_trainer.py` line 103 is `all_aux_losses.append(outputs.aux_loss)` — same shape. So this is axolotl hardening beyond upstream, which may be more than you want in a fix PR; equally it may be worth an upstream issue.
- I have **not** verified a real Llama4 GRPO run — this is the output-class contract plus the real method, not a training run. Whether Llama4 is a configuration you intend to support through this path is your call; if not, the missing-attribute case is theoretical for your users and only the `getattr` guard's cheapness argues for it.

The mixed-dtype question I also checked and it's a non-issue: `torch.stack` promotes a `bfloat16` chunk and the hardcoded `float32` fallback to `float32` cleanly, so the pinned dtype doesn't break aggregation.

Environment: macOS, Python 3.12, torch 2.14.0 (CPU), transformers 5.16.1, trl 1.9.0, axolotl 0.19.0.dev0, PR head `fd820f6`. Everything above is CPU-reproducible; the triton-gated paths and any real training run remain untestable here.

PR #3982, opened 2026-09-04
fix(trainer): stop forcing model_accepts_loss_kwargs=False for every model (breaks gradient accumulation)

WatchTree-19:
## What

`AxolotlTrainer.__init__` (`src/axolotl/core/trainers/base.py`, L113-131) tries to detect whether the model's `forward()` accepts `num_items_in_batch`, and sets `self.model_accepts_loss_kwargs = False` if it does not. The detection can never succeed, so the flag is forced `False` for essentially every model, silently reverting the transformers >= 4.46 gradient accumulation fix.

Two independent reasons the check cannot pass:

**1. `base_model` is not PEFT-only.** The code walks `.base_model` to unwrap PEFT. But `base_model` is a property on *every* `PreTrainedModel`:

```python
@property
def base_model(self) -> nn.Module:
    return getattr(self, self.base_model_prefix, self)
```

So for a plain `LlamaForCausalLM` the walk descends into `LlamaModel`, the inner module that never computes a loss.

**2. `num_items_in_batch` is not a named parameter of `forward()`.** It arrives through `**kwargs` and is consumed by `self.loss_function` (`ForCausalLMLoss`). So even on the right module, `inspect.signature(forward)` never lists it.

Net effect: the condition is unsatisfiable, the flag goes `False`, transformers stops counting items per accumulation window, and each micro-batch loss becomes a mean over its own supervised tokens. The optimizer step then computes a **mean of micro-batch means** instead of `sum(token losses) / total supervised tokens`.

## The in-code comment is not correct

The comment being replaced says the mis-normalisation inflates the *logged* loss and that "the gradient itself is still correct". Measured directly on `HuggingFaceM4/tiny-random-LlamaForCausalLM`, `gradient_accumulation_steps=2`, comparing the accumulated gradient against the correctly token-weighted gradient:

| supervised tokens per micro-batch | cosine(grad, correct) | grad norm ratio |
|---|---|---|
| [8, 8] | 1.000 | 1.000 |
| [6, 10] | 0.969 | 1.009 |
| [4, 12] | 0.867 | 1.203 |
| [2, 13] | 0.683 | 1.342 |
| [1, 15] | 0.492 | 1.760 |

Error is exactly zero when micro-batches carry equal supervised-token counts and grows with the imbalance. So it is invisible on uniform-length data and worst under completion-only masking (`train_on_inputs: false`) on mixed-length conversations, and under sample packing.

## Also worth flagging

`src/axolotl/monkeypatch/gemma4_loss_kwargs.py` already sets this flag back to `True` for Gemma 4, and its own docstring says `False` "wrongly makes the Trainer withhold `num_items_in_batch` and mis-normalize the loss under gradient accumulation". This block in `base.py` then flips it straight back for that same model.

## Fix

- Unwrap PEFT explicitly with `get_base_model()` rather than walking `.base_model`, so plain `PreTrainedModel`s are left alone.
- Inspect `loss_function`, which is what actually receives `num_items_in_batch`.
- Correct the comment.

## Verification and its limits

Verified by direct gradient comparison, which is the table above, not by the test suite: the relevant tests need a GPU and I do not have one. A tiny-random model on CPU reproduces the effect, so I am happy to add a CPU-runnable regression test if maintainers want one in-tree, and equally happy to be told the detection should be written a different way.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved training compatibility for models and adapters with custom loss functions.
  * Preserved support for batch-size information when the model’s loss function accepts it.

### DIFF
--- a/tests/core/trainers/test_loss_kwargs_detection.py
+++ b/tests/core/trainers/test_loss_kwargs_detection.py
@@ -11,6 +11,7 @@ from transformers.loss.loss_utils import (
     ForTokenClassification,
 )
 
+from axolotl.core.trainers import base as base_module
 from axolotl.core.trainers.base import model_loss_accepts_num_items_in_batch
 
 
@@ -101,8 +102,20 @@ class TestDetection:
 
         assert model_loss_accepts_num_items_in_batch(Model()) is True
 
-    def test_uninspectable_loss_left_alone(self):
+    @pytest.mark.parametrize("exc", [ValueError, TypeError])
+    def test_uninspectable_loss_left_alone(self, monkeypatch, exc):
+        """The fallback for a callable whose signature cannot be retrieved.
+
+        Which builtins expose a signature varies by CPython version, so rather
+        than pick one and hope, force the failure the fallback exists for.
+        """
+
+        def raises(*_args, **_kwargs):
+            raise exc("no signature found")
+
+        monkeypatch.setattr(base_module.inspect, "signature", raises)
+
         class Model:
-            loss_function = print  # builtin, no retrievable signature
+            loss_function = staticmethod(ForCausalLMLoss)
 
         assert model_loss_accepts_num_items_in_batch(Model()) is True
--- a/tests/core/trainers/test_loss_kwargs_detection.py
+++ b/tests/core/trainers/test_loss_kwargs_detection.py
@@ -1,0 +1,108 @@
+"""Tests for detection of whether a model's loss consumes ``num_items_in_batch``."""
+
+# pylint: disable=missing-class-docstring,too-few-public-methods
+
+import inspect
+
+import pytest
+from transformers.loss.loss_utils import (
+    ForCausalLMLoss,
+    ForSequenceClassificationLoss,
+    ForTokenClassification,
+)
+
+from axolotl.core.trainers.base import model_loss_accepts_num_items_in_batch
+
+
+class TestRealTransformersLossSignatures:
+    """Pin the upstream signatures the detection depends on.
+
+    If transformers changes which losses take ``num_items_in_batch`` these fail
+    loudly, which is the intent: the detection would then be making the wrong
+    call for those model types and needs revisiting.
+    """
+
+    def test_causal_lm_loss_takes_num_items_in_batch(self):
+        params = inspect.signature(ForCausalLMLoss).parameters
+        assert "num_items_in_batch" in params
+
+    @pytest.mark.parametrize(
+        "loss_fn", [ForSequenceClassificationLoss, ForTokenClassification]
+    )
+    def test_non_causal_losses_do_not_take_num_items_in_batch(self, loss_fn):
+        params = inspect.signature(loss_fn).parameters
+        assert "num_items_in_batch" not in params
+
+
+class TestDetection:
+    def test_causal_lm_loss_is_accepted(self):
+        class Model:
+            loss_function = staticmethod(ForCausalLMLoss)
+
+        assert model_loss_accepts_num_items_in_batch(Model()) is True
+
+    @pytest.mark.parametrize(
+        "loss_fn", [ForSequenceClassificationLoss, ForTokenClassification]
+    )
+    def test_losses_without_the_kwarg_are_forced_off(self, loss_fn):
+        """The case the maintainer asked about.
+
+        A sequence- or token-classification model's loss genuinely does not take
+        ``num_items_in_batch``, so the flag must be forced False for it.
+        """
+
+        class Model:
+            loss_function = staticmethod(loss_fn)
+
+        assert model_loss_accepts_num_items_in_batch(Model()) is False
+
+    def test_peft_wrapper_is_unwrapped(self):
+        class Inner:
+            loss_function = staticmethod(ForSequenceClassificationLoss)
+
+        class PeftLike:
+            # A PEFT wrapper's own loss_function would resolve to the causal
+            # default; only the wrapped model's loss is authoritative.
+            loss_function = staticmethod(ForCausalLMLoss)
+
+            def get_base_model(self):
+                return Inner()
+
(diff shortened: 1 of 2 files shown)


ved1beta (approved):
consider trimming comments a little rest looks good

NanoCode012:
Do you know if we can have some test for this, for ex: using this path, assert that a certain config which should be forced, does indeed lack the num items in batch kwarg? (Ofc it would break in future if transformers update, but that breaking would be intended)

WatchTree-19:
Yes, and it turns out transformers gives us a clean way to do exactly that.

`ForSequenceClassificationLoss` and `ForTokenClassification` genuinely do not take `num_items_in_batch`, while `ForCausalLMLoss` does. So a sequence- or token-classification config is a real "should be forced" case rather than a hand-rolled stub, and I have asserted both directions against the upstream functions themselves.

I pulled the inline check out into a module-level `model_loss_accepts_num_items_in_batch` so it can be tested without standing up a Trainer. `tests/core/trainers/test_loss_kwargs_detection.py` covers:

- `ForCausalLMLoss` takes the kwarg, so the flag stays True
- `ForSequenceClassificationLoss` and `ForTokenClassification` do not, so the flag is forced False
- a PEFT-style wrapper is unwrapped through `get_base_model()` before inspection, and the wrapped model's loss is what counts
- a causal LM exposing `.base_model` and `.model` properties is not descended into, which is the regression this PR is fixing
- a model with no `loss_function`, and one whose loss has no retrievable signature, both leave the transformers default alone

The first group also pins the upstream signatures directly, so if transformers changes which losses take `num_items_in_batch` these fail rather than silently going quiet. That is the breakage you flagged, and I agree it is the intended kind: if that changes, the detection is making the wrong call for those model types and should be revisited.

One thing on your side: the workflows for this push are sitting at `action_required`, so Tests, Preview and docker-e2e-tests have not executed on the new commit. Whenever you get a moment for "Approve and run workflows" I would rather the new test proved itself in CI than take my word for it.

WatchTree-19:
Yes, and transformers gives a clean way to do exactly that.

`ForSequenceClassificationLoss` and `ForTokenClassification` genuinely do not take `num_items_in_batch`, while `ForCausalLMLoss` does. So a sequence- or token-classification config is a real "should be forced" case rather than a hand-rolled stub, and I have asserted both directions against the upstream functions themselves.

I pulled the inline check out into a module-level `model_loss_accepts_num_items_in_batch` so it can be tested without standing up a Trainer. `tests/core/trainers/test_loss_kwargs_detection.py` covers:

- `ForCausalLMLoss` takes the kwarg, so the flag stays True
- `ForSequenceClassificationLoss` and `ForTokenClassification` do not, so the flag is forced False
- a PEFT-style wrapper is unwrapped through `get_base_model()` before inspection, and the wrapped model's loss is what counts
- a causal LM exposing `.base_model` and `.model` properties is not descended into, which is the regression this PR is fixing
- a model with no `loss_function`, and one whose loss has no retrievable signature, both leave the transformers default alone

The first group also pins the upstream signatures directly, so if transformers changes which losses take `num_items_in_batch` these fail rather than silently going quiet. That is the breakage you flagged, and I agree it is the intended kind: if that changes, the detection is making the wrong call for those model types and should be revisited.

One thing on your side: the workflows for this push are sitting at `action_required`, so Tests, Preview and docker-e2e-tests have not executed on the new commit. Whenever you get a moment for "Approve and run workflows" I would rather the new test proved itself in CI than take my word for it.

tests/core/trainers/test_loss_kwargs_detection.py
```diff
@@ -0,0 +1,108 @@
...
+        class Model:
+            pass
+
+        assert model_loss_accepts_num_items_in_batch(Model()) is True
+
+    def test_uninspectable_loss_left_alone(self):
+        class Model:
+            loss_function = print  # builtin, no retrievable signature
+
+        assert model_loss_accepts_num_items_in_batch(Model()) is True
```
winglian, line 121: Use a genuinely uninspectable callable in the test

CPython exposes inspect.signature(print) as (*args, sep=' ', end='\\n', file=None, flush=False). The helper therefore returns False, so this test always fails instead of exercising the exception fallback. Mock inspect.signature to raise ValueError or use a callable with an explicitly invalid signature.

WatchTree-19: You are right, and thank you for catching it before CI did.

`inspect.signature(print)` raises `ValueError` on Python 3.10, which is what I ran locally, so the test passed and I did not think to check it anywhere else. On newer CPython `print` carries a signature through Argument Clinic, the helper returns False, and the assertion fails. That test was pinned to the interpreter I happened to use rather than to the behaviour it was meant to cover.

Replaced it with a monkeypatched `inspect.signature` that raises, parametrised over `ValueError` and `TypeError` since the helper catches both, so the fallback is exercised regardless of which builtins expose a signature on the interpreter under test.

I confirmed the failure you describe rather than taking it on faith: with a stubbed `inspect` that gives `print` a signature, the old assertion does flip to False and fail, and the new one passes under both exception types.

WatchTree-19:
Trimmed, thanks for the read.

The docstring on `_model_loss_accepts_num_items` is down from 12 lines to 7 and the block in `__init__` from 11 to 4. What I kept is the two things that are not obvious from the code: that `num_items_in_batch` is a parameter of `loss_function` rather than of `forward()`, so inspecting `forward()` never finds it, and that `.base_model` is a property on every `PreTrainedModel`, so walking it descends out of the model that computes the loss. The rest was restating the PR body.

I left the test docstrings alone on the assumption you meant the source comments, but say the word and I will take those down too.

WatchTree-19:
Thanks again for the review. This is approved and the checks are green, so I think the only thing outstanding is the GPU test workflow needing a maintainer to approve the run on a fork PR. Anything you need from me before this goes in with the next scheduled release?

PR #3988, opened 2026-09-09
fix(perplexity): stop scoring padding in batched references

WatchTree-19:
# Description

`Perplexity.compute` tokenizes with `padding=True`, then discards the `attention_mask` and never masks pad positions out of the labels:

```python
references_tokenized = self.tokenizer(
    references, return_tensors="pt", padding=True, truncation=True
)
input_ids: Tensor = references_tokenized["input_ids"]
...
labels_slice = input_ids_slice.clone()
labels_slice[:, :-trg_len] = -100

outputs = model(input_ids=input_ids_slice, labels=labels_slice)
```

Only the sliding-window prefix is masked. Padding is not. So with more than one reference of differing length the model attends across pad positions and is then scored on predicting them, and the number returned is not the perplexity of the text.

The fix keeps the `attention_mask`, slices it alongside `input_ids`, sets pad positions to `-100` in the labels, and passes the mask to the model.

## Motivation and Context

Perplexity here is a scorer, so a wrong value does not surface as a crash. It surfaces as a confident number that is not the quantity it claims to be. The size of the error tracks how much padding the batch happens to carry, so it is not a fixed offset a user could calibrate away.

No open issue; found while reading the callback.

## How has this been tested?

`HuggingFaceTB/SmolLM2-135M`, the model `tests/test_perplexity.py` already uses, with the TinyStories text from `test_perplexity_longer_than_stride` and the shorter one from `test_perplexity_short`. `Perplexity(tokenizer, max_seq_len=512)`, as the tests construct it:

| | current | with the fix |
|---|---|---|
| long text alone | 7.406 | 7.406 |
| short text alone | 10.334 | 10.334 |
| **both together** | **423.049** | **7.560** |

Each text scores between 7 and 11 on its own. Passed together, current `main` reports 423. The batch is 2 x 382 and 356 of the 762 scored targets, 46.7%, are padding.

Both existing tests pass a single reference, so no padding is involved and neither pinned value moves: 7.406 and 10.334 round to the 7.41 and 10.33 they assert, on both branches.

> **Correction.** The table above replaces the one this description opened with, which read 7.663 / 268.425 / 7.956. Those came from an earlier run I did not repeat against the exact strings the repo's tests use, and 7.663 is visibly inconsistent with this repo's own pinned 7.41. Every figure here was produced by calling `Perplexity.compute` on a clean checkout of `main` and on this branch. The correction makes the reported error larger, not smaller.

## Tests added

Two, both in `tests/test_perplexity.py`, both failing on `main`.

`test_batched_references_are_not_inflated_by_padding` asserts the pooled score of two references lies between their individual scores: 7.406 < 7.560 < 10.334 holds here, and 423.049 does not.

`test_padding_is_never_scored` wraps the model's `forward` and asserts the `attention_mask` reaches the model and every pad position is `-100` in the labels.

Verified in both directions: 4 pass with the fix, and the two new ones fail without it while the two existing ones still pass.

## AI Usage Disclaimer

Yes. Claude was used throughout: reading the callback, drafting the patch and the tests, and writing this description. The figures in the table were produced by running the scorer on both branches rather than estimated.

## Types of changes

### DIFF
--- a/src/axolotl/utils/callbacks/perplexity.py
+++ b/src/axolotl/utils/callbacks/perplexity.py
@@ -53,6 +53,8 @@ class Perplexity:
         )
         input_ids: Tensor = references_tokenized["input_ids"]  # type: ignore
         input_ids = input_ids.to(model.device)
+        attention_mask: Tensor = references_tokenized["attention_mask"]  # type: ignore
+        attention_mask = attention_mask.to(model.device)
 
         sequence_length = input_ids.size(1)
 
@@ -64,12 +66,17 @@ class Perplexity:
             end_loc = min(begin_loc + self.max_seq_len, sequence_length)
             trg_len = end_loc - prev_end_loc
             input_ids_slice = input_ids[:, begin_loc:end_loc]
+            attention_mask_slice = attention_mask[:, begin_loc:end_loc]
             labels_slice = input_ids_slice.clone()
             labels_slice[:, :-trg_len] = -100
+            # Padding is not part of the text being measured.
+            labels_slice[attention_mask_slice == 0] = -100
 
             with torch.no_grad():
                 outputs: CausalLMOutput = model(
-                    input_ids=input_ids_slice, labels=labels_slice
+                    input_ids=input_ids_slice,
+                    attention_mask=attention_mask_slice,
+                    labels=labels_slice,
                 )
 
             losses.append(outputs.loss)
--- a/tests/test_perplexity.py
+++ b/tests/test_perplexity.py
@@ -45,3 +45,42 @@ def test_perplexity_short(model, metric):
     result = metric.compute(model, [sample_text])
     ppl = result["score"]
     assert round(ppl, 2) == 10.33
+
+
+def test_batched_references_are_not_inflated_by_padding(model, metric):
+    """Scoring several references at once pads the batch to the longest one.
+
+    Those pad positions are not part of any reference, so the pooled score must
+    still sit between the scores of the individual texts.
+    """
+    long_text = "Once upon a time, there was a little car named Beep. Beep loved to go fast and play in the sun. Beep was a healthy car because he always had good fuel. Good fuel made Beep happy and strong."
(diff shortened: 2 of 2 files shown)


WatchTree-19:
Two things.

**The table in the description was wrong, and I have corrected it.** It read 7.663 / 268.425 / 7.956. Re-running against the exact strings the repo's own tests use gives 7.406 / 423.049 / 7.560, and 7.406 is the value behind the pinned 7.41 in `test_perplexity_longer_than_stride`, so the old long-text figure contradicted this repo's own test and I should have caught that before opening. Every number now in the description came from calling `Perplexity.compute` on a clean checkout of `main` and on this branch. The direction is unchanged and the error is larger than I first reported. The two added tests are unaffected: 7.406 < 7.560 < 10.334 still holds, and 423.049 does not.

**On the window weighting @coderabbitai raises.** It is right, and it is a different defect from this one. `outputs.loss` is a mean over that window's valid targets, and `torch.stack(losses).mean()` weights every window equally, so windows holding different numbers of valid targets are combined wrongly.

Two things came out of measuring it.

It does not move any number in this PR. At `max_seq_len=512`, which is how the two existing tests and my two new ones construct the scorer, the batch is 382 tokens wide and the loop runs once, so equal weighting and token weighting are the same arithmetic.

It is not caused by padding, and it is already present on `main`. Forcing multiple windows with `max_seq_len=128, stride=128`:

| | equal weighting | token weighting |
|---|---|---|
| long text alone, no padding at all | 11.0101 | 10.9887 |
| both texts together, on this branch | 11.1485 | 10.9470 |

The first row is a single reference, so there is no padding anywhere in it, and the two weightings already disagree.

So I would rather not fold it in here. This PR is a masking fix with two tests that pin it, and changing how window losses are combined would move numbers for every multi-window run for an unrelated reason. Happy to open it as a follow-up with its own test, or to add it to this branch if a maintainer would rather have one PR.

PR #3995, opened 2026-09-11
fix(offload+kernels): make gradient checkpointing offload and AMP kernels device-agnostic

li-lizhe:
# Device-agnostic gradient checkpointing offload + AMP kernels

Two related device-agnostic fixes (offload AMP decorators + kernels AMP decorators), combined per maintainer review request.

## What changed

1. **Offload (CPU/disk gradient checkpointing)** — `src/axolotl/monkeypatch/gradient_checkpointing/offload_cpu.py` & `offload_disk.py`:
   - Detect the actual accelerator via `torch.accelerator.current_accelerator()` (falls back to `"cuda"` on old PyTorch) instead of hardcoding `device_type="cuda"` for the AMP `custom_fwd`/`custom_bwd` decorators.
   - Remember the real compute device (`ctx._device = hidden_states.device`) and move hidden states back to it in `backward` instead of a hardcoded `.to("cuda")`.

2. **Kernels AMP decorators** — `src/axolotl/kernels/utils.py`: use the detected accelerator for `custom_fwd`/`custom_bwd` instead of hardcoding `device_type="cuda"`.

## Verification (Ascend 910B2, torch 2.14.0a0 + torch_npu)

- Old `device_type="cuda"` path crashes with `AssertionError: Torch not compiled with CUDA enabled` on a non-CUDA backend.
- With this fix, `torch.accelerator.current_accelerator()` returns `npu` and the AMP decorators build with `device_type=npu`; forward/backward run correctly.

## Tested

Offload CPU + disk forward/backward on NPU — passes.
AMP kernels custom_fwd/custom_bwd constructed with `device_type=npu` — passes.

Not tested: CUDA-specific paths, multi-GPU NCCL/HCCL collectives (NPU env has no NCCL).

PR #3996, opened 2026-09-11
fix(kernels): make AMP custom_fwd/bwd device-agnostic

li-lizhe:
## Problem

`src/axolotl/kernels/utils.py` builds the AMP decorators with a hardcoded `device_type="cuda"`:

```python
torch_amp_custom_fwd = torch.amp.custom_fwd(device_type="cuda")
torch_amp_custom_bwd = torch.amp.custom_bwd(device_type="cuda")
```

`torch.amp.custom_fwd` requires `device_type` to be a valid accelerator for the running torch build. On Ascend NPU (torch_npu), `"cuda"` is not valid, so any custom autograd op that uses these decorators raises an error instead of running on NPU. The same problem affects XPU/MPS.

## Fix

Detect the active accelerator once at import time via `torch.accelerator.current_accelerator()`, falling back to `"cuda"` only when `torch` has no `accelerator` API. The decorators then target the real device.

This is device-agnostic with no `if npu:` branches; behavior on CUDA is unchanged.

## Verification

Same device-agnostic pattern as the verified Ascend NPU fix in LLaMA-Factory checkpointing. On Ascend 910B (torch 2.14.0a0 + torch_npu), `torch.accelerator.current_accelerator()` returns `npu`, so the decorators are built with `device_type="npu"` and the custom ops run correctly.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved automatic accelerator detection for mixed-precision operations.
  - Ensured supported non-CUDA accelerators use the appropriate device type when applying custom forward and backward operations.
  - Retained CUDA as a fallback when accelerator detection is unavailable, improving compatibility without changing existing CUDA behavior.


li-lizhe:
## ✅ NPU Live Verification (Ascend 910B2, torch 2.14.0a0 + torch_npu)

Verified on an Ascend NPU server. The `device_type="cuda"` → `torch.accelerator.current_accelerator()` change works:

```
torch: 2.14.0a0+git69231fe
accel: npu
custom_fwd(device_type=npu) created  [OK]
```

- `torch.accelerator.current_accelerator()` returns `npu` on this hardware
- `torch.amp.custom_fwd(device_type="npu")` constructs cleanly (old `device_type="cuda"` would break AMP context at runtime with "Torch not compiled with CUDA enabled")
- Same device-agnostic pattern already verified end-to-end in the sibling PR #3995 for the gradient-checkpointing offload paths.

Ready to merge.

ved1beta:
make this and 3995 in one pr , probably close this one

li-lizhe:
Closing per @ved1beta review request — the AMP kernels change from this PR has been merged into #3995, which now contains both the offload and kernels device-agnostic fixes in one PR. Thanks for the review!

PR #3997, opened 2026-09-11
properly capture final loss metrics to telemetry

winglian:
we've been seeing basically 0 loss and learning rate across all telemetry runs making it harder to help users find optimal lrs. this fixes that

## Summary by CodeRabbit

* **Bug Fixes**
  * Telemetry now reports the latest available value for each training metric, even when metrics are spread across multiple log entries.
  * Training metrics are preserved when evaluation or partial logs appear after training logs.
  * Genuine zero values are reported correctly.


ved1beta (approved):
looks solid

PR #4002, opened 2026-09-11
fix(sequence_parallel): reject batch_ring with pre-packed multi-doc sequences

### LINKED ISSUE #3608: Ring Attention w/ document packing produces different results

pujitha24:
# Description

`ring_attn_func: batch_ring` (the default ring-attention implementation for context
parallelism when `sample_packing` is not set) silently produces wrong training loss on
pre-packed multi-document sequences, because the underlying ring-flash-attn "batch" API
has no `cu_seqlens` channel and rotates K/V across the whole sequence, ignoring document
boundaries. This adds a runtime guard that detects packed multi-document sequences (via
`position_ids` resets) and raises a clear `ValueError` instead of silently training on
corrupted attention, pointing the user to `ring_attn_func: varlen_llama3`.

## Motivation and Context

Reported in https://github.com/axolotl-ai-cloud/axolotl/issues/3608, with two independent
reproductions (the reporter and a third party) showing ~1.83-1.84x loss inflation for
`cp=2, ring_attn_func=batch_ring` on a 5-document pack vs. the `cp=1` baseline, while
`varlen_llama3` lands within noise of baseline. A single non-packed document is bitwise
identical across all configs, isolating the bug to document-boundary handling.

`validate_ring_attn_func` in `utils/schemas/validation.py` only has the `sample_packing`
config flag to pick a default `ring_attn_func`, but externally pre-packed data (position
IDs reset per document, `sample_packing: false`) is invisible at config-validation time —
it only shows up in the batch at train time. That data silently falls through to
`batch_ring` and gets corrupted with no error or warning.

A prior attempt at this exact fix (#3867) was closed not because it was wrong, but
because a maintainer was mid-refactor of the CP/SP code in #3844, which was expected to
delete this code path outright. #3844 is still open and unmerged, so `batch_ring` (and
this bug) is still live on `main`; this PR ports that approach onto current `main`, where
it applies with no changes needed to the surrounding code.

This is a detection-and-refusal fix, not an auto-repair: it does not reroute
`batch_ring` to `varlen_llama3` on the user's behalf, since that would be a behavior
change deserving its own discussion. It only turns a silent wrong number into a loud,
actionable error.

## How has this been tested?

Added `tests/utils/ctx_managers/test_sequence_parallel_doc_boundaries.py` (CPU-only,
no GPU/distributed setup needed):
- `_has_multiple_documents` against single-doc, multi-doc, and padded variants (5 cases).
- `apply_sequence_parallelism` raises `ValueError` for `batch_ring` + multi-doc
  `position_ids`, and does not raise for the legitimate `batch_ring` + single-doc case.

Ran `pytest tests/utils/ctx_managers/test_sequence_parallel_doc_boundaries.py -v` in a
minimal CPU venv (torch, transformers==5.16.1, pytest) — all 7 tests pass. Confirmed the
test actually exercises the new code: with only the source change stashed, the same test
file fails at collection (`ImportError: cannot import name '_has_multiple_documents'`).

Also ran, against the two changed files, the exact CI-pinned tool versions from
`.pre-commit-config.yaml`: `ruff@v0.15.8 check` and `ruff format --check` (clean after
one auto-format pass), `mypy==1.19.1` (zero new errors — the file's 7 pre-existing mypy
errors are byte-for-byte identical with and without this diff applied), and
`bandit==1.9.4 --ini .bandit` (no issues).

I could not run the actual multi-GPU repro from the issue (needs 2+ GPUs and the
`ring_flash_attn` package; this environment has neither). The targeted test above is a
direct regression test for the code path this PR changes, and its fail-before/pass-after
behavior demonstrates the guard fires exactly when intended.

## AI Usage Disclaimer

Yes — Claude was used to investigate the issue, draft the fix (adapted from the
closed #3867 approach onto current `main`), write the test, and run validation.

N/A (no UI change).

## Types of changes

## Social Handles (Optional)

---
AI assistance: this change was drafted with Claude Code.

Fixes #3608

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented incorrect loss calculations when batch ring attention is used with pre-packed multi-document sequences.
  * Added a clear validation error for unsupported packed-document inputs.
  * Preserved normal processing for single-document sequences.

* **Tests**
  * Added coverage for document-boundary detection, padding scenarios, and validation behavior.


src/axolotl/utils/ctx_managers/sequence_parallel.py
```diff
@@ -19,6 +19,24 @@
...
+    A document boundary shows up as ``position_ids`` resetting to 0 mid-sequence.
+    Mirrors the boundary detection in ``get_cu_seqlens_from_pos_ids`` (reset to 0
+    starts a new document) after stripping right-side zero padding, so a single
+    padded document is not mistaken for a pack.
+    """
+    if position_ids.dim() == 1:
+        position_ids = position_ids.unsqueeze(0)
+    for row in position_ids:
+        padding_length = (row == 0).int().flip(dims=[0]).cumprod(dim=0).sum().item()
+        adjusted_row = row[:-padding_length] if padding_length else row
```
ved1beta: This false-positives on a padded single document. DataCollatorForSeq2Seq pads position_ids with range(remainder_len)
Suggesting masking out padding via attention_mask / labels == -100 (or the pad token in input_ids) before looking for resets

src/axolotl/utils/ctx_managers/sequence_parallel.py
```diff
@@ -52,6 +71,18 @@ def apply_sequence_parallelism(
 
     # Update ring attention params if needed
     if batch.get("position_ids") is not None and batch_size == 1:
+        # batch_ring ignores cu_seqlens: it rotates K/V across position_id resets, so
+        # tokens attend across documents and the loss is silently wrong. Only varlen
+        # respects boundaries within a packed sequence.
+        if ring_attn_func is RingAttnFunc.BATCH_RING and _has_multiple_documents(
+            batch["position_ids"]
+        ):
```
ved1beta: this will fail for micro_batch size >= 2

ved1beta (changes requested):
few comments

pujitha24:
@ved1beta I've pushed changes addressing your review — the branch is now at `a809b95` and CI is green. Could you take another look when you have a moment? Happy to keep iterating if anything is still off.

ved1beta (approved):
lookss good , manually tested

pujitha24:
This is approved with green CI and no outstanding comments as far as I can tell — ready whenever you have a moment. Happy to rebase first if you'd like it freshened.

NanoCode012 (commented):
Does this check only run once on apply or only on certain batch first or always per batch?

PR #4009, opened 2026-09-14
chore: bump transformers to 5.17.0 and drop the flash-attn2 compat patches

NanoCode012:
# Description

FA v3 kernel upstream has been updated to fix GQA backward. This PR removes the patches that defaults to v2.

In parallel, transformers 5.17 also fixes the other pytorch2.13 FA compatibility issue, so we're dropping the needed patch on our end.

## Summary by CodeRabbit

- **Improvements**
  - Flash Attention availability checks now support kernel fallback detection for Flash Attention 2 and 3, improving compatibility across supported environments.
  - Removed the legacy workaround that forced Flash Attention 2 to use a specific kernel version.

- **Chores**
  - Updated the pinned Transformers dependency from 5.16.1 to 5.17.0.


NanoCode012:
Superseded by #4000

PR #4011, opened 2026-09-14
fix(optimizer): pass weight_decay to the LoRA+ optimizer

### LINKED ISSUE #4010: LoRA+ Optimizer Ignores the Configured weight_decay

### Expected Behavior

With `loraplus_lr_ratio` set, the LoRA+ optimizer should still apply the configured `weight_decay` to the parameters that are normally decayed, like the default optimizer path does.

### Current behaviour

Every LoRA+ param group gets `weight_decay=0.0`, whatever `weight_decay` is in the config. Nothing is logged, so the run looks normal.

`OptimizerMixin.create_optimizer` calls `peft.optimizers.create_loraplus_optimizer(opt_model, optimizer_cls, loraplus_lr_ratio=..., loraplus_lr_embedding=..., **optimizer_kwargs)`. `optimizer_kwargs` comes from `Trainer.get_optimizer_cls_and_kwargs`, which only returns `lr`, `betas` and `eps` for AdamW (weight decay is normally applied per param group). PEFT reads the decay from a separate `loraplus_weight_decay` kwarg that defaults to `0.0`, so it is never set:

https://github.com/axolotl-ai-cloud/axolotl/blob/169df0bd5bb35968fe5f6e3d23fc465c51faf116/src/axolotl/core/trainers/mixins/optimizer.py#L142-L153

The same trainers built without `loraplus_lr_ratio` do use `weight_decay`, as do the `lr_groups` / `embedding_lr` groups in `create_optimizer_grouped_parameters`.

### Steps to reproduce

Run this on CPU from a source checkout (no model download needed):

```python
import tempfile

import torch
from peft import LoraConfig, get_peft_model
from transformers import Trainer

from axolotl.core.trainers.mixins.optimizer import OptimizerMixin
from axolotl.core.training_args import AxolotlTrainingArguments


class TinyModel(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.proj = torch.nn.Linear(8, 8)

    def forward(self, x):
        return self.proj(x)


class TinyTrainer(OptimizerMixin, Trainer):
    pass


for loraplus_lr_ratio in [None, 16.0]:
    model = get_peft_model(TinyModel(), LoraConfig(r=2, target_modules=["proj"]))
    with tempfile.TemporaryDirectory() as output_dir:
        args = AxolotlTrainingArguments(
            output_dir=output_dir,
            learning_rate=1e-4,
            weight_decay=0.1,
            loraplus_lr_ratio=loraplus_lr_ratio,
            report_to=[],
        )
        optimizer = TinyTrainer(model=model, args=args).create_optimizer()
        groups = [(g["lr"], g["weight_decay"]) for g in optimizer.param_groups if g["params"]]
        print(f"loraplus_lr_ratio={loraplus_lr_ratio}: (lr, weight_decay) per group = {groups}")
```

Output on `main`:

```text
loraplus_lr_ratio=None: (lr, weight_decay) per group = [(0.0001, 0.1)]
loraplus_lr_ratio=16.0: (lr, weight_decay) per group = [(0.0001, 0.0), (0.0016, 0.0)]
```

In a normal run this is just `adapter: lora`, `loraplus_lr_ratio: 16` and any non-zero `weight_decay`.

### Config yaml

```yaml
adapter: lora
loraplus_lr_ratio: 16
weight_decay: 0.1
```

### Possible solution

Pass `loraplus_weight_decay=self.args.weight_decay` to `create_loraplus_optimizer`. PEFT still keeps the no-decay LoRA B group (biases and norm weights) at `0.0`. LLaMA-Factory's LoRA+ optimizer does the same, taking the decay from `training_args.weight_decay`. With that change the second line above becomes `[(0.0001, 0.1), (0.0016, 0.1)]`. I have the one-line change and a test in `tests/core/test_builders.py` ready and will open a PR referencing this issue.

This has been the behaviour since LoRA+ support was added: the original helper read `optimizer_kwargs.get("weight_decay", 0.0)` from the same kwargs, and the switch to PEFT's helper in #1928 kept it.

- [x] Linux
- [ ] macOS
- [ ] Windows

### Python Version

3.12

### axolotl branch-commit

main/169df0bd5bb35968fe5f6e3d23fc465c51faf116 (torch 2.13.0+cpu, transformers 5.16.1, peft 0.20.0, trl 1.9.0)

### Acknowledgements

I used Claude Code to help trace this and write it up; the reproduction output above is from a run on my machine.

yupengtang:
# Description

Pass the configured `weight_decay` to PEFT's LoRA+ optimizer. `create_loraplus_optimizer` takes it as `loraplus_weight_decay` (default `0.0`), and the kwargs we forward from `get_optimizer_cls_and_kwargs` never contain `weight_decay`, so every LoRA+ run trained without weight decay.

Fixes #4010

## Motivation and Context

With `loraplus_lr_ratio: 16` and `weight_decay: 0.1`, all LoRA+ param groups currently end up with `weight_decay=0.0`, while the same config without LoRA+ uses `0.1`. Details and a reproduction are in #4010. PEFT still keeps its no-decay group (biases and norm weights) at `0.0`, so only the groups that are meant to be decayed change. Existing LoRA+ configs with a non-zero `weight_decay` will start applying it; configs with `weight_decay: 0` are unaffected.

## How has this been tested?

* Added `test_loraplus_optimizer_weight_decay` to `tests/core/test_builders.py`: builds the SFT trainer with `loraplus_lr_ratio: 16` and checks that the decayed param groups use the config's `weight_decay` (`0.01`). It fails on `main` (no param group has a non-zero weight decay) and passes with this change.
* `pytest tests/core/test_builders.py`: 21 passed.
* The reproduction from #4010 prints `[(0.0001, 0.0), (0.0016, 0.0)]` on `main` and `[(0.0001, 0.1), (0.0016, 0.1)]` with this change.
* `ruff check` / `ruff format --check` with the pinned ruff 0.15.8.

Environment: Linux (WSL2), Python 3.12, torch 2.13.0+cpu, transformers 5.16.1, peft 0.20.0, trl 1.9.0.

## AI Usage Disclaimer

Yes. I used Claude Code to help trace the bug and draft the change and this description. I checked the change, and ran the reproduction and tests myself.

## Types of changes

## Summary by CodeRabbit

* **Bug Fixes**
  * LoRA+ training now consistently applies the configured weight decay to optimizer parameter groups.


ved1beta (approved):
GG

tests/core/test_builders.py
```diff
@@ -165,6 +165,22 @@ def test_polora_optimizer(self, sft_cfg, model, tokenizer):
         assert "betas" not in optimizer_kwargs
         assert "eps" not in optimizer_kwargs
 
+    def test_loraplus_optimizer_weight_decay(self, sft_cfg, model, tokenizer):
```
NanoCode012: Can you add a test where you set weight_decay to 0 and assert it as well?

PR #3950, opened 2026-08-25
feat:gguf exports

ved1beta:
## Description

Adds `axolotl export`, which converts a trained checkpoint to
[GGUF](https://github.com/ggml-org/ggml/blob/master/docs/gguf.md) — the format
used by llama.cpp, Ollama, LM Studio, and llamafile.

Export shells out to a user-supplied, pre-built llama.cpp checkout
(`convert_hf_to_gguf.py` for conversion, `llama-quantize` for quant types);
llama.cpp is not added as an Axolotl dependency. The checkout is located via
`export.llama_cpp_dir` or `$LLAMA_CPP_DIR`.

```bash
axolotl merge-lora config.yml       # LoRA/QLoRA runs only
axolotl export config.yml --quantize Q4_K_M,Q8_0

export:
  outtype: f16          # f32 | f16 | bf16 | q8_0 | auto
  quantize: [Q4_K_M]    # additional llama.cpp quant types, one file each
  output_dir:           # default: {output_dir}/gguf
  llama_cpp_dir:        # default: $LLAMA_CPP_DIR
```

Each quant type is written alongside the unquantized conversion, so one run
can emit run-f16.gguf, run-Q4_K_M.gguf, run-Q8_0.gguf.

Adapters

LoRA/QLoRA runs must be merged first — export prefers {output_dir}/merged
and refuses an adapter-only directory with a pointer to axolotl merge-lora.
Exporting an adapter as a standalone GGUF (convert_lora_to_gguf.py, loadable
via --lora) is not wired up yet.

Not covered

imatrix calibration, gguf-split for large models, multimodal mmproj
files, standalone LoRA adapter GGUFs, and pushing to the Hub.

How has this been tested?

Unit tests for CLI wiring and arg precedence (tests/cli/test_cli_export.py),
End-to-end conversion against a real llama.cpp checkout is not in CI.

AI Usage Disclaimer

Claude.

PR #4020, opened 2026-09-18
feat(export): export LoRA adapters to GGUF

### LINKED ISSUE #3998: Export LoRA Adapters to GGUF via convert_lora_to_gguf.py

### 🔖 Feature description

axolotl export (#3950) only converts a full / merged checkpoint via convert_hf_to_gguf.py. llama.cpp also ships [convert_lora_to_gguf.py](https://github.com/ggml-org/llama.cpp/blob/master/convert_lora_to_gguf.py), which turns a PEFT adapter directory into a standalone GGUF LoRA that runtimes load on top of a base GGUF:
```
llama-cli -m base-f16.gguf --lora my-adapter.gguf
```

### ✔️ Solution

Dispatch: in axolotl.cli.export.resolve_model_dir, when output_dir holds an adapter (adapter_config.json present, no merged/), call convert_lora_to_gguf.py <adapter_dir> --base <base> --outfile … --outtype … instead of raising. The script reads base_model_name_or_path from adapter_config.json (which axolotl already writes); pass cfg.base_model as --base when it is a local path so no Hub fetch is needed.
Config: auto-detect from the adapter dir, or an explicit export.lora: true so a run that has both merged/ and an adapter can choose which to export.
Schema: reject export.quantize for adapter exports — llama-quantize only accepts full models. outtype choices for the LoRA converter are f32 f16 bf16 q8_0 auto (no tq1_0/tq2_0), default f32.
Preflight: fail early on adapters llama.cpp cannot convert:
lora_modules_to_save / any embed_tokens or lm_head tensors in the adapter — llama.cpp errors with "Embeddings is present in the adapter", typically from adding tokens during fine-tuning.
DoRA (peft_use_dora) — the converter only handles lora_A / lora_B tensors.
Docs: remove "exporting a LoRA adapter as a standalone GGUF" from the Known gaps list in docs/export.qmd; document that the adapter must be paired with a GGUF of the same base at runtime.

### ❓ Alternatives

na

### 📝 Additional Context

na

### Acknowledgements

MannXo:
# Description

`axolotl export` refused an adapter-only `output_dir` and pointed at `axolotl merge-lora`. It now dispatches to llama.cpp's `convert_lora_to_gguf.py` instead, writing a standalone GGUF LoRA that runtimes load on top of a GGUF of the same base model.

```bash
axolotl export config.yml
llama-cli -m base-f16.gguf --lora outputs/my-run/gguf/my-run-lora-f32.gguf
```

**Dispatch.** `resolve_model_dir` detects the adapter from `adapter_config.json` in `output_dir`. `merged/` still wins when both exist, so nothing changes for a run that was merged first. A new `export.lora` overrides the detection in either direction, and `--lora/--no-lora` overrides it from the CLI.

**Schema.** `export.quantize` is rejected for an adapter export, because `llama-quantize` only takes full models. `outtype` is limited to the five the LoRA converter accepts (`f32`, `f16`, `bf16`, `q8_0`, `auto`), so `tq1_0`/`tq2_0` are out. The default becomes `f32` rather than `f16`, matching `convert_lora_to_gguf.py`'s own default, and an explicitly set `outtype` still wins. The check runs in the model validator when `export.lora` is set, and again at export time for the auto-detected case, where the schema never saw an adapter.

**Preflight.** Three adapter shapes the converter cannot represent are caught before the conversion starts. All three currently fail late, inside `get_tensors`, with `sys.exit(1)` and a bare "Unexpected name" line:

- `peft_use_dora`. The converter has no DoRA handling at all, so the magnitude vectors are neither `lora_A` nor `lora_B`.
- `lora_modules_to_save`. PEFT writes those layers out whole rather than as a low-rank pair.
- Embedding tensors in the adapter (`embed_tokens`, `lm_head`), which is the case llama.cpp prints "Embeddings is present in the adapter" for.

Tensor names come from the safetensors header, read directly from the 8-byte length prefix and its JSON, so no weights are loaded. A legacy `adapter_model.bin` skips that one check and relies on `modules_to_save`, following the same "unparseable input skips the check" pattern as `_tokenizer_vocab_size`.

**Default filename** gains a `-lora` marker (`{output_dir}/gguf/{run}-lora-{ftype}.gguf`) so an adapter export cannot overwrite a full-model one in the same directory.

## Motivation and Context

Fixes #3998.

`axolotl export` (#3950) only converts a full or merged checkpoint. llama.cpp also ships `convert_lora_to_gguf.py`, which turns a PEFT adapter directory into a GGUF LoRA. Exporting an adapter as a standalone GGUF was listed under Known gaps in `docs/export.qmd`; that line is now removed, along with the paragraph saying export refuses an adapter directory.

Two notes where I departed from the issue text:

1. The issue lists the schema as `src/axolotl/utils/schemas/config.py`. It is `src/axolotl/utils/schemas/export.py`.
2. `src/axolotl/cli/args.py` and the generated `src/axolotl/cli/config_options.py` also needed the new key, or `export.lora` would not be overridable from the CLI.

## How has this been tested?

32 new tests, 42 cases with parametrization, across the three existing export test files:

- `tests/utils/test_gguf.py`, `TestLoraPreflight` and `TestExportLoraGGUF`, plus one case for `resolve_llama_cpp` now checking the script it is about to run.
- `tests/cli/test_cli_export.py`, five new `TestResolveModelDir` cases covering the resolution table and a new `TestDoExportLora` for the dispatch.
- `tests/utils/schemas/validation/test_export.py`, the new rejections and the `outtype` default resolution.

**I was not able to run pytest locally.** Every test in `tests/` imports `axolotl`, which imports torch, and the documented install pins a CUDA torch backend that will not build on my machine. CI's CPU matrix is the first real run of these, so please treat a red `tests` job as mine to fix.

What I did verify locally:

- `pre-commit run` on the nine changed files, at the versions pinned in `.pre-commit-config.yaml`, with `SKIP=check-cli-config-options` exactly as `lint.yml` sets it. ruff, ruff-format, mypy and bandit all pass.
- `config_options.py` was hand-written because regenerating it needs an installed axolotl. I reproduced `generate_config_options.py`'s formatting logic against the real `ExportConfig` fields and confirmed it emits the five existing `export.*` entries byte for byte before adding the sixth, so `axolotl generate-cli-config-options --check` should be clean.
- The implementation itself, driven directly against temporary fixtures with a stubbed logger, since `utils/gguf.py` needs nothing but stdlib. 16 of 16 assertions pass on the adapter paths, and 15 of 15 on the CLI resolution and dispatch, including two regression checks that a full-model export still resolves to `merged/`, still defaults to `f16`, and still calls `convert_hf_to_gguf.py`.

I also read `convert_lora_to_gguf.py` at llama.cpp master to confirm the CLI surface rather than infer it. `--base`, `--outfile`, `--outtype` with those five choices, `f32` default, and the `sys.exit(1)` paths the preflight is built around.

`--base` is passed only when `cfg.base_model` is a local directory. Otherwise the converter resolves the base config from the Hub using the `base_model_name_or_path` axolotl already writes into `adapter_config.json`.

## AI Usage Disclaimer

Yes, Claude Code. It was used for the implementation, the tests and this description. I reviewed every line, and the upstream converter's behaviour quoted above was checked against the real script rather than taken on trust.

Not applicable, no UI change.

## Types of changes

New feature (non-breaking). Existing full-model and merged-checkpoint exports are unaffected, and the previous "run merge-lora first" error still fires for an adapter directory when `export.lora: false` is set.

## Summary by CodeRabbit

* **New Features**
  * Added support for exporting LoRA and QLoRA adapters as standalone GGUF LoRA files.
  * Added automatic adapter detection, with options to explicitly enable or disable LoRA export.
  * Added supported output formats and clearer default filenames for LoRA exports.
  * Added validation for unsupported quantization settings and adapter configurations that require merging first.

* **Documentation**
  * Updated export documentation with configuration examples, limitations, and preflight checks.


ved1beta:
everything feels bloated , please lint and make it short and precise

MannXo:
Rebased onto `82109ef`, which picks up #4021. The three ruff errors were unused
imports in `tests/core/test_builders.py`, a file this PR does not touch, so
pre-commit should come back green without a change here.

Trimmed 61 lines, now +472/-24 from +533/-24:

- Dropped the embedding-tensor pre-check in `lora_preflight`, with its tests and
  its docs bullet. It hand-parsed the safetensors header to pre-empt an error
  `convert_lora_to_gguf.py` raises on its own. The DoRA and `modules_to_save`
  guards stay, since both are a line of `adapter_config.json`.
- Comments back to one line each, per the repo's comment style.
- Shorter error messages and docstrings.

Verified with the pinned ruff, `v0.15.8`:

```
$ uvx ruff@v0.15.8 check --no-fix
All checks passed!
$ uvx ruff@v0.15.8 format --check
959 files already formatted
```

I have not run mypy or bandit locally, only ruff, and the new tests still have
not been run here for lack of a GPU-free env.

Of what is left, 281 of the 472 added lines are tests. Say the word if there is
a specific part you want cut and I will take it out.

src/axolotl/utils/schemas/export.py
```diff
@@ -73,3 +96,15 @@ def validate_quantize(self):
                 "is set, e.g. `model-{ftype}.gguf`."
             )
         return self
+
+    @model_validator(mode="after")
+    def validate_lora(self):
+        if self.lora:
+            validate_lora_export(self.outtype, self.quantize)
+        return self
```
ved1beta, line 109: f32 LoRA default never applies from the CLI , load_cfg validates the whole config and writes defaults back,

src/axolotl/utils/gguf.py
```diff
@@ -195,3 +217,47 @@ def export_gguf(
...
+    lora_preflight(adapter_dir)
+
+    cmd = [
+        sys.executable,
+        str(llama_cpp / LORA_CONVERT_SCRIPT),
+        str(adapter_dir),
+        "--outfile",
+        str(converted),
+        "--outtype",
+        outtype,
```
ved1beta, line 253: --base: use the adapter dir, drop the base_model plumbing

ved1beta (changes requested):
Make lora a plain bool, not tri-state

MannXo:
Thanks, all three are in.

`export.lora` is a plain bool now. The auto-detect is gone with it, so an
adapter-only `output_dir` is refused and the error points at `--lora` alongside
`merge-lora`.

`--base` takes the adapter dir. Training pre-saves the base model's config
beside the adapter, which is all the converter wants, so `base_model` is no
longer plumbed through. It is still omitted when there is no `config.json`
there, which leaves the Hub fallback intact for an adapter axolotl did not
write.

On the f32 default, I read your note as the mechanism being broken rather than
the default being unwanted, so I fixed the mechanism: `outtype` defaults to
`None` and resolves to f16, or f32 for a LoRA export. That survives validation
writing defaults back, which `model_fields_set` could not. Say the word if you
would rather it just went.

One thing that fell out of the bool change and is worth a look. `do_export`
built its overrides from every CLI arg, so a flag nobody passed arrived as
`None` and overwrote the config block. That was harmless while the fields were
all optional and would have started raising on `lora`. It now skips `None`.

Still no GPU box here, so the suite is unrun as before.

ved1beta:
fix tests

ved1beta (approved):

